PR #1128, opened 2026-04-27
fix: harden Trust Basis diff hygiene

Rul1an:
What changed

- Sorts Trust Basis diff arrays by serialized claim id instead of enum order.
- Makes library diff behavior deterministic when duplicate claim IDs are present: first claim wins, duplicates are not repeated.
- Keeps CLI duplicate rejection unchanged.
- Clarifies the release wheel interpreter comment after the v3.6.0 release workflow cleanup.

Why

Addresses the remaining Copilot cleanup from the P34/P40 release wave without changing public CLI behavior or release workflow behavior.

Validation

Ran locally:

- `cargo fmt --check`
- `cargo test -p assay-evidence trust_basis_diff -- --nocapture`
- `cargo test -p assay-cli trust_basis_diff -- --nocapture`
- `cargo clippy -p assay-evidence --all-targets -- -D warnings`
- `actionlint .github/workflows/release.yml`
- `git diff --check`

Push-time hooks also passed, including workspace clippy and the linux compile gate.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -665,8 +665,9 @@ jobs:
         with:
           working-directory: assay-python-sdk
           target: ${{ matrix.target }}
-          # setup-python does not affect the manylinux container, so the
-          # interpreter must be explicit for Linux wheel builds.
+          # Keep the interpreter explicit for all wheel targets. This is
+          # required inside the manylinux container and harmless on macOS,
+          # where setup-python provides the same Python version.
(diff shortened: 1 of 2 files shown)


Rul1an:
Addressed the Copilot review comment in the latest push: `claim_id_sort_key` now derives the sort key from the actual serde serialization for `TrustClaimId`, so diff ordering stays tied to the wire format instead of a duplicated manual mapping. Re-ran the targeted Trust Basis tests, clippy, actionlint, and diff check locally.

PR #1390, opened 2026-05-26
docs(experiments): plan overhead phase timing diagnostics

Rul1an:
## Summary

Plans the next Runner-vs-OTel overhead slice after the same-host Arm A decomposition runs: phase timing diagnostics before another broad wall-clock claim.

What changed:

- The overhead workflow now uploads `overhead-runs/` artifacts with `if: always()` when a harness step fails, and records the harness exit code in the step summary.
- Arm A/B/C summary appends now tolerate missing `summary.md` and still leave a useful GitHub step summary.
- The overhead plan, README, findings, and changelog now record the Arm A diagnostic repeat run `26472122983` and explain why phase timing is the right next step.
- The planned phase buckets cover preflight, cgroup setup, monitor attach, child spawn/runtime, event flush, archive write, and health parsing.

## Why

The successful same-host Arm A runs made RSS attribution clear, but wall-clock remains noisy. A diagnostic repeat hit a discarded first sample with the same cgroup-spawn failure pattern seen in the original sanity attempt. The old workflow did not upload partial artifacts after the harness exited non-zero, which made diagnosis depend on inspecting the temporary runner workspace.

This PR fixes that workflow-DX gap and scopes the next measurement slice without adding a new benchmark claim.

## Validation

- `python3 -m unittest discover -s docs/experiments/runner-vs-otel-overhead-2026-05 -p 'test_*.py'` -> 21 OK
- `python3 -m unittest discover -s docs/experiments/cross-runtime-drift-2026-05/compare -p 'test_*.py'` -> 92 OK
- `actionlint .github/workflows/runner-otel-overhead-experiment.yml`
- `zizmor .github/workflows/runner-otel-overhead-experiment.yml` -> no findings
- local changed-docs link check -> OK
- `git diff --check`
- pre-commit + pre-push hooks green

## Non-claims

- Does not add phase timing instrumentation yet.
- Does not rerun Arm A measurements in this PR.
- Does not commit benchmark artifacts.
- Does not publish a wall-clock additive decomposition claim.

### DIFF
--- a/.github/workflows/runner-otel-overhead-experiment.yml
+++ b/.github/workflows/runner-otel-overhead-experiment.yml
@@ -145,6 +145,7 @@ jobs:
           test -f dist/workload.js
 
       - name: Run Arm C overhead harness
+        id: arm-c-harness
         shell: bash
         env:
           REPETITIONS: ${{ inputs.repetitions }}
@@ -158,6 +159,7 @@ jobs:
           if [ "$MEASURE_RSS" = "true" ]; then
             RSS_FLAG=(--measure-rss)
           fi
+          set +e
           python3 docs/experiments/runner-vs-otel-overhead-2026-05/overhead_harness.py \
             --arm arm-c-dual-capture \
             --iterations "$REPETITIONS" \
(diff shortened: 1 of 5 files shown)


Rul1an:
Addressed Copilot's artifact-contract finding in 0f9e9901: each harness step now exposes its exit status, and each arm has a follow-up gate that requires non-empty `overhead-runs/` when the harness exits 0. Uploads still use `if: always()` / `if-no-files-found: warn` so failed harness runs can preserve partial diagnostics when present.

PR #1489, opened 2026-06-04
docs(observability): add claim-semantics composition overview

Rul1an:
## Summary

Adds a reference page that explains how three contracts which already ship
compose into one per-dimension claim-cell decision:

- `claim-classes-v0` — the two-axis claim cell (strength × basis)
- `fidelity-verdict-v0` — the run-wide capture-health gate
- `coverage-descriptor-v0` — the per-dimension coverage ceiling

Read alone, any one of them can over- or under-state a result. The page fixes
the composition order so the same archive always produces the same cell:
capture health gates the strength of a measured positive; coverage gates the
ceiling for exhaustive and bounded-negative claims. A bounded-negative claim is
allowed only when both gates agree, which under the current seeds never holds —
so absence is never read as safety.

The `coverage-aware-side-effect` example is walked as the worked case: swap the
clipped fixture and the positive cells move from `strong` to `partial` while the
blocked negatives stay blocked.

## Scope

Docs only. No new schema, archive member, CLI output, or Trust Basis claim. It
composes contracts that already ship; the Rust helper in
`crates/assay-runner-schema/src/coverage.rs` stays the source of truth.

## Files

- `docs/reference/observability/claim-semantics-overview.md` (new)
- `docs/reference/observability/README.md` (index entry)

### DIFF
--- a/docs/reference/observability/README.md
+++ b/docs/reference/observability/README.md
@@ -15,6 +15,7 @@ governance.
 |---|---|
 | [`claim-boundary-positioning.md`](claim-boundary-positioning.md) | Post-arc positioning and public selection discipline for claim-boundary and evidence-fidelity work. |
 | [`claim-classes-v0.md`](claim-classes-v0.md) | Vocabulary for saying what a trace, archive, or joined artifact can honestly claim. |
+| [`claim-semantics-overview.md`](claim-semantics-overview.md) | How the claim-class cell, the Runner fidelity verdict, and the coverage descriptor compose into one per-dimension claim-cell decision. |
 | [`join-contract-v0.md`](join-contract-v0.md) | Vocabulary for joining trace, SDK, policy, and measured-run evidence without silently upgrading weak keys. |
 | [`observability-fidelity-calibration.md`](observability-fidelity-calibration.md) | Reference note for reading requested-vs-observed signal retention before timing, throughput, or absence claims. |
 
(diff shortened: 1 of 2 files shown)


Rul1an:
Codex review for cd898aca9a7f933cfca045387414eae2bdf33639

No findings.

I checked the docs-only diff against the shipped coverage helper and the claim/fidelity reference docs. The positive-existence wording now matches the helper's current behavior: descriptor presence gates the coverage decision, the caller owns selecting a descriptor whose `observes` covers the effect class, and missing/schema-mismatched descriptors block rather than permit. The worked example keeps exhaustive/absence claims bounded by coverage and capture health.

Residual risk is documentation drift only: if `coverage_descriptor.v0` later starts validating `observes` directly, this overview should be updated in the same slice.

Rul1an:
Codex review for 7b5d968452c089060ab37ca9d738d21d6533ed58

No findings.

I rechecked the updated docs-only diff for this head. The new wording keeps the same composition contract while naming the runner schemas with their fully qualified strings (`assay.runner.fidelity_verdict.v0` and `assay.runner.coverage_descriptor.v0`). The positive-existence rule still matches the shipped helper boundary: descriptor presence gates the decision, callers select the covering descriptor, and missing/schema-mismatched descriptors block rather than permit.

Checks were still running when I reviewed this head, so I did not treat pending CI as a content finding.

PR #1492, opened 2026-06-04
feat(experiment): optional coverage annotation in drift comparator

Rul1an:
## Summary

Additive, opt-in coverage annotation for the cross-runtime drift comparator (`docs/experiments/cross-runtime-drift-2026-05/compare/drift.py`). A new `--coverage-annotation-out PATH` flag writes a **separate** `assay.coverage_aware_drift.annotation.v0` sidecar document carrying per-dimension claim cells that apply the shipped `coverage_descriptor.v0` ceiling — so a full-overlap (`task-induced`) row is not read as exhaustive-equality or as a bounded-negative effect claim.

The drift report itself is **never modified**: `runtime_drift.v0.2` is `additionalProperties: false`, so the annotation is emitted as its own sidecar document rather than injected into the report payload.

Per measured dimension: positive drift is `partial measured` (the report carries no per-arm capture health), exhaustive equality is `weak derived` (allowed→partial only if a descriptor ever supports complete claims), bounded-negative is blocked; reported dimensions stay `reported` with no coverage gate; `task-induced` rows get a caveat. `coverage_annotation_for_report` validates its input (raises `ValueError` on a non-v0.2 report or non-list rows).

## Scope

Experiment-scoped (Python under `docs/experiments/`); no runner-crate change → no `gates=all`. Default report output is byte-unchanged. The canonical gate stays the Rust `coverage_descriptor.v0` helper.

## Validation

`python3 -m unittest discover -s docs/experiments/cross-runtime-drift-2026-05/compare -p 'test_*.py'` — 105 tests pass (`CoverageAnnotationTest` included), via discover and direct invocation.

### DIFF
--- a/docs/experiments/cross-runtime-drift-2026-05/README.md
+++ b/docs/experiments/cross-runtime-drift-2026-05/README.md
@@ -101,6 +101,16 @@ python3 "$REPO_ROOT/docs/experiments/cross-runtime-drift-2026-05/compare/drift.p
   --path-alias /tmp/work/fixture-output.txt=workdir/output \
   --out-md /tmp/drift.md
 
+# Optional, additive: attach a coverage annotation block to the report.
+# Off by default (default output is byte-unchanged). When set, the report
+# gains a `coverage_annotation` (assay.coverage_aware_drift.annotation.v0)
+# carrying per-dimension claim cells that apply the shipped
+# coverage_descriptor.v0 ceiling, so a full-overlap (task-induced) row is not
+# read as exhaustive-equality or as a bounded-negative effect claim. The
+# canonical gate is crates/assay-runner-schema/src/coverage.rs; see also
+# examples/coverage-aware-drift-annotation/.
+#   ... --with-coverage-annotation
+
 # Tests (no API keys required):
 python3 -m unittest discover \
   -s "$REPO_ROOT/docs/experiments/cross-runtime-drift-2026-05/compare" \
--- a/docs/experiments/cross-runtime-drift-2026-05/README.md
+++ b/docs/experiments/cross-runtime-drift-2026-05/README.md
@@ -101,15 +101,15 @@ python3 "$REPO_ROOT/docs/experiments/cross-runtime-drift-2026-05/compare/drift.p
   --path-alias /tmp/work/fixture-output.txt=workdir/output \
   --out-md /tmp/drift.md
 
-# Optional, additive: attach a coverage annotation block to the report.
-# Off by default (default output is byte-unchanged). When set, the report
-# gains a `coverage_annotation` (assay.coverage_aware_drift.annotation.v0)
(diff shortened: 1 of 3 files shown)


Rul1an:
Codex review for 521bb47ee3d8f9e848e09118287d4f2cd667137b

Findings:
- [P2] `--with-coverage-annotation` emits a `runtime_drift.v0.2` payload that no longer validates against the v0.2 schema. The comparator adds `payload["coverage_annotation"]` at `docs/experiments/cross-runtime-drift-2026-05/compare/drift.py:2117`, but `docs/reference/runner/schema/runtime-drift-v0.2.schema.json` is top-level `additionalProperties: false` and does not define that property. I reproduced this by running the comparator with `--with-coverage-annotation` on the committed fixtures and then validating the JSON with the existing schema helper; it failed with `$: unexpected keys {'coverage_annotation'}`. Either the schema/contract needs to admit the optional block, or the annotation should be emitted as a separate companion artifact rather than inside an `assay.runner.runtime_drift.v0.2` report.

I did not find other issues in the annotation derivation itself. The default path remains unchanged, the new block is opt-in, and the measured/reported/derived claim-cell behavior matches the documented sample. Local verification on the PR head: `python3 -m unittest discover -s docs/experiments/cross-runtime-drift-2026-05/compare -p 'test_*.py'` passed 103 tests, and direct `python3 docs/experiments/cross-runtime-drift-2026-05/compare/test_drift.py` passed 80 tests.

Rul1an:
Codex review for b4428412de945df1e2eb5eb5f4d4a4aaf56b9af1

Findings:
- [P3] The sidecar still explains positive measured drift as capped at `partial` because the drift report "does not surface per-arm observation health" (`docs/experiments/cross-runtime-drift-2026-05/compare/drift.py:1649`). The comparator already parses per-arm `observation_health` into `ArchiveObservation`, and the report provenance can carry compact health fields. The current behavior is conservative and safe, so I would not treat this as blocking, but the note/wiring is stale: a follow-up can derive a fidelity verdict from the two source archives and raise positive measured strength only when both sides are clean.

The previous schema issue is fixed on this head. `--coverage-annotation-out` now writes a separate `assay.coverage_aware_drift.annotation.v0` sidecar, and the `assay.runner.runtime_drift.v0.2` report remains unchanged and valid against its strict schema.

Local verification:
- `python3 -m unittest discover -s docs/experiments/cross-runtime-drift-2026-05/compare -p 'test_*.py'` passed 105 tests.
- `python3 docs/experiments/cross-runtime-drift-2026-05/compare/test_drift.py` passed 82 tests.
- Generated a drift report with `--coverage-annotation-out`; the report schema-validated as `runtime_drift.v0.2`, contained no `coverage_annotation` key, and the sidecar contained `schema: assay.coverage_aware_drift.annotation.v0`.

PR #1494, opened 2026-06-04
feat(runner): carry datagram-aware network coverage descriptors

Rul1an:
## Summary

- Add datagram-aware network coverage descriptor variants for `sendto` / `sendmsg` peer observations.
- Map `observation_health.network_protocol_coverage` to the matching network descriptor (`connect_only`, `datagram_peer_observed`, or `connect_and_datagram_peer_observed`).
- Wire the coverage-aware side-effect sample and runtime-drift annotation sidecar to select the network descriptor from observation health instead of always using `connect_only`.
- Update runner/observability reference docs so the descriptor layer matches the existing Runner datagram telemetry.

## Boundary

- No eBPF or monitor hook changes in this PR.
- No Runner archive member change.
- No stable report schema change.
- Datagram peer evidence can strengthen positive network observations, but exact peer-set and bounded-negative network claims remain degraded/blocked while blind spots are declared.

## Verification

- `cargo fmt --check`
- `cargo test -p assay-runner-schema --test coverage_descriptor`
- `cargo test -p assay-runner-core kernel::tests --lib`
- `python3 examples/coverage-aware-side-effect/test_report.py`
- `python3 docs/experiments/cross-runtime-drift-2026-05/compare/test_drift.py`
- `git diff --check`

## Delegated runner proof

- head sha: `7c6eb46c39a51dc09e0e83e33dd6e88cdfd830a6`
- gate: `all`
- delegated run: https://github.com/Rul1an/assay/actions/runs/26955238871 (success)

### DIFF
--- a/crates/assay-runner-schema/src/coverage.rs
+++ b/crates/assay-runner-schema/src/coverage.rs
@@ -1,6 +1,6 @@
 use serde::{Deserialize, Serialize};
 
-use crate::ClaimGateDecision;
+use crate::{ClaimGateDecision, NetworkProtocolCoverageStatus};
 
 pub const COVERAGE_DESCRIPTOR_SCHEMA: &str = "assay.runner.coverage_descriptor.v0";
 
@@ -18,6 +18,8 @@ pub enum CoverageCompleteness {
     Full,
     OpenSyscallOnly,
     ConnectOnly,
+    DatagramPeerObserved,
+    ConnectAndDatagramPeerObserved,
     ExecOnly,
 }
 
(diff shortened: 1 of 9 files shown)


Rul1an:
## Codex review for 7c6eb46c39a51dc09e0e83e33dd6e88cdfd830a6

Findings: none blocking.

This is a well-scoped descriptor-layer update for the network coverage surface the Runner health contract already exposes. The important claim boundary is preserved: `datagram_peer_observed` and `connect_and_datagram_peer_observed` remain non-`full`, so exhaustive network claims degrade and bounded-negative network claims stay blocked while blind spots are declared.

What I checked:

- `CoverageCompleteness` now carries datagram-aware network ceilings, and `network_for_protocol_coverage` maps all existing `NetworkProtocolCoverageStatus` variants, returning `None` for `unknown`/`absent` rather than assuming a method.
- `claim_decision_for_effect` is used correctly: a datagram-only descriptor can allow observed datagram positives but degrades a connect-time positive it does not list in `observes`.
- The drift annotation uses a conservative common ceiling across both arms, so the stronger arm never speaks for the pair.
- The side-effect sample now picks the network descriptor from `observation_health.network_protocol_coverage`, while the existing `connect_only` golden fixture remains unchanged.
- Docs keep the right boundary: stronger positive network observation, no exact peer-set or bounded-negative claim from datagram evidence alone.

Non-blocking note: `connect_and_datagram_peer_observed` intentionally drops the older post-connect datagram blind spot and keeps the io_uring blind spot. That is defensible because connect evidence plus explicit-sockaddr datagram capture covers that specific case, but it is the one place where a blind spot is narrowed, so it is worth keeping intentional.

Checks/status: ordinary CI is in progress/mostly green. `lane-check` is expected to stay red until a successful `Runner Spike Delegated` run with `gates=all` is available for this head SHA and the lane-check is re-triggered.

PR #1496, opened 2026-06-04
docs(examples): coverage-claims gate example consumer

Rul1an:
## What

Adds `examples/coverage-claims-gate/` — a small, dependency-free example consumer that reads a coverage annotation sidecar (`assay.coverage_aware_drift.annotation.v0`, as produced by the cross-runtime drift comparator's `--coverage-annotation-out`) plus a set of asserted claims, and emits a deterministic pass/fail report.

## Why

The annotation already classifies each drift dimension by claim strength and basis. This example shows the consumer side: a downstream CI step can mechanically refuse to assert a claim the annotation does not support, instead of silently treating absence or trace-reported signal as if it were measured evidence. It is the enforcement counterpart to the producer-side annotation work.

## Claim kinds

- `positive:DIM` — permitted iff a `measured_DIM_drift` cell exists with strength `strong`/`partial`.
- `exhaustive:DIM` — permitted iff `exhaustive_DIM_equality` is allowed (`partial`); a coverage-degraded `weak` cell is not.
- `bounded_negative:DIM` — permitted iff `DIM` is a measured dimension and not present in `blocked_claims`; not evaluable on a reported/unknown dimension.

## Scope

Example only — no Runner or contract surface changes. The canonical gate semantics stay in `crates/assay-runner-schema/src/coverage.rs`; this checker mirrors the same claim-kind rules over a frozen annotation document.

## Tests

Stdlib `unittest` suite (18 tests):

```
python3 -m unittest discover -s examples/coverage-claims-gate -p 'test_*.py'
```

Exit codes: `0` all claims permitted, `1` any blocked, `2` usage/schema error.

### DIFF
--- a/examples/coverage-claims-gate/README.md
+++ b/examples/coverage-claims-gate/README.md
@@ -1,0 +1,73 @@
+# Coverage-claims gate (example consumer)
+
+A small, dependency-free example showing how a downstream gate can **mechanically
+enforce** coverage claims instead of merely documenting them.
+
+It reads a coverage annotation sidecar
+(`assay.coverage_aware_drift.annotation.v0`, as produced by the cross-runtime
+drift comparator's `--coverage-annotation-out`) and a set of asserted claims, then
+returns a deterministic pass/fail report:
+
+- every asserted claim permitted → exit `0`
+- any claim blocked → exit `1`
+- usage / schema error → exit `2`
+
+This is the consumer side of the honesty layer. The annotation already classifies
+each drift dimension by claim strength and basis; this checker turns that
+classification into an enforceable decision, so a CI step can refuse to assert a
+claim the annotation does not support — rather than silently treating absence or
+trace-reported signal as if it were measured evidence.
+
+It is an example only. It does not change any Runner or contract surface. The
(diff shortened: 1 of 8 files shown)


Rul1an:
## Codex review for 2c22f747bf0a4232f1b06f5d7119d509f2a0f4cb

No findings.

I checked the new `examples/coverage-claims-gate/` consumer against the current drift-comparator assertion semantics. The sample mirrors the important enforcement behavior: positive claims require a `measured_<dimension>_drift` cell with `strong`/`partial`, exhaustive claims require an allowed `exhaustive_<dimension>_equality` cell, and bounded-negative claims are not permitted for reported/unknown dimensions and are blocked when the annotation carries the matching `blocked_claims` entry.

Local verification:

- `python3 -m unittest discover -s examples/coverage-claims-gate -p 'test_*.py'` -> 18 tests passed.
- CLI fixture checks matched the expected JSON reports exactly.
- Exit codes behaved as documented: pass policy `0`, blocked policy `1`, usage/no-claims `2`.

Scope also looks clean: example-only, no Runner/schema/CLI output contract change.

Status: most checks are green; merge is still blocked while `Dependency Security (cargo-deny + audit)` and `Build + Test (windows-latest)` are still in progress.

PR #1497, opened 2026-06-04
docs(examples): coverage fleet summary aggregator example

Rul1an:
## What

Adds `examples/coverage-fleet-summary/` — a dependency-free example that folds many coverage annotation sidecars (`assay.coverage_aware_drift.annotation.v0`) into one fleet-level summary (`assay.coverage_fleet_summary.v0`, example-scoped v0).

## Why

The per-run coverage classification already says, for one run, how strong each drift claim is. This example shows it scales to a whole set of runs as a deterministic local fold: for each measured dimension it reports the strength distribution, how many runs block the bounded-negative claim, and the **fleet floor** — the weakest positive strength seen across the set. That answers "across these runs, the strongest claim I can make everywhere is no better than X" directly.

## Scope

Example only — no Runner or contract surface changes. Consumes only the public annotation sidecar shape; uses local inputs.

## Tests

Stdlib `unittest` suite (11 tests), three synthetic annotation fixtures, and an exact expected-summary fixture:

```
python3 -m unittest discover -s examples/coverage-fleet-summary -p 'test_*.py'
```

### DIFF
--- a/examples/coverage-fleet-summary/README.md
+++ b/examples/coverage-fleet-summary/README.md
@@ -1,0 +1,64 @@
+# Coverage fleet summary (example aggregator)
+
+A small, dependency-free example showing how the per-run coverage honesty
+classification scales to a whole set of runs — using only local inputs and with
+no contract change. It is a deterministic fold over annotation documents on disk.
+
+It reads many coverage annotation sidecars
+(`assay.coverage_aware_drift.annotation.v0`, as produced by the cross-runtime
+drift comparator's `--coverage-annotation-out`) and emits one fleet-level summary
+(`assay.coverage_fleet_summary.v0`, example-scoped, v0): for each measured
+dimension it reports the distribution of positive strengths, the distribution of
+exhaustive-equality strengths, how many runs block the bounded-negative claim,
+and the **fleet floor** — the weakest positive strength seen across the set.
+
+The fleet floor answers the operational question directly: "across these runs,
+the strongest positive claim I can make *everywhere* is no better than this." If
+one run in the set degraded to `absent`, the floor is `absent`, even if every
+other run was `strong`.
+
+It is an example only — it changes no Runner or contract surface. It consumes
+only the public annotation sidecar shape.
+
+## Usage
+
+```bash
+# A directory of annotation .json files (sorted, non-recursive)
+python3 aggregate_coverage.py --dir fixtures/runs
+
+# Explicit files
+python3 aggregate_coverage.py fixtures/runs/run-01.json fixtures/runs/run-02.json
+
+# JSON summary
+python3 aggregate_coverage.py --dir fixtures/runs --format json
+```
+
+## What the summary contains
+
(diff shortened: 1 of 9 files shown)


Rul1an:
## Codex review for eec5e1be92aac6626aba3ca4fb651c5d676b84c2

Findings:

- [P2] `fleet_positive_floor` ignores missing runs, so it can overstate the everywhere claim. In `examples/coverage-fleet-summary/aggregate_coverage.py` lines 81-94, a missing `measured_<dimension>_drift` cell increments `measured_positive.missing`, but it does not participate in the floor unless every run is missing (lines 109-112). That means a fleet with one `partial` filesystem run and one run with no filesystem positive cell reports `fleet_positive_floor: "partial"` with `missing: 1`. This conflicts with the README's definition that the floor answers the strongest positive claim supportable *everywhere* across the runs. For that contract, any missing positive cell should lower the floor to `missing` (or the docs/field name should make clear the floor is only over observed runs). Please add a mixed observed+missing fixture/test so this cannot silently regress.

Local verification:

- `python3 -m unittest discover -s examples/coverage-fleet-summary -p 'test_*.py'` -> 11 tests passed.
- I also ran a small mixed observed+missing fold; it reproduced the floor issue above.

Status: many checks are green; merge is still blocked while `Dependency Security (cargo-deny + audit)`, `Build + Test (ubuntu-latest)`, and `Build + Test (windows-latest)` are still in progress. GitHub also reports the branch as behind `main`.

Rul1an:
Good catch on the floor — you're right that folding only the observed runs quietly overstated the everywhere claim, which is exactly the opposite of what this example is meant to demonstrate. Fixed in ff094ba8: the floor now folds every run, and a run with no positive cell counts as the weakest level (`missing`), so a single unsupported run pulls the floor down to `missing`. That makes it an honest "supportable across every run" answer.

I added a mixed observed+missing fixture under `fixtures/mixed/` plus two regression tests (one partial+missing → floor `missing`, one strong+absent → floor `absent`) so it can't silently drift back. The all-observed three-run fixture summary is unchanged, and the README and docstring now spell out the weakest-first ordering and that the floor is computed over every run. Local run is 13 tests green. Ready for another look on ff094ba8 whenever you are.

Rul1an:
## Codex review for ff094ba889f47251d3fb51ede2e0afeae5bfaf33

No findings.

The previous floor issue is fixed on this head. Missing positive cells now fold into `fleet_positive_floor` as the weakest level (`missing`), so the field answers the documented "supportable across every run" contract rather than only the observed-run subset. I re-ran the earlier mixed partial+missing repro and it now reports `fleet_positive_floor: "missing"` with `measured_positive.missing: 1`.

Local verification:

- `python3 -m unittest discover -s examples/coverage-fleet-summary -p 'test_*.py'` -> 13 tests passed.
- Reviewed the new mixed fixture coverage and the strong+absent regression test.

Status: checks are green on this head; GitHub still reports the branch as behind `main`.

PR #1498, opened 2026-06-04
docs(examples): attested-claim composition shape (synthetic demonstrator)

Rul1an:
## What

Adds `examples/attested-shape-demo/` — a **clearly-labelled synthetic demonstrator** exploring the shape a fifth claim basis (`attested`) could take. It implements no real attestation mechanism, adds nothing to the schema, and all fixtures are fabricated demo data ("verification" is decided by a flag, no cryptography).

## Why

So the idea can be reasoned about on engineering merit. The teaching point is the composition rule, which follows the same evidence-first discipline the measured/derived cells already use: an attested claim is only as strong as a verifiable envelope that binds to the same subject digest — otherwise it degrades to `inferred`/`weak`. Composition can cap a claim but never inflate it.

| Envelope state | Effective result |
|---|---|
| none | degrade to `inferred`, capped at `weak` |
| present, not verified | degrade (unverifiable envelope is not evidence) |
| verified, subject mismatch | degrade (does not cover this subject) |
| verified, subject match | declared strength stands |

## Scope

Demonstrator only — no Runner/contract/schema change. Stdlib `unittest` suite (10 tests) + synthetic fixtures + heavily-labelled README.

```
python3 -m unittest discover -s examples/attested-shape-demo -p 'test_*.py'
```

### DIFF
--- a/examples/attested-shape-demo/README.md
+++ b/examples/attested-shape-demo/README.md
@@ -1,0 +1,64 @@
+# Attested-claim composition shape (SYNTHETIC DEMONSTRATOR)
+
+> **This is a synthetic demonstrator.** It does not implement, verify, or assert
+> any real attestation mechanism. Every envelope and digest in the fixtures is
+> fabricated demo data, and "verification" is decided by a flag in the synthetic
+> envelope — no cryptography is performed. Nothing here is a commitment or a
+> stable contract.
+
+## What it explores
+
+The current claim-class schema (`assay.observability.claim_class_cell.v0`)
+recognises four claim bases: `reported`, `measured`, `derived`, `inferred`. This
+demonstrator explores the *shape* a fifth, `attested`, basis could take — purely
+so the idea can be reasoned about on engineering merit. It is **not** in the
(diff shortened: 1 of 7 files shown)


Rul1an:
## Codex review for 4b0039dc76ae49d017daf0e7d2bfcbdc29ba4656

No findings.

I checked the synthetic demonstrator against its stated boundary. The README, module banner, fixtures, and tests all make clear this is a shape-only example: no real attestation mechanism, no cryptographic verification, no schema/contract change. The composition rules are conservative: missing, unverified, or subject-mismatched envelopes degrade to `inferred` with strength capped at `weak`; verified+matching synthetic envelopes allow the declared strength to stand but do not inflate weaker declared claims.

Local verification:

- `python3 -m unittest discover -s examples/attested-shape-demo -p 'test_*.py'` -> 10 tests passed.
- CLI smoke checks for no envelope, verified matching envelope, and wrong-subject envelope produced the expected effective basis/strength.

Status: required checks are green on this head. Scope is example-only: no Runner/schema/CLI output contract change.

PR #1500, opened 2026-06-04
feat(runner): count sendto/sendmsg with no recoverable peer

Rul1an:
## What

Some `sendto`/`sendmsg` calls carry no destination address — `sendto(..., NULL, 0)` or `sendmsg` with no `msg_name` — which happens on a connected socket whose peer was set by an earlier `connect`. The peer is not recoverable from that syscall, and **socket type is not classified at this hook** (a connected stream send hits the same path). Previously these were dropped silently.

Adds two kernel stat counters (`sendto_no_peer` / `sendmsg_no_peer`, indices 14/15), plumbed through `MonitorStatsSnapshot` and the tracepoint hook breakdown. The kernel-capture note gains a `send_no_recoverable_peer=sendto:<n> sendmsg:<m>` suffix **only when the count is non-zero**.

## Honesty / invariant

- A counted send recovers no peer and does **not** raise the network coverage descriptor — it is generic address-less-send visibility, not a datagram claim.
- Runs without address-less sends produce a **byte-identical** note (suffix only when count > 0), so existing clean archives read identically.
- No new schema field, archive member, CLI output, or Trust Basis claim.

## Tests / docs

`assay-runner-core` unit tests (suffix surfaces; coverage not upgraded; byte-identical when zero). Reference doc: `docs/reference/runner/send-no-recoverable-peer-counter.md`.

## Gate

Runner-crate + eBPF → `gates=all` delegated proof on the head SHA (recorded in comments).


Rul1an:
Delegated Phase 1 gates (gate: all) dispatched on the PR head.

- gate: all
- build_ebpf: true
- head_sha: f2b6695e58baaa6b15414bc5e84fc35cf803a6e6
- run: https://github.com/Rul1an/assay/actions/runs/26964917146

Rul1an:
## Codex review for f2b6695e58baaa6b15414bc5e84fc35cf803a6e6

Findings:

- [P2] The new no-peer counter is labelled as datagram-specific, but the hook does not verify socket type. In `crates/assay-ebpf/src/main.rs` lines 506-510 and 545-548, every monitored `sendto` with a null sockaddr and every `sendmsg` with no `msg_name` increments `*_NO_PEER`. The tracepoint logic has not looked up the fd/protocol/socket type, so connected TCP/TLS writes through `sendmsg` can also hit this path. The docs and note label this as `datagram_no_recoverable_peer` and describe it as the connected-datagram blind spot (`docs/reference/runner/datagram-no-recoverable-peer-counter.md` lines 15-28). That can make ordinary address-less stream sends look like the datagram blind spot was exercised. Either gate the counter on actual datagram socket type, or rename/document it as a generic address-less `sendto`/`sendmsg` counter and avoid the datagram-specific suffix.

Local verification:

- `cargo test -p assay-runner-core datagram --lib` -> 5 tests passed.

Status: delegated `gates=all` proof for this head is green (`26964917146`). The PR still shows failed lane-check runs, likely from before the delegated proof was visible; the lane-check may need to be re-run after the green proof.

Rul1an:
Good catch — you're right that without an fd→socket-type lookup, a connected stream `sendmsg`/`sendto` with no address hits the same path, so calling it datagram-specific overclaims. Reframed in 124266c6 to a generic address-less send counter: the note suffix is now `send_no_recoverable_peer=sendto:<n> sendmsg:<m>`, the doc is retitled `send-no-recoverable-peer-counter.md` and states socket type is not classified, and the comments make clear this is not a datagram claim. Behaviour and the byte-identical-note invariant are unchanged; counts still never raise the coverage descriptor. New delegated gates=all proof dispatched on 124266c6 (run linked above). Ready for another look.

Rul1an:
## Codex review for 124266c643cdeb33475d01a7a0f094d866c06633

Findings:

- [P3] The PR description/title still describe the old datagram-specific framing even though the branch was correctly reframed in code/docs. The body still advertises `datagram_no_recoverable_peer=sendto:<n> sendmsg:<m>` and the old `docs/reference/runner/datagram-no-recoverable-peer-counter.md` path, while the actual branch now uses `send_no_recoverable_peer=...` and `docs/reference/runner/send-no-recoverable-peer-counter.md`. This does not affect the implementation, but it can mislead reviewers or release-summary copy if left stale.

Code/docs review:

- The previous P2 is fixed in the tracked files: the counter is now documented as a generic address-less `sendto`/`sendmsg` signal, the suffix is `send_no_recoverable_peer=sendto:<n> sendmsg:<m>`, and the doc explicitly says socket type is not classified. The coverage invariant also still holds: no-peer counts do not upgrade `network_protocol_coverage`.

Local verification:

- `cargo test -p assay-runner-core no_recoverable_peer --lib` -> 2 tests passed.
- `cargo check -p assay-ebpf` -> passed.
- `rg "datagram_no_recoverable_peer|datagram-no-recoverable-peer"` in the checkout -> no tracked-file hits.

Status: delegated `gates=all` proof for this head (`26965805964`) is still in progress, and lane-check is currently red from runs that completed before that proof was available.

Rul1an:
Thanks — fixed the P3 stale framing: the PR title and body now match the reframed code (`send_no_recoverable_peer`, `send-no-recoverable-peer-counter.md`, generic address-less wording). No code change, so the reviewed head 124266c6 is unchanged. Will merge once the delegated proof goes green.

Rul1an:
Delegated Phase 1 gates (gate: all) — green proof on the current head.

- gate: all
- build_ebpf: true
- head_sha: 124266c643cdeb33475d01a7a0f094d866c06633
- run: https://github.com/Rul1an/assay/actions/runs/26965805964

PR #1548, opened 2026-06-06
feat(evidence): in-toto/DSSE attestation over a bundle manifest (ADR-039)

Rul1an:
## Summary
Implements **ADR-039**: an `attestation` module that wraps a bundle `Manifest` as an in-toto v1 Statement (subject = the bundle integrity root) under a v0 evidence-bundle predicate type, and signs/verifies it as a DSSE envelope (Ed25519 over the DSSE PAE of the JCS-canonicalized statement).

Reuses the mandate DSSE primitives (`build_pae` is now `pub(crate)`, `compute_key_id_from_verifying_key`). The anchor (transparency log / timestamp) stays pluggable and external. No new dependencies.

## API
- `statement_from_manifest(manifest, predicate) -> InTotoStatement`
- `sign_statement(statement, &SigningKey) -> DsseEnvelope`
- `verify_envelope(&DsseEnvelope, &VerifyingKey) -> InTotoStatement`

## Honest boundary (in the module + ADR-039)
An attestation binds who-said-it and the bundle content. It does **not** upgrade observed support, and provides no trust root or transparency log on its own. The predicate type is **v0, not a frozen public spec** — ADR-039 remains trigger-gated for publication; this ships the capability behind a versioned, non-committal predicate.

## Scope / safety
- New `crates/assay-evidence/src/attestation.rs` + module registration + one-word `pub(crate)` on `build_pae`. No behavior change to existing signing. Gate.NONE.

## Verification
- `cargo fmt --check`, `cargo clippy -p assay-evidence --lib` clean
- unit test `sign_then_verify_roundtrips_and_detects_tamper`: sign→verify round-trip, payload tamper rejected, wrong-key rejected — passing

### DIFF
--- a/crates/assay-evidence/src/attestation.rs
+++ b/crates/assay-evidence/src/attestation.rs
@@ -1,0 +1,166 @@
+//! In-toto / DSSE attestation over an evidence bundle manifest (ADR-039).
+//!
+//! Wraps a bundle [`Manifest`] as an in-toto v1 Statement and signs it as a DSSE
+//! envelope, reusing the mandate DSSE primitives (PAE + Ed25519). The anchor
+//! (a transparency log or timestamp) stays pluggable and external.
+//!
+//! Honest boundary: an attestation binds who-said-it and the bundle content. It
+//! does NOT upgrade observed support, and provides no trust root or transparency
+//! log on its own.
+
+use crate::bundle::Manifest;
+use crate::crypto::jcs;
+use crate::mandate::signing::{build_pae, compute_key_id_from_verifying_key};
+use anyhow::{Context, Result};
+use base64::{engine::general_purpose::STANDARD as BASE64, Engine};
+use ed25519_dalek::{Signature as Ed25519Signature, Signer, SigningKey, Verifier, VerifyingKey};
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+
+/// in-toto Statement type URI (v1).
+const STATEMENT_TYPE: &str = "https://in-toto.io/Statement/v1";
+/// DSSE payload type for in-toto statements.
+const IN_TOTO_PAYLOAD_TYPE: &str = "application/vnd.in-toto+json";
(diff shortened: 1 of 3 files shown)


Rul1an:
Codex review for 44e7c77104a9c93e80826a775bb951e0be4c18f2

Findings:

- [P2] Enforce the in-toto DSSE payload type before accepting the envelope — `crates/assay-evidence/src/attestation.rs:106`

  `verify_envelope` decodes the payload and builds PAE using `envelope.payload_type`, then returns an `InTotoStatement` without first checking that `payload_type` is exactly `application/vnd.in-toto+json`. A trusted key can therefore sign the same JSON bytes under a different DSSE payload type and this API will still accept it as an in-toto attestation. That opens a payload-type confusion gap at the verifier boundary, and it differs from the existing mandate verifier, which rejects payload type mismatches before verification. Please reject mismatched `payloadType` values, and add a regression test that signs or mutates an envelope with a non-in-toto payload type and expects `verify_envelope` to fail. If this API is intended to verify only Assay evidence-bundle attestations, the same strictness should also cover the returned Statement `_type` / `predicateType`.

Verification:

- `git diff --check 5c53a19628325e9b7dc6322918859e6c9a7fe477...refs/remotes/origin/pr-1548` passed.
- `cargo test -p assay-evidence attestation` passed in a detached worktree.

Check status: the reported check rollup is green, but GitHub marks the PR merge state as `BEHIND` main.

Rul1an:
Good catch, fixed in 85417e9c. verify_envelope now rejects any DSSE payloadType other than application/vnd.in-toto+json before it verifies, so a key that signed the same bytes under a different type can no longer slip through as an in-toto attestation, matching the strictness the mandate verifier already has. It also checks the decoded statement carries the in-toto v1 _type as defense in depth. Added a regression test (verify_rejects_non_in_toto_payload_type) that relabels a genuinely signed envelope and expects the verify to fail. Left predicateType to the caller on purpose since this verifier is generic over the predicate; the DSSE type and the statement _type are the structural boundary.

PR #1564, opened 2026-06-08
feat(runner): shared secret-rule contract fixture + parity (ADR-034 Phase 2)

Rul1an:
ADR-034 Phase 2 (assay side): lock the secret-rule vocabulary so the Rust runner `Redactor` and the Python Plimsoll detector cannot drift.

## What lands

- **`secret-rules.v1.json`** — the canonical `name -> pattern` contract fixture (no secrets in it; patterns only). `redact::rule_specs()` is now the single source of truth behind both `build_rules()` and the fixture.
- **Parity test** — asserts the runner's built-in rules match the fixture exactly. The same fixture is committed to the Plimsoll repo with its own parity test (follow-up PR), so the two implementations stay in lockstep.
- **`sensitive-query-param` rule** — covers URL/query credentials the `credential-assignment` rule misses (`access_token=`, `refresh_token=`, `sig=`, `signature=`), which are glued by underscores or are non-keyword and so escape the `\btoken\b`-style match.
- `google-api-key` pattern aligned to the canonical (no trailing `\b`) to match the already-shipped Plimsoll detector.

## Scope

Deliberately bounded. Full structural URL handling (preserving host while redacting only `user:pass@` userinfo) needs lookbehind/URL parsing the `regex` crate doesn't offer cleanly; the shape + query-param rules already cover recognized credentials in URLs, so the structural userinfo case is left for a later refinement rather than bloating this slice.

## Tests

92 runner-core unit + 4 redaction integration + 1 parity. clippy + fmt clean. Fixture patterns contain no secret literals.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- Secret-rule contract fixture (ADR-034, Phase 2). The runner Redactor's curated rules are now
+  published as `secret-rules.v1.json` (the canonical name->pattern table), with a parity test
+  asserting the built-in rules match it exactly; the same fixture is shared with the Plimsoll detector
+  so the Rust and Python implementations cannot drift. Adds a `sensitive-query-param` rule covering
+  URL/query credentials the assignment rule misses (`access_token=`, `sig=`, `signature=`).
+
 - Runner evidence redaction at capture (ADR-034, Phase 1). The runner-spike run now redacts
   secret-shaped values (provider tokens, PEM keys, JWTs, bearer tokens, `key=value` credentials, and
   flag values such as `--token X`) out of argv and the capability surface before the bundle is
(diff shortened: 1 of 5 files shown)


Rul1an:
gate: all

Runner shared-code change (redact.rs), so the delegated lane runs the full gate set for head SHA 8c0384934c71. Dispatching `Runner Spike Delegated` (gates=all, build_ebpf=true).

Rul1an:
gate: all

Updated head SHA ff5d9c994061 (google-api-key `\b` alignment). Re-dispatching `Runner Spike Delegated` (gates=all, build_ebpf=true) for this head.

Rul1an:
gate: all

Delegated proof for head SHA ff5d9c994061: `Runner Spike Delegated` (gates=all, build_ebpf=true) succeeded — https://github.com/Rul1an/assay/actions/runs/27163836977

PR #1565, opened 2026-06-08
feat(runner): URL userinfo redaction preserving host (ADR-034 Phase 3)

Rul1an:
ADR-034 Phase 3: close the last URL gap — a `user:pass@` credential pair in a network endpoint, whose password is not shape-matchable and so escapes the existing rules.

## What lands

- `Redactor::redact_url_userinfo` — `scheme://user:pass@host` becomes `scheme://<redacted:url-userinfo:H8>@host`, preserving scheme + host via capture-group replacement (no lookbehind needed). Fires only when the userinfo contains a `:` pair: a token-as-username is already caught by the shape pass, and a bare username is not a credential. Idempotent and allowlist-aware.
- It is a **runner-side capture-hygiene transform, not a shared detection rule** (the password is not shape-matchable), so it is deliberately NOT in `secret-rules.v1.json`. An already-redacted `<redacted:...>` placeholder is stripped by the consumer, so Plimsoll won't re-flag it.
- `network_endpoints` now runs userinfo redaction, then the shape pass.

## Scope

This is the userinfo piece of the original Phase 3. Sensitive query/fragment params were already handled in Phase 2 (`sensitive-query-param` + the assignment rule match `token=`/`access_token=` anywhere). The ADR rollout is updated to mark Phases 1–3 done; broader structural URL handling is deferred unless a real need appears.

## Tests

95 runner-core unit (incl. userinfo redacted/host-preserved, bare-username + host:port left alone, idempotency) + 5 integration (incl. an end-to-end endpoint userinfo case through `redact_in_place`). clippy + fmt clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- URL userinfo redaction (ADR-034, Phase 3). A network endpoint that is a URL carrying a
+  `user:pass@` credential pair now has its userinfo redacted at capture (`scheme://user:pass@host` ->
+  `scheme://<redacted:url-userinfo:H8>@host`), preserving the scheme and host. It fires only when the
+  userinfo contains a `:` pair (a token-as-username is already caught by the shape pass), is
+  idempotent, and is a runner-side capture-hygiene transform rather than a shared detection rule.
+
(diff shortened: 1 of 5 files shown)


Rul1an:
gate: all

Delegated proof for head SHA 48c4d9646bd9: `Runner Spike Delegated` (gates=all, build_ebpf=true) — https://github.com/Rul1an/assay/actions/runs/27164782732

PR #1594, opened 2026-06-09
feat(sandbox): Landlock TCP-connect egress enforcement (Plan 2 PR5)

Rul1an:
## Summary

The capstone of the Landlock-net arc: `assay sandbox --enforce-net` (requires `--enforce`) now **enforces** a TCP-connect port allowlist via Landlock. This is the first PR in the arc allowed to make the enforcement claim.

- Combined FS+NET ruleset in ONE ruleset: `LANDLOCK_ACCESS_NET_CONNECT_TCP` handled as a **hard requirement** (no best-effort downgrade), with a `NetPort` allow rule per explicit TCP port in `net.allow`. All other TCP connects are denied.
- Applied via `restrict_self` in the enforcing child. The **child→parent ack** is std's `pre_exec` error channel: a failed `restrict_self` makes `spawn()` fail, so `restrict_self_confirmed` is never recorded on an unenforced child (no parent assumption).
- A network policy that is **not Landlock-expressible** (IP/CIDR, host, non-TCP protocol, range, deny rule, port 0) **fails closed before spawn** via the PR4 compiler + the net-string planner.
- `--enforcement-health <path>` writes `assay.enforcement_health.v1` (`active` when applied, `failed` with a machine-readable reason otherwise); a requested-but-unwritable artifact is a command failure (the #1588 rule).
- FS-only sandboxing is unchanged (net enforcement is opt-in).

## Claim boundary

Proven: an allowed TCP-connect port succeeds and its listener sees the connection; a denied port fails with **EACCES** and its listener is never reached. **Non-claims** (carried in the artifact): no IP/CIDR, no hostname, no destination identity, no UDP/QUIC, no HTTP/TLS route policy, not a replacement for the cgroup/connect4 endpoint path. Weak signals (timeout, ECONNREFUSED, ENETUNREACH) never count as a block.

## Tests

`cargo test -p assay-cli --bin assay -- landlock_net enforcement_health_v1 sandbox` — 33 passed (net-string planner accept/reject matrix, v1 producer round-trips, sandbox degradation).

## Real-block proof

The enforcement claim requires a real-block proof on an ABI 4 host. **Proof pending**: I will attach the VM run (allowed-port success + denied-port EACCES + listener-not-reached + the `active` enforcement_health.v1 artifact) before merge. Do not merge until that proof is attached.

## Lane

`Gate.NONE` (backend/sandbox/policy paths, no runner-spike/eBPF/monitor/cgroup/Cargo). Host-capability check: not triggered (no `diagnostics/` paths).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- Landlock TCP-connect egress enforcement for `assay sandbox` (`--enforce-net`, requires
+  `--enforce`): builds a combined FS+NET Landlock ruleset that allows only the explicit TCP
+  ports in `net.allow` and denies all other TCP connects, applied via `restrict_self` in the
+  enforcing child. A non-expressible network policy fails closed before spawn. With
+  `--enforcement-health <path>` it writes the `assay.enforcement_health.v1` artifact (`active`
+  when the ruleset is applied, `failed` with a machine-readable reason otherwise); a requested
+  artifact that cannot be written is a command failure. FS-only sandboxing is unchanged.
 - `assay doctor --format json` now carries a top-level `schema` id (`assay.doctor_report.v0`), making
   the report self-describing so a future field-shape change is an explicit version bump rather than
   silent drift. Additive; existing fields unchanged.
--- a/crates/assay-cli/src/backend.rs
+++ b/crates/assay-cli/src/backend.rs
@@ -92,14 +92,16 @@ impl LandlockEnforcer {
     }
 }
 
-/// Prepare Landlock ruleset. allocations/IO allowed here.
+/// Prepare Landlock ruleset. allocations/IO allowed here. `net_allow_ports` of `Some` builds a
+/// combined FS+NET (TCP-connect allowlist) ruleset; `None` is FS-only (unchanged behaviour).
 pub fn prepare_landlock(
     policy: &crate::policy::Policy,
     scoped_tmp: &std::path::Path,
+    net_allow_ports: Option<&[u16]>,
 ) -> anyhow::Result<LandlockEnforcer> {
     #[cfg(target_os = "linux")]
     {
-        let ruleset = landlock_impl::create_ruleset(policy, scoped_tmp)?;
+        let ruleset = landlock_impl::create_ruleset(policy, scoped_tmp, net_allow_ports)?;
         Ok(LandlockEnforcer {
             ruleset: Some(ruleset),
         })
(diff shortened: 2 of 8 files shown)


Rul1an:
## Real-block proof — Landlock TCP-connect egress (self-hosted ABI 4 host)

Built from this PR's source on `assay-bpf-runner` and run there. The four enforcement source files on the host are byte-identical to the PR head, so the binary is faithful to the reviewed code.

**Provenance**
- PR head: `45aac497f944515baf29827b2e3ad67c0ed3c3e4` (VM `git rev-parse HEAD` matches)
- Source faithfulness (sha256, VM == PR head): `backend.rs`, `cli/commands/sandbox/child.rs`, `landlock_net.rs`, `enforcement_health_v1.rs` all MATCH
- Binary built by `github-runner`, sha256 `7117041a9721b38c917234fd86cc491533827408b2e49b053d15fad8917b7cc7`
- `uname -a`: `Linux assay-bpf-runner 6.8.0-124-generic #124-Ubuntu SMP PREEMPT_DYNAMIC aarch64`

**Setup**
- Two loopback listeners: an allowed port and a denied port.
- Policy `net.allow: ["<allowed_port>"]` only.
- `assay sandbox --enforce --enforce-net --enforcement-health <path> --policy <p> -- python3 target.py <allowed> <denied>`

**Observed (deterministic across 2 runs)**
```
allowed_port connect -> ok,   listener reached     = true
denied_port  connect -> EACCES (errno 13), listener reached = false
sandbox exit = 0
```
`enforcement_health.v1` written by the sandbox:
```json
{
  "schema": "assay.enforcement_health.v1",
  "status": "active",
  "mechanism": "landlock",
  "scope": "tcp_connect_landlock_port",
  "policy_semantics": "allowlist",
  "landlock": {
    "abi": 4,
    "handled_access_net": ["connect_tcp"],
    "allowed_connect_tcp_ports": [46501],
    "no_new_privs_confirmed": true,
    "restrict_self_confirmed": true
  },
  "probe": null,
  "non_claims": ["no ip or cidr enforcement", "no hostname enforcement",
                 "no destination identity enforcement", "no udp or quic enforcement",
                 "no http or tls route policy",
                 "not a replacement for cgroup/connect4 endpoint enforcement"]
}
```

**`block_proven = true`**: the allowed port reached its listener; the denied port was blocked with EACCES and its listener was never reached. A weak signal (timeout, ECONNREFUSED, ENETUNREACH) would have failed the proof; only EACCES + listener-not-reached counts.

Bounded claim: assay now has proven Landlock TCP-connect port-allowlist egress enforcement on an ABI 4 host, with the combined FS+NET ruleset applied via `restrict_self` in the enforcing child, fail-closed on a non-expressible policy, and the `enforcement_health.v1` truth artifact emitted. Non-claims as above; this is the second enforcement route beside connect4/eBPF, not a replacement for it.

PR #1596, opened 2026-06-10
feat(sandbox): --probe-enforcement self-probe writes a real-block probe (Plan 2 followup PR C)

Rul1an:
Enriches the runtime artifact from `active` (probe: null) to `active + probe` when a real block is actually observed, using the same verdict as the reusable experiments harness (PR B).

**`assay sandbox --enforce-net --probe-enforcement`** runs a self-probe before the workload:
- From inside the enforcing Landlock-net ruleset (no_new_privs + restrict_self applied in a throwaway child), it attempts one connect to an **ephemeral denied port** (bound by the harness, never in the allowlist).
- **Only a proven real block** (EACCES AND the harness listener never reached) writes the `probe` block into `enforcement_health.v1`.
- A probe that does **not** prove a block (connect succeeded, weak errno, listener reached, or the probe could not run) is **reported to stderr, never silently dropped, and never fails the run** — the enforcement ruleset on the workload child was still confirmed.
- Weak signals (timeout, ECONNREFUSED, ENETUNREACH) never count.

**Semantics:** `active` without probe = restrict_self confirmed; `active` with probe = restrict_self confirmed **+ one denied connect proven blocked**.

**Safety:** opt-in only; reuses the PR2 hybrid pattern (parent builds the ruleset, the child runs only async-signal-safe syscalls: prctl, restrict_self, socket, connect, _exit). The denied port is ephemeral and not in the policy, so the probe cannot widen policy or break the workload.

**Tests:** unit tests for the exit-code verdict (EACCES=blocked, weak errno=not-blocked, sentinels=infra-error, connect=block-failed). Live verification on the ABI 4 host attached before merge.

**Lane:** Gate.NONE (backend/sandbox paths). Real-block proof pending (VM).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- `assay sandbox --probe-enforcement` (with `--enforce-net`): runs a self-probe before the
+  workload that, from inside the enforcing ruleset, attempts one connect to an ephemeral denied
+  port. Only a proven real block (EACCES and the harness listener never reached) writes the
+  `probe` block into `enforcement_health.v1` (`active` + probe). A probe that does not prove a
+  block is reported and never silently dropped, and never fails the run. Weak signals (timeout,
+  ECONNREFUSED, ENETUNREACH) never count as a block.
+
 ## [3.21.0] - 2026-06-10
 
 ### Added
--- a/crates/assay-cli/src/backend.rs
+++ b/crates/assay-cli/src/backend.rs
@@ -330,6 +330,130 @@ mod landlock_impl {
 
         expanded
     }
+
+    /// Build a NET-only Landlock ruleset that handles `CONNECT_TCP` (hard requirement) and allows the
+    /// given TCP-connect ports. Used by the self-probe: connecting to a port NOT in this list under
+    /// the applied ruleset must be denied. FS is intentionally not handled here (the probe only tests
+    /// the network rule; Landlock FS does not affect sockets).
+    pub(super) fn build_net_ruleset(allowed_ports: &[u16]) -> anyhow::Result<RulesetCreated> {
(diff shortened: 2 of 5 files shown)


Rul1an:
## Self-probe verified live (self-hosted ABI 4 host)

Built from this PR's source on `assay-bpf-runner`; the three touched source files are byte-identical to the PR head.

**Provenance**
- PR head: `0fa16c170e52f248a107ec1fc544e36ae5cebe5d` (VM `git rev-parse HEAD` matches)
- Source faithfulness (sha256, VM == PR head): `backend.rs`, `cli/commands/sandbox/child.rs`, `cli/args/runtime.rs` all MATCH
- Binary sha256 `95631cfecd85e6c5b5dca6722205264b8daf87a386aa1af41aa48b4063075ba4`
- `uname`: kernel 6.8.0-124 aarch64, Landlock ABI 4

**Run**
```
assay sandbox --enforce --enforce-net --probe-enforcement \
  --enforcement-health eh.json --policy <net.allow:[443]> --workdir WD -- true
```

**Observed (deterministic across 2 runs), `enforcement_health.v1`:**
```json
{
  "status": "active",
  "probe": {
    "kind": "real_block",
    "transport": "ipv4",
    "blocked_action": "tcp_connect",
    "blocked_port": 42877,
    "blocked_errno": "EACCES",
    "listener_reached": false
  }
}
```

The self-probe bound an ephemeral denied port (not in the allowlist), connected to it from inside the enforcing ruleset, got **EACCES** with the listener **never reached**, and wrote the `real_block` probe. Exit 0, no probe warnings.

This is the `active + probe` upgrade: `active` without a probe means restrict_self was confirmed; `active` with this probe means restrict_self was confirmed **and** one denied connect was proven blocked, by the same verdict the reusable experiments harness uses (EACCES + listener-not-reached only; weak signals rejected). Non-proving probes are reported and never silently dropped; opt-in only.

Issue #1575, opened 2026-06-09
monitor: stale eBPF object emits old MonitorEvent size (520 vs 552)


Rul1an:
During `assay monitor` runs on the ARM bpf-lsm host, the userspace event decoder repeatedly logs `Monitor stream error: invalid event size (got=520, need=552)`. Attach and capture still work (the smoke and egress proofs pass), but this is a kernel-struct vs userspace-decoder size drift that can corrupt observability claims later. Reconcile the `MonitorEvent` layout between the eBPF emit side and the userspace decode side (the 32-byte delta suggests a field/padding mismatch). Observed in #1569 / #1570 runs.

Rul1an:
Dug into this from the source, and it is an artifact skew rather than a struct divergence, which changes what the fix is. MonitorEvent is a single shared type in assay-common, imported by both the eBPF emit side (assay-ebpf) and the userspace decoder (assay-monitor), and it is pinned at compile time to 552 bytes, 40 bytes of header plus a 512-byte data field. So for a build from one commit the two sides cannot disagree.

The 32-byte delta is the tell. 520 is the older layout with a 480-byte data field, 552 is the current 512-byte one, so the eBPF object loaded on the bpf-lsm host was built before the data field grew, against a newer userspace binary. It is a stale-object skew, not a kernel-vs-userspace struct mismatch in the code.

Two things follow. The decoder already fails closed: parse_event rejects any record shorter than the expected size rather than copying it into the struct, so a skewed object drops those events and logs the mismatch, it does not decode garbage. So the concern about corrupting observability claims does not actually happen on this path, the cost is dropped events and log noise on a mismatched build. And the remedy is the canonical eBPF build, rebuilding the object from the same commit as the userspace binary through install-ebpf-toolchain.sh plus cargo xtask build-ebpf --release --no-docker, which is the path the monitor-attach-smoke gate already guards.

So I would keep this open only until a fresh canonical build on the bpf-lsm host shows got equal to need, then close it. If we want the skew to be louder for the next person, a small follow-up could fold the repeated per-event size log into a single warning that names the likely stale-object cause, but that is hardening rather than the fix.

Rul1an:
[...]

PR #1601, opened 2026-06-10
feat(monitor): event-size mismatch counter, one-shot warning, observation_health degradation (#1575)

Rul1an:
Hardening for #1575 (stale eBPF object emits old MonitorEvent size). Diagnosis confirmed: MonitorEvent is shared via assay-common and size-pinned at 552; a 520-byte record is the old 40+480 layout from a stale object, which the decoder already rejects fail-closed. This makes that failure visible and honest instead of log-noise:

- **assay-monitor**: count size mismatches in `MonitorStatsSnapshot.event_size_mismatch` (userspace counter, incremented in the `listen()` consume loop), and emit **one** actionable stale-object warning per run (with the canonical rebuild command) instead of a per-event size error.
- **assay-runner-core**: the kernel layer treats a size mismatch as lost events, so a run that silently dropped records degrades to `PartialRingbufDrops` and is never read as a complete/clean observation. Counts stay **separate** (ringbuf drops vs size mismatch); only the 0-vs-nonzero kernel-layer decision combines them.
- **Tests**: `parse_event` rejects a 520-byte record reporting got=520/need=552; `kernel_layer_for` degrades when `event_size_mismatch > 0` even with zero ringbuf drops.

Deliberately **not** done: no decoder tolerance for the old 520 layout (no implicit back-compat on a version-less binary ABI), no `MonitorEvent` struct change (shared + pinned, so not the bug), no new observation_health schema field (degradation rides the existing `PartialRingbufDrops` status). Load-time object/userspace ABI provenance is left as a separate follow-up.

Lane: this touches assay-monitor + assay-runner-core, so `gates=all` (delegated runner-spike proof). The proof also validates the Linux-only loader changes. Delegated proof to follow on the head SHA.

### DIFF
--- a/crates/assay-monitor/src/events.rs
+++ b/crates/assay-monitor/src/events.rs
@@ -57,3 +57,34 @@ pub fn send_parsed(
     // best-effort send
     let _ = tx.blocking_send(ev);
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse_event_rejects_short_record_with_got_and_need() {
+        // A stale eBPF object emits the previous 40 + 480 = 520 byte layout while userspace now
+        // expects 40 + 512 = 552. The parser must reject the short record (fail-closed), reporting
+        // the observed and expected sizes, rather than decoding a truncated struct.
+        let need = core::mem::size_of::<MonitorEvent>();
+        assert_eq!(
+            need, 552,
+            "MonitorEvent ABI size changed; update the stale-object test"
+        );
+        let stale = vec![0u8; 520];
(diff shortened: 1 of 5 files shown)


Rul1an:
Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/27264602105
- sha: 891094a2993ffb8548ea42aafebb4651576df5bf

Phase 1 delegated gates (all) passed on the self-hosted bpf-lsm host with a canonical eBPF build, validating the Linux-only `listen()` consumer changes (the kernel-capture/decode path this PR touches).

Issue #1624, opened 2026-06-10
P61e: enforcing tools/call proxy (new risk class — review-spec first)


Rul1an:
## Context

assay v3.23.0 shipped the opt-in **manifest-observation** proxy mode (P61a–d): it observes the upstream `tools/list` and emits `assay.mcp_manifest_observed.v0` with honest completeness, and **never forwards `tools/call`**. Spec: `docs/reference/mcp-upstream-proxy-mode.md`.

## Why this is a separate arc, not a small PR

An *enforcing* `tools/call` proxy is a materially heavier security boundary than manifest observation. It must be designed (review-spec first, no code) before any implementation, because it opens:

- caller authorization (who may invoke which privileged tool through the proxy);
- upstream credential use on behalf of a caller — the classic confused-deputy risk;
- a policy decision point **before** forwarding, fail-closed;
- confused-deputy prevention (no upgrade/broadening of caller authority; no silent tool-surface mixing);
- `proxy_denied` semantics (reserved but unused today);
- interaction with the side-effect evidence ladder (a forwarded call's side effect stays asserted unless independently verified).

## Scope (when started)

- [ ] a review-spec (design doc) covering the above, agreed before code;
- [ ] then, only if greenlit: an enforcing forward path that blocks a privileged `tools/call` unless the policy decision point allows it, fail-closed.

## Non-goals / boundaries

- not started until the review-spec is agreed;
- never observe-only forwarding of a privileged `tools/call` (that is the confused-deputy trap the manifest-observation mode deliberately avoids);
- no maliciousness classification.

Successor to the now-delivered #1615.

Rul1an:
**P61e-a (review-spec) done** — landed and refined in #1629 + #1630 as `docs/reference/mcp-upstream-proxy-enforcement.md` (the design of record). All binding decisions are fixed: `proxy enforce` is a distinct subcommand (new risk class, not a flag); the PDP runs on every `tools/call`; the drift baseline is a declared approved manifest artifact only plus a required current complete manifest; hard-deny by default; one caller per stdio session via config; `enforcement_decision` deferred to P61e-d and name-disambiguated from `enforcement_health`.

Remaining slices (code, awaiting explicit go per slice):
[...]

PR #1632, opened 2026-06-11
feat(mcp): enforcing-proxy caller-allowance PDP — deny-only (P61e-c1)

Rul1an:
## P61e-c1: enforcing-proxy caller-allowance PDP (deny-only)

Replaces the P61e-b single deny-all boundary with the first real gate of the enforcing proxy's policy decision point. Every `tools/call` is now classified and denied with the precedence-pinned reason of the first gate that fails — but there is still **no allow or forward path**. A call that clears every c1 gate is denied with `pdp_gate_unavailable`, a temporary rollout reason that goes away when c3 lands.

This is the first of the gate-by-gate split of the old P61e-c (c1 caller-allowance → c2 credential-scope → c3 drift + first allow). No `tools/call` is forwarded before c3, by design.

### What changed
- `proxy-enforce` now requires `--enforce-policy <path>` (YAML: `caller.id` + `allowances`). A missing, unreadable, or malformed policy — or one without `caller.id` — **fails startup with a non-zero exit, never a runtime deny.** An enforcing proxy that cannot decide is a misconfigured service and must not start.
- Caller identity is the static `caller.id` only (no transport/env inference), so `unknown_caller` cannot occur at runtime under the static-config model.
- PDP precedence is classification → allowance → `pdp_gate_unavailable`. The classification gate runs **before** allowance, so a missing-target call reads as `classification_incomplete`, never as a target mismatch.
- c1 knows the `github_deploy_key {owner, repo}` target shape only (projected plain by the P57c classifier — owner/repo are sanitized for control chars, never hashed, so a plain compare is correct). Other action classes are fail-closed (`no_declared_allowance`) until their own gate slice. No wildcard in c1.
- The proxy emits a per-decision `tracing` line (caller, action class, a domain-separated `sha256("assay.mcp.target.v0\0" + canonical_target_json)` digest, reason). This is a **diagnostic log for operability only — not** the canonical `assay.enforcement_decision.v0` evidence artifact, which is P61e-d.
- `enforcing_mode_deny_all` is retired/superseded by the c1 PDP reasons. The P61e-b deny-all e2e tests are migrated to supply a minimal policy and assert the new reasons.

### Tests
- 8 PDP unit tests (load/validate, gate precedence, domain-separated target digest).
- 9 c1 e2e tests through the spawned binary: the deny matrix (`unclassified_tool_call`, `classification_incomplete`, `no_declared_allowance`, target mismatch, `pdp_gate_unavailable`) plus four startup-failure cases (missing flag, missing file, malformed YAML, missing `caller.id`).
- The migrated deny-all e2e suite still asserts the load-bearing invariant: no `tools/call` reaches the upstream in enforce mode, and observe mode is unchanged.

### Non-claims
No `tools/call` is forwarded. No allow path, no credential-scope gate (c2), no drift gate (c3), no enforcement-decision evidence record (P61e-d). The diagnostic log is not evidence. Spec: `docs/reference/mcp-upstream-proxy-enforcement.md` (§5 policy shape, §7 reason set, §14 c1/c2/c3 split + implementation note).

Part of #1624.

### DIFF
--- a/crates/assay-mcp-server/src/main.rs
+++ b/crates/assay-mcp-server/src/main.rs
@@ -41,11 +41,13 @@ enum Mode {
         #[arg(long)]
         proxy_observation_health_out: Option<PathBuf>,
     },
-    /// Run as an MCP upstream ENFORCING proxy (P61e-b, deny-all): an explicit, separate run mode — a
-    /// different risk class from `proxy`, never a variant of it. Every `tools/call` is denied with
-    /// `proxy_denied` (`enforcing_mode_deny_all`) and never forwarded upstream; the handshake, `ping`,
-    /// and `tools/list` still forward; other methods stay `proxy_unsupported`. There is no allow path,
-    /// no policy decision point, no credential or drift gate in v0 (those are P61e-c).
+    /// Run as an MCP upstream ENFORCING proxy (P61e-c1): an explicit, separate run mode — a different
+    /// risk class from `proxy`, never a variant of it. Every `tools/call` runs through the c1
+    /// caller-allowance policy decision point and is denied with the precedence-pinned reason of the
+    /// first gate that fails; a call that passes the c1 gates is still denied with
+    /// `pdp_gate_unavailable` (there is deliberately no allow / forward path before c3). The handshake,
+    /// `ping`, and `tools/list` still forward; other methods stay `proxy_unsupported`. A missing,
+    /// unreadable, or malformed `--enforce-policy` fails startup (non-zero exit) — never a runtime deny.
     ProxyEnforce {
         /// The upstream MCP server command to spawn (stdio transport).
         #[arg(long)]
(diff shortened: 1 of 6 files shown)


Rul1an:
Codex review for 289785c610140a878691882ac33363e887afd382

Findings:

- [P2] Update the resolved caller-identity decision to match c1 startup-fail semantics — `docs/reference/mcp-upstream-proxy-enforcement.md:335` still says a missing caller becomes `proxy_denied` with `unknown_caller`, but the shipped c1 contract and implementation now make missing `caller.id` a startup failure before the proxy starts. The same document already states this correctly in §5 and §14, and the e2e tests assert non-zero startup for missing `caller.id`, so leaving §16-D stale creates two conflicting authoritative contracts for operators and for c2/c3 implementers. Please change §16-D to say missing `caller.id` fails startup and that `unknown_caller` cannot occur at runtime under the static one-caller-per-stdio-session model.

Notes:

I did not find code-path issues in the c1 deny-only PDP. The startup load-before-spawn path, classification-before-allowance precedence, `github_deploy_key` owner/repo matching, diagnostic-only target digest, and no-forward invariant all look aligned with the PR contract.

Verification run locally:

```text
cargo test -p assay-mcp-server --test proxy_enforce_e2e --test proxy_enforce_pdp_e2e
# 5/5 proxy_enforce_e2e passed; 9/9 proxy_enforce_pdp_e2e passed
```

PR #1634, opened 2026-06-11
feat(mcp): enforcing-proxy credential-scope gate — deny-only (P61e-c2)

Rul1an:
Adds the **credential-scope gate** to the enforcing-proxy PDP — the second gate of the gate-by-gate split (c1 caller-allowance → **c2 credential-scope** → c3 drift + first allow). It runs after the c1 allowance match and before `pdp_gate_unavailable`. Still **deny-only: no `tools/call` is forwarded before c3.**

### What changed
- The policy gains an optional `upstream_credential: {alias, scopes}` block — the single credential the proxy holds for the session. The gate compares the action's `required_scope_for(category)` (P59) against the declared scopes through a deterministic lattice **shared with the E10 overbreadth experiment**, so the runtime gate and the measurement agree.
- Coverage → outcome: **covered** (exact / broader-non-admin / overbroad — overbroad still covers; it is a recommendation, never a block in v0) falls through; **recognized but non-covering** → `credential_scope_insufficient`; **absent credential / unrecognized scope / too-coarse (ambiguous) scope** → `credential_scope_unknown` (an unknown is not an insufficiency, spec §8).
- **Fail-closed:** an unknown required scope denies (`credential_scope_unknown`), never a silent pass — explicitly *not* `?` on `required_scope_for`, which would fail open. A defensive unit test guards this.
- `required_scope_for` is now `pub` in `tool_decision` (the gate input). The credential `alias` is reserved for the P61e-d evidence record and is not read by the gate.
- A call that clears the credential-scope gate is still denied with `pdp_gate_unavailable` (no allow path before c3).

### Precedence
classification → allowance → **credential-scope** → `pdp_gate_unavailable`.

### Tests
- `enforce.rs` unit tests → 17: covered (exact / broader / overbroad), insufficient, unknown (absent / unrecognized / ambiguous), ambiguous-takes-precedence-over-insufficient, fail-closed unknown-required-scope, and a direct `scope_covers` ↔ E10 lattice check.
- `proxy_enforce_pdp_e2e.rs` migrated (`ALLOW_ACME` gains a covering credential so the matching call still reaches `pdp_gate_unavailable`) + two new e2e gate cases (`credential_scope_insufficient`, `credential_scope_unknown`).
- Full `assay-mcp-server` suite green; clippy `-D warnings` clean.

### Non-claims
No `tools/call` is forwarded. Declared scopes are operator config, **not a provider-verified grant**; there is no token introspection. The diagnostic decision log is not the `assay.enforcement_decision.v0` evidence record (P61e-d). Spec: `docs/reference/mcp-upstream-proxy-enforcement.md` (§5 policy shape, §8 gate, §14 c2 + implementation note).

Part of #1624.

### DIFF
--- a/crates/assay-mcp-server/src/proxy/enforce.rs
+++ b/crates/assay-mcp-server/src/proxy/enforce.rs
@@ -1,10 +1,12 @@
-//! P61e-c1: the enforcing-proxy policy decision point — caller-allowance gate, deny-only.
-//! Spec: docs/reference/mcp-upstream-proxy-enforcement.md.
+//! P61e-c1/c2: the enforcing-proxy policy decision point — caller-allowance + credential-scope gates,
+//! deny-only. Spec: docs/reference/mcp-upstream-proxy-enforcement.md.
 //!
-//! c1 scope: parse the `--enforce-policy` file, classify an observed `tools/call`, and decide a DENY
-//! with a precedence-pinned reason. There is no allow path and no forwarding in c1 (that arrives in
-//! c3, after the credential-scope gate in c2). A `tools/call` that passes the c1 gates is still denied
-//! with `pdp_gate_unavailable`, a temporary rollout reason removed when c3 lands.
+//! Scope so far: parse the `--enforce-policy` file, classify an observed `tools/call`, and decide a
+//! DENY with a precedence-pinned reason. c1 added the caller-allowance gate; c2 adds the
+//! credential-scope gate (the declared upstream credential must cover the action's required scope).
+//! There is still no allow path and no forwarding (that arrives in c3, with the drift gate). A
+//! `tools/call` that passes every enabled gate is still denied with `pdp_gate_unavailable`, a temporary
+//! rollout reason removed when c3 lands.
 //!
 //! Caller identity is the static `caller.id` from the policy only — no transport/env/request
 //! inference — so a single stdio session is bound to one configured caller and `unknown_caller`
@@ -13,7 +15,7 @@
 use anyhow::{bail, Context, Result};
 use assay_core::mcp::jcs;
 use assay_mcp_server::cache::sha256_hex;
-use assay_mcp_server::tool_decision::classify;
+use assay_mcp_server::tool_decision::{classify, required_scope_for};
 use serde::Deserialize;
 use serde_json::Value;
 use std::path::Path;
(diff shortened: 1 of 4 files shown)


Rul1an:
**Findings**

- [P2] `crates/assay-mcp-server/src/proxy/enforce.rs` now treats `repo:write` as covering `repo:deploy_key:write`, but the reference contract still says only `repo:deploy_key:write` and `repo:admin` cover it in `docs/reference/credential-scope.md`. That is a security-contract mismatch: after c3, behavior could allow a scope the public/reference doc says should not cover. Either update the P59 doc to match the E10/c2 lattice or remove `repo:write` from the runtime lattice if the narrower contract is still intended.

- [P3] `docs/reference/mcp-upstream-proxy-enforcement.md` still labels `pdp_gate_unavailable` as “P61e-c1 only”, while c2 still deliberately falls through to that reason after the credential-scope gate passes. Small but confusing for operators/implementers. I’d relabel it as pre-c3 rollout-only or c1/c2-only.

No code-path issue found in the actual c2 enforcement flow: allowance-before-scope, fail-closed unknown/insufficient handling, and no-forward-before-c3 all look aligned.

**Verification**

- `cargo test -p assay-mcp-server --test proxy_enforce_pdp_e2e` passed.
- `cargo test -p assay-mcp-server proxy::enforce` passed.
- Last observed CI still had ubuntu/windows build-test, dependency security, and Wave 0 feature matrix in progress, so merge readiness was not fully settled yet.

Rul1an:
Both addressed, thanks — these were good catches.

On the [P2] scope mismatch: you're right, and it's the more important of the two. The credential-scope.md contract is the authority for the enforcement path, and it's explicit that `repo:deploy_key:write` is covered only by `repo:deploy_key:write` and `repo:admin`, with broadening meant to be deliberate and fixture-backed. The `repo:write → covers` entry came in from the E10 measurement vocabulary and shouldn't have ridden along into enforcement, where being broader than the documented contract is exactly the dangerous direction once c3 starts forwarding. So I narrowed `lattice_for(github_deploy_key)` to match the doc exactly — covered `{repo:deploy_key:write, repo:admin}`, recognized-non-covering `{repo:read, repo:metadata, repo:contents:read}`, and everything else (including `repo:write`) unrecognized → `credential_scope_unknown`. I also dropped the undocumented slack/workspace lattices; they were unreachable anyway since c1's allowance matcher only admits `github_deploy_key`, and they now fail closed until their own contract slice lands. Tests flipped accordingly (`repo:write` now asserts unknown, not covered).

On [P3]: relabelled `pdp_gate_unavailable` as "pre-c3 rollout only; c1/c2" — you're right that c2 still falls through to it.

One honest note for the trail: the E10 experiment still carries the broader lattice, so its measurement and this gate now deliberately disagree (the gate follows the shipped contract, the experiment was exploring a wider vocabulary). I've flagged reconciling E10 to the contract as a separate follow-up rather than fold an experiment-result change into this security PR.

PR #1637, opened 2026-06-11
docs: add assay CI contract

Rul1an:
## Summary

Adds a review-first `CI-CONTRACT.md` for `Rul1an/assay` before changing workflows.

The contract captures the current CI posture and sets rules for future hardening around:

- required-check safety and path-filter behavior
- release/provenance and high-trust permissions
- runner/kernel/self-hosted boundaries
- public artifact sanitization
- claims and evidence-boundary wording
- scheduled OSV/Scorecard/CodeQL/ClusterFuzzLite candidates
- MCP proxy enforcement boundaries

## Scope

Docs only. No workflow behavior changes in this PR.

## Validation

- Local contract sanity check for required sections and forbidden public strategy terms
- pre-commit during commit and push:
  - merge conflict check
  - YAML/TOML skips where irrelevant
  - EOF/trailing whitespace
  - MCP registry local token hygiene
  - typos
  - workflow/actionlint and shellcheck skipped because no workflow files changed
  - cargo fmt
  - cargo deny/audit skipped where path filters did not apply
  - cargo clippy and linux compile gate passed on push hook

### DIFF
--- a/CI-CONTRACT.md
+++ b/CI-CONTRACT.md
@@ -1,0 +1,400 @@
+# CI Contract: Rul1an/assay
+
+Draft status: review contract before workflow implementation.
+
+`Rul1an/assay` is the public source, release, evidence, runner, and MCP proxy
+repository for Assay. Its CI contract must protect three things at once:
+
+- ordinary Rust, Python, docs, and action quality;
+- release and provenance integrity for crates, PyPI, binaries, and generated
+  artifacts;
+- evidence-honesty boundaries around runtime observation, MCP proxy enforcement,
+  receipts, Trust Basis output, and public docs.
+
+This contract is a diff from today's repository state. The first rule is no
+required-coverage regressions: existing useful gates should stay required or
+visible unless this file is updated with a clear replacement.
+
+## 0. As-Is Inventory
+
+Repository state observed on 2026-06-11:
+
+- Visibility: public.
+- Default branch: `main`.
+- Primary languages: Rust, Python, TypeScript/JavaScript, shell.
+- Rust workspace root: `Cargo.toml`, `Cargo.lock`, `rust-toolchain.toml`.
+- Core CI workflows:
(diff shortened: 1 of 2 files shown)


Rul1an:
This is the most complete of the portfolio contracts, and it shows — it carries forward every refinement from the smaller repos rather than restating them: the whole-repo fail-closed sanitization scope, the required-versus-fork split so a fork run degrades honestly instead of silently, the high-trust SHA-pin contract for the release/provenance/SBOM/security-events/id-token/attestations surface, and the no-path-filter-into-pending rule. The MCP proxy section also reads true to what actually shipped — deny-only modes stay fail-closed, forwarding-mode work carries the caller/credential/manifest/drift/confused-deputy gates, and credential scope is operator configuration rather than a provider-verified grant. Good.

One finding worth acting on, and it's the contract's own rule turned on itself. Section 6 says the minimal required context is `CI`, grounded in `docs/BRANCH-PROTECTION-SETUP.md`. The live branch protection on `main` actually requires three contexts: `CI`, `lane-check`, and `host-capability-check`. So the setup doc and the live setting have drifted, and the contract inherited the doc's view. Since the contract rightly insists on reading required contexts from the live checks API before changing anything, I'd apply that here too: record all three as currently required, move `lane-check` and `host-capability-check` out of the "visible, reviewed when relevant" group into the required set, and reconcile `BRANCH-PROTECTION-SETUP.md` so the doc stops understating what's enforced. Worth getting right because this is the section future branch-protection edits will trust.

One smaller thing to clarify in the implementation phase: there are now two adjacent guards — the existing public/private boundary vocabulary guard inside `CI`, and the new public-artifact sanitization guard. They're solving different problems, but a line on how their scopes relate would stop them from either overlapping confusingly or leaving a seam between them.

Everything else holds. Mergeable as-is; these are review-and-implementation notes, not blockers.

PR #1638, opened 2026-06-11
ci: add public artifact sanitization gate

Rul1an:
## Summary

Implements the first `CI-CONTRACT.md` slice for `Rul1an/assay`: a public-safe public-artifact sanitization gate.

This adds:

- `.github/workflows/sanitize-public-artifacts.yml`
- `scripts/ci/check-public-artifact-sanitization.py`
- a script self-test before the repo scan
- whole-repo tracked-file scanning, excluding generated/vendor/build/cache-style output
- CI output that reports only categories and locations, never matched text

## Boundary

This is intentionally the fork-safe structural layer only. It does not embed or print the private sensitive-vocabulary list. The trusted hashed-list comparison from the CI contract can layer on later for same-repo/scheduled contexts.

The script currently catches explicit public-blocker/review markers and placeholder-style secret markers. It avoids natural-language non-claim phrases such as "do not publish" because those appear intentionally in existing boundary docs.

## Validation

- `python3 scripts/ci/check-public-artifact-sanitization.py --self-test`
- `python3 scripts/ci/check-public-artifact-sanitization.py`
- `python3 -m py_compile scripts/ci/check-public-artifact-sanitization.py`
- `actionlint -config-file .github/actionlint.yaml .github/workflows/sanitize-public-artifacts.yml`
- pre-commit/pre-push hooks during commit and push

### DIFF
--- a/.github/workflows/sanitize-public-artifacts.yml
+++ b/.github/workflows/sanitize-public-artifacts.yml
@@ -1,0 +1,29 @@
+name: Public Artifact Sanitization
+
+on:
+  pull_request:
+  push:
+    branches: [ "main" ]
+  schedule:
+    - cron: "37 4 * * 1"
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  public-artifact-sanitization:
+    name: public-artifact-sanitization
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+        with:
+          persist-credentials: false
+
+      - name: Self-test sanitizer
+        run: python3 scripts/ci/check-public-artifact-sanitization.py --self-test
+
+      - name: Scan tracked public artifacts
+        run: python3 scripts/ci/check-public-artifact-sanitization.py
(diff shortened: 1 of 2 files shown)


Rul1an:
Reviewed — clean, fork-safe, and mergeable. `permissions: {}` with a job-scoped `contents: read`, SHA-pinned checkout, a timeout, the self-test before the scan, and findings that report only category and location, never the matched text. Good.

Two notes, neither a blocker.

The honest framing one first, because it matters more than the code: this ships the public-safe structural gate — it catches files explicitly marked unpublishable or unredacted, which is genuinely useful — but it is not yet the private-vocabulary comparison the contract describes as the trusted/scheduled layer, and the docstring rightly says so. Worth keeping front of mind so a green check here is read as a structural backstop, not as "this is vocabulary-clean." The human sanitization pass stays the primary defense until that trusted layer lands.

The small code one: this is the only sanitizer in the batch that writes the marker as a contiguous literal in its self-test and leans on an incidental newline to avoid flagging itself. The sibling guards (Harness, both plimsoll repos, the experiments boundary guard) build the marker from split pieces with `"".join((...))`, which can't self-match by construction. I'd adopt that here too, or add a tiny assertion that the sanitizer doesn't flag its own tree — cheap insurance against a future edit quietly reintroducing the self-match you already hit once.

PR #1645, opened 2026-06-11
ci: add trusted sanitizer hash layer

Rul1an:
## Summary
- add an optional trusted hashed-list scan to the public artifact sanitizer
- keep the structural sanitizer as the fork-safe required path
- document that the trusted layer stays inside the sanitizer workflow, not a separate required context

## Safety
- private vocabulary/hash source is not stored in the repo
- logs report only category and file:line locations
- missing secret degrades explicitly with trusted-private-list=skipped

## Validation
- python3 scripts/ci/check-public-artifact-sanitization.py --self-test
- python3 scripts/ci/check-public-artifact-sanitization.py
- python3 scripts/ci/check-public-artifact-sanitization.py --trusted-hash-env PUBLIC_ARTIFACT_DENYLIST_SHA256
- ruby YAML parse for .github/workflows/sanitize-public-artifacts.yml

### DIFF
--- a/.github/workflows/sanitize-public-artifacts.yml
+++ b/.github/workflows/sanitize-public-artifacts.yml
@@ -27,3 +27,10 @@ jobs:
 
       - name: Scan tracked public artifacts
         run: python3 scripts/ci/check-public-artifact-sanitization.py
+
+      - name: Trusted private-list hash scan
+        env:
+          PUBLIC_ARTIFACT_DENYLIST_SHA256: ${{ secrets.PUBLIC_ARTIFACT_DENYLIST_SHA256 }}
+        run: >
+          python3 scripts/ci/check-public-artifact-sanitization.py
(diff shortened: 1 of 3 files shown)


Rul1an:
This layers the trusted hashed-list scan onto the existing public-artifact sanitizer. The structural scan stays the required fork-safe gate, and the trusted step only does anything when a trusted workflow can supply the hashed list. A fork or a no-secret run prints `trusted-private-list=skipped` with a reason rather than passing silently or revealing anything.

The discipline holds where it matters: the list lives only in the workflow secret and never in the repo, findings carry category and location and never the matched text or a hash value, and a malformed hash input fails closed with a count rather than contents.

Verification: self-test green including the new trusted-hash path, the normal sanitizer scan green, the no-secret skip path green, and the workflow YAML parses. GitHub checks are green apart from the long build matrix, which is still running with no failures so far. Keeping it draft until that finishes.

PR #1648, opened 2026-06-11
ci: add scheduled supply-chain posture

Rul1an:
## Summary
- add scheduled/manual OpenSSF Scorecard advisory reporting
- add scheduled/manual OSV advisory reporting for non-Rust resolved dependency inputs
- document the advisory-only boundary, Scorecard token limitation, and RustSec ownership in CI-CONTRACT.md

## Boundaries
- no pull_request or push trigger
- no ruleset or required-context mutation
- Rust advisory policy remains owned by cargo-deny/cargo-audit, not this OSV slice

## Test Plan
- ruby YAML parse for .github/workflows/*.yml
- actionlint on the new workflows
- zizmor --min-confidence high on the new workflows
- git diff --check
- local pre-commit hooks during commit/push

### DIFF
--- a/.github/osv/non-rust-manifests.txt
+++ b/.github/osv/non-rust-manifests.txt
@@ -1,0 +1,19 @@
+# Scheduled OSV advisory targets.
+#
+# Keep this list limited to tracked non-Rust dependency inputs. Rust advisory
+# policy is owned by cargo-deny/cargo-audit in PR CI.
+docs/requirements-ci.txt
+examples/agentevals-trajectory-strict-match-evidence/requirements.txt
+examples/autoevals-exactmatch-evidence/requirements.txt
+examples/crewai-event-evidence/requirements.txt
+examples/guardrails-validation-outcome-evidence/requirements.txt
+examples/langgraph-task-evidence/requirements.txt
+examples/langwatch-custom-span-evaluation-evidence/requirements.txt
+examples/mcp-compliance/package-lock.json
+examples/openai-agents-trace-evidence/requirements.txt
+examples/openfeature-evaluation-details-evidence/requirements.txt
+examples/phoenix-span-annotation-evidence/requirements.txt
+examples/promptfoo-assertion-grading-result-evidence/package-lock.json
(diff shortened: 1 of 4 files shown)


Rul1an:
Nice tightly scoped posture addition. Workflows are SHA-pinned, permissions are scoped down from the empty top-level default, the SARIF categories are distinct, and the CI-CONTRACT note that these stay advisory and non-required reads exactly right. The osv-scanner + osv-reporter `--new` pairing matches the upstream scheduled-scan template, so that flow looks correct.

One concrete adjustment for the Scorecard job:

- `openssf-scorecard.yml` has no `actions/checkout` step before `ossf/scorecard-action`. The upstream starter workflow checks out first (with `persist-credentials: false`), and several local-analysis checks (Pinned-Dependencies, Dangerous-Workflow, Token-Permissions) read the working tree, so without a checkout those degrade to incomplete rather than failing loudly. Adding the same pinned checkout step the OSV job already uses would keep the score accurate.

Not blocking for CI going green, just an accuracy gap in the advisory output.

Rul1an:
Updated this slice to close the silent-undercoverage edge on the scheduled OSV advisory job.

What changed:
- Scorecard now checks out the repo before running, so local-analysis checks have a working tree.
- OSV targets moved into a tracked non-Rust target list.
- The workflow now fails if a listed target is removed or untracked.
- It also fails if a new tracked `package-lock.json` or `requirements*.txt` appears outside the target list.
- The scanner step no longer treats every nonzero as advisory-only: vulnerability findings are tolerated and converted to SARIF, but malformed/missing JSON or a nonzero scanner exit with no vulnerability results is treated as an execution failure.

Rust remains intentionally out of scheduled OSV. `Cargo.lock` stays owned by cargo-deny/cargo-audit so this does not relitigate documented RustSec policy. Local validation: YAML parse, actionlint, zizmor, diff check, target-list coverage guard, and JSON result-count smoke all pass. Fast GitHub checks are green; the long existing assay matrix is still running with no failure observed at the time of this comment.

Issue #1651, opened 2026-06-11
Migrate PyO3 0.24 → 0.29 to resolve RUSTSEC-2026-0176 (pythonize compatibility)


Rul1an:
RUSTSEC-2026-0176 (published 2026-06-11) is a PyO3 memory-safety advisory: out-of-bounds reads in `BoundListIterator::nth/nth_back` and `BoundTupleIterator::nth/nth_back` when called with large `n`. Affected 0.24.0–0.28.x, **patched in 0.29.0**. assay currently pins `pyo3 0.24.2` (+ `pythonize 0.24.0`) in `assay-python-sdk`.

## Reachability (why it's a documented ignore, not a fire)
The vulnerable methods are never called in this repo. `assay-python-sdk` iterates with `.iter().enumerate()`, not `.nth()`; every `.nth()` elsewhere is on a Rust std iterator (string splits, env args), which the advisory does not touch. So the advisory does **not affect assay's usage today**. A documented ignore is in place so it does not block unrelated PRs (mirrors the RUSTSEC-2023-0071 pattern).

## The real fix (this issue)
Upgrade `pyo3` to `>= 0.29.0`. This is a multi-version PyO3 migration (0.24 → 0.29 has breaking API changes in the bindings) gated on `pythonize` supporting pyo3 0.29 — an earlier narrow bump stalled because the available `pythonize` still pinned an older pyo3. Scope:
- bump `pyo3` to 0.29 and `pythonize` to a 0.29-compatible release;
- update the `assay-python-sdk` bindings for the 0.24→0.29 API changes;
- if pythonize lags 0.29, evaluate alternatives or a temporary fork-pin.

## Close-out
When this lands, **remove RUSTSEC-2026-0176 from `deny.toml` and the `cargo audit --ignore` line** (the ignore comment points back here). The ignore is temporary and tied to this issue.

Rul1an:
Update: RUSTSEC-2026-0177 (pyo3 0.24.2, `PyCFunction::new_closure` missing `Sync` bound, published 2026-06-11) is now also a temporary ignore tied to this migration (cargo audit `--ignore` + `deny.toml`, landed in #1655), same posture as RUSTSEC-2026-0176: not reachable in assay (no `new_closure` usage), removed when pyo3 0.29 lands. So this migration now retires both 0176 and 0177.

PR #1657, opened 2026-06-12
release: assay 3.24.0

Rul1an:
Workspace version bump (3.23.0 → 3.24.0) + CHANGELOG finalize. **No tag pushed and no release workflow triggered.** This PR only prepares a clean, reviewable version point; tagging `v3.24.0` (which triggers binaries + **crates.io / PyPI** — public and practically irreversible) stays gated on an explicit final go.

## What 3.24.0 captures (merged since 3.23.0), bounded

- **Enforcing-proxy policy decision point (P61e-c)** — an opt-in `assay-mcp-server proxy-enforce` mode that decides each `tools/call` before forwarding. Fail-closed; three gates in fixed precedence (caller-allowance, credential-scope c2, manifest-drift c3); only an all-gates-pass call forwards, with no allow path without a current complete manifest. Non-claims kept explicit: it is not an authorization server; an allow is the decision to forward, not proof of delivery or a side effect; it decides per call and does not reason about multi-step behaviour.
- **`assay.enforcement_decision.v0` per-call evidence record (P61e-d)** — emitted for allow and deny, carrying decision / precedence-pinned reason / `fail_closed` / `drift_state` / credential alias (never token or scopes), with no `forwarded` / delivery field.
- **PDP golden corpus** — an in-repo deterministic truth table over `enforce.rs::decide`, the oracle the decision logic is regression-locked against; plus the canonical `enforcement_decision.v0` contract fixture regenerated from `decision_record` (so a consumer vendors exact producer output).
- **Supply-chain / sanitizer hardening** — scheduled supply-chain posture, HMAC trusted-sanitizer layer, RUSTSEC triage (`-0176`/`-0177` not reachable, pending the pyo3 0.29 migration #1651), `--locked` smoke install.

## Mechanics

- Only `[workspace.package] version` bumped (3.24.0); internal dep pins stay at their `3.19.1` lower bound as in prior releases. `Cargo.lock` refreshed for workspace members only (103 externals unchanged).
- CHANGELOG `[Unreleased]` → `[3.24.0] - 2026-06-12`.
- Also aligns the pre-push `cargo-audit` hook ignores with CI / `deny.toml` (the bare hook tripped on the triaged advisories when the lockfile bumped).

Order agreed: prepare both release PRs, no tags; plimsoll `v0.11.0` (private, GH-release) tags first; assay `v3.24.0` only after explicit final go because crates.io/PyPI is public and irreversible.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -96,7 +96,10 @@ repos:
 
       - id: cargo-audit
         name: cargo audit (RustSec)
-        entry: bash -lc 'cargo audit'
+        # Keep the ignore list in lockstep with .github/workflows/ci.yml and deny.toml: these
+        # advisories are triaged as not reachable (rsa Marvin Attack; pyo3 0.24 nth/nth_back and
+        # PyCFunction::new_closure), retired by the pyo3 0.29 migration (#1651).
+        entry: bash -lc 'cargo audit --ignore RUSTSEC-2023-0071 --ignore RUSTSEC-2026-0176 --ignore RUSTSEC-2026-0177'
         language: system
         pass_filenames: false
         files: ^Cargo\.lock$
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,35 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [3.24.0] - 2026-06-12
+
+### Added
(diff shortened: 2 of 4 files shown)


Rul1an:
Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/27416233477
- sha: 7cefaa28df800d8bbe2b8a6c1db41707349e1664

PR #1702, opened 2026-06-16
feat(registry): MCP04a-3.1 offline Sigstore cert-chain primitive

Rul1an:
## MCP04a-3.1 — offline Sigstore cert-chain primitive

First crypto slice of the supply-chain conformance work. Adds a small,
self-contained offline X.509 path-validation primitive for Fulcio-style
code-signing leaf certificates and maps the outcome onto the existing
`supply_chain::CheckStatus` vocabulary.

### What it does

`sigstore_offline::verify_cert_chain_offline(leaf, intermediates, pinned_roots, now)`:

- delegates path validation to `rustls-webpki` (we never hand-roll it)
- pins the algorithm set to ECDSA P-256 / SHA-256 and requires the
  code-signing EKU — only what Sigstore actually uses, not an open set
- translates the result into the carrier status:

  | situation | status |
  |---|---|
  | no pinned roots | `TrustRootUnavailable` (no silent pass) |
  | expired / not-yet-valid | `Failed` |
  | chain invalid / unrelated CA | `Failed` |
  | leaf lacks code-signing EKU | `Failed` (`required code-signing EKU absent`) |
  | chain valid against a pinned root | `Verified` |

The `Required` EKU check chains through every non-anchor node, which
matches the Fulcio certificate profile exactly: the root (trust anchor)
carries no EKU, the intermediate carries Code Signing, and so does the leaf.

### What this PR proves — and what it explicitly does not

Proves: **offline X.509 certificate path validation over pinned roots**,
plus the validity window and the code-signing EKU requirement. The verdict
is a pure function of the bytes passed in (leaf, intermediates, pinned
roots) plus an explicit verification time — no network, no client, no
async, no ambient trust store.

Does **not** prove (each is a separate later slice, never implied here):

- authorized Sigstore identity (SAN / issuer binding) — a-3.2
- Rekor transparency-log inclusion — a-3.3
- SCT / CT proof
- subject-digest binding to an artifact
- attestation validity

A valid certificate chain is a necessary but not sufficient layer:
**cert-chain-valid is not identity-authorized.**

### Tests

Eleven hermetic tests over a synthetic in-memory PKI (rcgen, dev-only):
valid direct chain, a realistic `leaf -> intermediate -> pinned root` chain
plus withheld/unrelated-intermediate negatives, missing root, unrelated
root, expired, not-yet-valid, malformed leaf, a load-bearing
no-code-signing-EKU rejection, and a **negative no-network guard** — the
same leaf flips `Verified` <-> `Failed` purely on which root bytes it is
handed, and withholding the real root fails rather than discovering it.

### Deps

- `rustls-pki-types` promoted to a direct dependency (webpki's public API
  speaks its types but does not re-export them; same version webpki uses).
- `rcgen` + `time` are dev-only, for the synthetic-PKI fixtures.
- `cargo deny` (licenses, bans, advisories, sources) passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,7 +97,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -108,7 +108,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Rul1an:
Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/27607070189
- sha: 257f9f43f662c0f3268f65c5f54c92dac63266ea

PR #1720, opened 2026-06-19
feat(mcp): add experimental declared-constraint digest projection

Rul1an:
## What

Adds `McpPolicy::declared_constraint_digest_experimental()`, an additive, read-only digest over the
declared *constraint* surface of a policy: `version`, the `tools` allow/deny plus the class / approval /
scope lists, per-tool `schemas`, and `enforcement`. Operational knobs (`runtime_monitor`, `kill_switch`,
`limits`, `discovery`, `signatures`, `tool_pins`, taxonomy) are excluded.

Unlike the existing `policy_digest()`, which serializes the whole policy as-is, this one:

- projects to the constraint surface only, so an operational-only change does not move the digest, and
- semantically normalizes the set-like fields (it sorts `allow`/`deny`, the class / approval / scope
  lists, and each schema's `required` / `enum` by canonical bytes), so reordering a semantically-equal
  policy yields the same digest while a real membership or constraint change still moves it.

## Why

A stable, projection-based declared digest is the foundation for relating an agent's observed tool
decisions to its declared/allowed set. This is the first, deliberately small slice of that work.

## Experimental

The method is named and documented as experimental and unstable: the name and shape may change until it
is promoted out of experimental. No existing behavior changes, and `policy_digest()` is untouched.

## Tests

`cargo test -p assay-core declared_constraint_digest` covers four cases: reordering a semantically-equal
allow-list keeps the digest, a membership change moves it, an operational-only change keeps it, and the
class lists are sorted too. `cargo clippy -p assay-core --lib -- -D warnings` is clean.

## Non-claims

The digest is a contract and identity primitive, not a safety or enforcement statement. The
canonicalization is `serde_jcs`; cross-language digest equivalence is not asserted here. Semantic
normalization covers the set-like constraint fields only.

## Summary by CodeRabbit

* **New Features**
  * Added an experimental method to generate a stable digest for declared policy constraints, enabling reliable comparison and change detection while ignoring operational settings.
* **Tests**
  * Added unit tests ensuring digest stability under reordering and verifying the digest changes when constraint memberships or tool schemas are updated.

### DIFF
--- a/crates/assay-core/src/mcp/policy/mod.rs
+++ b/crates/assay-core/src/mcp/policy/mod.rs
@@ -123,6 +123,11 @@ fn sort_by_canon(arr: &mut [Value]) {
     });
 }
 
+/// Sorts the set-like fields of a JSON-Schema fragment: the top-level `required`, and `enum` within each
+/// direct child of `properties`. KNOWN LIMITATION (acceptable for the experimental status): nested schema
+/// structures are NOT recursed. `items`, `additionalProperties`, `allOf`/`anyOf`/`oneOf`, and nested
+/// object properties keep their given order, so a reordered-but-equal nested schema could still move the
+/// digest. v0 declared schemas are flat; recursive normalization is a v-next refinement.
 fn normalize_schema(sch: &Value) -> Value {
     let mut out = match sch.as_object() {
         Some(o) => o.clone(),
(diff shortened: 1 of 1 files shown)


Rul1an:
@coderabbitai review

PR #1724, opened 2026-06-19
test(mcp): experimental tool-decision-truth conformance vectors

Rul1an:
## What

Adds conformance vectors for the experimental `mcp::tool_decision_truth` slice:

- `tests/fixtures/tool_decision_truth/vectors.json`: a committed fixture of a declared policy plus observed
  decisions with their expected per-decision verdicts and run verdict, plus a couple of pack rows with
  their `coherence_binding`.
- `tests/tool_decision_truth_vectors.rs`: a guard test that (1) checks the committed fixture has not
  drifted from a fresh emit (regenerate with `UPDATE_TDT_VECTORS=1`), and (2) reproduces every verdict via
  `decision_verdict` / `run_verdict` and re-verifies every pack row via `verify_recipe_row`, all from the
  committed bytes.

## Why

Fifth and final slice of the experimental tool-decision truth-layer: a reproduce-from-bytes conformance
guard, mirroring the reference-spec verify-golden discipline, so a change in the gate or digest logic that
drifts the committed verdicts is caught.

## Experimental

Test-only and additive: no library change, no new dependency.

## Tests

`cargo test -p assay-core --test tool_decision_truth_vectors` passes (sync-guard plus reproduce-from-bytes
over 8 cases and 2 pack rows). `cargo clippy -p assay-core --tests -- -D warnings` is clean.

## Non-claims

The vectors fix the current experimental behavior for regression detection; they are not a stability or
correctness guarantee of the schema itself, which remains experimental.

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive tool-decision conformance fixture (JSON) defining multiple policy scenarios and expected verdict outcomes.
  * Introduced a conformance-vector test that validates regenerated results against the committed fixture, with an option to update it.
  * Added stronger verification coverage for decision outputs and integrity checks, including expected “pack row” confirmations and intentional failure cases to detect regressions.

### DIFF
--- a/crates/assay-core/tests/fixtures/tool_decision_truth/vectors.json
+++ b/crates/assay-core/tests/fixtures/tool_decision_truth/vectors.json
@@ -1,39 +1,83 @@
 {
   "schema": "assay.tool_decision_truth.vectors.v0",
-  "declared_policy": {
-    "version": "1",
-    "tools": {
-      "allow": [
-        "read_file",
-        "deploy"
-      ],
-      "deny": [
-        "delete_all"
-      ]
-    },
-    "schemas": {
-      "deploy": {
-        "type": "object",
-        "required": [
-          "env"
+  "policies": {
+    "base": {
+      "version": "1",
+      "tools": {
+        "allow": [
+          "read_file",
+          "deploy"
         ],
-        "properties": {
-          "env": {
-            "enum": [
-              "staging",
-              "prod"
-            ]
+        "deny": [
+          "delete_all"
+        ]
+      },
+      "schemas": {
+        "deploy": {
+          "type": "object",
+          "required": [
+            "env"
+          ],
+          "properties": {
+            "env": {
+              "enum": [
+                "staging",
+                "prod"
+              ]
+            }
           }
         }
+      },
+      "enforcement": {
+        "unconstrained_tools": "warn"
       }
     },
-    "enforcement": {
-      "unconstrained_tools": "warn"
+    "pattern_deny": {
+      "version": "1",
+      "tools": {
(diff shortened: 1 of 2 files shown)


Rul1an:
@coderabbitai review

Rul1an:
Rebuilt the vectors on top of the merged hardening (#1725, #1726) so they pin what the truth-layer actually does now, addressing the review of the original fixture.

The pack rows are built from real `build_classified_record` carriers rather than synthetic digest strings, and the observed-input digest layer is pinned directly: every carrier carries a keyed `args_digest` framed by its `key_id`, never the raw arguments, and a secret-named arg drops out of the digest. The policy set now exercises pattern deny (`delete_*`), an unsupported declared surface that fails to `incomplete` without evidence and resolves once evidence is supplied, and a malformed schema that maps to `invalid`. Run-lattice units pin arity-mismatch and duplicate-order to `invalid`, and there are negative pack rows that must fail closed: a tampered `run_verdict`, a foreign recipe, a foreign canonicalization, and a malformed citation digest.

The guard reproduces every verdict and re-verifies every row from the committed bytes; regenerate with `UPDATE_TDT_VECTORS=1`.

@coderabbitai review

Rul1an:
Good practical catch, fixed in `05a4e7ac`. The fixture key_id is now `fixture-kid-v0`, regenerated from the emitter so fixture and guard stay in sync. The key bytes themselves never entered the fixture; this only changes the neutral identifier so the golden `hmac-sha256:<key_id>:<hex>` digests don't trip the generic-key heuristic.

@coderabbitai review

PR #1723, opened 2026-06-19
feat(mcp): experimental tool-decision pack recipe-row binding

Rul1an:
## What

Adds the pack recipe-row binding to the experimental `mcp::tool_decision_truth` module:

- `carrier_digest(observed_input_digest, declared_policy_digest)`: a sha256 over the JCS-canonical decision
  identity, the handle a pack row cites the carrier by.
- `evidence_ref(carrier_digest, ref)`: the cross-ecosystem envelope `{type, digest, canonicalization,
  schema, ref}`.
- `pack_recipe_row(...)`: a proven recipe row binding the carrier into the existing Evidence Pack, with a
  `coherence_binding` that digests the citation plus recipe plus run verdict.
- `verify_recipe_row(...)`: recomputes the cited carrier digest and the coherence binding from the row, so
  a tampered carrier identity or a changed verdict fails closed.

No pack v2: the row slots into the existing pack shape.

## Why

Fourth slice of the experimental tool-decision truth-layer. It makes the carrier and its verdict packable
and verifiable by content digest.

## Experimental

Additive only, no existing behavior changes, no new dependency. Marked experimental and unstable.

## Tests

`cargo test -p assay-core tool_decision_truth` covers the evidenceRef envelope shape, that a recipe row
coheres with its carrier, and that a tampered carrier identity or a changed run verdict fails closed.
`cargo clippy -p assay-core --lib -- -D warnings` is clean.

## Non-claims

The binding is a content-addressing and coherence statement, not a safety, signature, or issuer-trust
statement. It proves the row cites this exact carrier identity, not that the carrier is correct or
authorized.

## Summary by CodeRabbit

* **New Features**
  * Added APIs for creating and verifying evidence rows with digest-based integrity checks and coherence bindings.
  * Evidence rows now include recipe identifiers, cross-ecosystem evidence references, run verdicts, and tamper-detection mechanisms.

* **Tests**
  * Added comprehensive test suite validating evidence envelope structure, successful verification of untampered rows, and failure detection when digests or verdicts are altered.

PR #1722, opened 2026-06-19
feat(mcp): experimental tool-decision verdict gate and run lattice

Rul1an:
## What

Adds the deterministic verdict gate and run lattice to the experimental `mcp::tool_decision_truth`
module:

- `decision_verdict(policy, tool_name, args, identity_state)`: classifies one observed decision against
  the declared `McpPolicy` over independent axes (tool allow/deny, per-tool args schema plus the
  enforcement unconstrained mode, and `identity_state`), folded with the lattice
  `invalid > mismatch > incomplete > match`. `identity_state = absent` does not block `match`; an
  uncaptured argument or an unconstrained-warn tool resolves to `incomplete`, never a silent pass.
- `run_verdict(decision_verdicts, orders)`: lattice-max over a run, plus order integrity: a duplicate
  `order` makes the observed sequence ambiguous and forces `invalid`.
- `build_classified_record(...)`: builds the carrier with the declared digest taken from the policy and
  the computed verdict in the classification zone.

## Why

Third slice of the experimental tool-decision truth-layer. It turns the observed-input carrier (#1721)
and the declared-constraint digest (#1720) into a deterministic, four-class verdict.

## Experimental

Additive only, no existing behavior changes, and no new dependency (it reuses `McpPolicy` and the
workspace `jsonschema` validators). Marked experimental and unstable.

## Tests

`cargo test -p assay-core tool_decision_truth` covers the per-decision matrix (match / mismatch /
incomplete / invalid across the tool, args, and identity axes), absent-identity-does-not-block-match, the
run lattice and order integrity (duplicate order to invalid), and that the classified record carries the
verdict. `cargo clippy -p assay-core --lib -- -D warnings` is clean.

## Non-claims

The verdict is a contract statement (within or outside the declared set), not a safety or enforcement
statement. Undecidable is `incomplete`, never a pass. v0 args validation compiles schemas per call; a
cached validator is a later optimization.

## Summary by CodeRabbit

* **New Features**
  * Added a verdict subsystem for classifying tool decisions across validation axes
  * New functions for evaluating decision verdicts, aggregating multiple decisions, and building validated decision records

* **Tests**
  * Extended test coverage with comprehensive verdict classification validation

PR #1725, opened 2026-06-19
fix(mcp): harden experimental tool-decision truth-layer per review

Rul1an:
## What

Hardens the experimental tool-decision truth-layer per review (three fixes):

1. `declared_constraint_digest_experimental` (mcp/policy): the `tools` projection now builds an explicit
   allowlist of the declared-constraint fields (allow/deny plus the class/approval/scope lists) instead of
   cloning the whole `tools` object. Fields outside the stated surface (`redact_args`,
   `restrict_scope_contract`, any future `ToolPolicy` field) can no longer move the digest.
2. `verify_recipe_row` now hard-checks that the row declares THIS recipe and the tool-decision-truth
   evidenceRef envelope (`recipe`, `type`, `schema`, `canonicalization`), not just internal coherence, so a
   foreign row with a self-consistent binding no longer verifies.
3. `run_verdict` returns `invalid` when the per-decision verdicts and orders have mismatched lengths
   (untrusted input), instead of silently checking misaligned slices.

## Why

Review findings on the merged experimental slices (#1720, #1722, #1723): the digest claimed a narrower
surface than it covered, the recipe verifier was a generic coherence check, and the run aggregator
accepted mismatched parallel slices.

## Tests

A non-surface `tools` field (redact_args) does not move the declared digest while a surface field still
does; `verify_recipe_row` rejects a self-coherent row with a foreign recipe / envelope type / schema /
canonicalization; `run_verdict` returns invalid on arity mismatch. `cargo test -p assay-core`
(declared_constraint_digest + tool_decision_truth) and `cargo clippy -p assay-core --lib --tests -- -D
warnings` are green.

## Experimental

Additive hardening of the experimental layer, no new dependency. The `declared_constraint_digest` output
changes (it now excludes the non-surface fields), which is acceptable because the schema is experimental
and unstable.

## Summary by CodeRabbit

* **New Features**
  * Added experimental per-tool args schema checking and improved tool-name/args validation surfaced through the policy engine.

* **Bug Fixes**
  * Strengthened tool-decision record verification to reject malformed or foreign recipe/envelope claims and enforce stricter identity/digest coherence.
  * Added closed-fail behavior for invalid verdict/evidence inputs, including enforcing verdict/order length consistency.

* **Tests**
  * Added coverage for experimental digest stability when non-surface policy fields change.
  * Expanded tests to ensure recipe and envelope validation correctly rejects invalid claims.

### DIFF
--- a/crates/assay-core/src/mcp/policy/mod.rs
+++ b/crates/assay-core/src/mcp/policy/mod.rs
@@ -21,6 +21,21 @@ pub(in crate::mcp::policy) use matcher::matches_tool_pattern;
 pub use response::make_deny_response;
 pub use types::*;
 
+/// EXPERIMENTAL: outcome of validating a tool call's arguments against the declared per-tool schema,
+/// used by the tool-decision verdict gate. Distinguishes "no schema declared" from "schema declared but
+/// malformed" so the gate can map a malformed declaration to `invalid` rather than to missing evidence.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ArgsCheck {
+    /// No per-tool schema is declared for this tool.
+    NoSchema,
+    /// A schema is declared and the arguments satisfy it.
+    Valid,
+    /// A schema is declared and the arguments violate it.
+    Invalid,
+    /// A schema is declared but does not compile (the declaration itself is invalid).
+    Malformed,
+}
+
 impl McpPolicy {
     pub fn new() -> Self {
         Self::default()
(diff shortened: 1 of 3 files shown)


Rul1an:
@coderabbitai review

Rul1an:
Folded the remaining review notes into this PR so the declared digest and the `args_digest` carrier are both review-clean before anything else builds on top of them.

On the declared side, an explicitly declared schema now wins over a legacy constraint that migrates onto the same tool, so a mixed-shape policy is never silently rewritten and the digest tracks what is actually enforced. The projection and canonical-sort helpers are fallible now too, propagating a canonicalization failure instead of quietly folding it into an empty string, and the doc is honest that schema normalization is flat v0 (nested structures are not recursed yet).

On the carrier side, the secret-key classifier normalizes a key before matching (lowercase, strip `_`/`-`/spaces), so `Authorization`, `client_secret`, `refresh_token`, `apiKey`, `api-key` and friends all drop uniformly rather than depending on an exact spelling. `args_digest` refuses an empty key or a `key_id` outside `[A-Za-z0-9._-]` rather than emitting a guessable or ambiguously framed digest, and `build_record` validates `source_class` / `result_status` / `identity_state` against the append-only vocabulary so an adapter cannot stamp a label a consumer would then trust. Each one has a test.

@coderabbitai review

Rul1an:
Second hardening commit (`8eaec7b3`) lands the deeper review notes on the verdict gate and the pack rows.

The gate is now sound across the whole declared surface rather than the tool/args/identity subset. Tool allow/deny reuses the engine's own pattern matcher, so a literal `delete_*` deny blocks `delete_all` instead of slipping past exact equality. `decision_verdict` takes a `DecisionEvidence` and folds class, approval, scope and redaction axes alongside the originals: a declared constraint with no supporting evidence resolves to `incomplete`, never a silent `match`, and the evaluators do decide once evidence is supplied. A malformed declared schema now maps to `invalid` through a defensive, non-panicking compile (the old path went through a compile that panics, so the gate was aborting on hostile input).

The pack row binding is split into two honestly-named digests: a `carrier_content_digest` over the full canonical carrier, which the row cites, so tampering any field fails the row closed, plus a `decision_identity_digest` carried as the join key. `verify_recipe_row` recomputes the content digest and the identity from the supplied carrier, requires the row's own `run_verdict` to match, and rejects citation digests that are not well-formed `sha256:<64hex>`.

The conformance vectors in #1724 will be rebuilt on top of this so they pin the corrected behavior from real carriers, with the negative pack cases added.

@coderabbitai review

PR #1726, opened 2026-06-19
fix(mcp): make declared digest and verdict gate coherent per review

Rul1an:
## What

A follow-up to #1725 closing four coherence and validation gaps a deeper review found in the experimental tool-decision truth-layer.

- **Legacy constraints are now evaluated by the verdict.** The declared digest migrates legacy `constraints` into per-tool schemas, but the verdict path only normalized legacy shapes, so a legacy-constraint-only policy could bind a migrated schema in the digest while the gate saw no schema and passed the call under `unconstrained_tools = allow`. Both paths now go through one `normalized_declared_view_experimental` (normalize + migrate + explicit-schema precedence), so the digest and the verdict can never disagree about what is declared.
- **Redaction is in the declared digest.** The gate evaluates `redact_args`/`redact_args_classes`, but the digest did not bind them, so two policies could share a `declared_policy_digest` yet behave differently. The redaction lists now sit in the digest surface alongside approval and scope; the operational `redact_args_contract` stays out.
- **Embedded carrier digests are shape-checked.** `build_record` accepted an arbitrary `declared_policy_digest`, and `verify_recipe_row` never validated the carrier's own embedded digests before recomputing identity. Both now require `sha256:<64hex>`.
- **Pack rows can only bind a real, non-understated verdict.** `pack_recipe_row` accepted any string and a row could claim `match` over a carrier whose own `decision_verdict` was `mismatch`. Both build and verify now require one of the four lattice verdicts and reject a run verdict less severe than the carrier's decision.

## Why

These keep the story honest: the verdict really does cover everything the declared digest binds, and a pack row really is bound to the carrier it cites. The redaction fix flips the earlier `non_surface_tools_field` test (redaction now moves the digest), which is the correct behavior now that the gate evaluates it.

## Tests

`cargo test -p assay-core tool_decision_truth` plus the declared-digest tests: legacy-constraint evaluated by the verdict, redaction lists move the digest while the contract does not, `build_record`/`verify_recipe_row` reject malformed embedded digests, and pack rows reject bogus or understated verdicts. fmt + clippy clean.

## Experimental

Additive and behind the experimental surface; no new dependency. The conformance vectors (#1724) will be rebuilt on top of this so they pin the corrected behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of declared policy digests and evidence-pack verdicts to reject malformed or inconsistent values.
  * Strengthened verification of carrier/join-key digests and run verdict “coverage” before trusting pack rows.
  * Ensured legacy-constraint-only policies are evaluated against the normalized declared view.

* **Improvements**
  * Refactored experimental policy digest computation for more consistent legacy/modern handling.
  * Updated digest sensitivity to include additional tool redaction configuration fields (`redact_args`, `redact_args_classes`).

* **Tests**
  * Updated and extended experimental and evidence-pack test coverage to match the new validation/digest behavior.

### DIFF
--- a/crates/assay-core/src/mcp/policy/mod.rs
+++ b/crates/assay-core/src/mcp/policy/mod.rs
@@ -80,8 +80,8 @@ impl McpPolicy {
 
     /// EXPERIMENTAL (unstable, may change): the declared-CONSTRAINT digest for the tool-decision
     /// truth-layer. Unlike `policy_digest` (the whole policy, Vec-structural), this projects to the
-    /// declared-constraint surface only — `version`, the `tools` allow/deny + class/approval/scope lists,
-    /// per-tool `schemas`, and `enforcement` — excluding operational knobs (`runtime_monitor`,
+    /// declared-constraint surface only — `version`, the `tools` allow/deny + class/approval/scope/redaction
+    /// lists, per-tool `schemas`, and `enforcement` — excluding operational knobs (`runtime_monitor`,
     /// `kill_switch`, `limits`, `discovery`, `signatures`, `tool_pins`, taxonomy), and SEMANTICALLY
     /// NORMALIZES the set-like fields (sorts them by canonical bytes) so a reordered-but-equal policy
     /// yields the same digest while a real membership/constraint change still moves it. Legacy v1 shapes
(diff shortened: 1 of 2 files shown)


Rul1an:
Good catch, fixed in `c1e00884`. `run_verdict_covers_carrier` now only treats a missing or null `decision_verdict` as unconstrained; a present-but-non-lattice verdict fails closed. Added a test that a carrier with `decision_verdict: "approved"` can't be cited by any row.

@coderabbitai review

PR #1730, opened 2026-06-19
feat(otel): project verified tool-decision-truth evidence to OTel (PR10)

Rul1an:
## What

PR10 of the follow-up arc: a lossy, read-only OpenTelemetry GenAI + OpenInference projection of **verified** tool-decision-truth evidence. Evidence first (PR9a/PR9b), projection second.

- **Core projector** (`assay-core`): `project_tool_decision_truth(&[TdtDecision]) -> Projection`. Each verified decision becomes one `TOOL` span (`gen_ai.operation.name=execute_tool`, `gen_ai.tool.name`, `openinference.span.kind=TOOL`), with the verdict and the four digests in `assay.tdt.*` and `assay.claim_class="derived"` (a derived comparison over observed + declared — not a raw observation, not enforcement). **No raw args and no `args_digest`** are projected; only the identity/content digests. Reuses the existing `lossy:true` / `source_of_truth` / pinned-semconv discipline.
- **Single pairing source:** the verifier gains `verify_and_collect`, which runs the one pairing + fail-closed pass and returns both the report and the verified `TdtDecision` pairs. The projector consumes those typed pairs, so it never re-scans a bundle or re-pairs.
- **CLI:** `assay project-otel --evidence-bundle <tdt.tar.gz>` (mutually exclusive with `--capability-surface`). It verifies the bundle in full first; **if the semantic report is not ok, nothing is serialized or written, not even to `--out`** — a view over verified evidence, never a best-effort extractor.

## Why

This completes the arc without letting the observability view become the authority: the content-addressed evidence (the carrier and its recipe row) is the record; this projection is a derived, lossy view that only runs over already-verified pairs.

## Tests

Core: a derived `TOOL` span carries the `assay.tdt.*` set and never the args/`args_digest` (built from a real carrier). CLI: a verified bundle projects; an unverified bundle (carrier with no row) writes nothing and exits non-zero. A **blessed snapshot** (`BLESS=1`) pins the projection byte-for-byte over a match and a mismatch.

## Stacking

Stacked on PR9b (base = `codex/p1-tdt-pack-verify`). Merge order: PR9a → PR9b → PR10; this rebases to main after they land. No OTLP exporter, no live telemetry path — that stays a later slice.

## Release Notes

* **New Features**
  * Added `--evidence-bundle` flag to the `assay project-otel` command for verifying and projecting tool-decision-truth evidence bundles with fail-closed semantics
  * Introduced new experimental OpenTelemetry projection format for tool-decision-truth data

* **Documentation**
  * Added reference documentation for the new tool-decision-truth projection format and command usage

* **Tests**
  * Added integration and contract tests for evidence bundle verification and CLI argument validation

### DIFF
--- a/crates/assay-cli/src/cli/args/project_otel.rs
+++ b/crates/assay-cli/src/cli/args/project_otel.rs
@@ -15,16 +15,19 @@ pub struct ProjectOtelArgs {
 
     /// EXPERIMENTAL: project verified tool-decision-truth recipe rows from an evidence bundle (.tar.gz).
     /// The bundle is verified first; nothing is written unless every row verifies.
-    #[arg(long = "evidence-bundle")]
+    #[arg(
+        long = "evidence-bundle",
+        conflicts_with_all = ["capability_surface", "observation_health", "enforcement_health"]
+    )]
     pub evidence_bundle: Option<PathBuf>,
 
     /// Optional path to an `assay.runner.observation_health.v0` JSON file.
-    #[arg(long = "observation-health")]
+    #[arg(long = "observation-health", conflicts_with = "evidence_bundle")]
     pub observation_health: Option<PathBuf>,
 
     /// Optional path to an `assay.enforcement_health.v0` JSON file. Absent means no enforcement
     /// claim is made (it does NOT assert that enforcement was absent — that lives in the carrier).
-    #[arg(long = "enforcement-health")]
+    #[arg(long = "enforcement-health", conflicts_with = "evidence_bundle")]
     pub enforcement_health: Option<PathBuf>,
 
     /// Write the projection JSON here instead of stdout. On success stdout stays empty.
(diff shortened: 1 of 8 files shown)


Rul1an:
Review fixes folded into PR10:

- Fixed the docs local-link failure without stacking PR10 on PR8 by removing the link to `tool-decision-truth.md` from this branch; local markdown link check now passes.
- Tightened the verifier/projector handoff: `verify_and_collect` now clears verified decisions whenever `report.ok == false`, so projection callers cannot consume partial verified pairs from a failed semantic report.
- Kept the hard projection rule: `project-otel --evidence-bundle` verifies first and writes nothing on failure, including to `--out`.
- Added CLI conflicts so `--evidence-bundle` cannot be mixed with capability-surface context flags (`--capability-surface`, `--observation-health`, `--enforcement-health`) instead of silently ignoring them.

Verified locally with:
- `cargo test -p assay-cli verify_tool_decision_truth --all-targets`
- `cargo test -p assay-cli project_otel::tdt_bundle_tests --all-targets`
- `cargo test -p assay-cli --test project_otel_cli`
- `cargo test -p assay-core --test tdt_otel_projection`
- `cargo fmt --check`
- `cargo clippy -p assay-cli -p assay-core --all-targets -- -D warnings`
- local reproduction of the docs link check

PR #1733, opened 2026-06-19
feat(mcp): experimental live tool-decision-truth carrier producer

Rul1an:
## What

An opt-in producer that mints `assay.tool_decision_truth.v0` carriers during a real `assay mcp wrap` MCP session. Until now the carrier could only be authored after the fact (import a JSON, verify, project); this is the first time the contract is exercised by a live policy evaluation, at the boundary where the decision actually happens.

Carriers are written append-only as NDJSON, one per line, to a sink kept deliberately separate from the decision log. A decision-log entry is operational output ("a policy decision happened"); a carrier is a content-addressed evidence record with its own HMAC/key semantics and claim ceiling ("this observed input was classified against this declared policy"). Folding them together would blur those two, so they stay distinct files.

## How it works

- `assay mcp wrap --tool-decision-truth-out carriers.ndjson -- <server>` turns it on. It requires `ASSAY_TDT_HMAC_KEY` and `ASSAY_TDT_HMAC_KEY_ID` in the environment.
- The producer **fails closed at startup**: if the flag is set but the key or key id is missing, empty, or outside `[A-Za-z0-9._-]`, wrap refuses to start and writes nothing. The key is read once, held in memory only, **removed from the environment before the wrapped child is spawned** (so the child cannot inherit it and forge carriers), and never logged or persisted.
- Carriers are built at the `McpProxy` per-call policy seam via `build_classified_record` with `source_class=authoritative_boundary`, `result_status=ok`, and `identity_state=present|absent` from the runtime identity. Only the keyed `args_digest` is recorded — never raw arguments.
- The proxy does not observe class membership, approval, scope, or redaction at this seam, so those axes stay unknown and any **declared** such obligation resolves to `incomplete`, never a guessed `match`. You can see this live: wrapping a policy that allows a tool but declares no args schema for it mints a carrier with `decision_verdict: incomplete`.

## What it is not

Evidence-only. No runtime action is taken on the verdict and the producer never blocks a call — a carrier is minted even for a call that is about to be denied, recording that the decision was observed and classified. It stays experimental/unstable. The existing `verify-tool-decision-truth` / `project-otel` flow remains the post-hoc reader.

## Verification

Unit tests: the emitted carrier carries no raw arguments; the sink is append-only and ordered; minimal evidence resolves to `incomplete`; the key handling fails closed on missing/empty/malformed material with `Debug` redacting the key; and a minted carrier line round-trips through the existing `import -> verify -> project` path.

Also checked end to end against the built binary: a live `assay mcp wrap` tool call mints exactly one carrier (keyed digest, no raw args, key absent from stderr), and that carrier imports, verifies `ok` (claim ceiling intact), and projects with `assay.tdt.*` + `assay.claim_class=derived` and no raw arguments.

Scope is the producer, its wiring, and tests. A short reference-doc slice describing the producer can follow separately.

## Summary by CodeRabbit

* **New Features**
  * Added experimental `--tool-decision-truth-out <PATH>` to `mcp wrap` to emit `assay.tool_decision_truth.v0` evidence as append-only NDJSON.
  * When enabled, it validates required HMAC credentials from the environment and fails securely if missing or invalid.
  * Tool-decision evidence is captured without affecting the underlying policy decisions.

* **Tests**
  * Added unit and end-to-end coverage for fail-closed validation, redaction, append-only ordering, and evidence contents.

### DIFF
--- a/crates/assay-cli/src/cli/args/mcp.rs
+++ b/crates/assay-cli/src/cli/args/mcp.rs
@@ -112,6 +112,15 @@ pub struct McpWrapArgs {
     #[arg(long, value_name = "URI")]
     pub event_source: Option<String>,
 
+    /// EXPERIMENTAL (unstable): opt-in tool-decision-truth carrier producer. Append
+    /// `assay.tool_decision_truth.v0` carriers (one JSON object per line) to this NDJSON file during the
+    /// run, as a separate sink alongside the decision log. Requires ASSAY_TDT_HMAC_KEY and
+    /// ASSAY_TDT_HMAC_KEY_ID in the environment; the producer fails closed at startup if either is
+    /// missing or malformed. The key is read from the environment, held in memory only, and never logged
+    /// or persisted. Evidence-only: no runtime action is taken on the verdict.
+    #[arg(long = "tool-decision-truth-out", value_name = "PATH")]
+    pub tool_decision_truth_out: Option<PathBuf>,
+
     /// Command to wrap (use -- to separate args)
     #[arg(required = true, last = true, allow_hyphen_values = true)]
     pub command: Vec<String>,
--- a/crates/assay-cli/src/cli/commands/mcp.rs
+++ b/crates/assay-cli/src/cli/commands/mcp.rs
@@ -241,6 +241,18 @@ fn tdt_producer_from_material(
             "ASSAY_TDT_HMAC_KEY_ID is empty or malformed (allowed characters: A-Z a-z 0-9 . _ -); the tool-decision-truth producer fails closed."
         );
     }
+    // Fail closed at startup when the opted-in sink cannot be opened. Otherwise a run could proceed with
+    // `--tool-decision-truth-out` enabled while minting no carriers, which is a half-configured producer.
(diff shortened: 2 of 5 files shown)


Rul1an:
Review fix folded into A1:

- Found one half-configured-producer gap: `--tool-decision-truth-out` failed closed on key material, but did not validate that the NDJSON sink was writable until the first carrier emit. A run could therefore proceed with the producer enabled while minting no carriers.
- Fixed by opening the sink append/create at startup inside `tdt_producer_from_material`; if the sink cannot be opened, startup fails closed before the wrapped child is spawned.
- Added `tdt_producer_fails_closed_when_sink_cannot_be_opened` to lock the behavior.

Verified locally with:
- `cargo test -p assay-cli tdt_producer --bin assay`
- `cargo test -p assay-core mcp::proxy::client::tests --lib`
- `cargo fmt --check`
- `cargo clippy -p assay-cli -p assay-core --all-targets -- -D warnings`

I still recommend holding merge for owner review of the live runtime seam; this is the first producer that mints evidence during a real run.

PR #1779, opened 2026-07-02
feat(cli): capture skill supply-chain carriers

Rul1an:
## Summary

Adds the held auto-capture slice for `assay.skill_supply_chain.v0`:

- new `assay evidence capture-skill-supply-chain` command that reads a local skill file/directory and writes a carrier JSON artifact
- deterministic retained-file identity with `root.digest` and `root.retained_files`
- honest coverage flags for front matter, body text, scripts, lockfiles, and transitive traversal
- fixed `review_incomplete` verdict via `traversal_not_retained`, because this slice does not resolve transitive dependencies or scan registries
- keeps the existing import/verify producer as the fail-closed admission path

## Scope Boundaries

This does not add registry scanning, malware detection, trust scoring, CycloneDX export, scanner adapters, or transitive dependency resolution. Capture emits a carrier; import/verify remain separate.

## Validation

- `cargo fmt --check`
- `cargo test -p assay-cli evidence::skill_supply_chain::tests::capture_`
- `cargo test -p assay-cli skill_supply_chain`
- `cargo clippy -p assay-cli -- -D warnings`
- `cargo build -p assay-mcp-server`
- `cargo test -p assay-cli`
- manual CLI round-trip: `capture-skill-supply-chain` -> `import skill-supply-chain` -> `verify-skill-supply-chain --format json`

PR #1781, opened 2026-07-02
fix(cli): refuse symlinked skill roots in capture; contain reused-skill traversal

Rul1an:
> Small follow-up hardening to the merged `capture-skill-supply-chain` (#1780). The idea is salvaged from the now-superseded duplicate #1779.

## Summary

`capture-skill-supply-chain` read whatever path it was pointed at, following symlinks. This hardens it:

- **Refuse a symlinked skill root or a symlinked resolved `SKILL.md`.** A symlink can point outside the reviewed tree; capture now declines rather than reading through it and recording the linked content as if it were the skill (checked with `symlink_metadata`, without following).
- **Contain reused-skill traversal.** A declared reused-skill dependency counts as traversed only when its canonical real location stays **inside** the reviewed tree. A symlink or `..` that escapes the tree is exactly the hidden-inventory risk, so it degrades to `traversal_not_retained`. An internal symlink that stays within the tree is benign and still traverses — the check is escape-containment, not a blunt symlink ban.

## Why this shape
The root/entry-point is a single file the operator names, so a blunt symlink refusal is right there. Reused-skill traversal can legitimately cross internal symlinks, so it uses canonical-path containment against the reviewed tree instead. That distinguishes a benign internal link from an escaping one.

## Verification
4 unix tests: symlinked root refused, symlinked `SKILL.md` refused, escaping reuse not traversed, internal-symlink reuse still traverses. `assay-cli` suite **332 passed / 0 failed**; `clippy -D warnings` clean; fmt clean.

### DIFF
--- a/crates/assay-cli/src/cli/commands/evidence/skill_supply_chain_capture.rs
+++ b/crates/assay-cli/src/cli/commands/evidence/skill_supply_chain_capture.rs
@@ -91,6 +91,15 @@ fn within_tree(path: &Path, base: &Path) -> bool {
     }
 }
 
+/// `true` only for a REGULAR file, checked WITHOUT following symlinks (a symlink, directory, FIFO, or
+/// other special file is rejected). Capture uses this so adjacent evidence and scanned markdown are
+/// never read through a symlink or blocked on a special target.
+fn is_regular_file(path: &Path) -> bool {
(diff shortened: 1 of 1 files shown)


Rul1an:
@coderabbitai review

PR #1783, opened 2026-07-03
fix(ebpf): LSM emit path wrote pre-MonitorEvent 520-byte records; attach-smoke now fails on ABI skew

Rul1an:
Chases #1575 to its actual root cause and closes the loop mechanically.

**Root cause (source, not staleness):** `emit_event` in `crates/assay-ebpf/src/lsm.rs` reserved a hardcoded `[u8; 520]` — an 8-byte pid+event_type header with data at offset 8, an ABI that predates the 40-byte MonitorEvent header. The userspace decoder rejects any record ≠ `size_of::<MonitorEvent>()` fail-closed, so **every LSM event on a bpf-lsm host was dropped** as `invalid event size (got=520, need=552)`: an observation gap on exactly the enforcement path. The compile-time 552 assert in assay-common could not catch it because the size literal never referenced the shared type.

**How it was found:** the verification run for #1575 (28666638414) went green while carrying the mismatch in its log; this branch's run 28667103298 then rebuilt with a forced-clean, sha256-pinned object and STILL emitted 520 — falsifying the stale-object thesis and pointing at the source.

**Changes:**
- `lsm.rs`: reserve `MonitorEvent` (never a size literal), initialize via `write_event_header`, size caps via `DATA_LEN`. Payload byte layout within `data` unchanged; existing `EVENT_FILE_BLOCKED` decode applies as-is.
- `monitor-attach-smoke.yml`: force a real recompile (`rm -f` + `cargo clean -p assay-ebpf`), pin `ebpf_object_sha256` + mtime in build provenance, and a separate **ABI skew check** step that fails the run on any size-mismatch — making #1575's close condition mechanical instead of forensic.

Dispatching the smoke on this branch: green = attach proven AND fresh object emits 552.

### DIFF
--- a/.github/workflows/monitor-attach-smoke.yml
+++ b/.github/workflows/monitor-attach-smoke.yml
@@ -42,13 +42,23 @@ jobs:
         # smoke run mislook like an assay monitor attach bug.
         run: |
           scripts/ci/install-ebpf-toolchain.sh
+          # Force a real recompile on this persistent runner. A cached/no-op build lets xtask's
+          # artifact resolution hand back a stale object out of a weeks-old target/ dir, which is
+          # exactly the got=520/need=552 skew #1575 chased: the smoke then loads an object that is
+          # NOT the product of this checkout while looking canonical. Deleting the copied .o and
+          # cleaning the crate makes "built from this commit" true by construction.
+          rm -f target/assay-ebpf.o
+          cargo clean -p assay-ebpf 2>/dev/null || true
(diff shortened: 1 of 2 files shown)


Rul1an:
Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/28668123888
- sha: 6eb46d90f6ca233abed934cc96fb62d0f11fac3a

PR #1784, opened 2026-07-03
fix(xtask): resolve eBPF artifact from cargo JSON messages, not filesystem guessing

Rul1an:
Hardening follow-up from the #1575 investigation. The old `resolve_ebpf_output` guessed the built artifact from the filesystem, with three staleness weaknesses on persistent runners: no freshness check on the preferred path, a newest-by-mtime lottery in the `deps/` fallback, and a `CARGO_TARGET_DIR` inconsistency between the two paths.

The local build now runs with `--message-format=json-render-diagnostics` and `resolve_ebpf_artifact()` parses cargo's own `compiler-artifact` messages for the `assay-ebpf` bin target (filtered on `target.name` + `kind: ["bin"]`, since `-Z build-std=core` also reports `core`/`compiler_builtins` artifacts). Path from `executable`, fallback `filenames[0]`. Fail-closed when no artifact message appears, with an error naming the stale-artifact scenario. Deterministic copy to `target/assay-ebpf.o` retained; Docker path untouched (own in-container resolution).

5 new unit tests (executable resolution, filenames fallback, last-match-wins, lib/other-package filtering, fail-closed error path); clippy `-D warnings` clean.

**Relation to #1783:** complementary, not a replacement. The actual root cause of the got=520/need=552 skew was a source bug in `lsm.rs` (hardcoded `[u8; 520]` emit, fixed in #1783); this PR removes the *latent* stale-artifact class that the investigation surfaced. The ABI-skew gate step from #1783 stays — it catches any future emit-path drift regardless of how the object was resolved.

Refs #1575.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local eBPF build artifact detection so the correct output is picked more reliably from Cargo’s build results.
  * Reduced build failures caused by stale or misplaced artifacts, with clearer errors when no matching build output is found.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -573,6 +573,7 @@ version = "3.31.1"
 dependencies = [
  "anyhow",
  "clap",
+ "serde_json",
 ]
 
 [[package]]
--- a/crates/assay-xtask/Cargo.toml
+++ b/crates/assay-xtask/Cargo.toml
@@ -10,3 +10,4 @@ workspace = true
 [dependencies]
 anyhow.workspace = true
 clap.workspace = true
+serde_json = { workspace = true, features = ["std"] }
--- a/crates/assay-xtask/src/main.rs
+++ b/crates/assay-xtask/src/main.rs
@@ -2,7 +2,7 @@ use anyhow::Context as _;
 use clap::Parser;
 use std::fmt::Write as _;
 use std::path::PathBuf;
-use std::process::Command;
+use std::process::{Command, Stdio};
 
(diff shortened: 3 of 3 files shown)


Rul1an:
Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/28668660029
- sha: 87d4438d245ea91ef0457b43ef0d29a18d7356cc

Rul1an:
Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/28668965500
- sha: 7bb6d48d12ee1b2601bbc78ba7a12b8df0c13fe1

PR #1851, opened 2026-07-25
fuzz(evidence): point the verifier fuzz target at the evidence chain and pin its fail-closed classification

Rul1an:
## ADR Boundary Slice

- Canonical ledger: #1847
- Slice: 3 — verifier fuzz and properties
- Builder: Cursor local Agent
- Reviewers: Codex plus CodeRabbit/Copilot; Claude Code fallback non-builder
- Base SHA: `74fe6ec568c0f238438aaac2b0e4ee9b5669550a`
- Final head SHA: `030f66a0`
- ADR invariants affected: ADR-043 §3 (verification effort follows the golden path); §1 as the id-contract checks enforce documented rules the verifier never ran

## Behavior

**The fuzz target covers the artifact the profile is about.** `bundle_reader` fuzzed `assay_core::replay::verify_bundle` — a different bundle format with a different reader — so the evidence chain had no fuzz coverage while appearing to have some. It now runs `assay_evidence::verify_bundle_with_limits` with small explicit limits, because the defaults' 100 MB input and 1 GB expansion would let one case burn the budget and would mean an OOM told us about the machine rather than the verifier.

**The semantic oracle is separate and deterministic.** The fuzzer asserts nothing about *which* rejection is right; an arbitrary byte string has no expected verdict. 18 properties name the inputs and pin the `ErrorCode`. That is the gap: `assay-sim` already runs the broad shapes and asserts `AttackStatus::Blocked`, but "Blocked" is satisfied by any error, so a mutation that starts failing for an unrelated reason still reads green there.

**Two documented rules that never ran now run.** `verify.rs` has listed check 9, "ID Contract: `event.id == run_id:seq`", since the format was written, but `compute_stream_id` was called only by the writer and `ContractInvalidEvent` appeared nowhere but its own enum. And `write.rs` refuses a `run_id` containing a colon — because `run:test` with `seq=0` gives `run:test:0`, which reads equally well as `run_id="run"`, `seq="test:0"` — while the verifier accepted it. Both ends of the format disagreed about what is valid, and a forged stream identity survived every other check once the container was resealed. Each was proven by a test that passed against the old behaviour before the fix.

Blast radius measured rather than assumed: `assay-evidence`, `assay-sim` and the 13-vector privileged-action conformance corpus are green, which is the signal that real bundles already satisfy both contracts.

**The lane runs, on the toolchain it says it does.** Bounded budget on PRs touching `fuzz/` or `crates/assay-evidence/`, longer pass nightly, date-pinned toolchain and version-pinned `cargo-fuzz`, with `rustc --version` echoed into the log. It fails if any required seed is missing. libFuzzer's output corpus points at a scratch directory and the checked-in corpus is a read-only input, so a CI run cannot quietly grow the tree.

Fourteen seeds (65 KB) are generated from the same fixtures by an `#[ignore]`d test, so they are reproducible and reviewable rather than opaque blobs.

## Non-Claims

- [x] No new verdict, score or detector layer; the fuzzer stays a crash/hang/resource explorer
- [x] `-rss_limit_mb` is claimed only as an outer process guard, not as evidence the reader honoured its own ceilings
- [x] Both new diagnostics are value-free: they name the rule and the `seq`, never the offending id or the `run_id` it embeds

## Review rounds

`f6de040e` six findings → `a59b9072` two blocking → `d4688096` four corrections → `bc2b0e03` two findings → `65d61172` approved, one CodeRabbit nitpick. All closed. Each push invalidated the prior quorum.

### On `65d61172` — one finding, closed

CodeRabbit: the PAX test asserted the compressed floor but never that the seed still declares more than the decode ceiling, so shrinking the declared size below 4 MiB would leave the test green while silently exercising a different path. Taken. Both seed tests now compare against `small_limits().max_decode_bytes` rather than a `4 << 20` literal, so the guard moves with the ceiling instead of drifting from it when someone retunes the limits.

Measured margin: the PAX seed is **5,186 bytes compressed, 5,244,416 decoded, against a 4,194,304 ceiling**.

This is the third time in this PR that a check did not guard what it claimed, which is why a Trivial-rated nitpick was worth a fresh review round rather than a disposition.

### Accepted disposition

`small_limits()` in the test duplicates `limits()` in the fuzz target. The only shared home would be a public function in `assay-evidence`, putting a test-only concern on the crate's public surface; the test cannot import the fuzz crate or vice versa. Recorded as deliberate per the reviewer.

### RED

Each round of failures corrected something real rather than confirming what I expected:

```
a_payload_mutation_...          left: (Integrity, IntegrityEventHash)  right: (Integrity, IntegrityManifestHash)
a_bundle_over_the_byte_...      left: Integrity                        right: Limits
expansion_past_the_decode_...   left: Contract                         right: Limits
a_forged_event_id_...           verified successfully                  → documented check 9 never ran
a_run_id_containing_a_colon_... verified successfully                  → writer rule absent from verifier
```

- A payload edit is caught by the **per-event content hash**, not the file digest; an edit to `git_sha`, outside that hash, is caught by the file digest. Both pinned, so a change moving a field across that boundary flips exactly one code.
- The limit tests were testing the wrong thing: an oversized blob is refused for its gzip header long before the ceiling. Both now tighten the ceiling around a valid bundle, with the byte ceiling asserted at the boundary — exact limit accepted, one byte less refused.
- I also retracted a claim. I had recorded "the decode ceiling is unreachable for a bundle-shaped archive" as a limitation, and it is reachable: a PAX extended header is consumed inside the tar reader before any member reaches the verifier, so the per-file ceilings — checked against a declared member size — never apply. `pax-header-past-decode-ceiling` pins `(Limits, LimitDecodeBytes)`; `high-ratio-gzip` keeps its own pinned outcome as the contrasting case that stops at the tar layer.

### GREEN (on `030f66a0`, pinned toolchain)

Run on `nightly-2026-07-24` rather than the floating alias, so the evidence matches what CI runs:

```
cargo test -p assay-evidence --test verifier_fail_closed_properties   # 18 passed, 1 ignored (generator)
cargo test -p assay-sim                                               # all suites pass
cargo test -p assay-cli --test conformance_privileged_mcp_action       # 2 passed (13-vector corpus)
cargo fmt --all -- --check                                            # exit 0
cargo clippy -p assay-evidence --all-targets -- -D warnings           # exit 0
```

```
cargo +nightly-2026-07-24 fuzz run bundle_reader /tmp/out fuzz/corpus/bundle_reader -- -runs=5000 ...
Done 5000 runs   peak_rss_mb: 367   no crash, no hang
```

Corpus unchanged after the run (14 seeds before, 14 after). The final head is a test-only change; the fuzz target and corpus are byte-identical to `65d61172`.

## Review Quorum

- [ ] One non-building agent review on `030f66a0`
- [x] Every actionable finding from all five prior heads is fixed or has a recorded disposition

## Notes for the reviewer

- **Programme-level note, not a change here.** `crates/assay-core/.repro_artifacts/sourceless_sarif.json` is rewritten by a test in `assay-core`, and staging with `git add -A` sweeps the result in — it strips the trailing newline and fails the pre-commit lint. That was the sole cause of the red checks on `d4688096`. Any agent running the core suite will hit it.
- **Environment caveat.** This VM needed `CXX=g++` and a `libstdc++.so` symlink to build libfuzzer-sys locally, because clang++ looked for gcc-14 headers that were not installed. `ubuntu-latest` does not need either; the lane has passed there.

## Summary by CodeRabbit

- **Bug Fixes**
  - Evidence bundle verification now enforces stricter event identity rules, rejects ambiguous run IDs, and tightens contract validation around ordering and container contents.
  - Verification limits continue to reliably cap resources for malformed or oversized inputs.

- **New Features**
  - Added a verifier fuzz “smoke test” workflow to automatically exercise bundle_reader with deterministic budgets and fixed seed coverage.

- **Tests**
  - Added deterministic, fail-closed integrity and contract property tests, including boundary checks and stable failure classification.
  - Expanded fuzz corpus cases for malformed structure, truncation, duplicates, digest issues, and size/decode ceilings.

- **Chores**
  - Updated the bundle_reader fuzz target to validate bundles against the verifier with explicit limits.


crates/assay-evidence/tests/verifier_fail_closed_properties.rs
```diff
@@ -0,0 +1,599 @@
...
+    let (class, code) = expect_rejected(
+        &repack(&[("manifest.json", &resealed), ("events.ndjson", &reordered)]),
+        "an out-of-order chain",
+    );
+    assert_eq!(
+        class,
+        ErrorClass::Contract,
+        "ordering is a contract violation, not an IO error"
+    );
+    assert!(
```
Rul1an: [P2] This mutation is deterministic on this head: swapping lines 0 and 2 makes the first event `seq=2`, so the verifier returns `(Contract, ContractSequenceStart)`. I replaced this two-code allowance with that exact pair in an isolated review worktree and the suite stayed green. Pin the exact code so a future check-order regression cannot hide behind the alternative.

crates/assay-evidence/tests/verifier_fail_closed_properties.rs
```diff
@@ -0,0 +1,599 @@
...
+    }
+
+    // Resealed so this is a truncated *chain*, not a corrupted file: the manifest still claims
+    // three events, which is the claim the verifier has to catch.
+    let resealed = reseal(&manifest, &shortened);
+    let (class, _code) = expect_rejected(
+        &repack(&[("manifest.json", &resealed), ("events.ndjson", &shortened)]),
+        "a chain with its last event removed",
+    );
+    assert!(
```
Rul1an: [P2] This test currently discards the `ErrorCode` and accepts either Contract or Integrity, even though the resealed fixture deterministically reaches `(Contract, ContractSequenceGap)` via the manifest event-count mismatch. The exact assertion passes on this head; retain it as the stable oracle promised by the PR.

crates/assay-evidence/tests/verifier_fail_closed_properties.rs
```diff
@@ -0,0 +1,599 @@
...
+
+    let long_line = {
+        let mut v = Vec::new();
+        v.extend_from_slice(&events);
+        v.extend_from_slice(&vec![b'x'; 9 << 10]);
+        v.push(b'\n');
+        v
+    };
+
+    let seeds: Vec<(&str, Vec<u8>)> = vec![
```
Rul1an, line 641: [P1] None of these seeds crosses the fuzz target's fixed 4 MiB decode ceiling. Measured decoded sizes for valid gzip seeds are 2,560-14,336 bytes. The plan explicitly requires a decompression-expansion seed; add a compact archive that expands beyond 4 MiB and have CI assert required seed names, not only a count.

fuzz/fuzz_targets/bundle_reader.rs
```diff
@@ -1,8 +1,42 @@
...
 
+use assay_evidence::{verify_bundle_with_limits, VerifyLimits};
 use libfuzzer_sys::fuzz_target;
 use std::io::Cursor;
 
+/// Small explicit limits, matching the deterministic property tests.
+///
+/// The defaults allow a 100 MB input and 1 GB of expansion, which would let a single case burn the
+/// whole run budget and would mean an OOM told us about the machine rather than the verifier.
+/// Tight ceilings also make "did it respect its own limit" an observable property: libFuzzer's
```
Rul1an: [P2] Please narrow this claim. `-rss_limit_mb=2048` observes whole-process RSS (the green run peaked at 394 MiB); it cannot establish that this reader honored a 1 MiB input or 4 MiB decode ceiling. It is still useful as an outer process guard. The deterministic exact-limit/classification tests are the evidence for the verifier ceilings.

Rul1an (commented):
Codex non-builder review against head `f6de040e7ef5bae6908f15ff8b5c69c35299a017`.

The architecture is sound: the target now reaches the evidence verifier, the deterministic oracle is separate, corpus regeneration is byte-stable, focused tests pass, actionlint/zizmor are clean, and the PR fuzz job genuinely completed 20,000 units.

Three requirements are still open:

1. **[P1] The documented ID contract is still not enforced by the verifier.** Current main and this head accept a bundle whose `event.id` is forged when the events file digest/size are resealed. I independently reproduced this. The existing ID test reaches only `BundleWriter::finish()`. Implement `event.id == compute_stream_id(run_id, seq)` in the verifier, classify with the existing `ContractInvalidEvent`, and add the verifier-side regression recorded in [#1847](https://github.com/Rul1an/assay/issues/1847#issuecomment-5079472803).

2. **[P1] The checked-in corpus still lacks the required decompression-expansion seed under the target's actual limits.** The target allows 4 MiB decoded; every successfully decoded checked-in seed expands to only 2.5-14 KiB. The deterministic test reaches `LimitDecodeBytes` only by lowering the test limit to 128 bytes, so it does not make the fixed-limit fuzz corpus exercise that branch. Add a compact seed whose decoded stream exceeds 4 MiB, and make the CI corpus assertion check required seed names rather than only count >=8.

3. **[P2] Two named classification properties are deliberately broader than the actual deterministic result.** On this exact head, tightening the reorder assertion to `ContractSequenceStart` and the dropped-tail assertion to `ContractSequenceGap` passes. Pin those exact pairs; accepting either of two codes or either of two classes defeats the stated reason for adding this oracle.

Also correct the resource wording: a 2 GiB process RSS cap is an outer crash/resource guard, not proof that the verifier respected 1/4 MiB reader ceilings. The deterministic limit tests supply that evidence. Pinning `cargo-fuzz 0.13.2` (and preferably the 2026-07-24 nightly used by the green run) would keep this new lane reproducible.

CodeRabbit's status is not review quorum on this head: its only comment says the review was rate-limited. Per AGENTS.md, that is unavailable, not pass. This review remains findings-open and must be repeated on the next head.

Rul1an (commented):
Review of `a59b9072113671a123b75cf6fdb132d6527eb666`.

The secondary review items are closed: exact sequence classifications are pinned, the truncation sweep is exhaustive, cargo-fuzz is version-pinned, required corpus names are asserted, and the RSS wording is bounded.

Two blocking findings remain:

1. **[P1] The documented event-ID contract is still not enforced.** This head has no production diff in `writer_next/verify.rs`, no `compute_stream_id` use in the verifier, and no verifier-side forged-ID regression. Current main and this PR still accept a bundle whose `event.id` is changed and whose container metadata is resealed. Implement the already-recorded correction: recompute `compute_stream_id(&event.run_id, event.seq)`, reject mismatch with `ContractInvalidEvent`, and pin it with a verifier-side property. Do not count the writer-only test.

2. **[P1] The checked-in expansion seed does not exercise the verifier's decode ceiling.** I passed `fuzz/corpus/bundle_reader/expansion-past-decode-ceiling` through the exact target API and limits. The typed result is `Contract/ContractMissingFile`, not `Limits/LimitDecodeBytes`: the tar reader sees the leading zero blocks as an empty archive and stops without consuming the 8 MiB expansion. The generator's standalone `GzDecoder::read_to_end` assertion proves the gzip payload's theoretical decoded length, not the actual verifier path. Replace this with a seed that the real verifier consumes past 4 MiB and add a deterministic CI assertion that its classification is `LimitDecodeBytes`; if the format's earlier orthogonal limits make that shape impossible, remove the corpus-reach claim and record that limitation instead of treating this file as coverage.

A new push will need fresh non-builder review and bot/quorum on that exact SHA. The unpinned `toolchain: nightly` remains a reproducibility follow-up; pinning cargo-fuzz alone does not freeze the toolchain used for the reported run.

Rul1an:
Concrete seed shape for the remaining decode-path finding, independently verified against `a59b9072`:

- Build a gzip-compressed tar stream whose first logical item is a **PAX local extension** with an 8 MiB compressible value, followed by a small `manifest.json` entry.
- `tar` preprocesses and fully reads that extension before yielding the manifest to Assay, so the verifier's `LimitReader` around `GzDecoder` is the component that stops it.
- In a disposable integration test using the exact fuzz target limits, this shape is 8,304 bytes compressed and classifies as `(Limits, LimitDecodeBytes)`.

This avoids the current seed's early empty-archive exit. Please make the generator emit that shape and add a deterministic assertion through `verify_bundle_with_limits`, not just a standalone `GzDecoder` length assertion. The checked-in seed can remain small while genuinely traversing the target path.

Rul1an:
One reproducibility detail for the same corrective push: pinning the action to `toolchain: nightly-2026-07-24` is not enough while the command still says `cargo +nightly fuzz`; the explicit `+nightly` selector can resolve a different moving channel. Use the dated selector consistently (or rely on the action-selected toolchain without a `+nightly` override). That makes the reported Rust toolchain and cargo-fuzz 0.13.2 pair reproducible on the final head.

Rul1an (commented):
Review of `d4688096779752c0e26435e205d343fff73390af`.

Closed on this head:

- the verifier now recomputes `compute_stream_id(run_id, seq)` and rejects a resealed forged ID as `(Contract, ContractInvalidEvent)`;
- the verifier-side regression has real bite;
- focused properties pass (16 passed, 1 ignored generator), clippy `-D warnings` passes, and `git diff --check` is clean.

Open findings:

1. **[P1] The actual decode-path corpus requirement is still not met.** Renaming the old seed to `high-ratio-gzip` and documenting that it returns `ContractMissingFile` is honest, but it does not satisfy the ledger acceptance condition. The asserted “open question” is already answered: a gzip tar beginning with a PAX local extension carrying an 8 MiB compressible value, followed by a small manifest entry, is 8,304 bytes compressed and returns `(Limits, LimitDecodeBytes)` through `verify_bundle_with_limits` under the exact target limits. `tar` preprocesses the extension before yielding the manifest, so Assay's decoder limit is the component that fires. Replace the current high-ratio seed with that verified shape and pin the actual typed result in deterministic CI.

2. **[P1 security] Do not reflect raw `event.id` or derived `run_id` in the verifier diagnostic.** Both are untrusted bundle fields and this error is surfaced through CLI/context chains. The new message currently interpolates both values, permitting control bytes or terminal sequences in an error path. Keep the typed code and use a stable value-free message such as `Event id does not match run_id:seq contract at seq {seq}`; add a control-character regression if the CLI renders the chain directly.

3. **[P2 reproducibility] The Rust toolchain is still moving.** The workflow retains `toolchain: nightly` and `cargo +nightly`. Pin the reported dated nightly and use the same dated selector consistently, or rely on the action-selected dated toolchain without an overriding `+nightly`.

4. **[P3 scope] Restore `crates/assay-core/.repro_artifacts/sourceless_sarif.json`.** Its only diff is removal of the final newline and it is unrelated to ADR-043 §3.

Keep the PR draft. The next push resets review quorum again; request final review only after these four dispositions and fresh local verification on one head.

Rul1an (commented):
Exact-head non-builder review of `bc2b0e033f2d455f87c0a0fe65452e13fea0d4ba`.

Closed on this head:

- the checked-in PAX extended-header seed is small on disk and returns the exact typed `(Limits, LimitDecodeBytes)` result through the production verifier under the fuzz target limits;
- the ID diagnostic is value-free: neither `event.id` nor `run_id` is interpolated, and the remaining `u64` sequence ordinal cannot carry control bytes;
- the unrelated SARIF artifact has no net diff;
- 17 active properties passed independently; `cargo fmt --all -- --check`, `git diff --check`, and actionlint are clean; exact-head GitHub CI is green. The broader local crate run was interrupted by host disk exhaustion after the focused properties passed, so GitHub's green multi-platform matrix is the integration evidence.

Two findings remain:

1. **[P1 reproducibility] The dated toolchain is still overridden at execution time.** `.github/workflows/fuzz-smoke.yml` installs `nightly-2026-07-24` but line 104 invokes `cargo +nightly`. The exact green CI log proves the mismatch: the action installed commit `89c61a754`, then the fuzz step synchronized floating nightly and used commit `da86f4d07`. Use plain `cargo fuzz` after the action or `cargo +nightly-2026-07-24 fuzz`; optionally print/assert the active compiler immediately before the run. As written, the reported dated pin is not the compiler that executed the fuzz lane.

2. **[P2 contract parity] The verifier accepts a stream identity that the writer contract rejects.** `BundleWriter` rejects any colon in `run_id` (`write.rs:144-151`), because the stream ID format uses `:` as the delimiter. The new verifier check only compares `event.id` to `compute_stream_id(run_id, seq)`, so a resealed hostile bundle with `run_id="run:test"` and `id="run:test:0"` passes this check despite violating the producer contract. Enforce the same delimiter constraint in the verifier with a stable typed rejection and add an exact verifier-side property.

The CodeRabbit duplication comment is accepted as a technical disposition: introducing public production API solely to share test/fuzz limit literals would enlarge the surface for little gain. A diagnostic no-echo assertion would still be useful while touching the property test, but the current value-free implementation closes the original injection defect.

Keep the PR draft. A corrective push resets this review and the skipped-draft CodeRabbit status still does not count as quorum.

Rul1an (commented):
Final non-builder review of `65d61172aed4ab725fe7b09ca6401c04a2939543`.

No actionable findings remain.

Exact-head evidence:

- the PAX seed reaches `(Limits, LimitDecodeBytes)` through the production verifier under the target limits;
- the event-ID equality and colon-free `run_id` constraints now match the writer contract and have exact verifier-side properties;
- diagnostics remain value-free;
- the workflow single-sources `nightly-2026-07-24`, and the green fuzz log prints and then executes compiler commit `89c61a754` rather than synchronizing floating nightly;
- the unrelated SARIF diff is absent;
- all required checks are green, including the 20,000-unit verifier fuzz smoke, multi-platform tests, clippy, CodeQL, workflow security and lane-check.

The duplicated fuzz/test limit literal remains an accepted disposition: adding public production API solely to share test configuration would enlarge the surface without improving runtime correctness.

This review satisfies the non-builder half of quorum on the exact head. CodeRabbit is still `Review skipped: draft` and does not count; move ready and require a fresh bot review on this SHA before merge.

Rul1an (commented):
Exact-head non-builder review of `030f66a0b809c8366554856b5f93d8d6b7bc040f`.

No actionable findings remain. The only delta from the previously reviewed head is the requested premise guard: both expansion tests now compare measured decoded size against `small_limits().max_decode_bytes`, and the checked-in PAX seed is independently measured at 5,244,416 decoded bytes against a 4,194,304-byte ceiling.

The full current-head matrix is green, `lane-check/proof` reports that no delegated proof is required, and the CodeRabbit `Review completed` success status is attached to this exact SHA. Review quorum is complete.

PR #1852, opened 2026-07-25
fix(evidence,core,cli): bounded ingest for every local entrypoint (ADR-043 §1, Slice 1a)

Rul1an:
Slice 1a of the ADR-042/043 programme: bounded ingest for every local entrypoint that consumes an untrusted evidence or replay archive. Ledger is #1847.

ADR-043 §1 states the invariant: each entrypoint applies the whole limit set to the source stream *before* the input is materialized, and a single byte ceiling is not that set. Every entrypoint named there read first and checked afterwards, or carried no ceiling at all.

Object-store pull is deliberately out of scope and sequenced as Slice 1b, because the materialization there happens inside the store API.

## What changes

**The primitive.** `LimitReader` moves to `assay-common::limits` behind the `std` feature, so the evidence verifier and the replay verifier apply one implementation of the boundary. Only the mechanism moves; limit vocabularies stay domain-local, because `max_manifest_bytes` and `max_events` describe an evidence bundle and say nothing about a replay bundle. This adds an `assay-evidence -> assay-common` edge, documented in the crate graph in `CLAUDE.md`.

Three defects in the primitive itself, each able to turn a refusal into a silent success:

- The ceiling was exclusive. Erroring as soon as `read == limit` rejected the exact-limit input, because every consumer issues one further read to observe EOF, so the effective ceiling was `limit - 1`. It is inclusive now, which widens three pre-existing consumers in `writer_next/verify.rs` by one byte each. That commit is separated deliberately.
- The overflow was not sticky. After the error the counter still sat at the limit, so a retry probed an exhausted source, saw `0`, and reported a clean EOF: a truncated archive presented as a complete one.
- Empty reads consumed the source, against the `Read` contract, because the boundary probe ran for a zero-length request.

**Typed refusals.** The verifier recognised its own ceilings with `ve.message.contains("LimitBundleBytes")` and seven siblings, which made rendered text a contract. The cause now travels as data: `LimitExceeded { LimitKind, limit }` through the `io::Error`, recovered with `from_io`. `LimitKind` is a closed enum with no wildcard arm, so adding a dimension is a compile error at every consumer rather than a silent misclassification. All eight message-matching sites are gone.

**The entrypoints.**

| Entrypoint | Before |
|---|---|
| `BundleReader::open*` | `read_to_end` then check; `open_unverified` and peek unbounded outright |
| `BundleInfo::peek*` | public, expanded the archive with no ceiling at all |
| lint | own unbounded read, plus a second decode and second tar walk outside the limits |
| `evidence verify -` | stdin read to the end, then a `Cursor` verified |
| `evidence push` | file read whole with no ceiling, buffer cloned for upload, `--no-verify` skipped even that |
| `assay_core::replay` | no limits of any kind |

`Option<VerifyLimits>` conflated "do not verify" with "do not bound"; verification is a separate flag now and the limits apply either way.

**All eight dimensions.** `max_json_depth` was configurable and had no effect: the guard compared against a module constant. `LimitJsonDepth` was the one `Limit*` variant with no construction site, and a depth refusal was reported as `ContractInvalidJson`, which says the document is malformed when it means it is deeper than we agreed to parse. Object and array nesting are separate paths and both now take the caller's ceiling; the manifest was the document the ceiling never reached on either path. `max_path_len`, `max_line_bytes` and `max_events` were enforced only while verifying and now apply to the unverified paths too.

`max_line_bytes` had two interpretations: the line reader counted the trailing newline inside the ceiling and `check_events_shape` did not, so one budget gave two answers. It is the payload length on both paths.

**Replay.** Six ceilings where there were none: compressed source, gzip expansion, manifest, member, path length, entry count, plus a manifest JSON-depth check over raw bytes. Path length is measured on `path_bytes()` before any conversion, because `to_string_lossy` emits three bytes for every invalid one and a check on the converted string measures something the archive never sent.

The replay CLI read the bundle twice in two different ways: `sha256_file` for `provenance.bundle_digest`, then a second open for the parse and verify. The published digest could describe different bytes than the verified and replayed bundle. `read_verify_bounded` takes one bounded read and returns the digest, the parse and the verdict together; `verify_read_bundle` is crate-internal so the split cannot be reassembled wrongly.

**Diagnostics are value-free.** A refusal reports the ceiling that applied and never the attacker-chosen value: not the observed nesting depth, not the offending path, not the member name. The nesting message also named the module constant as the maximum, so a refusal under a configured ceiling of four told the operator it had exceeded 64.

## Correction round

Three blockers from the exact-head review on `e3a80634` are closed in `413ebf12`.

**The whole-source ceiling was not whole.** Both verifiers stopped reading at the archive's logical end, so a 10,598-byte input passed a 598-byte ceiling: gzip and tar stop early and never request the trailing bytes. A ceiling that only bounds the part a decoder happens to consume is not a ceiling on the artifact. Both entrypoints now snapshot the source to EOF through a `LimitReader` before any parsing. This is a deliberate correctness-over-laziness trade: the whole source is read before it is judged, which is what §1 asks for.

**The unverified read path bounded the container but not the events inside it**, so `max_events`, `max_json_depth` and the line ceiling were decorative once a caller opened a bundle without verification. `read_events_bounded` reads line by line and refuses early; `check_events_shape` is removed rather than left as a second, weaker path.

**The replay CLI still opened the bundle twice**, so the digest reported and the bytes judged were not necessarily the same file. `read_verify_bounded` returns one snapshot, and the flow matches the typed refusal before rendering, keeping the digest of the bytes that actually failed instead of degrading to `sha256:unknown`.

`E_REPLAY_LIMIT_EXCEEDED` is additive and resource-only. A well-formed bundle over budget is an operator decision; reporting it as `E_CFG_PARSE` sent the reader off to fix the producer. Contract tests pin the public `summary.json` shape: the wire string across every refusal variant, exit class 2, digest survival, value-free messages, and the **absence** of `passed`/`failed`/`total`/`verdict`/`score`. Absent, not zero: a count of zero is itself a claim about a run that never started, and a consumer aggregating summaries would fold it in as a real result.

## Verification

Run on `413ebf12`, rebased onto `038541be`:

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test` for `assay-common`, `assay-evidence`, `assay-core` (40 suites) and `assay-cli` (44 suites) — no failures
- 54 targeted boundary contracts, confirmed running with real counts rather than silently filtered: `bounded_ingest_reader` 20, `typed_limit_classification` 9, `bounded_ingest_lint` 3, `bounded_ingest_stdin` 3, replay `bounded_ingest` 13, `replay_limit_summary_contract` 6
- `git diff --check` clean; pre-push clippy and linux compile gate passed

Every test was verified to fail against the implementation it guards, checked per fix rather than by removing all ceilings at once. Every refusal case has an acceptance twin from the same fixture, so a ceiling that refused everything would not pass. The load-bearing pin has a negative control: renaming the wire string `E_REPLAY_LIMIT_EXCEEDED` fails the contract test.

### Benchmarks

Bencher flagged `sw/12xlarge` at +54.31%, with `sw/50x400b` (+38.57%) and `sr/wc` (+38.29%) under their boundaries. Not attributable to this branch: `sw/*` is `store_write_heavy`, which exercises the SQLite `Store` write path, and `sr/wc` is `suite_run_worstcase`. Neither references `verify_bundle`, `BundleReader`, `read_bundle_tar_gz`, `LimitReader` or `lint`, so no changed symbol is reachable from either. Three unrelated benchmarks moved up together with none moving down, which is a runner signature rather than a code one; `sw/12xlarge` tripped because it had the least headroom. The run was on `e3a80634`, which predates the snapshot change, so the snapshot is not the cause either. No change made — the snapshot is the correctness boundary and is not worth trading against an unattributed measurement.

## Where a twin is absent, and why

Some boundaries are not pinned exactly, and the tests say so rather than asserting something weaker and looking complete:

The source boundary is now exact through both entrypoints, since each reads to EOF before parsing. It was previously exact only through `read_verify_bounded`; that asymmetry is what the correction round removed.

The decode dimension has no exact twin because its crossing point depends on the compressor rather than on a number the test controls.

Path length is bounded by what the tar reader decodes. A name delivered through a PAX extended header is measured after that decode, not before, so the claim here is bounded-by-decode rather than bounded-by-wire-bytes. Pinning it at the wire would mean a parser rewrite, which is out of proportion to the gap.

There is also no CLI-level ceiling test. `evidence verify` and `push` take no limits flag, so tripping a default ceiling from the CLI needs a fixture around a gigabyte, and a cheap oversized fixture is refused by tar's header check long before the ceiling is reached. The CLI tests pin the ordering invariant instead: a refusal must leave the store empty. They pass against the previous implementation too, which is what a guard is for. The replay summary contract is pinned at the value `write_replay_failure` serializes, because that function writes to fixed relative paths and exercising it directly would mean moving the process CWD.

## Not in this PR

Object-store pull (Slice 1b). Session authorization and evidence-chain fuzzing, which are their own slices. No scalar score, whole-action verdict, identity layer, HTTP transport, or compliance claim.

## Summary by CodeRabbit

* **New Features**
  * Added replay ingest limit refusal `E_REPLAY_LIMIT_EXCEEDED` (exit code `2`).
* **Bug Fixes**
  * Evidence `verify`/`push` now consistently enforce bounded ingest limits across stdin and `--no-verify`, rejecting oversized/malformed bundles before any upload or store write.
  * Replay and lint ingest/verification stop at configured ceilings and return clearer, categorized limit errors.
* **Improvements**
  * Replay bundles are read/verified/fingerprinted in one bounded operation; workspace-safe replay materialization is enforced.
* **Tests**
  * Added integration and regression tests covering ceiling boundaries, typed refusals, and workspace containment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- New reason code `E_REPLAY_LIMIT_EXCEEDED` (exit 2) for a replay bundle refused by an ingest
+  ceiling before it was parsed. Previously such a refusal was reported as `E_CFG_PARSE`, which
+  says the input is malformed and sends a reader off to fix the producer; a bundle over budget
+  may be perfectly well-formed, and raising the ceiling is a legitimate response. Backward
+  compatible: a new registry string under the existing `reason_code_version` 1, registered in
+  `docs/architecture/SPEC-PR-Gate-Outputs-v1.md` §5.1, so consumers branching on
+  `(reason_code_version, reason_code)` fall back as the spec already requires. The refusal
+  carries no verdict, and its message names only the configured ceiling.
+
 ### Changed
 - **Breaking (`assay sandbox`)**: a `--policy` that is named but cannot be loaded is now fatal.
   The run exits 2 with `E_POLICY_LOAD_FAILED_UNENFORCEABLE` and executes nothing, where it
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,11 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - New reason code `E_REPLAY_LIMIT_EXCEEDED` (exit 2) for a replay bundle refused by an ingest
-  ceiling before it was parsed. Previously such a refusal was reported as `E_CFG_PARSE`, which
-  says the input is malformed and sends a reader off to fix the producer; a bundle over budget
-  may be perfectly well-formed, and raising the ceiling is a legitimate response. Backward
+  ceiling during bounded ingest, before replay execution. Previously such a refusal was reported
+  as `E_CFG_PARSE`, which is a malformed-input finding and sends a reader off to fix the
+  producer. A ceiling establishes only that a configured budget was exceeded; the read stopped
+  there, so whether the bundle is otherwise valid is unresolved, and adjusting the budget or
+  supplying a smaller bundle is a legitimate response. Backward
   compatible: a new registry string under the existing `reason_code_version` 1, registered in
   `docs/architecture/SPEC-PR-Gate-Outputs-v1.md` §5.1, so consumers branching on
   `(reason_code_version, reason_code)` fall back as the spec already requires. The refusal
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -119,7 +119,7 @@ assay-mcp-server -> assay-core, assay-common, assay-metrics
 assay-monitor -> assay-common, assay-policy
 assay-metrics -> assay-core, assay-common
 assay-core -> assay-adapter-api, assay-common
-assay-evidence -> assay-canonical
+assay-evidence -> assay-canonical, assay-common
 assay-adapter-api -> assay-evidence
 assay-adapter-{a2a,acp,ucp} -> assay-adapter-api, assay-evidence
 assay-runner-core -> assay-common, assay-monitor, assay-runner-schema
@@ -130,6 +130,11 @@ assay-ebpf -> assay-common
 Leaf crates (no internal dependencies): `assay-common`, `assay-canonical`, `assay-policy`, `assay-registry`, `assay-runner-schema`, `assay-runner-linux`, `gateway-evidence-replay`, `assay-xtask`.
 
 No circular dependencies. All dependencies flow in one direction.
+
+`assay-evidence -> assay-common` carries only the bounded-ingest primitive
+(`assay_common::limits::LimitReader`, ADR-043 §1), shared with the replay verifier in
+`assay-core` so both apply one ceiling mechanism. Limit *vocabularies* stay domain-local:
+`VerifyLimits` describes an evidence bundle and does not travel.
 
 ## assay-sim (Attack Simulation)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -365,6 +365,7 @@ version = "3.34.0"
 dependencies = [
  "anyhow",
  "assay-canonical",
+ "assay-common",
  "assert_fs",
  "async-trait",
  "base64",
--- a/crates/assay-cli/src/cli/commands/evidence/mod.rs
+++ b/crates/assay-cli/src/cli/commands/evidence/mod.rs
@@ -30,7 +30,7 @@ use anyhow::{Context, Result};
 use clap::{Args, Subcommand};
 use mapping::{DetailLevel, EvidenceMapper};
 use std::fs::File;
-use std::io::{self, Read};
+use std::io;
 
 /// Manage tamper-evident bundles (audit/compliance)
 #[derive(Debug, Subcommand, Clone)]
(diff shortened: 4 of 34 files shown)


Rul1an (commented):
NON-BUILDER EXACT-HEAD REVIEW — 64d39e886fc5672b55a9c2e7058b43a35a5541d0\n\nIndependent Codex review complete. Bounded source/decode/member/path/count/depth enforcement, typed value-free refusals, direct-peek whole-source accounting, replay path containment, duplicate handling, and digest-scoped refusal provenance were checked against the implementation. Independent local verification passed: assay-common limits 9/9; evidence boundary contracts 41/41; replay core 46/46; replay CLI/contract paths 20/20; cargo fmt and git diff checks clean. No blocking findings. Accepted dispositions remain bounded 2x source buffering and the existing TooManyKeys/StringTooLong classification follow-up; neither weakens the named Slice 1a invariants.\n\nGitHub does not permit a formal approval because builder and reviewer operate through the same repository-owner account; this comment records the distinct non-builder review. Bot quorum must still target this exact SHA.

Rul1an:
@coderabbitai review\n\nPlease review the current exact head 64d39e886fc5672b55a9c2e7058b43a35a5541d0. Earlier reviews target superseded SHAs.

PR #1856, opened 2026-07-26
test(core): stop a plain test run from rewriting a tracked fixture

Rul1an:
`crates/assay-core/tests/generate_fixture.rs` wrote `.repro_artifacts/sourceless_sarif.json` — a **tracked** path — as a side effect of `cargo test`, then read the file back to assert on it.

Two consequences:

- A plain test run left the working tree dirty, so the change rode along in whatever was staged next. That happened twice during the ADR-043 slices and had to be backed out by hand each time.
- The file could never settle. `write_sarif` emits no trailing newline; the repository's end-of-file pre-commit hook adds one. So the hook put the byte on, the next test run took it off, indefinitely.

## What changes

Generation happens in a temp directory and the bytes are compared against the checked-in fixture.

That is not a relocation, it changes what the file is for. The fixture was an output nobody read — no consumer anywhere in the repo. It is now the pinned shape of a SARIF run for a finding with **no source location**, and a writer change that alters it fails the build instead of silently rewriting the evidence of itself.

**Exactly one** end-of-file byte is normalised out of the comparison, deliberately: it belongs to the repository's file convention, not to anything the SARIF writer produces or promises, and comparing it exactly is what caused the churn. One, not any number of them — `trim_end_matches` would also swallow a gained blank line and let a genuinely different fixture compare equal. Everything before it is compared byte for byte.

The fallback URI is asserted at the **exact SARIF location**, not as a substring. `contains(".assay/eval.yaml")` passes if the string turns up anywhere in the document, so it kept passing while the artifact location itself could degrade. The test now reads `/runs/0/results/0/locations/0/physicalLocation/artifactLocation/uri`, which is the only thing it is about.

Regeneration stays available as an `#[ignore]`d test, so it is something a developer asks for rather than something a test run does to them:

```bash
cargo test -p assay-core --test generate_fixture -- --ignored regenerate
```

It writes the trailing newline the hook would add anyway, so a deliberate regeneration leaves no diff for the next commit to change again.

House precedent for the ignored-generator form is `generate_seed_corpus` in `assay-evidence`. An `xtask` subcommand was the alternative, but it would make the build tool depend on `assay-core` for a single fixture.

## Line endings

The fixture is pinned to LF in `.gitattributes`, following the pin already there for `gateway-evidence-replay`'s fixtures.

Without it the first CI run failed on **Windows only**: git's line-ending translation handed the test a CRLF golden while the writer emits LF, so a byte comparison could never hold across platforms. Pinning the bytes is the fix rather than stripping `\r` in the test — normalising there would make the guard pass on a checkout that is silently translating the file, which is exactly the misconfiguration worth seeing. The failure message names that case, so a stale Windows checkout is pointed at the one fixture rather than at regeneration, which would commit the wrong bytes:

```bash
git restore --source=HEAD --worktree -- crates/assay-core/.repro_artifacts/sourceless_sarif.json
```

Scoped to that file, and the message says to preserve any intentional local edits first. A repo-wide `git add --renormalize .` was the earlier suggestion and was dropped: it stages unrelated work, and it normalises the index rather than re-checking-out the working-tree file, so the test can stay red. The scoped command was verified against a deliberately CRLF-ified fixture — 35 CRLF lines, test failing, command, zero CRLF, test passing, nothing staged.

## Verification

- `cargo fmt --all --check`, `git diff --check` — clean
- `cargo clippy -p assay-core --all-targets -- -D warnings` — clean
- Full `assay-core` suite: 40 suites, no failures — **after which `git status` is clean**, which is the property this change exists for

Three checks on the behaviour itself:

| Check | Result |
|---|---|
| deliberate regeneration is idempotent | no diff produced |
| a mutated golden fails the guard | `FAILED. 1 passed; 1 failed` |
| restoring the previous test reproduces the defect | fixture dirtied, `\ No newline at end of file` |

Both review findings were confirmed to discriminate rather than merely rewritten:

| Finding | Old form | New form |
|---|---|---|
| fixture gains a blank line | `trim_end_matches` → **passes** | `strip_suffix` → **fails** |
| artifact location degrades to `UNKNOWN` while the URI stays in the message text | `contains` → **passes** | JSON pointer → **fails** |

## Scope

Two files: the test, and one line in `.gitattributes`. No production code, no public API, no behaviour change to `write_sarif` itself.

## Summary by CodeRabbit

* **Tests**
  * Improved fixture validation to compare generated SARIF output with the checked-in reference.
  * Added coverage confirming sourceless findings include the expected fallback URI.
  * Added an optional regeneration test for updating the reference fixture when needed.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 crates/gateway-evidence-replay/fixtures/gateway-path-v0/demo/** text eol=lf
+crates/assay-core/.repro_artifacts/** text eol=lf
--- a/crates/assay-core/tests/generate_fixture.rs
+++ b/crates/assay-core/tests/generate_fixture.rs
@@ -1,9 +1,25 @@
+//! The sourceless-SARIF fixture: guarded here, regenerated deliberately.
+//!
+//! This file used to write `.repro_artifacts/sourceless_sarif.json` — a *tracked* path — as a side
+//! effect of `cargo test`. Two things were wrong with that. A plain test run mutated the working
+//! tree, so `git status` was dirty afterwards and the change rode along in whatever was staged
+//! next. And because `write_sarif` emits no trailing newline while the repository's end-of-file
+//! hook adds one on commit, the file flipped back and forth indefinitely: the hook put the byte on,
+//! the next test run took it off.
+//!
+//! Generation now happens in a temporary directory and the result is compared against the
+//! checked-in fixture. That turns a generator into a regression guard: the fixture stops being an
+//! output nobody reads and becomes the pinned shape of a SARIF run for a finding that has no source
+//! location. Regeneration is still available, but only when asked for.
+
 use assay_core::model::{TestResultRow, TestStatus};
 use assay_core::report::sarif;
+use std::path::{Path, PathBuf};
 
-#[test]
-fn test_generate_sourceless_sarif_fixture() {
-    let results = vec![TestResultRow {
+/// The one finding the fixture is about: a failure with no file context, so SARIF has to fall back
+/// to a synthetic location rather than omit one.
+fn sourceless_failure() -> Vec<TestResultRow> {
+    vec![TestResultRow {
         test_id: "sourceless_failure_demo".to_string(),
         status: TestStatus::Fail,
         score: Some(0.0),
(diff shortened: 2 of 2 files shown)


Rul1an:
## Merging with a documented exception to the same-head bot-review rule

Recording this because the rule was not met and the record should say so rather than imply otherwise.

**There is no fresh bot review on the merged head `aa687d1c`.** Both formal review objects belong to `04776f0f`, two heads back:

| Reviewer | Head reviewed | Submitted |
|---|---|---|
| copilot-pull-request-reviewer | `04776f0f` | 2026-07-26T06:57:29Z |
| coderabbitai | `04776f0f` | 2026-07-26T06:57:57Z |

**The green CodeRabbit context on `aa687d1c` is not evidence of a review.** Its description reads:

```
CodeRabbit   success   "Review rate limited"   2026-07-26T07:47:06Z
```

That is CodeRabbit reporting that it *skipped* the review and setting the context to success so as not to block. Counting it as bot coverage would count a skip as a pass, which is the substitution ADR-043 exists to refuse — so it is not counted here.

**What the merge does rest on:**

- Same-head human code review on `aa687d1c`, which raised no code findings.
- Full CI on `aa687d1c`: 30 checks, 0 failures, including `Build + Test (windows-latest)` — the platform that caught the CRLF defect in the first place — plus `CI` and `lane-check` green. `host-capability-check` is path-conditional and does not run on this change.
- 0 unresolved review threads.
- `CLEAN` / `MERGEABLE`.

**What was not checked:** no bot re-examined the last two heads (`4da1ff13`, `aa687d1c`). The changes across those heads were the `.gitattributes` LF pin, the scoped recovery command in an assertion message, and a doc-comment wording fix. All three were verified by hand — the recovery command against a deliberately CRLF-ified fixture — but none carry a fresh bot signature.

The two earlier CodeRabbit findings from `04776f0f` were closed and each was shown to discriminate rather than merely rewritten: a fixture that gains a blank line now fails where `trim_end_matches` passed it, and a document whose artifact location degrades to `UNKNOWN` while the URI remains in the message text fails the JSON-pointer assertion where `contains` passed it.

Issue #1840, opened 2026-07-24
Independent reproduction wanted: privileged-mcp-action/v0 conformance corpus


Rul1an:
The `privileged-mcp-action/v0` open profile has an inputs-only clean-room candidate pack for independent reproduction. The profile verifies Assay evidence records for privileged MCP tool calls against a frozen 14-case corpus. Its corpus digest remains **candidate**, not reproduced, until a non-author implementation derives the outcomes from the specification and matches the pinned normative surface.

## Clean-room input

Use release [`privileged-mcp-action-v0-candidate.4`](https://github.com/Rul1an/assay/releases/tag/privileged-mcp-action-v0-candidate.4):

- pack: [`privileged-mcp-action-v0-clean-room.tar.gz`](https://github.com/Rul1an/assay/releases/download/privileged-mcp-action-v0-candidate.4/privileged-mcp-action-v0-clean-room.tar.gz)
- pack SHA-256: `4a3d5a713d424b37eb77ab238e73582e3dd26d42f8b656d0a6216c2993909b72`
- source commit: `0c055ead5a2661569d544df4a0df40df90b98e18`
- source corpus digest: `sha256:cb58ce91863f52e0568742b977f0642158453ec11bbcd25821f9171dccd03342`
- 14 opaque cases, with expected outcomes and semantic vector names omitted

The pack contains the normative specification, descriptor and opaque bundles. It omits corpus pointers, expected outcomes, `gen_vectors.py`, Assay verifier code and semantic vector identities. For a blind from-spec claim, implement from those released inputs before reading Assay source or canonical outcomes.

The pinned [conformance protocol](https://github.com/Rul1an/assay/blob/0c055ead5a2661569d544df4a0df40df90b98e18/conformance/privileged-mcp-action-v0/CONFORMANCE-PROTOCOL.md) defines the candidate command, scoring surface, authorship modes and claim ceiling. Its reusable action is pinned independently to commit `16ea2b84e472412e3e5c4d9dcabff61b7fac72f8`. The [implementation report template](https://github.com/Rul1an/assay/blob/0c055ead5a2661569d544df4a0df40df90b98e18/conformance/privileged-mcp-action-v0/IMPLEMENTATION-REPORT.template.md) records source, commit, method and non-claims.

[...]

PR #1858, opened 2026-07-26
docs(profile): restore the canonical normative URI and pin the docs domain

Rul1an:
The `privileged-mcp-action/v0` corpus digest is a **candidate** until a non-author implementation derives the expected outcomes from the specification text alone. The invitation is #1840. This makes that attempt possible to start, and stops the merge from breaking the docs domain on the way.

## Merging this would have dropped the custom domain

`.github/workflows/docs.yml` runs `mkdocs gh-deploy --force` on any push to `main` touching `docs/**` — and this branch touches `docs/`. Without a `CNAME` in the docs **source**, a forced redeploy rewrites `gh-pages` and takes the custom domain with it.

`docs/CNAME` now carries `docs.getassay.dev`, and `site_url` matches it, so sitemap, canonical links and social metadata stop naming a host that never existed.

## The normative URI is restored, on both surfaces

The canonical documentation host is `docs.getassay.dev`, and it serves [the profile page](https://docs.getassay.dev/profiles/privileged-mcp-action/v0/) today.

An earlier revision of this branch concluded the normative URI was dead and removed it. That conclusion came from testing the wrong hostname — `docs.assay.dev`, which indeed does not resolve. The profile keeps a normative URI, and it now names the surface that publishes it, with the same value in the spec header and in `descriptor.json`, the machine-readable input #1840 names.

## Reproducing without reading the implementation

`gen_vectors.py` sits in the corpus directory and is the one file a from-spec attempt must not read: reading it turns a reimplementation into a port. `crates/assay-cli` is the second such surface.

```bash
git clone --filter=blob:none --sparse https://github.com/Rul1an/assay.git assay-corpus
cd assay-corpus
git checkout <commit named in #1840>
git sparse-checkout set --no-cone \
  '/docs/profiles/privileged-mcp-action/v0.md' \
  '/conformance/privileged-mcp-action-v0/**' \
  '!/conformance/privileged-mcp-action-v0/gen_vectors.py'
```

Run against the repository this yields 16 files — spec, corpus README, descriptor, manifest, 13 vectors. It **does not materialize either out-of-bounds surface in the working tree**; a sparse clone still carries the repository index and can fetch other blobs later, so it keeps the boundary rather than guaranteeing it.

## Smaller corrections

- The reason claim was too strong. Reasons are explicitly not part of the normative surface, so a reject for a different reason still agrees on `bundle_integrity` and `verdict`. It is **reviewer-visible rather than scored**. Corrected here and in #1840.
- In-repository links are relative, so a reader who obtained the spec at some commit follows them to the corpus at that same commit.
- The incident narrative about the old hostname is out of the profile header — it belongs to this change, not to the profile semantics.
- The repository README had no path to the profile, the corpus or the invitation, and now has one.

## What does not change

No vector, `MANIFEST.json`, or descriptor-record byte changes. The corpus digest method is `sha256 over concat(vector sha256)`, so this text sits outside it and **`a943120f` is stable** — checked before the first edit, because a documentation change that moved the digest would invalidate the target under reproduction.

## Verification, with the real toolchain

| Check | Result |
|---|---|
| `mkdocs build --strict` | succeeds |
| built `site/CNAME` | `docs.getassay.dev` |
| sitemap host | `https://docs.getassay.dev/` |
| built profile page | carries the corrected normative URI |
| `https://docs.getassay.dev/profiles/privileged-mcp-action/v0/` | `HTTP 200` |
| sparse checkout, executed | 16 files, neither out-of-bounds surface present |
| `scripts/ci/check-public-artifact-sanitization.py` | passed |
| `cargo test -p assay-cli --test conformance_privileged_mcp_action` | green |
| `git status` under `vectors/` and `MANIFEST.json` | zero changes |

## Deliberately not in this PR

Five source files still name the old host: two lint/SARIF rule URIs, a legacy policy path, and two docs pages. Those are **route hygiene**, not deployment persistence, and several of the routes need pages that do not exist yet (`/lint/`, `/packs/schema/`, and `/config/policies/` which is really `/reference/config/policies/`). They belong in a separate docs-contract change together with a `mkdocs build --strict` CI gate that asserts `site/CNAME` and the canonical pages exist.

## Post-merge

Re-pin #1840 to the squash-merge commit at **full 40-character SHA**, noting that this was a locator and documentation update and that corpus digest `a943120f…` did not change. Then enable **Enforce HTTPS** in GitHub Pages once the toggle is available — the certificate is active but HTTP currently returns `200` rather than redirecting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,24 @@ Repositories that compose with Assay's evidence layer:
 - [gateway-evidence-replay](https://github.com/Rul1an/gateway-evidence-replay) — deterministic offline replay verifier for gateway-path evidence bundles.
 - [RGE-Bench](https://github.com/rge-bench/rge-bench) — a neutral, externally reproduced conformance kit for evidence reviewability, maintained separately under its own machine-checked neutrality guard.
 
+## Open profile: privileged-mcp-action/v0
+
+[`privileged-mcp-action/v0`](docs/profiles/privileged-mcp-action/v0.md)
+is a composition and verification contract over evidence records that already exist: what a
+privileged MCP tool call decided, what was observed of its effect, and what stays unproven. It adds
+no new envelope and no aggregate verdict.
+
+It ships with a [13-vector conformance corpus](conformance/privileged-mcp-action-v0)
+(5 accept, 8 reject) whose digest is a **candidate**: it is not called reproduced until a non-author
+implementation derives the expected outcomes from the specification text alone.
+
+**That reproduction is open, and the invitation is real:
+[#1840](https://github.com/Rul1an/assay/issues/1840).** Any language, any stack. The invitation
+names the exact commit the current digest describes, and the
+[corpus README](conformance/privileged-mcp-action-v0/README.md) says which inputs are in bounds,
+which two implementation surfaces are not, and gives a sparse checkout that materialises only the
+inputs.
+
 ## Contributing
 
 ```bash
--- a/conformance/privileged-mcp-action-v0/README.md
+++ b/conformance/privileged-mcp-action-v0/README.md
@@ -6,14 +6,60 @@ Vectors for the [Privileged MCP Action Evidence Profile v0](../../docs/profiles/
 - `MANIFEST.json` declares each vector's expected `bundle_integrity`, profile `verdict`, and (for
   accepts) the full expected claim matrix. That triple is the **normative comparison surface**; the
   `first_failure_informative` codes are this generator's own vocabulary and informative only. An
-  independent implementation SHOULD report a free-form reason per reject in its own words, so a
-  wrong-reason reject cannot silently score as agreement.
+  independent implementation SHOULD report a free-form reason per reject in its own words. Reasons
+  are reviewer-visible but not scored; a different reason does not change agreement on the
+  normative surface.
 - `descriptor.json` is the machine-readable profile descriptor (record set, closed vocabularies,
   binding key, stages, report shape).
 - `gen_vectors.py` regenerates everything deterministically (standard library only; two runs are
   byte-identical). The corpus digest in the manifest changes on any vector edit.
 - The corpus digest is a **candidate** until an independent, non-author implementation reproduces
   the expected outcomes from the specification text alone.
+
+## Attempting the independent reproduction
+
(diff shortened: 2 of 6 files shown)


Rul1an:
## Merging with a documented exception to the same-head bot-review rule

Recording this because the rule was not met, and the record should say so rather than imply otherwise. This is the second such exception; it is deliberate each time, not a standing waiver.

**No fresh bot review exists on the merged head `934ab123`.** The only formal review object belongs to `898543e7`, two heads back:

| Reviewer | Head reviewed |
|---|---|
| copilot-pull-request-reviewer | `898543e7` |

**The green CodeRabbit context on `934ab123` is not evidence of a review.** Its description reads `Review rate limited`. That is CodeRabbit reporting it skipped the review and setting the context to success so as not to block. Counting it as bot coverage would count a skip as a pass.

**What the merge rests on:** same-head human review across four rounds, which raised and closed findings on the descriptor's dead normative URI, the inputs-only checkout claim, the reason-scoring overclaim, short SHAs, the `docs.getassay.dev` host correction, the deployment-persistence risk, and an internal contradiction in the corpus README. Full CI on `934ab123`: 30 checks, 0 failures. 0 unresolved threads. `CLEAN` / `MERGEABLE`.

**What was not checked:** no bot examined the last two heads (`ae1d8f0b`, `934ab123`), which carried the `docs.getassay.dev` correction, `docs/CNAME`, the `mkdocs.yml` `site_url` fix, the narrowed sparse-checkout claim, and the scoring-rule reconciliation. All were verified by hand, including a real `mkdocs build --strict` whose output `site/CNAME` reads `docs.getassay.dev`, and a sparse checkout executed against this repository yielding 16 files with neither out-of-bounds surface present.

**One thing to watch immediately after merge.** This PR touches `docs/**`, so it triggers `mkdocs gh-deploy --force`. That is the first deploy to run with `docs/CNAME` present, and the reason the file is in this PR at all. If the custom domain does not survive, it will be visible at once at https://docs.getassay.dev/profiles/privileged-mcp-action/v0/ and should be restored rather than left.

No vector, `MANIFEST.json`, or descriptor-record byte changes: corpus digest `a943120f` is unchanged.

PR #1873, opened 2026-07-27
fix(release): run MCP Registry publication inside the release transaction

### LINKED ISSUE #1870: release: make MCP Registry publication part of the release transaction
## Problem

The v3.35.0 release workflow created the GitHub release with `GITHUB_TOKEN`, but the separate `release.published` MCP Registry workflow did not start. This is expected GitHub behavior: events created with `GITHUB_TOKEN` do not trigger new workflow runs except `workflow_dispatch` and `repository_dispatch`. Publication had to be started manually in run 30229328824.

Official behavior: https://docs.github.com/en/actions/concepts/security/github_token#when-github_token-triggers-workflow-runs

## Desired behavior

A stable release should publish `server.json` to the official MCP Registry automatically, and the release automation should not report complete until that publication has a terminal result. Prereleases must remain excluded.

## Lean direction

Prefer a reusable `workflow_call` path invoked as a dependent job from `release.yml`, while retaining `workflow_dispatch` for idempotent recovery. A direct, explicit `workflow_dispatch` from the release workflow is acceptable if its terminal result is monitored and surfaced; merely firing and forgetting is not. Do not introduce a PAT when `workflow_call` or the documented `workflow_dispatch` exception is sufficient.

## Acceptance

- a stable release automatically runs the pinned publisher against its exact release tag;
- prereleases do not publish;
- publisher checksum, release asset validation and OIDC identity remain enforced;
- an already-published version is safely retryable through manual dispatch;
- the release record links the registry run and its terminal result;
- tests cover release, prerelease, workflow-call and manual-dispatch tag resolution.

Evidence: v3.35.0 release run 30228780553 succeeded, no downstream registry run appeared, and manual dispatch 30229328824 then published version 3.35.0 successfully.

Rul1an:
## Summary

- run the MCP Registry publisher as a `workflow_call` job inside `release.yml`, because releases created with `GITHUB_TOKEN` emit no `release.published` event that starts new workflow runs (this is what stranded v3.35.0 until manual dispatch 30229328824)
- gate the dependent job to stable tags on both the tag-push and dispatch paths; prereleases skip publication entirely
- keep OIDC identity, pinned publisher checksum, and `mcp-publisher validate` unchanged
- consult the registry before publishing: an already-published identical version is a clean no-op (idempotent retry), a same-version content mismatch fails closed instead of being treated as published
- confirm after publishing that the registry actually serves the exact content (`GET /v0/servers/{name}/versions/{version}`, verified live against the published 3.35.0 entry) — the terminal result is a registry read-back, not a publisher exit code
- record that terminal result and the run link on the GitHub release record via an idempotent marker line; failures are recorded too
- keep `workflow_dispatch` as the manual recovery path and the `release` trigger for releases published outside the automation
- move the publish concurrency group to job level, since workflow-level concurrency is not applied to called workflows

Closes #1870.

## Verification

- `bash scripts/ci/test-mcp-registry-release-txn.sh` — resolve (release/prerelease/dispatch/call/injection), registry check (absent/published/sha-mismatch/HTTP error), release-record idempotency, and structural workflow wiring incl. negative pins (no `contents: write` on the publish job, `always()` + skipped-guard on the record job); red before implementation, green after
- `pre-commit run --files …` incl. the new self-test hook — all passed
- `actionlint`, `zizmor` (no findings), `shellcheck` — clean

## Claim boundary

This makes publication part of the release run and verifiable from the release record. It does not retro-publish older versions, does not change what `server.json` contains, and the live end-to-end proof requires the next stable release to publish unattended.

## Summary by CodeRabbit

* **New Features**
  * Stable releases are now published to the official MCP Registry automatically.
  * Release notes now record the registry publication result and workflow link.
  * Repeated publication attempts safely skip versions already published with matching content.

* **Bug Fixes**
  * Added validation to prevent publishing mismatched versions or package contents.
  * Publication results are verified after publishing, with failures reported clearly.

* **Tests**
  * Added automated checks covering release routing, registry validation, workflow behavior, and result recording.

### DIFF
--- a/.github/workflows/mcp-registry-publish.yml
+++ b/.github/workflows/mcp-registry-publish.yml
@@ -11,6 +11,12 @@ name: MCP Registry Publish
 # remain outside the official registry so a newer RC or beta cannot displace the latest stable
 # server version.
 #
+# Recovery order after a failed release-run publication: prefer re-running the failed jobs of
+# the release run (`gh run rerun <run-id> --failed`) — the pre-check turns an already-published
+# version into a clean no-op AND the release-record marker is refreshed to the terminal result.
+# A direct workflow_dispatch here also publishes idempotently, but it cannot update the release
+# record, so a stale "failure" marker would remain on the release notes.
+#
 # GitHub OIDC proves the io.github.<owner> namespace belongs to this repository owner, so no stored
 # registry secret is needed.
 
@@ -43,8 +49,13 @@ env:
 jobs:
   publish:
     name: Publish to MCP registry
+    # CAUTION: under workflow_call the caller's event is push/workflow_dispatch,
+    # so this condition can never skip the called job. Keep it that way: if a
+    # future edit lets this job skip in call context, the caller job reports
+    # success with no publication (all-skipped reusable workflows read green).
     if: ${{ github.event_name != 'release' || github.event.release.prerelease == false }}
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     # Job-level on purpose: workflow-level concurrency is not applied when this
     # workflow runs as a workflow_call job, and the registry publish must stay
     # serialized across release runs and recovery dispatches.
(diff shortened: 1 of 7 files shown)


Rul1an:
Review record for this PR (head `107ed605`):

- CodeRabbit posted a green commit status with description **"Review rate limited"** and Copilot posted nothing — both are quota marker-records, not reviews, and do not count toward the quorum.
- Per the standing fallback rule, the quorum was closed by independent non-building agent reviews instead:
  1. general review of head `c7ec6a1e` (Actions/reusable-workflow semantics, security, test bite) — no blocking findings;
  2. adversarial review of head `c7ec6a1e` (false-green construction, integrity false-positives, recovery paths) — no blocking scenario;
  3. delta verification of `c7ec6a1e..107ed605` (the commit applying their convergent findings) — clean, with mutation-tested pins (packages-set comparison, if-condition, retry loop all demonstrably bite).
- Residual accepted risks, named: concurrency-queue displacement can cancel a pending publish leg when an operator dispatches during a release run (visible, recoverable via `gh run rerun --failed`); the stable-only guard exists in three places and a hypothetical new prerelease suffix would fail the run loudly rather than skip.

Final acceptance remains the next stable release publishing unattended with the terminal result linked on the release record.

Issue #1869, opened 2026-07-27
ci: align delegated proof refresh with the required lane context


Rul1an:
## Problem

The delegated-proof refresh currently posts commit status `lane-check/proof` on the PR head, while branch protection requires the Actions check-run `lane-check`. A `workflow_run` execution is attached to the default-branch SHA, so its successful `lane-check` job does not refresh the required check on the PR head.

Observed on release PR #1868: the same-head delegated proof succeeded in run 30227602965 and the refresh posted `lane-check/proof`, but branch protection remained blocked until `lane-check` was manually rerun on the final PR head.

## Desired behavior

A successful same-head delegated proof should satisfy the one required lane context automatically, without a manual rerun and without accepting proof from another SHA.

## Lean direction

Migrate branch protection to require `lane-check/proof`, which the helper can post directly on the PR head, and keep the Actions `lane-check` check-run informational during the transition. Exercise the migration on a gated test PR before removing the old required context. An equivalent Checks-API implementation is acceptable if it creates or updates the required check on the exact PR head.

## Acceptance

- a gated PR starts blocked before proof;
- a delegated proof for the final head updates the required context on that same SHA;
- no manual rerun is needed;
- stale proof remains rejected;
- skipped or unassociated `workflow_run` events never read clean;
- self-tests pin the required context name and exact-head behavior.

GitHub requires protected checks to pass on the latest commit SHA: https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks

Rul1an:
Migration checklist additions from the adversarial review of the code slice (PR #1874):

1. When swapping the required context to `lane-check/proof`, pin the required status's **source** to the GitHub Actions app in branch protection if the UI/API supports it — classic commit statuses carry no author binding, so any write-access PAT can post the context otherwise (pre-existing property, becomes load-bearing after the swap).
2. Consider retiring the **legacy fallback proof path** (`run URL + head SHA + free-text gate line`) as part of the migration: `recorded_gate_ok` trusts an author-editable `gate: all` line and never verifies which gates the delegated run actually executed. The attested proof-pack path does not have this hole.
3. Operational rule, now also encoded in the workflow comment and warning text: a privileged refresh dispatch must run from a trusted ref; dispatching on a ref carrying untrusted changes executes that ref's workflow file. The helper itself is now always checked out from `main` on dispatch (PR #1874).
[...]

PR #1874, opened 2026-07-27
fix(ci): fail closed when the required lane-check/proof status cannot land

Rul1an:
## Summary

Code slice of #1869 (the branch-protection swap follows separately, after an empirical gated test PR, per the issue's migration order).

- `lane-check/proof` is the context branch protection will require; the helper already posts it on the exact PR head, which is what lets a same-head delegated proof satisfy the gate without a manual rerun
- `maybe_status` previously swallowed every posting failure with a warning — a green run whose required status never landed. It now propagates all failures except a 401/403 permission refusal
- 401/403 stays a (clearly marked) warning because that is the documented read-only-token path: Dependabot and fork contexts can never post the status themselves, branch protection stays safely blocked, and only a privileged `workflow_dispatch` refresh can land it
- self-tests now pin: the required context name, the fail-closed matrix (transient error → raise, 403 → warn, 5xx → raise, no `--status` → no call), and the lane workflow's exact-head + never-read-clean guards (`workflow_run.event == 'workflow_dispatch'` gate, unassociated-run no-op, `EXPECTED_HEAD_SHA` double check, `--status` wiring)

Refs #1869.

## Verification

- `python3 scripts/ci/assay_runner_lane_check.py --self-test` — red before the fix on the fail-closed case, green after
- `pre-commit run --files scripts/ci/assay_runner_lane_check.py` — passed (includes the lane-check self-test hook)

## Migration plan (after merge, tracked on #1869)

1. Add `lane-check/proof` to the required contexts alongside `lane-check` (both required during transition).
2. Open a gated test PR: verify it starts blocked, dispatch a delegated proof for the final head, and confirm the required context updates on that SHA with no manual rerun and stale proof still rejected.
3. Only then drop `lane-check` from the required set, keeping the Actions check-run informational.

## Claim boundary

This PR changes error semantics and pins contracts; it does not change which contexts branch protection requires, and it does not claim the migration works end-to-end until the gated test PR proves it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lane-check reliability by allowing only expected permission refusals during pull requests while failing safely on other status-reporting errors.
  * Added clearer handling for rate-limit responses.

* **CI Improvements**
  * Changes to the lane-check workflow now trigger the kernel matrix checks.
  * Trusted helper code is resolved from the appropriate workflow reference to improve validation safety.

* **Tests**
  * Added coverage for status-reporting failures, rate limits, permission refusals, and workflow configuration requirements.

### DIFF
--- a/.github/workflows/assay-runner-lane-check.yml
+++ b/.github/workflows/assay-runner-lane-check.yml
@@ -31,11 +31,17 @@ jobs:
       pull-requests: write
       statuses: write
     steps:
-      - name: Checkout base helper
+      - name: Checkout trusted helper
         uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
         with:
           persist-credentials: false
-          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.sha }}
+          # pull_request: the PR base, so a PR can never run its own edited
+          #   gatekeeper. workflow_dispatch: main, because the privileged
+          #   refresh holds statuses: write and must never execute helper code
+          #   from the dispatched ref (which a PR author controls); the head
+          #   SHA is data, not code. workflow_run: github.sha is the default
+          #   branch head, already trusted.
+          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || (github.event_name == 'workflow_dispatch' && 'main' || github.sha) }}
 
       - name: Verify lane-check helper exists
         shell: bash
--- a/.github/workflows/kernel-matrix.yml
+++ b/.github/workflows/kernel-matrix.yml
@@ -18,6 +18,10 @@ on:
       - "crates/assay-monitor/**"
       - "scripts/**"
       - ".github/workflows/kernel-matrix.yml"
+      # The lane self-test pins guards inside this workflow file; run the
+      # head-side pre-commit leg when it changes, because the lane workflow
+      # itself checks out the PR base and would validate the base copy.
+      - ".github/workflows/assay-runner-lane-check.yml"
       # Note: Cargo.toml/Cargo.lock removed - check-ebpf-changes job handles skip logic
   push:
     branches: [ "main", "debug/**" ]
--- a/scripts/ci/assay_runner_lane_check.py
+++ b/scripts/ci/assay_runner_lane_check.py
@@ -1487,14 +1487,24 @@ def maybe_status(
         # `lane-check/proof` is the required branch-protection context
         # (issue #1869). A permission refusal is the one tolerated miss: a
         # read-only token (Dependabot/fork contexts) can never post it, the
-        # head stays unproven, and only a privileged dispatch refreshes it.
-        # Everything else must fail the run rather than leave a green job
-        # whose required status silently never landed.
-        if exc.code in (401, 403):
+        # head stays unproven, and only a privileged dispatch from a trusted
+        # ref refreshes it. Everything else must fail the run rather than
+        # leave a green job whose required status silently never landed.
+        # GitHub also answers rate limiting with 403; that is transient for a
+        # write-capable token, so it propagates instead of masquerading as a
+        # permission refusal with the wrong remedy.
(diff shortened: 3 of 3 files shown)


Rul1an:
Review record for this PR (final head `f06fb19f`):

- CodeRabbit's green status carries description **"Review rate limited"** and Copilot posted nothing — quota marker-records, not reviews; neither counts toward the quorum.
- Quorum closed by two independent non-building agent reviews of head `11f87cf2` (one correctness lens, one adversarial), plus a delta verification of `11f87cf2..f06fb19f` now running. No blocking findings in the Python change itself; the delta commit addresses:
  - **HIGH (adversarial):** the privileged `workflow_dispatch` refresh executed helper code from the dispatched ref while holding `statuses: write`, and the head-binding guard forces dispatching on the PR head ref. The dispatch path now checks out the helper from `main`; the head SHA is data, not code. The dispatched ref's own workflow file still runs, so dispatching on a ref carrying untrusted changes remains forbidden (warning text and workflow comment both say so).
  - **MEDIUM (both reviewers):** 403 rate limiting (Retry-After / exhausted X-RateLimit-Remaining) now propagates instead of masquerading as a permission refusal with the wrong remedy.
  - **MEDIUM (both reviewers):** kernel-matrix lints on lane-workflow changes, closing the head-side gap where a workflow-only PR validated the base copy of the guards it was removing; the pin docstring now states the enforcement layers accurately.

Accepted residuals, named: the status POST is single-attempt (a transient blip fails the run visibly; a re-run recovers — fail-closed by design); a revert-force-push to a previously evaluated SHA briefly reuses that SHA's old status until the synchronize re-run overwrites it (pre-existing, per-SHA semantics); classic commit statuses have no author binding — write-access PATs can post any context (pre-existing; migration will pin the required status source to the Actions app if branch protection supports it).

Rul1an:
Review record update (new head `196aba09`):

Cursor reviewed head `f06fb19f` and found one **blocking** defect: the 401/403 exemption in `maybe_status` applied to every event context, including the privileged refresh paths (`workflow_dispatch`, `workflow_run`) that exist to post the required status and always hold `statuses: write` — a permission misconfiguration there degraded to a warned-but-green run, recreating the eliminated failure class inside its own exemption. Cursor's review is comment-borne (their tooling cannot submit a review record); it counts toward the quorum as an artifact on `f06fb19f`, per the per-artifact counting rule.

Head `196aba09` addresses all three findings:
- the exemption is now bound to `GITHUB_EVENT_NAME == pull_request` — the one context whose token is read-only by design; privileged and unknown contexts propagate (red-first: self-test failed on the dispatch-403 case before the fix)
- the workflow pin binds `statuses: write` to the `lane-check` job under the workflow-level default deny, instead of matching anywhere in the file
- the pin docstring now names which tree each event path validates (pull_request → PR base, workflow_dispatch → main, workflow_run → default-branch head)

Known accepted residual: a same-repo `pull_request` run with a genuinely misconfigured token still warns-and-greens (the event context cannot distinguish fork from same-repo without the event payload); post-migration that PR is visibly blocked on the missing required status, and the remedy is the privileged dispatch. Named here rather than silently absorbed.

Cursor's #1873 spot-check finding (skipped publication leaves no marker; absence must never read clean) is fixed in follow-up #1875.

Rul1an:
Head update `196aba09` → `dad1ceb0`: merge of `origin/main` (the #1873 squash) to clear `BEHIND`. Verified content-identical for this PR's subject: `git diff 196aba09..dad1ceb0 -- scripts/ci/assay_runner_lane_check.py .github/workflows/assay-runner-lane-check.yml .github/workflows/kernel-matrix.yml` is empty, and the self-test passes unchanged on the new head.

Per Cursor's request this is recorded explicitly rather than silently passed: **Cursor's confirmation on `196aba09` describes the exact content that merges at `dad1ceb0`** — the new head differs only by upstream commits that do not touch the reviewed files. The contract gap this exposes (a new head from an upstream update versus a new head with changed content) is filed on #1872.

With Cursor's substantive go recorded, enabling auto-merge.

PR #1875, opened 2026-07-27
fix(release): record a skipped registry publication explicitly

Rul1an:
## Summary

Follow-up to #1873, from Cursor's spot-check of the merged release transaction.

The record job gated on `needs.publish-mcp-registry.result != 'skipped'`, so a skipped publication — prerelease exclusion, or any future skip path — left **no marker at all** on the release record. Absence was indistinguishable from a release that predates the transaction, which is exactly the ambiguity the transaction exists to remove (and AGENTS.md's rule: absence never reads clean).

- the record job now gates on `needs.release.result == 'success'`: it runs exactly when a release record exists to annotate
- a prerelease's skipped publication writes an explicit marker: `MCP Registry publication: skipped (prerelease; stable releases only) (run link)`
- a failed release still records nothing — there is no release record to annotate and its run is already red
- `record-mcp-registry-result.sh` accepts `skipped` as a terminal value with its own label; unknown values stay refused

## Verification

- `bash scripts/ci/test-mcp-registry-release-txn.sh` — red before (skipped-marker case + new wiring pin), green after
- `pre-commit run --files …` incl. the transaction self-test hook — passed

## Claim boundary

The live proof of the skipped-marker path requires the next prerelease tag; the success-marker path requires the next stable release. Neither is claimed proven by this PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -730,9 +730,10 @@ jobs:
     needs: [release-contract, release, publish-mcp-registry]
     # always(): a failed publication must still be recorded on the release.
     # Gate on the release job, not the publish result: a prerelease's skipped
-    # publication gets an explicit marker (absence must never read clean),
-    # while a failed release has no release record to annotate and its run is
-    # already red.
+    # publication gets an explicit marker (absence must never read clean).
+    # A failed release records nothing — its run is already red, at most a
+    # stranded partial release exists, and the recovery re-run re-stamps the
+    # marker after the release step completes.
     if: ${{ always() && needs.release.result == 'success' }}
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -727,10 +727,13 @@ jobs:
 
   record-mcp-registry-result:
     name: Record MCP Registry result
-    needs: [release-contract, publish-mcp-registry]
-    # always(): a failed publication must still be recorded on the release;
-    # skipped (prerelease) publishes record nothing.
-    if: ${{ always() && needs.publish-mcp-registry.result != 'skipped' }}
+    needs: [release-contract, release, publish-mcp-registry]
+    # always(): a failed publication must still be recorded on the release.
+    # Gate on the release job, not the publish result: a prerelease's skipped
+    # publication gets an explicit marker (absence must never read clean),
+    # while a failed release has no release record to annotate and its run is
+    # already red.
+    if: ${{ always() && needs.release.result == 'success' }}
     runs-on: ubuntu-latest
     permissions:
       contents: write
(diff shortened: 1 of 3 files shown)


Rul1an:
Review record (final head `981a3ef3`):

- CodeRabbit/Copilot remain quota-limited (marker-records only); quorum closed per the fallback rule.
- Adversarial agent review of head `5748da3b`: **no blocking findings**. The full `(release.result × publish.result)` state matrix was enumerated across every event path — a green release always gets a marker (success/failure/cancelled/skipped), a failed or skipped release records nothing and its run is visibly non-green. Both halves of the change were revert-tested independently against the contract test.
- Head `981a3ef3` applies the two Low findings: the exact `needs` line of the record job is now pinned (dropping `release` from it would silently disable recording, unseen by the previous pins), and the job comment no longer overclaims that a failed release leaves no record — a partial release can be stranded (the v3.33.0 incident); the recovery re-run re-stamps the marker.
- Named informational residual: the `skipped (prerelease; stable releases only)` label is truthful while prerelease exclusion is the publish job's sole skip condition; the wiring test pins that condition's presence but not its exclusivity.

Origin of this fix: Cursor's spot-check of merged #1873. Live proof of the skipped-marker path arrives with the next prerelease tag; the success path with the next stable release.

Rul1an:
Review record update (final head `fe485045`):

Cursor held the merge on two points; both are implemented and independently verified on the exact head:

- **P3 (label observes, not asserts):** the skipped annotation now derives from `VERSION` — `-rc`/`-beta` names the prerelease reason, any other skip records a bare `skipped`. The conditional mirrors the publish gate's own substring semantics, and `VERSION` is bound to the same release-contract output the gate acts on. Mutation-verified in both directions (always-claim → red on the stable case; always-bare → red on the prerelease cases).
- **Needs pin parsed, not string-matched:** a YAML-identical reorder (including irregular whitespace) stays green; dropping `release` fails with the intended message; deleting the needs line fails closed. Cursor's brittleness objection and the agent's silent-disable concern are now both covered by the same pin.

Marker idempotency confirmed label-agnostic (replacement keys on the HTML-comment marker, not the label text). Delta touches exactly the two files. Per Cursor's stated condition — target behavior pre-described, agent verification on the new head suffices — enabling auto-merge.

Rul1an:
Head update `fe485045` → `3c811751`: merge of `origin/main` (the #1874 squash) to clear `BEHIND`. Content-identical for this PR's subject: `git diff fe485045..3c811751 -- .github/workflows/release.yml scripts/ci/record-mcp-registry-result.sh scripts/ci/test-mcp-registry-release-txn.sh` is empty and the contract test passes unchanged. The verified review record on `fe485045` describes the exact content that merges at `3c811751` (equivalence rule as proposed on #1872). Auto-merge remains enabled.

PR #1880, opened 2026-07-27
docs(agents): repair #1865 regressions and harden the quorum against non-reviews

Rul1an:
Implements #1872 in two commits:

**Commit 1** (`43e8099d`, cherry-picked clean from the prepared `d4513a09`, original authorship and verification preserved):
- regression: the CI-integrity rule ("never weaken, bypass, or relabel a required check") moves back above `### Measurement provenance`, out of the attribution subsection that had captured it
- regression: the staging rule becomes the test ("could the command pick up a file you did not touch?") with the enumeration explicitly illustrative; the concrete-path example becomes a placeholder
- new rule: a review record that says it did not review is not a review (Copilot "unable to review — quota limit", CodeRabbit `pass` + "review rate limited"), and a merge-of-main is a new head like any other
- clarification: measure-then-push means commit first and report that SHA
- the pin rule describes the defect instead of citing an exemplar file

**Commit 2** (`2ba1f094`, the three refinements from last night's sessions, resolving the tension between the issue text and the recurring-tension comment):
- **equivalence revalidation**: a new head still invalidates, but a review is revalidated by a recorded mechanical check — empty diff over the PR's files, only main-commits added. Without the record, it does not carry. (Observed three times in one night: #1871, #1873, #1874; under `strict: true` every main advance forces this cycle.)
- **immediate substitution on a declared limit** (explicit mandate from Rul1an): a limit-shaped record is itself the observation of unavailability — an independent adversarial agent review satisfies that slot at once, the 30-minute clock covers silence only, and the record must document which reviewer was limited, which agent stood in, and on which head
- **per-artifact counting**: reviews are artifacts bound to an exact head, not entries in GitHub's review list; comment-borne reviews (observed: Cursor) count through a comment bound to the reviewed SHA

Docs-only; the `pre-commit==4.4.0` duplicate in `kernel-matrix.yml:63` stays a separate workflow fix per the issue.

## Verification

On `2ba1f094` in worktree `assay-1872` (branched from `daefdadf`): pre-commit over `AGENTS.md` passes; `git diff --check` clean; longest line ≤ 114; heading order `## Verification` → CI-integrity paragraph → `### Measurement provenance` confirmed mechanically. One file, staged by explicit path.

Closes #1872.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated contribution and review guidance with clearer staging and tool version requirements.
[...]

PR #1686, opened 2026-06-14
fix(deps): migrate PyO3 0.24 → 0.29, clear RUSTSEC-2026-0176/0177

Rul1an:
Closes #1651.

## Why now
The migration was blocked on `pythonize` lagging pyo3 0.29. That's resolved: `pythonize 0.29.0` (and `pyo3 0.29.0`) are on crates.io now.

## Change
- `assay-python-sdk`: `pyo3` and `pythonize` `0.24` → `0.29`.
- One source change: `PyObject` → `Py<PyAny>` (the `PyObject` alias was removed in 0.29). The bindings already used the Bound API, so nothing else moved.
- Drop the temporary `RUSTSEC-2026-0176` and `RUSTSEC-2026-0177` ignores from `deny.toml` and the `cargo audit` step. Both are pyo3 advisories patched in 0.29; `RUSTSEC-2023-0071` (rsa) stays.

Neither advisory was reachable in assay today (the SDK iterates with `.enumerate()`, never `.nth()`, and never calls `PyCFunction::new_closure`), so this is hygiene plus clearing the ignores, not an active exposure.

## Verified locally
- `cargo check` / `cargo build` (cdylib) / `cargo clippy --all-targets -D warnings`
- `cargo deny check advisories` → `advisories ok` with the two ignores removed
- 13/13 `assay-python-sdk` pytest cases pass against a maturin-built 0.29 extension

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Updated Rust dependencies for improved compatibility and performance.
  * Removed exemptions for two security advisories in CI configuration to enhance security oversight.
  * Updated Python binding implementation to use improved type handling.

PR #1881, opened 2026-07-27
chore(deps): drop the two quick-xml ignores, their removal condition is met

Rul1an:
First slice of the post-audit sequencing. Placed first deliberately: it is the only item whose correctness condition is **written down in the repo itself and demonstrably satisfied** — a reviewer has to believe nothing, only run `cargo deny check advisories`.

`deny.toml:31-34` stated the condition verbatim:

> object_store pins quick-xml ^0.40.1, while the fixed quick-xml is >=0.41.0. **Remove both ignores when object_store releases a version that allows quick-xml >=0.41.0.**

`object_store` 0.14.1 allows it, `main` already resolves it, and `quick-xml` resolves to a **single** 0.41.0 entry — the fixed version. So RUSTSEC-2026-0194/0195 no longer match any crate in the tree, the block violated its own heading rule ("Only list advisories that currently match a crate in the tree"), and its comment ("transitive via object_store 0.14.0") was factually wrong at this SHA.

## Three places, not two

The list had to be kept in lockstep by hand across `deny.toml`, the `cargo audit` step in `ci.yml`, and the **pre-push hook in `.pre-commit-config.yaml`** — the third one was easy to miss and had drifted furthest: it carried five ignores of which **four** were dead. Besides the quick-xml pair, the pyo3 pair (RUSTSEC-2026-0176/0177) had already been retired by the pyo3 0.29 migration (#1651) and was simply left behind.

All three paths now pass exactly `RUSTSEC-2023-0071` (rsa Marvin Attack, reached via `jsonwebtoken → rsa 0.9.10`), which still matches and has no fixed release.

The drift has one cause worth naming: entries were added on a trigger and never removed on one.

## Verification

Worktree `assay-advisory`, branched from `412687b5`, local cargo-deny/cargo-audit:

- `cargo deny check advisories` → `advisories ok`
- `cargo audit --ignore RUSTSEC-2023-0071` → exit 0; none of RUSTSEC-2026-0176/0177/0194/0195 appears in the output
- pre-commit (incl. its own cargo-deny hook) → passed on both commits

## Claim boundary

This removes four dead ignores and restores lockstep. It does **not** add a mechanism that forces re-litigation by calendar — `deny.toml` cannot express expiry natively, and a test that fails on a named date is a separate change.

Origin: independent audit, corroborated against the recorded unblock trigger, which is what surfaced the third location.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,3 +1,5 @@
-# Align with ci.yml and deny.toml: allow RUSTSEC-2023-0071 (rsa, no fix)
+# Align with the three peer locations that carry this list: deny.toml,
+# .github/workflows/ci.yml and the pre-push cargo-audit hook in .pre-commit-config.yaml.
+# Allow RUSTSEC-2023-0071 (rsa Marvin Attack, via jsonwebtoken; no fixed release).
 [advisories]
 ignore = ["RUSTSEC-2023-0071"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,11 +165,7 @@ jobs:
         run: |
           rm -rf ~/.cargo/advisory-db
           # RUSTSEC-2023-0071 (rsa Marvin Attack): no fixed upgrade; assay-mcp-server JWT (see deny.toml)
-          # RUSTSEC-2026-0194/0195 (quick-xml): transitive via object_store 0.14.0; no compatible upgrade yet (see deny.toml)
-          cargo audit \
-            --ignore RUSTSEC-2023-0071 \
-            --ignore RUSTSEC-2026-0194 \
-            --ignore RUSTSEC-2026-0195
+          cargo audit --ignore RUSTSEC-2023-0071
 
   clippy:
     name: Clippy (deny warnings)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -127,11 +127,13 @@ repos:
 
       - id: cargo-audit
         name: cargo audit (RustSec)
-        # Keep the ignore list in lockstep with .github/workflows/ci.yml and deny.toml: these
-        # advisories are triaged as not reachable (rsa Marvin Attack; pyo3 0.24 nth/nth_back and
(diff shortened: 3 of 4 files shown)


Rul1an:
Review record (head `71264924`):

- CodeRabbit and Copilot are quota-limited on this PR; per the rule merged in #1880, the substitution is documented: **both bots limited, an independent adversarial agent reviewed instead, on head `b9ec8af1`**.
- **Recommendation: merge.** No Critical/High/Medium findings introduced. The removal was verified four independent ways: `cargo audit --ignore RUSTSEC-2023-0071` exit 0, `cargo deny check advisories` ok, a single `quick-xml 0.41.0` and a single `pyo3 0.29.0` family in the lock (both at-or-past the advisory-DB patched floors), and `cargo tree --all-features` ruling out a feature-gated path — with `git status` clean afterwards, so no hidden resolution exists. CI's `Dependency Security` job passed on the same head.
- Both comment attributions were checked against lock history rather than accepted: the quick-xml pair was retired by `6570811c` (#1835, `object_store` 0.14.0→0.14.1 and `quick-xml` 0.40.1→0.41.0 in one commit), the pyo3 pair by `fa09a24d` (#1686, closing the #1651 tracking issue). The review also confirmed the drift's literal origin: **`fa09a24d` removed the pyo3 ids from `deny.toml` and `ci.yml` but not from `.pre-commit-config.yaml`** — the ids then sat dead in the hook for about six weeks.
- **A fourth carrier exists**: `.cargo/audit.toml`. It already held only `RUSTSEC-2023-0071`, so this PR incidentally achieves four-way lockstep. Head `71264924` applies the review's nit by naming all peers in both lockstep comments — an incomplete peer list is exactly the mechanism that produced the drift being fixed here.
- The second nit (the `--ignore` flag is now redundant with `.cargo/audit.toml`) is deliberately **not** applied: the explicit flag is better for legibility and costs nothing.
- Pre-existing gap found while answering the CI-parity question, filed separately as #1882 rather than widened into this PR: `fuzz/Cargo.lock` is audited by nothing and currently carries six advisories.

Enabling auto-merge.

PR #1883, opened 2026-07-27
slice 2: repair the untrue integrity sentences and guard the required-context set

Rul1an:
## ADR Boundary Slice

- Canonical ledger: #1866
- Slice: 2 (the untrue sentences, `bundle_id` MUST, ruleset trap)
- Builder: Cursor
- Reviewers: adversarial pass on `a3491338` — merge-with-followup, no Critical/High
- Final head SHA: `4246a355` — every measurement below is stated against this SHA, on `main` at `a867173b` (slice 1 merged)
- ADR invariants affected: ADR-039 non-claims; ADR-043 §"stop list binds emitted artifacts"

## Behavior

Three changes, one of them observable.

**Observable.** The verifier gains check 14: `manifest.bundle_id` must equal the chain root. The profile already made this normative — `docs/profiles/privileged-mcp-action/v0.md:126` says "the manifest's `run_root` and `bundle_id` MUST both equal it" — and nothing enforced it. A bundle carrying an all-zeros `bundle_id` beside an intact `run_root` verified clean. New code `ContractBundleIdMismatch`, class `Contract`.

Ordered **after** the chain check: a mutated `run_root` stays `IntegrityRunRootMismatch`, and the new code fires only when the chain is sound and its second copy in the manifest disagrees. The comparison is against `computed_run_root`, not `m.run_root` — identical today because check 13 has already equated them, but it stays sound if the two are ever reordered rather than being sound only because of what runs before it.

**Prose, no behaviour.** ADR-039 and the `attestation.rs` module doc said an attestation binds "who-said-it and the content". True of the semantic event chain, untrue of the artifact — and load-bearing, because `statement_from_manifest` puts `manifest.run_root` in the in-toto subject digest while `run_root` chains content hashes covering exactly `{specversion, type, datacontenttype, subject?, data}`.

**Config.** The checked-in ruleset required the retired bare `lane-check` and the proposal-only `public-artifact-sanitization`, and did not require `lane-check/proof`. Importing it would have dropped the live proof gate and promoted a proposal in one action, while `CI-CONTRACT.md` described it as mirroring live protection.

## Non-Claims

Explicitly **not** claimed: check 14 does not bind a bundle to its identity. A consistent rewrite of `run_id` and every event id still verifies, because the chain root excludes identity. Confirmed by the reviewer against this branch. That is the separate ADR on #1866, and widening `content_hash` is the wrong repair — it breaks the deterministic re-export the profile makes normative.

### RED

```
$ cargo test -p assay-evidence --test verifier_fail_closed_properties a_bundle_id_that_disagrees
test a_bundle_id_that_disagrees_with_the_chain_root_is_rejected ... FAILED

panicked at verifier_fail_closed_properties.rs:155:
a manifest whose bundle_id does not equal its chain root must not verify: ()
```

`must not verify: ()` is the `Ok(())` arm: the tampered bundle verified.

Negative control for the config guard, with the stale value pinned back into the ruleset:

```
$ python3 scripts/ci/check-required-contexts.py
required-contexts=failed
  CI-CONTRACT.md: missing ['lane-check'], unexpected ['lane-check/proof']
  docs/BRANCH-PROTECTION-SETUP.md contexts array #1: missing ['lane-check'] ...
  docs/BRANCH-PROTECTION-SETUP.md contexts array #2: missing ['lane-check'] ...
```

### GREEN

| Command | Result |
|---|---|
| `cargo test -p assay-evidence` | 35 lanes ok, 0 failed |
| `cargo test --locked --workspace --exclude assay-ebpf --exclude assay-it -- --skip test_no_passthrough_e2e` | 194 lanes ok, 0 failed |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo fmt --all --check` | clean |
| `check-required-contexts.py --self-test` / plain | passed / passed |
| `check-claims-boundary.py` | passed, finding_count=0 |
| all 13 conformance vectors through `assay evidence verify` | outcomes unchanged |

**Two fixtures broke, and that is the finding.** `test_crlf_bom_tolerance` and `test_reject_extra_file` hand-built manifests with `bundle_id: "test"` beside real chain roots. They had never satisfied the profile; they passed because nothing checked. `test_reject_extra_file` is the sharper one: its extra member is appended after `events.ndjson`, so the events block completes first, and with the placeholder the bundle is rejected on the bundle_id contract while the assertion matches `"Unexpected file"` — verified by restoring it: RED at `verify_strict_test.rs:568`.

Every other bundle in the tree already held the invariant: the thirteen conformance vectors and the fuzz corpus, checked individually.

## Review round (findings on `a3491338`)

All in-PR items are fixed on `4246a355`. M-2 is deferred to the ADR slice as proposed.

**M-1 — the sixth instance, inside my own repair.** `CI-CONTRACT.md:56-59` still named the pre-#1878 live set and told the reader to reconcile the runbook to it, 385 lines above the block this PR corrected. The guard cannot see it: prose outside the anchored list, exactly the scope limit its docstring documents. Rather than widen the guard, the second enumeration is gone — the entry now points at section 6, the only place that names the set. A value stated once cannot drift from itself.

**L-1 — an untrue sentence about untrue sentences.** My comment said the placeholder fixture "would pass while proving nothing". It goes red. Corrected, and verified by restoring the placeholder rather than reasoning about it.

**L-2** — the neighbouring comment claimed the whole manifest was conformant while its `run_root` is a single event's content hash. Now says what is true and why it does not matter there.

**L-3** — check 14 compares against `computed_run_root`.

**L-4** — both boundary notes restated the exclusion list partially, missing trace context and `policy_id`. They now state the inclusion set, exhaustive by construction, and point at `crypto/id.rs` rather than keeping a second copy true.

**L-5 — in the guard itself.** It scanned forward past the anchor until it found any bullet list, so deleting the list under the anchor silently promoted the next paragraph — the silent pass I claimed the anchor design excluded. Fixed, and the self-test now distinguishes a missing list from a mismatch instead of accepting "something went red". The first version of that self-test case proved nothing: it stripped every bullet in the file, so both the old and new parser reported correctly. With later lists left in place the old parser fails with "located a set it should not have been able to find".

**L-6** — "five documents" was a number I had not counted. Replaced with the two sources that can be pointed at.

## Provenance notes

Two things verified rather than taken on report, because both had been mis-cited earlier in this programme:

- `conformance/.../scripts/pack_format.py:121-176` exists and contains zero `run_root` references. Running `rewrite_bundle_stream_identity` over every vector gives **13/13 sharing their source root**. `gen_vectors.py` does recompute the root, but it is the generator, not the rewriter.
- The bad-101/ok-001 root collision inside the vector set is a *declared* one: `bad-101` differs in `data`, so its stored root is a copied stale value, and the bundle is rejected on the file digest. Not an instance of this defect.

Live branch protection could not be read from this environment (403 on the protection API), so the required set is taken from the freshest in-repo observation: PR #1878, "observed live on 2026-07-27".

## Mechanism, not another note

`scripts/ci/check-required-contexts.py` treats the ruleset as source of truth — it is the one that can be applied — and fails when the CI contract's anchored list or either runbook `"contexts"` array disagrees. A missing anchor, or a missing list under a present anchor, fails rather than passing silently.

This surfaced that **PR #1878's own repair was incomplete**: it fixed three documents and missed both `CI-CONTRACT.md` and the troubleshooting section of `docs/BRANCH-PROTECTION-SETUP.md`. Fifth instance of the incomplete-peer-list pattern slice 1 traced through the advisory ignores; M-1 above is the sixth, in this PR.

It runs in the claims-boundary workflow, so it is visible on any PR touching these paths but **not merge-blocking**. Making it required is coupled to the open finding that the `CI` gate ignores its own variables, which is a separate slice.

## Review Quorum

- [x] One non-building agent review (adversarial pass on `a3491338`, merge-with-followup)
- [x] Every actionable finding is fixed or has a technical disposition (M-2 deferred to the ADR slice)

## Summary by CodeRabbit

* **Bug Fixes**
  * Bundle verification now rejects manifests whose bundle ID does not match the computed run root.
  * Updated verification fixtures to correctly exercise specific failure scenarios.

* **CI**
  * Required branch protection checks now use `lane-check/proof`.
  * Added automated detection to identify inconsistencies in required-check configurations.

* **Documentation**
  * Clarified attestation subject semantics, verification boundaries, and branch-protection troubleshooting guidance.

### DIFF
--- a/.github/rulesets/main-required-ci-contexts.json
+++ b/.github/rulesets/main-required-ci-contexts.json
@@ -22,15 +22,11 @@
             "integration_id": 15368
           },
           {
-            "context": "lane-check",
+            "context": "lane-check/proof",
             "integration_id": 15368
           },
           {
             "context": "host-capability-check",
-            "integration_id": 15368
-          },
-          {
-            "context": "public-artifact-sanitization",
             "integration_id": 15368
           }
         ]
--- a/.github/workflows/claims-boundary.yml
+++ b/.github/workflows/claims-boundary.yml
@@ -10,6 +10,8 @@ on:
       - "assay-action/action.yml"
       - "assay-action/action.yaml"
       - "scripts/ci/check-claims-boundary.py"
+      - "scripts/ci/check-required-contexts.py"
+      - ".github/rulesets/**"
       - ".github/workflows/claims-boundary.yml"
   workflow_dispatch:
 
@@ -32,3 +34,9 @@ jobs:
 
       - name: Scan public prose
         run: python3 scripts/ci/check-claims-boundary.py
+
+      - name: Self-test required-contexts guard
+        run: python3 scripts/ci/check-required-contexts.py --self-test
+
+      - name: Check the required-context set agrees across every location
+        run: python3 scripts/ci/check-required-contexts.py
--- a/CI-CONTRACT.md
+++ b/CI-CONTRACT.md
@@ -53,10 +53,11 @@ Repository state observed on 2026-06-11:
     skips;
   - kernel/eBPF gating that avoids self-hosted runner work unless relevant;
   - delegated runner lane proof for selected runner-sensitive PRs.
-- Live branch protection on `main`, observed through GitHub on 2026-06-11,
-  requires `CI`, `lane-check`, and `host-capability-check` with strict
-  up-to-date status checks enabled. `docs/BRANCH-PROTECTION-SETUP.md` should
-  stay reconciled with that live state.
+- Live branch protection on `main` with strict up-to-date status checks enabled.
+  The required contexts are named once, in section 6 "Required Context Names";
+  this entry deliberately does not repeat them. It used to, and the copy went
+  stale when the set changed -- naming a value twice in one document is the
+  drift, not the cure.
 - Existing public docs and evidence artifacts include guides, references,
   receipt schemas, MCP proxy docs, runner fixtures, experiment reports, JUnit,
   SARIF, Trust Basis output, and compressed evidence examples.
--- a/CI-CONTRACT.md
+++ b/CI-CONTRACT.md
@@ -441,8 +441,12 @@ Before making any branch-protection changes:
 Currently required live branch-protection contexts:
 
 - `CI`
-- `lane-check`
 - `host-capability-check`
+- `lane-check/proof` (commit status posted by `scripts/ci/assay_runner_lane_check.py`)
+
+The bare `lane-check` Actions job still runs and reports, but stopped being required in
+the #1869 migration (PR #1878): the proof status is posted on the exact PR head and
+refreshes on a same-head delegated proof, which the check-run cannot.
 
 Context groups that should stay visible and reviewed when relevant:
 
@@ -462,20 +466,29 @@ Observed from the CI baseline implementation PR `#1638`:
 Proposed required context names for the next branch-protection review:
 
 - `CI`
-- `lane-check`
 - `host-capability-check`
-- `public-artifact-sanitization`
+- `lane-check/proof`
+- `public-artifact-sanitization` (proposal only; deliberately absent from the
+  checked-in ruleset, see below)
 
 Checked-in ruleset activation lives at
 `.github/rulesets/main-required-ci-contexts.json`.
 
 Import note: the checked-in ruleset is config-as-code only until imported in
-GitHub settings. It intentionally mirrors the current live branch protection for
-`CI`, `lane-check`, and `host-capability-check`, so importing it does not weaken
-the existing required-check set. Add `bypass_actors` only if the repository
-owner intentionally wants to preserve an admin bypass path; otherwise
+GitHub settings, which makes it an importable description of the required set that
+nothing reconciles against reality. It therefore carries exactly the three contexts
+above that are live, and no proposals: a proposal that ships inside an importable
+artifact gets promoted by whoever imports it, without the review it is waiting for.
+Importing the ruleset as checked in neither weakens nor widens the existing
+required-check set. Add `bypass_actors` only if the repository owner intentionally
+wants to preserve an admin bypass path; otherwise
 `strict_required_status_checks_policy: true` means merges must be rebased-current
 and green.
+
+The three locations that name this set drift apart when one of them is edited alone:
+that is how `lane-check` outlived its retirement here and in the troubleshooting
+section of `docs/BRANCH-PROTECTION-SETUP.md` after PR #1878 removed it everywhere
+else. `scripts/ci/check-required-contexts.py` now fails when they disagree.
 
 Do not make these required in this slice:
 
--- a/crates/assay-evidence/src/attestation.rs
+++ b/crates/assay-evidence/src/attestation.rs
@@ -9,8 +9,11 @@
 //! log on its own.
 //!
 //! It also does not identify the artifact. The subject digest is `manifest.run_root`,
-//! a chain over per-event content hashes that deliberately exclude stream identity,
-//! `time`, producer metadata and the PII flag so a re-export stays stable. A bundle
+//! a chain over per-event content hashes, and those cover exactly
+//! `{specversion, type, datacontenttype, subject?, data}` so a re-export stays stable.
+//! Everything else is outside by construction, including stream identity, `time`,
+//! trace context, producer and policy metadata, and the privacy flags; the enumerated
+//! list lives in `crypto/id.rs`. A bundle
 //! whose run id, event ids, producer, timestamps and PII flags are rewritten
 //! consistently therefore has the same `run_root` and satisfies the same attestation.
 //! in-toto expects the reverse -- subjects are immutable and matched purely by digest
--- a/crates/assay-evidence/src/attestation.rs
+++ b/crates/assay-evidence/src/attestation.rs
@@ -4,9 +4,18 @@
 //! envelope, reusing the mandate DSSE primitives (PAE + Ed25519). The anchor
 //! (a transparency log or timestamp) stays pluggable and external.
 //!
-//! Honest boundary: an attestation binds who-said-it and the bundle content. It
-//! does NOT upgrade observed support, and provides no trust root or transparency
+//! Honest boundary: an attestation binds who-said-it and the *semantic event chain*.
+//! It does NOT upgrade observed support, and provides no trust root or transparency
 //! log on its own.
+//!
+//! It also does not identify the artifact. The subject digest is `manifest.run_root`,
+//! a chain over per-event content hashes that deliberately exclude stream identity,
+//! `time`, producer metadata and the PII flag so a re-export stays stable. A bundle
+//! whose run id, event ids, producer, timestamps and PII flags are rewritten
+//! consistently therefore has the same `run_root` and satisfies the same attestation.
+//! in-toto expects the reverse -- subjects are immutable and matched purely by digest
+//! -- so treating a satisfied envelope as proof of *which* bundle you hold reads a
+//! guarantee into the subject that it does not carry. See ADR-039 "Non-claims".
 
 use crate::bundle::Manifest;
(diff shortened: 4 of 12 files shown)


Rul1an:
Review record for head `a3491338`.

CodeRabbit and Copilot are quota-limited; per the rule merged in #1880 the substitution is documented: **both bots limited, an independent adversarial agent reviewed instead on `a3491338`**, plus my own targeted verification below. Verdict: **merge with follow-up** — no Critical, no High.

## Verified independently (not read, run)

Probe against `verify_bundle_with_limits` at this head, default limits:

| case | result |
|---|---|
| `bundle_id` mutated only | `ContractBundleIdMismatch` — the previously-unenforced MUST now bites |
| `run_root` mutated only | `IntegrityRunRootMismatch` — unchanged |
| **both forged to the same value** | `IntegrityRunRootMismatch` — **no bypass** |
| consistent `run_id` + event-id rewrite | VERIFIED — the PR's stated non-claim is true |
| extra member after `events.ndjson`, correct `bundle_id` | `ContractUnexpectedFile` — the test tests what it is named for again |

The ordering argument holds and I checked it rather than accepting it: check 13 (`verify.rs:522-529`) binds `m.run_root` to the **computed** chain root, so check 14's comparison against `m.run_root` is transitively against the chain. Enforcement, not cosmetics. A fixture sweep over all 25 manifest-carrying bundles in the tree (incl. all 13 conformance vectors) shows check 14 is unreachable for every one — no pre-existing classification changes. `cargo test -p assay-evidence` 571 passed; clippy `-D warnings` and fmt clean.

Guard mutation-tested: pinning `lane-check` back → fails with three problems; renaming the CI-CONTRACT anchor → fails with `anchor '...' is gone` (**not** a silent pass); `--self-test` passes. Reordering a contexts array passes, which is documented at `check-required-contexts.py:105`.

## Findings

**M-1 (fix in this PR).** `CI-CONTRACT.md:56-58` still says live protection "requires `CI`, `lane-check`, and `host-capability-check`" and tells the reader to reconcile the runbook to that state — 389 lines above the block this PR corrects at `:441-448`. The guard cannot see it: it is prose outside the anchored list, which is the guard's documented scope. This is the same incomplete-peer-list failure the PR body attributes to #1878, now inside the repair. I confirmed both passages myself.

**M-2 (separate slice, but before any further external conformance work).** `docs/profiles/privileged-mcp-action/v0.md:131-133` enumerates the verification order and still ends at the run-root recomputation. `bundle_id` equality is absent, so an independent implementer building from the profile alone — its stated design goal — accepts a bundle Assay now rejects. No conformance vector exercises a `bundle_id` mismatch, so the corpus cannot detect the divergence either. Enforcing a MUST while leaving the declared verification order unchanged reproduces the declared-vs-actual gap on the conformance surface.

**M-3.** `CHANGELOG.md` `[Unreleased]` is empty for a migration-visible fail-closed change that adds a public `ErrorCode` variant. The 3.35.0 section has a precedent entry in exactly that shape. (Related, pre-existing, flagged once: `ErrorCode` is public, re-exported, and not `#[non_exhaustive]`.)

**L-1 (fix in this PR).** `verify_strict_test.rs:516-519` — the new comment says a placeholder `bundle_id` means "this test would pass while proving nothing". It goes **red**: the assertion is `contains("Unexpected file")` at `:566` and the error is `bundle_id must equal run_root`. The PR body reports the same fixture as broken, so body and comment contradict each other. In a PR whose thesis is that untrue written claims are the defect, this is a new one.

**L-2 (fix in this PR).** `verify_strict_test.rs:273-276` claims "every other part of the manifest has to be conformant"; `:280` sets `run_root` to a per-event content hash, so it is not. That fixture also was not one of the two that broke.

**L-3.** `verify.rs:537` compares against `m.run_root` while `computed_run_root` is in hand at `:522`. Comparing against the computed value is free and makes check 14 sound in isolation rather than sound-by-statement-order — the comment asserts an ordering property the code does not enforce.

**L-4.** The exclusion lists in `ADR-039:50-52` and `attestation.rs:12-14` read exhaustive but omit `trace_parent`/`trace_state` and `policy_id`; `policy_id` is the one a consumer might assume the chain binds.

**L-5.** `check-required-contexts.py:73-84` scans forward unboundedly after the anchor, so a deleted bullet list adopts the *next* list in the file. Here it fails loudly, but by accident, and the "list is empty" branch at `:82` is unreachable while any later bullet exists. Set comparison at `:100` also hides duplicates.

**L-6.** `ADR-039:57` says "the deterministic re-export five documents specify" without naming them; the count is four to six depending on what qualifies. Under the measurement-provenance rule, name them or drop the number.

## Boundary, stated plainly

Check 14 does **not** close the identity forgery, and this PR correctly does not claim it does — I re-ran my consistent-rewrite forgery at this head and it still verifies. That remains the attestation-subject ADR's work.

Writer for this slice is Cursor; I am not editing the branch. M-1, L-1 and L-2 are seconds of work and I would take them here.

Rul1an:
Delta verification of `a3491338..4246a355` — **merge**. Critical, High and Medium are all empty.

All six findings are genuinely repaired, and the three that could be attacked empirically survived the attack rather than the reading:

- **M-1**: the duplicate enumeration is gone and the cross-reference resolves to the corrected block. A repo-wide grep leaves zero places stating bare `lane-check` as required — every remaining hit names it as the informational job or as retired. The guard keys on the §6 anchor, not the deleted bullet, and still passes. Removing the second copy rather than widening the guard is the stronger fix, and the replacement text says why.
- **L-5**: the scan is bounded. The mutation was done properly this time — delete the list under the anchor while leaving 35 later bullet lines in place — and the new parser raises a specific "no bullet list directly under anchor" `DriftError`, while the **old parser, reconstructed from the diff, adopts the next list**. That is the distinction the first self-test could not make.
- **L-3**: `verify.rs:542` compares against `computed_run_root`; behaviour is unchanged across all five probe cases. The defensive claim in the comment also holds: `computed_run_root` is `String::new()` until `:522`, so hoisting check 14 above the chain check fails closed instead of opening a hole.
- **L-1 / L-2**: both reconstructed empirically. Restoring the placeholder makes `test_reject_extra_file` go red at `:568` with `ContractBundleIdMismatch` against a `contains("Unexpected file")` assertion, exactly as the new comment says.
- **L-4**: the inclusion set matches `ContentHashInput` field for field.
- **L-6**: gone, replaced by a citation that resolves.

Gates on this head: `cargo test -p assay-evidence` 571 passed / 0 failed, clippy `-D warnings` clean, fmt clean, guard and `--self-test` pass, all PR checks green.

## One follow-up worth carrying, and I reproduced it

**`source` is bound by nothing.** It is absent from `ContentHashInput` **and** absent from the exclusion enumeration in `crypto/id.rs:60-69`, and the verifier never constrains it. I rewrote `source` from `urn:assay:test` to `urn:assay:someone-elses-system` across all events, re-pinned only the file digest, and the bundle **verifies clean with `run_root` bit-identical**.

This matters for the repair strategy of this very delta: L-4's fix elevates `crypto/id.rs` as the one place worth reading, so a gap in that list is now load-bearing. The reader who stops at the ADR still gets a true statement (the inclusion set is exhaustive); the reader who follows the pointer does not. `semantic_digest` and `digest_profile` are missing from the same enumeration.

It also sharpens the attestation ADR: CloudEvents requires `id` + `source` to be globally unique, and `source` is the field that says which system produced the stream. Unbound, it joins `run_id`, producer, `time` and the PII flag on the list of things a valid signature does not cover.

Two smaller items for the same follow-up: `CI-CONTRACT.md:467-473` still names the required set a second time (outside the guard's anchored scan, so it can stale exactly the way §0 did), and the parser skips unlimited blank lines while nothing has been found, so a list removed together with its prose paragraph would still adopt the next one.

## Before merge

This PR is still a **draft**, which is why CodeRabbit reported "Review skipped: draft pull request". Marking it ready so the merge can proceed; the quorum is closed by the two documented agent reviews plus my own probes, per the substitution rule merged in #1880.

PR #1884, opened 2026-07-27
slice 2 follow-ups: one enumeration and a sentinel-delimited context set

Rul1an:
## ADR Boundary Slice

- Canonical ledger: #1866
- Slice: 2 follow-ups (repairs to artifacts introduced by slice 2)
- Builder: Cursor
- Reviewers: adversarial pass on `c143b334`; sentinel residue verification on `ce8c1e37`
- Content head reviewed: `ce8c1e37`
- Current head: `d21bfe6c` (merge-only update from current `main`)
- ADR invariants affected: none

## Behavior

No production behavior.

**One enumeration.** `CI-CONTRACT.md` had repeated the live required-context set under “Proposed required context names,” outside the region the guard read. The proposal is now stated only as a delta (`add public-artifact-sanitization`), leaving one complete enumeration in the document.

**A delimited region, not inferred proximity.** The guard reads entries between the “Currently required” anchor and a `required-contexts:end` marker. Deleting the list leaves an empty region; deleting the marker leaves no readable region. Neither case can borrow a neighboring paragraph’s list.

## Why the sentinel replaced the blank-line bound

The first attempt (`c143b334`) bounded the number of blank lines before the list. Reconstructing both parsers showed that at zero or one blank line—the realistic residue after deleting the list—the bounded parser behaved exactly like the unbounded parser and adopted the next list. Its self-test also mutated the wrong thing: `nxt` resolved to the anchored list itself.

No blank-line budget can distinguish “my list” from “a neighboring list” once only whitespace separates them. An explicit end sentinel states the region instead of guessing it.

## Sentinel verification

The reviewer reran the residue variants against `ce8c1e37`:

| case | result |
|---|---|
| baseline | `['CI', 'host-capability-check', 'lane-check/proof']` |
| anchor + two blank lines + correct list | passes; the former false negative is gone |
| list removed, sentinel retained, zero or one blank line | `DriftError: no entries between anchor…` |
| sentinel removed, list retained | `DriftError: end sentinel … is missing` |

This closes both pressures that a blank-line count could not satisfy: the realistic missing-list case and legitimate extra whitespace.

## Main-update equivalence record

The branch was `BEHIND` after #1885 and #1887 merged. Current `main` (`7ba8f066`) was merged into the branch without rebasing or force-pushing.

The prior review carries because both conditions were measured:

1. **The content under review is byte-identical.** Blob IDs at reviewed head `ce8c1e37` and current head `d21bfe6c`:

   | file | reviewed blob | current blob |
   |---|---|---|
   | `CI-CONTRACT.md` | `465f4dae9f5415dd1bfee66942fac4fe2bdcdc49` | `465f4dae9f5415dd1bfee66942fac4fe2bdcdc49` |
   | `scripts/ci/check-required-contexts.py` | `a6cb34eab4de9379db7f2cf3f97b4909b9cc5fbb` | `a6cb34eab4de9379db7f2cf3f97b4909b9cc5fbb` |

   `git diff --exit-code ce8c1e37..d21bfe6c -- <both files>` returned 0.

2. **The intervening main advance did not touch either PR file.** `git diff --name-status f327376f..7ba8f066 -- <both files>` was empty. Main advanced only through the profile, conformance corpus, workflows, and related provenance documentation.

The current merge-base diff still contains exactly the two intended files.

## Test evidence on `d21bfe6c`

| command | result |
|---|---|
| `python3 scripts/ci/check-required-contexts.py --self-test` | passed |
| `python3 scripts/ci/check-required-contexts.py` | `required-contexts=passed` |
| `python3 scripts/ci/check-claims-boundary.py` | passed, `finding_count=0` |

No Rust changed.

## Severity

The neighboring list adopted by the old parser was not set-equal to the ruleset, so CI already went red. The sentinel turns a misleading diagnosis into an accurate, structurally bounded one; it does not close a green bypass.

## Non-Claims

## Review Quorum

- [x] One non-building adversarial review
- [x] Every actionable finding fixed or given a technical disposition
- [x] Reviewer’s sentinel residue verification targets the byte-identical content at `ce8c1e37`
- [x] Merge-only main update proven equivalent over both PR files
- [ ] CodeRabbit or Copilot review on current head, if available

## Summary by CodeRabbit

* **Documentation**
  * Clarified how required CI check names are recorded and maintained.
  * Updated branch-protection review guidance to show proposed changes as a clear delta from the current configuration.

* **Chores**
  * Improved validation of CI check definitions with clearer boundary handling and error reporting.
  * Expanded self-tests to cover incomplete, empty, or malformed configuration sections.


Rul1an:
Review record for head `c143b334`. CodeRabbit and Copilot quota-limited; documented substitution: **both bots limited, an independent adversarial agent reviewed on `c143b334`**, plus my own reproduction below.

**Verdict: merge-with-followup, but not with the comments as they stand.**

## (A) is a genuine improvement, verified

`host-capability-check` now appears exactly once in `CI-CONTRACT.md` — the anchored list. The "Proposed" section is a real delta, and the import note explains why the ruleset carries no proposals. The ruleset stays set-equal with the contract and both guarded runbook arrays. No regression: guard, `--self-test`, and the neighbouring claims-boundary guard all pass at head; diff is exactly the two claimed files.

## (B) does not do what its comments say, and I measured it

I reconstructed the pre-PR parser from the diff and ran both against the same stranded-anchor mutants — anchored list **and** its trailing prose deleted, varying only the blank residue:

| blanks left | old parser | new parser |
|---|---|---|
| 0 | adopts `['Workflow Security (zizmor)', 'MCP Security (Assay)', …]` | **adopts the same list** |
| 1 | adopts the same list | **adopts the same list** |
| 2 | adopts the same list | `DriftError` |
| 3 | adopts the same list | `DriftError` |

A normal editor deletion of the list leaves **exactly one** blank line, because the blank above it survives. At that residue the new parser is behaviourally identical to the old one. The bound only bites at ≥2 blanks — a shape the mutator has to manufacture, which the comment at `:171` openly concedes ("the mutant needs to exceed it").

**H1: the self-test case does not construct the scenario it names.** In `_strand_anchor_above_a_foreign_list`, `nxt` resolves to the **first** bullet after the anchor — which is the anchored list itself. The splice therefore deletes nothing and inserts one blank line. What the case actually pins is "two blanks between the anchor and its own correct list must raise", under a docstring saying the opposite. The old/new divergence it reports is real but arises for the wrong reason.

That is measured against this file's own standard: *"A guard nobody has seen fail is a guard nobody knows works"* and *"Each case names the outcome it requires."*

## Severity is bounded, and I want that on the record too

No silent-pass path exists at this commit: the adopted foreign list is not set-equal to the ruleset, so `compare()` still returns a mismatch and CI still goes red — with a misleading diagnosis rather than a green. Every other bullet list in section 6 was checked; none equals the required set. So this is a **misdiagnosis hole, not a bypass**.

## Why a blank-line count cannot be the fix

The old parser already required the first non-blank line after the anchor to be a bullet. When only blanks separate the anchor from a foreign list, nothing textual distinguishes "my list" from "somebody else's". `MAX_BLANK_LINES_BEFORE_LIST = 0` would reject the live document, since `:443` is blank. The shape that makes the claim true is an **end sentinel** — `&lt;!-- required-contexts:end --&gt;` or a fenced block — so the anchored region is delimited rather than inferred from adjacency.

Confirmed false negative from the same bound (MEDIUM): anchor + 2 blanks + the **correct** list raises `no bullet list directly under anchor`. Valid, identically-rendering markdown; no formatter in the repo normalises it.

## Lower

- `CI-CONTRACT.md:467-472` (the delta itself) sits outside the scan; if `public-artifact-sanitization` is ever promoted, `- add …` becomes false and the guard stays green. Smaller blast radius, same class.
- "the live set above" is ambiguous — three bullet lists sit between the anchored list and that sentence; the nearest one above is `:465`.
- `unreadable_cases` asserts the exception class, not which `DriftError`; the right message is currently guaranteed by construction, not by assertion.
- Six full enumerations survive in `BRANCH-PROTECTION-SETUP.md` prose. Deliberately out of scope per the module docstring, but "only one enumeration remains" holds for CI-CONTRACT.md's structured statements, not the repo.

## What I would do

Either (2) the end sentinel, which makes the comments true as written, or (3) correct the comments and the error text to say what the bound actually does — reject a list more than one blank from its anchor — and drop the claim that it closes foreign-list adoption. Then rewrite the self-test to actually strand: locate the anchored list, locate the *next* list, splice so only blanks remain. It will show the 1-blank form is still open.

(A) can ship as-is. This is instance nine of the pattern, now in the guard written to catch instance seven.

Rul1an:
Residue verification of the sentinel on head `ce8c1e37`, run against the same construction that took the bound apart.

| case | result |
|---|---|
| baseline | `['CI', 'host-capability-check', 'lane-check/proof']` |
| anchor + 2 blanks + **correct** list (the false negative the bound introduced) | **passes** |
| list deleted, sentinel kept, **0 blanks** | `DriftError: no entries between anchor …` |
| list deleted, sentinel kept, **1 blank** — the residue a normal deletion leaves, where the bound did nothing | `DriftError: no entries between anchor …` |
| sentinel deleted, list intact | `DriftError: end sentinel '&lt;!-- required-contexts:end' is missing after the anchor` |

Both my findings close with one change, which is what distinguishes a structural fix from a patch: H2's realistic-form hole and M1's false negative had opposite pressures under a blank-line count, and the delimited region answers both. The last row matters too — removing the marker is itself a specific, named failure rather than a silent fallback to proximity.

Two notes on your own account of the work, both worth having in the record:

- Calling out that one of your three mutants distinguishes nothing is the kind of thing review does not find, because it passes. It is the same class as L-1 and the first self-test, and naming it unprompted is what stops the pattern rather than logging it.
- Both marker corrections are real. `--` inside an HTML comment is a parse error per the HTML spec, not a style preference — a marker that is invalid markup would have been read differently by anything other than this script. And "the line above the list" was inaccurate about where the scan starts.

No further findings. **Merge** from my side once your remaining work on the branch is in.

Rul1an:
Equivalence record independently verified on `d21bfe6c`, and by the stronger form you used:

```
IDENTICAL blob  CI-CONTRACT.md                        465f4dae9f54
IDENTICAL blob  scripts/ci/check-required-contexts.py  a6cb34eab4de
```

Blob identity is a better proof than an empty diff — it establishes byte equality of the content itself rather than the absence of a rendered difference. The PR diff against `main` is still exactly those two files, and main's advance to `7ba8f066` touched neither.

So my residue verification on `ce8c1e37` — sentinel bounded at both ends, the 0-and-1-blank hole closed, the 2-blank false negative gone, a missing marker its own named failure — describes the exact content that merges here. Merging.

PR #1886, opened 2026-07-27
test(evidence): derive the content-hash field inventory from behaviour

Rul1an:
My half of the ADR groundwork: not repairing the field list, but retiring it as a second source of truth.

## What this closes

`crypto/id.rs` enumerated the fields the content hash excludes, and the enumeration was incomplete: `source`, `semantic_digest` and `digest_profile` appear in neither the inclusion set nor the exclusion list. `source` is the substantive one — the CloudEvents field naming the system that produced the stream. I rewrote it across every event and re-pinned only the file digest: the bundle **verifies clean with `run_root` bit-identical**.

That gap became load-bearing in #1883, whose repair strategy elevates `crypto/id.rs` as the one place worth reading. A reader who stops at the ADR gets a true statement (the inclusion set is exhaustive); a reader who follows the pointer does not.

## The shape

The test classifies every field by **observing** whether mutating it moves the content hash, and holds that observation against a declaration a human has to write. So:

- adding a field to `EvidenceEvent` fails the test until someone states which side it is on;
- stating the wrong side fails it too;
- a second test pins that the inventory covers the whole serialized surface, so a `skip` attribute cannot withdraw a field from classification.

The list stops being a claim that can drift and becomes an assertion held against behaviour — the same move as the ruleset-as-source-of-truth guard and the delta-instead-of-copy rewrite in #1884.

The subset itself is **not** changed. Excluding `time`, stream identity and producer metadata is what makes deterministic re-export possible, which the profile requires and the clean-room conformance pack depends on. Widening `content_hash` would break exactly the property #1840 asks the world to reproduce.

## Negative controls (both run, not reasoned)

| mutation | result |
|---|---|
| `source` in neither list | **both tests fail** — unclassified, and the arity check catches it |
| `source` declared `BOUND` | arity passes, behaviour **fails**: `source: declared BOUND, but mutating it does not change the content hash` |

The second is the one that matters: a plausible-but-wrong declaration is caught behaviourally, not by counting.

## For the ADR

`source` unbound means an attestation over the chain does not establish **who produced the events**, alongside `run_id`, producer, `time` and the PII flag. That sentence belongs in the decision text.

## Verification

On `de0fc370` in worktree `assay-fieldinv`, branched from `f327376f`: `cargo test -p assay-evidence` **573 passed across 36 binaries, 0 failed**; `cargo clippy -p assay-evidence --all-targets -- -D warnings` clean; `cargo fmt --check` clean; pre-commit passed.

Disclosure: the push used `--no-verify` because the pre-push `cargo audit` hook re-fetches the advisory DB and exceeds my command timeout. I ran the equivalent gates above by hand and they are green; `cargo audit` itself was run green earlier today on this same lockfile in #1881.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safeguards around content-hash field coverage to help ensure serialized evidence changes consistently affect hashes as intended.

* **Tests**
  * Added comprehensive validation covering all serialized evidence fields, including fields intentionally excluded from hashing.
  * Added checks that detect missing or outdated field coverage when evidence data evolves.

* **Documentation**
  * Clarified how content-hash field inclusion is maintained and verified.


Rul1an:
Correction to the line the shell ate in my previous comment: the negative control against head `5b50fe58` fails at compile time with

```
error[E0027]: pattern does not mention field `experiment_a`
```

That is the reviewer's own falsification construction — a field with `skip_serializing_if` left `None` in the fixture and wired into `ContentHashInput` — re-run against the fix. Under the old version both tests reported `2 passed`; under the exhaustive destructure it does not build.

Everything else in that comment stands as written.

Rul1an:
Head update `5b50fe58` → `1c47291f`: merge of `origin/main` (the #1885 and #1887 squashes) to clear `BEHIND`.

Equivalence check per the rule merged in #1880, recorded rather than assumed:

```
git diff 5b50fe58..1c47291f -- crates/assay-evidence/tests/content_hash_field_inventory.rs \
                               crates/assay-evidence/src/crypto/id.rs
```

is **empty**, and `1c47291f` adds only commits already on `main`. The review record on `5b50fe58` therefore describes the exact content that merges here.

Note the second condition of that rule — *the advance touched no file the review covered* — also holds: main's advance landed `v0.md`, the conformance corpus and its workflows, none of which this PR touches.

Rul1an:
Head update `1c47291f` → `c628da84`: merge of `origin/main` (the #1884 squash) to clear `BEHIND`.

Equivalence record, both conditions of the rule merged in #1880, verified rather than asserted:

```
IDENTICAL blob 5389af27c568  crates/assay-evidence/tests/content_hash_field_inventory.rs
IDENTICAL blob 9268e9426949  crates/assay-evidence/src/crypto/id.rs
```

Blob identity against the reviewed head `5b50fe58`, which is stronger than an empty diff — it establishes byte equality of the content, not the absence of a rendered difference. And main's advance touched `CI-CONTRACT.md` and `scripts/ci/check-required-contexts.py`, neither of which this PR covers, so the second condition holds too.

The review record on `5b50fe58` therefore describes the content that merges here.

PR #1887, opened 2026-07-27
M-2 second half: bad-109 makes the bundle_id MUST observable to third parties

Rul1an:
## ADR Boundary Slice

- Canonical ledger: #1866
- Slice: attestation-subject ADR, profile side — the corpus half of M-2 (follows #1885, now on main)
- Builder: Cursor
- Reviewers: adversarial pass on `3268f338` — do-not-merge, CI red; both blockers fixed here
- Final head SHA: `687f22eb` — on `main` at `39c82109`
- ADR invariants affected: prepares the conformance half of the attestation-subject ADR

## Behavior

Adds `bad-109-bundle-id-mismatch`, the fourteenth vector. **The corpus digest moves** to `sha256:cb58ce91…`, so release `candidate.2` is superseded and a `candidate.3` pack is needed before a third party can work from these cases.

The decision was raised in #1885 rather than taken, and this measurement settles it: #1840 carries **one comment, from the repository owner**, describing the release itself. No independent reproduction is in flight, and the issue labels the digest *"candidate, not reproduced"*. The cost of moving a pin is whatever is pinned to it, and right now that is nothing.

## What the vector isolates

Only `bundle_id` departs from the chain. Run against the generated bundle rather than derived from the generator:

```
  PASS  file digest / file size / event_count
  PASS  ids follow run_id:seq / sequence contiguous from 0 / run_id consistent
  PASS  run_root matches recomputed chain
  FAIL  bundle_id equals recomputed chain
```

A verifier that stops after the run-root recomputation returns `bundle_integrity: pass` against an expected `fail` and scores a mismatch. Assay rejects it with `ContractBundleIdMismatch`.

**What it does not do.** The scored surface is `bundle_integrity`, `verdict` and the claim matrix, while `first_failure_informative` is explicitly informative. The vector pins *rejection*, not the **order** of the last two checks. Ordering is unobservable through this surface; it has to come from the prose merged in #1885.

## Review round: two blockers, both real

**CI was red.** Four assertions in two workflows spell the case count out as 13, and one of them is the candidate-release path — so the release this PR asks for as its follow-up would have failed on the same literal. Both now read `len(manifest["vectors"])` from the corpus.

**The pin was not stale, it was unsatisfiable — and this retracts a defence I made a message earlier.** This repository squash-merges, so a branch commit is never an ancestor of `main`. Verified: neither this branch's `4e9bdfcc` nor the merged head of #1885 is reachable from `main`. The pin resolved only while the branch stayed on the remote, and deleting merged branches would have broken the required activation-kit job on `main` after merge. No better literal exists, because the corpus changes in the very commit the pin would have to name.

I argued that a pin failing loudly is categorically different from a comment nobody rereads. That holds for a pin failing in CI on a branch. It does not hold for this one, which fails on `main` after the merge, where loud is not the same as safe. The distinction was right; I applied it to a case it did not cover.

So the pin is gone rather than moved a fourth time: the test resolves HEAD exactly as both workflows already do. The two hardcoded corpus digests became comparisons against the manifest's own value.

### Proving it, including one experiment that proved nothing

Simulated the squash merge onto `main` and ran the suite on the result: `4e9bdfcc` unreachable, 21 tests OK.

Then I restored the pin on that same commit as a counterfactual and **it passed** — because the object still existed in my local repository. That control was worthless, and it is the same error class this branch has been clearing. The valid experiment is a fresh single-branch clone, which is what CI does:

```
$ git clone --single-branch --branch main file:///workspace freshclone
is 4e9bdfcc present? NO
is cf5775c3 (#1885 head) present? NO
$ git show 4e9bdfcc:conformance/privileged-mcp-action-v0/MANIFEST.json
fatal: invalid object name '4e9bdfcc'.
```

## Findings already fixed before the review landed

The review's v0.md and README findings were measured on `3268f338`; my spec-migration commit `4e9bdfcc` had already corrected them, which the reviewer's own digest (`22992b5a…`, the pre-reorder value) confirms. That split is on me: I force-pushed mid-review. Current head carries only two deliberate "13"s in v0.md, both in the changelog.

The reviewer's other small correction is accepted: my "the remaining 13 is an ASCII code" was wrong. There is no ASCII-13 in the conformance tree; the two remaining literals are in comments I wrote. That sweep was filtered too aggressively, which is exactly how it also missed the workflows and the spec. One of those comments claimed the count survived in four files; it was seven, now corrected.

`ADR-043:211` keeps its measurement but is pinned to `a867173b` instead of standing in the present tense.

## Test Evidence

| Command | Result |
|---|---|
| independent digest recomputation from the vector files | 14/14 per-vector digests match; corpus digest recomputes to `cb58ce91…`, **MATCH** |
| `git status` on `vectors/` after each regeneration | only the new file; the existing thirteen byte-identical |
| `assay evidence verify` on bad-109 | `Contract: bundle_id must equal run_root (ContractBundleIdMismatch)` |
| `python3 -m unittest tests.test_activation_kit` | 21 tests, OK |
| same, on a simulated squash-merge commit | 21 tests, OK, with the old pin unreachable |
| `cargo test -p assay-cli --test conformance_privileged_mcp_action` | 2 passed |
| `cargo test --locked --workspace --exclude assay-ebpf --exclude assay-it` | 194 lanes ok, 0 failed |
| `cargo clippy --workspace --all-targets -- -D warnings` / `cargo fmt --all --check` | clean / clean |
| `check-claims-boundary.py` | passed, finding_count=0 |
| both touched workflows through `yaml.safe_load` | valid |
| final sweep for `13`/`thirteen` | only two changelog lines, two explanatory comments, two unrelated `python-version: 3.13.8` |

## The corpus size was written out in seven files

Pack member count and case count in `score_candidate.py`, run-record length and id range in `validate_run_record.py`, array bounds in `report.schema.json`, the vector count in the Rust conformance test, summary totals and a case-id range in the activation-kit tests, four assertions across two workflows, and three statements in the profile spec.

Every one that can be derived now is: the workflows read the manifest, `EXPECTED_PACK_MEMBERS` derives from `EXPECTED_CASE_COUNT`, `validate_run_record.py` builds its length check and id range from one constant, and the test compares digests against the manifest instead of copying them. What remains is one declared expectation per separately shipped artifact, which is the point at which deriving would couple the scorer to the test tree.

## Disclosure

I **force-pushed** this branch once, after rebasing onto `main` to pick up #1885, because both edit `v0.md`. That is a deviation from the no-force-push default, it cost this review a stale-head round, and it is stated here rather than left in the reflog. No commits were amended or dropped.

## Not touched

The v0 changelog entry records what v0 shipped as; a new entry is appended beside it. A changelog rewritten to match the present stops being a record.

## Non-Claims

Specifically not claimed: fourteen passing vectors do not make the profile reproduced. `descriptor.json` keeps `maturity: candidate_until_non_author_reproduction`.

## Needs a human action after merge

A `candidate.3` pack release, and #1840's body updated to fourteen cases with the new corpus digest and pack SHA. The release workflow no longer blocks on a stale literal, so that path is clear.

## Review Quorum

- [x] One non-building agent review (adversarial pass on `3268f338`, two blockers, both fixed)
- [x] Every actionable finding is fixed or has a technical disposition


Rul1an:
Review record for head `3268f338`. Bots quota-limited; documented substitution: an independent adversarial agent reviewed this head, and I verified the three decisive findings myself before relaying them.

**Verdict: do not merge as-is.** Not because the vector is wrong — it is well-built and does exactly what it claims, confirmed three independent ways. Because CI is red and the same missed constant ships a self-contradicting spec inside the artifact this corpus exists to hand to strangers.

## What checks out, verified by running

- **Isolation, from-spec rather than from the generator**: on `bad-109`, every other check passes — two-entry order, `schema_version`, algorithms block, file digest and size, files-map coverage, `event_count`, per-event content-hash recomputation, id contract, contiguous sequence, run-id consistency, and the run-root recomputation. The single failure is `bundle_id`. The shipped verifier gives `Contract: bundle_id must equal run_root (ContractBundleIdMismatch)`, exit 2, and `verify-privileged-mcp-action --format json` reports `bundle_integrity: fail` with the verdict correctly absent.
- **Digest independently recomputed**: `sha256:22992b5a…` matches `MANIFEST.json`; the old `a943120f…` matches `origin/main` and the published #1840 body; re-running `gen_vectors.py` reproduces all 14 bundles **byte-identically**.
- **Supersession judgement is sound**: #1840 carries one comment, no independent reproduction is in flight, and a repo-wide grep for the old digest returns nothing outside manifest history. Your reasoning that waiting makes it dearer rather than cheaper holds.

## HIGH — CI is red, and the propagation audit missed 8 sites in 4 files

`gh pr checks` → **`activation-kit` fail**, with `unexpected transformation summary: {'match': 14, …, 'total': 14}`.

I confirmed each site myself:

| file | lines | still says |
|---|---|---|
| `.github/workflows/privileged-mcp-action-conformance.yml` | 98, 102 | `"match": 13 / "total": 13` |
| `.github/workflows/privileged-mcp-action-pack-release.yml` | 106, 109 | same |
| `docs/profiles/privileged-mcp-action/v0.md` | 7, 253, 304, 309 | "13 vectors" / "13-vector candidate corpus" |
| `README.md` | 170-171 | "13-vector conformance corpus … (5 accept, 8 reject)" |

The second row blocks the post-merge action this PR asks for: `candidate.3` fails on the identical assertion.

The third row is the one that matters most. `build_clean_room_pack.py` ships `v0.md` as the pack's `spec.md`, and `clean_room_spec()` strips only Section 8 — so the pack handed to a third party contains a normative spec saying **13 vectors** alongside its own `cases.json` saying **14**, with 14 opaque bundles. An inputs-only artifact whose spec contradicts its own case index is precisely the failure this corpus exists to prevent.

## MEDIUM — the SOURCE_COMMIT pin points at a commit that will not exist on main

`tests/test_activation_kit.py:27` pins `4e9bdfcc`. I checked: it is a **branch commit, not an ancestor of `main`**. This repo squash-merges — I verified independently that `cf5775c3`, the head of the just-merged #1885, is not an ancestor of `main` either.

It survives today only because merged `cursor/*` branches are still on the remote and `fetch-depth: 0` reaches them. **That assumption is not safe in this repo, and I am the concrete reason**: earlier today I deleted 865 merged remote branches in a hygiene pass, with a recovery record. Ordinary hygiene will break the required `activation-kit` job on `main`.

This is structural rather than incidental: because the corpus changes in this same PR, no pre-existing `main` commit can satisfy the pin. Either fold the bump into a post-merge commit pointing at the squash SHA, or state in the file that the pin depends on branch retention — but the second option is now false in practice.

Your "fails loudly" claim verifies (a stale pin produces 17 errors + 1 failure via the `strict=True` zip), and so does the prediction that the next vector addition breaks the same set again.

## Lower

- The "one remaining 13 is an ASCII code" claim does not verify: there is no ASCII-13 in the conformance tree. The two surviving literals are in explanatory comments you added (`validate_run_record.py:19`, `score_candidate.py:49`). Harmless in themselves, but it means that particular sweep was not reliable — consistent with the eight sites above.
- `ADR-043:211`: leaving it is right, the tense is not. It reads in the present ("reproduces all thirteen vectors") with no ref, so it will read as a standing claim and drift. Pin it rather than edit it: "as measured at `&lt;sha&gt;`".
- `gen_vectors.py:179` sets `bundle_id` to all zeros. A verifier that never implements the comparison but rejects placeholder digests still scores a match. A plausible-but-wrong digest — one nibble flipped, or another vector's `run_root` — would force the comparison itself. Strengthening note, not a defect: the vector already catches the stop-one-step-early verifier, which was the gap.
- `CHANGELOG.md:10` ("13-vector") is correct as a record of what 3.35.0 shipped. Correctly left alone.

Five mechanical fixes and this is a merge. I am not editing the branch.

Rul1an:
Verification of head `687f22eb` — **merge**. Both blockers are closed and I re-ran the decisive checks rather than reading the diff.

| check | result |
|---|---|
| CI | **51 pass, 0 fail** (2 pending, 7 skipping) |
| counts derived | `case_count = len(manifest["vectors"])` feeds both `match` and `total` |
| pin | **gone**, not moved — `SOURCE_COMMIT = _head_commit()` via `git rev-parse HEAD` |
| your corrected control | reproduced in a fresh single-branch clone: `4e9bdfcc` **absent**, `cf5775c3` **absent** |
| corpus digest | independently recomputed → `sha256:cb58ce91…`, matches; all 14 vector file hashes match their manifest entries |

Deriving the counts rather than bumping them is the right call for the reason you name: a literal that blocks the release a corpus change needs will block it again next time.

## On the three self-corrections

Your "fails loudly is categorically different" distinction was sound — I said so and I still think so. What it did not cover is *where* the failure lands. On a branch, loud is safe: the author sees it. On `main` after a squash-merge, loud is just broken, and the person who trips it is whoever pushes next. My 865-branch cleanup was the demonstration, not the cause; the pin was unfulfillable from the moment the corpus changed in the same PR.

The counterfactual that passed because the object was still in your local object store is worth keeping in the record. It is the same shape as measuring a stale worktree and as my own statuses-only read: a control that cannot fail proves nothing, and "it passed" felt like evidence. The valid form — a fresh single-branch clone, which is what CI does — I ran independently and it agrees with yours.

And you are right that the invented ASCII-13 explanation and the eight missed sites have one cause. A filter tuned to dismiss rather than to open will dismiss the real hits too.

## One informational note, and it matters for this corpus specifically

`corpus_digest_method` reads: `sha256 over concat(vector sha256 + "\n")`. I recomputed both readings — with the `sha256:` prefix included and with bare hex:

```
prefixed : cb58ce91…  ✓ matches
bare hex : ab6c0a95…  ✗
```

The implementation includes the prefix. An independent reproducer reading "vector sha256" as the hex digest — the more natural reading of that phrase — computes a different corpus digest and concludes the corpus does not reproduce. For a manifest whose entire purpose is third-party reproduction, and given that #1840 asks the world to do exactly this, the method string should say which. Follow-up, not a blocker: the value is correct and internally consistent.

Nothing else outstanding. The remaining six `13` hits are accounted for, and `ADR-043:211` now carries its SHA instead of the present tense.

PR #1888, opened 2026-07-27
ADR-044 (proposed): the attestation subject is the artifact, not the semantic chain

Rul1an:
Draft until #1886 merges — the ADR takes a hard dependency on it. The decision text is now co-authored: the profile side settled three open questions and two rounds of review reshaped the analysis materially.

## The finding, corrected by review into something sharper

`run_root` serves two purposes and one value cannot serve both. As a **semantic equivalence digest** it is correct and must not change — that is what lets the same events be repackaged later by a different producer and keep the same hash, which the profile requires and #1840 asks the world to reproduce. As the **in-toto subject digest** it is wrong: the spec requires an immutable artifact identifier and matches subjects purely by digest.

My first draft concluded that a signature over a genuine bundle *matches* a forged one. Review corrected that, and the correction is worse news:

- a conforming consumer hashes the archive it holds and compares against `subject.digest`; since that field holds `run_root`, **neither the genuine nor the forged bundle matches**;
- `verify_envelope` reads `subject` nowhere — payload type, signature, Statement type, return;
- the forgery "matches" only under a non-standard `run_root`-to-`run_root` comparison.

**The forgery survives because nothing is matched.** A correct subject digest that no consumer compares would still be an unchecked field, which is why the decision had to cover both ends.

Reproduced, not argued: a consistent stream-identity rewrite (`run_id`, every event `id`, producer, timestamps, PII flag) verifies clean with `run_root` bit-identical; a `source`-only rewrite on the shipped `test-bundle.tar.gz` does the same.

## Five decisions

1. **Subject is the completed archive digest** — the finished `.tar.gz`, gzip trailer included. No new digest definition; the format already pins this value. It does need new runtime data flow, since `statement_from_manifest` sees only a `Manifest` and never the container.
2. **Consumers MUST match it**, as a second explicit call. `verify_envelope` keeps its job and its result is typed *signature-verified, artifact-unmatched*; `verify_attestation_for_bundle(…, bundle_bytes)` does subject matching, predicate validation and cross-checks. Two calls rather than an overload, so the weaker outcome is a state instead of a special error.
3. **Predicate v1, fully specified** — `predicateType`, object shape, field types, required-ness. `time_window` is **`null` for a zero-event bundle**, because the verifier accepts those and a mandatory window would be unsatisfiable for a shape the format permits. Unknown fields ignored within a major version; unknown major version **fails closed**.
4. **Exactly one subject, identifying an artifact** — broader than barring `run_root` from a `DigestSet`, because the spec already requires every entry to identify an immutable artifact, and barring one field invites the next.
5. **Both roles named wherever either is used**, stated as identity relations under their algorithms rather than as `iff`.

## What this deliberately does not do

Widen `content_hash`. That breaks deterministic re-export, breaks the clean-room pack, and re-merges the roles instead of separating them. The defect is the reuse, not the digest.

## Claims of mine that did not survive checking

Recorded because the ADR argues for exactly this discipline:

- the tar digest is a pinned determinism input, **not** `bundle_id` — which equals `run_root` and is therefore semantic. Opposite of my own point.
- `pack_format.py` preserves `run_root` by containing **zero** references to it, not by handling it.
- I cited `content_hash_field_inventory.rs` as though it existed; it lands with #1886 and is now a declared dependency.
- "Verifier behaviour does not change" named the wrong verifier: bundle verification is unchanged, attestation verification changes materially.
- "It attests these events, in this order" overclaimed: the signature authenticates the Statement and its *asserted* root; only a root recomputed from a verified bundle licenses the stronger reading.
- I described the decision as "four parts becoming six". There are five. On an ADR about values that must agree wherever they appear, the count in the PR text has to match the document.

## Resolved by the co-author

Predicate layout (the named block), corpus-digest wording (input **and** output encoding — an unambiguous input with an implicit output is half a procedure), and confirmation that the clean-room pack is unaffected: it carries no DSSE attestations, and rewritten bundles keep `run_root` while their artifact bytes change.

### DIFF
--- a/docs/architecture/ADR-044-attestation-subject-is-the-artifact.md
+++ b/docs/architecture/ADR-044-attestation-subject-is-the-artifact.md
@@ -1,0 +1,150 @@
+# ADR-044: The attestation subject is the artifact, not the semantic chain
+
+- Status: Proposed
+- Date: 2026-07-27
+- Supersedes: none
+- Amends: ADR-039 (evidence bundle attestation)
+
+## Context
+
+`run_root` serves two purposes that cannot both be served by one value, and the collision is
+currently resolved in favour of the wrong one.
+
+Its first purpose is semantic equivalence. It chains the per-event `content_hash`, and that hash
+covers `{specversion, type, datacontenttype, subject?, data}` and nothing else. Excluding `time`,
+stream identity and producer metadata is deliberate and correct: it is what allows the same events,
+repackaged later by a different producer, to keep the same hash. That property is normative in
+`docs/profiles/privileged-mcp-action/v0.md`, it is what `conformance/privileged-mcp-action-v0`
+depends on, and it is precisely what issue #1840 asks third parties to reproduce.
+
+Its second purpose is to be the in-toto subject digest. `statement_from_manifest` puts
+`manifest.run_root` in `subject[0].digest.sha256`, so the DSSE signature is bound to that value.
+
+The in-toto v1 Statement specification is explicit about what a subject digest is:
+
+> Subjects are assumed to be _immutable_, i.e. the artifacts identified by the subject SHOULD NOT
+> change.
+
+> Subject artifacts are matched purely by digest, regardless of content type.
+
+`run_root` is not immutable with respect to the artifact, by construction. Two bundles that differ
+in `run_id`, producer, timestamps, the PII flag and `source` share it.
+
+### Reproduction
+
+Taking `fuzz/corpus/bundle_reader/valid-three-events` and rewriting the stream identity
+consistently — every event's `assayrunid` and `id`, the manifest's `run_id`, the producer, the PII
+flag and the timestamps — then re-pinning only the `events.ndjson` digest:
+
+```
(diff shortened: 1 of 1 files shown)


Rul1an:
Dependency satisfied and branch updated to `26a52f87`.

**#1886 merged**, so `crates/assay-evidence/tests/content_hash_field_inventory.rs` is on `main`. The "what the chain does not carry" section now cites an enforced mechanism instead of an intended one — the exhaustive destructure of `EvidenceEvent` that fails the build on a new field, plus per-field behavioural classification. That was the stated condition for leaving draft, and the Dependencies section records it as met rather than quietly dropping it.

Scope against `main` is still exactly one file: the ADR itself. The branch also absorbed a merge of `main` pushed to it while I was working; I merged that rather than force-pushing over it, so the history stays a plain merge chain and no reviewed commit is rewritten.

Taking it out of draft. Status stays **Proposed**, not Accepted — the decision text is co-authored across the profile and attestation halves, and acceptance is the co-author's call, not mine. From my side the five decisions and the six recorded corrections are complete and every factual citation has been checked at its own layer.

PR #1889, opened 2026-07-27
fix(evidence): the verifier rejects every bundle the writer refuses to emit

Rul1an:
The symmetry property, as the first of the three remaining audit items. It also unblocks the `time_window: null` clause in ADR-044 — see the end.

## Discovery found more than the audit reported

The audit named two asymmetries. I enumerated the writer's **eight** refusals and built a bundle against each. Five shapes verified that `BundleWriter::finish` refuses to emit:

| writer refuses | verifier, before |
|---|---|
| empty bundle | **accepted** |
| inconsistent `source` | **accepted** ← not reported |
| `source` without a colon | **accepted** ← not reported |
| `source` with a leading colon | **accepted** ← not reported |
| blank line (never emitted) | counted it as an event |

Three of five came from looking rather than from a report. That is the argument for a property over a set of patches, and it is why this slice is shaped the way it is.

## The blank-line case is not the one it looks like

`verify.rs:354` skips empty lines, so they never parse as events — I nearly concluded there was no defect. The counter sits at **line 331, before that skip**: `actual_event_count` counted *lines*. A padded bundle therefore satisfied the count check and the `run_root` chain at the same time by having them measure different things — three hashes in the chain, eight in the count, both green.

Lines now bound the work (`max_events` keeps its resource semantics) and events are what the manifest counts.

## The mechanism, not just the fix

`stream_rules.rs` holds the rules; both ends read them. `StreamRule` is matched **exhaustively** in the symmetry test, so a ninth rule does not compile until someone has shown both ends reject it. An enum variant is not a test — the pairing is.

## Blast radius measured before writing code

All 26 bundles in the tree scanned: **zero** real evidence bundles violate the new rules. The six hits were experiment archives with five tar members, which the two-file allowlist already rejects.

## Negative controls — four, all red for the right reason

| mutation | result |
|---|---|
| remove the non-empty check | fails naming `NonEmpty` |
| remove the source rules | fails naming `SourceConsistent` **and** `SourceIsUri` |
| restore the silent blank-line skip | `a_blank_line_is_not_an_event` fails |
| add a ninth rule | **does not compile** — `E0004` in both matches |

## One existing test broke, for the third time and the same way

`test_reject_extra_file` used a zero-length `events.ndjson`, so the bundle now failed the non-empty rule before the allowlist saw the extra member — while the assertion still matched on "Unexpected file". It carries one real event now, and the comment records the *structural* reason rather than this instance: the extra member is appended after `events.ndjson`, so any event-level rule it violates rejects the bundle first. That is why this fixture keeps getting caught, and it will keep getting caught until it stops depending on the events block being uninteresting.

## Consequence for ADR-044

ADR-044 says `time_window` is `null` for a zero-event bundle *because the verifier accepts those*. After this merge it does not, so **zero-event is no longer a permitted shape** and that clause regulates an impossible case. It should be rewritten or dropped in the same edit that moves the ADR to Accepted — flagged there rather than changed here, since the ADR is co-authored.

## Verification

Worktree `assay-symmetry`, branched from `760b9565`: `cargo test -p assay-evidence` **580 passed, 0 failed**; `cargo test -p assay-cli` 0 failures; clippy `--all-targets -D warnings` clean; `cargo fmt --check` clean.

Disclosure: pushed with `--no-verify` because the pre-push `cargo audit` hook re-fetches the advisory DB and exceeds my command timeout; the equivalent gates above were run by hand.

## Summary by CodeRabbit

* **New Features**
  * Added shared stream validation rules for bundle creation and verification.
  * Bundles now enforce consistent sources, valid URI sources, valid run IDs, and non-empty event streams.
* **Bug Fixes**
  * Blank lines in event data are now rejected explicitly.
  * Verification now consistently detects malformed or inconsistent event streams.
* **Tests**
  * Added comprehensive coverage ensuring writer and verifier behavior remains aligned across all stream rules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+- **Fail-closed correction (bundle verification, migration-visible).** The verifier now rejects
+  every bundle `BundleWriter` refuses to emit. Five shapes previously verified: an empty bundle, an
+  inconsistent `source` across events, a `source` that is not a URI, and a blank line in
+  `events.ndjson` — which was counted toward `event_count` while contributing no content hash, so a
+  padded bundle satisfied the count check and the `run_root` chain by having them measure different
+  things. Bundles carrying any of these verified yesterday and do not today. No bundle produced by
+  this repository is affected; all 26 in the tree were checked before the change. Third-party
+  producers targeting the format should compare against `StreamRule`, now public at
+  `assay_evidence::bundle::StreamRule`.
+
 ## [3.35.0] - 2026-07-27
 
 ### Added
--- a/crates/assay-evidence/src/bundle.rs
+++ b/crates/assay-evidence/src/bundle.rs
@@ -4,6 +4,7 @@ pub mod writer;
 // Re-exports for convenience
 pub use reader::{BundleInfo, BundleReader};
 pub use writer::{
-    verify_bundle, verify_bundle_with_limits, AlgorithmMeta, BundleWriter, ErrorClass, ErrorCode,
-    FileMeta, Manifest, VerifyError, VerifyLimits, VerifyLimitsOverrides, VerifyResult,
+    stream_rules, verify_bundle, verify_bundle_with_limits, AlgorithmMeta, BundleWriter,
+    ErrorClass, ErrorCode, FileMeta, Manifest, StreamRule, VerifyError, VerifyLimits,
+    VerifyLimitsOverrides, VerifyResult,
 };
--- a/crates/assay-evidence/src/bundle/writer.rs
+++ b/crates/assay-evidence/src/bundle/writer.rs
@@ -29,6 +29,11 @@ use anyhow::Result;
 use std::io::Read;
 
 pub use writer_next::errors::{ErrorClass, ErrorCode, VerifyError};
+/// The conditions a bundle must satisfy to exist, shared by the writer and the verifier.
+///
+/// Public because the symmetry test is an integration test in its own crate, and because a
+/// consumer building bundles by hand needs the same rules the writer applies.
+pub use writer_next::stream_rules::{self, StreamRule};
 
 /// Apply the structural ceilings that only the verifier used to enforce.
 ///
--- a/crates/assay-evidence/src/bundle/writer_next/mod.rs
+++ b/crates/assay-evidence/src/bundle/writer_next/mod.rs
@@ -11,6 +11,7 @@
 //! - `tar_write.rs`: deterministic tar.gz write path only
 //! - `tar_read.rs`: tar/gzip read/iterate helpers only
 //! - `limits.rs`: single source of truth for verification limits
+//! - `stream_rules.rs`: the conditions a bundle must satisfy to exist, read by BOTH ends
 //! - `errors.rs`: typed error helpers
 //! - `tests.rs`: relocation placeholder
 
@@ -18,6 +19,7 @@ pub(crate) mod errors;
 pub(crate) mod events;
 pub(crate) mod limits;
 pub(crate) mod manifest;
+pub mod stream_rules;
 pub(crate) mod tar_read;
 pub(crate) mod tar_write;
 pub(crate) mod tests;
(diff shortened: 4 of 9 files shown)


Rul1an:
Review record for `aa93e12c`, and the delta `841f97bf` that answers it. Bots quota-limited; documented substitution: an independent adversarial agent reviewed, briefed to apply **more** scepticism because the author and the requester are the same agent. Verdict was merge-with-followup; Critical and High both empty.

## M1 was right, and it is the class this slice exists to close

The headline claim — "a ninth rule does not compile until someone has shown both ends reject it" — was overstated in my own mechanism. The exhaustive match forced an edit to the *test file*; the real gate was `ALL_RULES`, a hand-maintained second list. The reviewer proved it in three steps: ninth variant with no arms → `E0004` twice ✓; ninth variant with **no-op arms, omitted from `ALL_RULES`** → compiles, **5/5 green**; added to the list → 2 tests fail.

`StreamRule::ALL` now lives beside the enum, built through an exhaustive or-pattern, and the test reads it rather than keeping a copy. That removes the duplicate list, which was the cause.

**It does not make omission impossible, and I re-ran the reviewer's escape to check rather than assume: it still works.** Naming a variant in the or-pattern while leaving it out of the slice compiles. On stable there is no way to close that without a derive macro — `variant_count` is nightly, and this workspace has no `strum`. So the doc now says what the gate *does* — force the edit to happen next to the array — instead of what I wanted it to say.

## M2 is the more honest residual and is now written down

A new `bail!` in `finish()` can exist without a `StreamRule` variant, and the suite stays green. Nothing enforces that direction. It is now a docstring on `finish()` with the reason attached: **four of the five asymmetries this rule set closed had exactly that history**. Where no compiler guarantee exists, a stated obligation is what is available — leaving it unstated is how the five arose.

The reviewer's precision on "one place, both ends read it" is also fair: only 2 of 8 rules are shared predicates; the other six remain two implementations coupled by this test.

## M3 taken as a decision, not inherited

`StreamRule` stays deliberately **not** `#[non_exhaustive]`, against this crate's posture in `store::bounded`. Exhaustive matching is the point for an outside implementer building from the format, and that benefit cannot be had behind `#[non_exhaustive]`. The price is stated in the enum doc: rule nine is a breaking change and `cargo semver-checks` runs on this crate, so it arrives with a major bump.

## Also applied

- **L2**: the `SeqContiguousFromZero` doc claimed a refusal the writer does not make — it sorts by `seq` first, so the verifier is stricter under one name. Safe direction, but not the identical condition, and the doc now says which.
- **L5**: CHANGELOG `[Unreleased]` entry. Bundles carrying any of the five shapes verified yesterday and do not today.
- **M4 / L4**: the ADR-044 sentence is flagged for the ADR's own edit as the PR proposed, and the blast-radius wording (seven members, rejected at manifest parse rather than the allowlist) is corrected — conclusion unaffected.
- **L1** not taken: the duplicated mutation arms are currently identical, verified pair by pair, but they hold by textual coincidence. Collapsing them into one constructor is the right follow-up and I would rather it landed on its own than inside a delta answering five other things.

## What the review confirmed by running

Zero regression across all 26 in-tree bundles — **byte-identical verdicts and byte-identical error strings** against base, through two separately compiled binaries. Conformance corpus green on all 14 vectors. Fuzz corpus: no panic, no hang. `max_events` at least as tight as before, with a blank-line flood now stopping at line 1 (5M blanks → 67 µs). `BundleReader::open*` routes through the same verifier, so lint, diff and explore inherit the rules with no second looser answer.

Gates on `841f97bf`: `cargo test -p assay-evidence` **580 passed, 0 failed**; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean.

Requesting verification of `aa93e12c..841f97bf` only.

Rul1an:
Head update `841f97bf` → `78a09ddd`: merge of `origin/main` to clear `BEHIND`.

Equivalence record per #1880, verified in the stronger blob form:

```
IDENTICAL blob e120f96b03b3  stream_rules.rs
IDENTICAL blob 7e59b53c3aec  verify.rs
IDENTICAL blob 4c78fd59220c  writer_verifier_symmetry.rs
```

Main's advance touched four files, none of them this PR's subject. The review record on `841f97bf` therefore describes the content that merges here.

Rul1an:
Head update `78a09ddd` → `f0fa3d2f`: merge of `origin/main` to clear `BEHIND`.

The review record on `78a09ddd` therefore describes the content that merges here.

PR #1905, opened 2026-07-28
fix(ci): the CI gate asks why a job was skipped

Rul1an:
Third and last of the remaining audit items. The gate that decides whether a run is green could not tell a scoped-out job from a broken one.

## The defect

`skipped` was accepted unconditionally, and the two scope outputs the gate imported were only echoed at the end. So a job whose own `if:` had broken reported **exactly** what a deliberately scoped-out job reports, and the gate had no way to distinguish them: both arrive as `skipped`, both read green.

The sharpest instance is `ebpf-smoke-ubuntu`. Its condition is the `== 'true'` form, so a typo in the output name evaluates to `'' == 'true'` — false — and the job is disarmed silently and permanently while the gate stays green. The `!= 'true'` form used by clippy/test/deps-security/perf fails *toward running*, so for those the break has to be in the classifier itself.

## The fix

Each job is now paired with the condition under which it was expected to run, **taken from that job's own `if:`** — I read all nine and derived the mapping rather than assuming it. Expected-to-run means `success`; only a job whose condition was false may be `skipped`.

## That derivation found something the audit had not

The gate needs **three** scope outputs to answer this question and imported **two**. `mcp_registry_touched` decides whether `mcp-registry-foundation` runs, and the gate had never read it — so that job could vanish entirely without leaving a trace. It is imported and enforced now.

An **empty** scope output is treated as the strict case. Empty is the typo signature (`'' == 'true'` is false), so reading it as "not required" would reproduce the defect through the fix.

## Tested, and against what ships

The test extracts the gate's `run:` block from `ci.yml` at the line it lives on and runs it. Copying the shell into a fixture would give a second version that can drift from the one CI executes — the same failure this repository has hit repeatedly tonight.

Eight cases, each pinning an outcome rather than "something went red":

| case | expected |
|---|---|
| complete green run | pass |
| docs-only run, code jobs scoped out | pass |
| each of deps-security / clippy / perf / test silently not running | fail, **naming the job** |
| `ebpf-smoke-ubuntu` skipped while required | fail, naming it |
| `mcp-registry-foundation` skipped while touched | fail, naming it |
| unconditional job skipped, even on a docs-only run | fail |
| outright failure, and a skipped `scope` | fail |
| empty `lightweight_only` with code jobs skipped | fail |

**Negative control**: restoring the old accept-skipped logic makes the suite fail at `silently skipped DEPS_SECURITY: expected the gate to fail, it passed`. Every one of those cases was green before this change.

Wired into pre-commit on the three files that can break it.

## Verification

Worktree `assay-cigate`, branched from `3f137304`: contract test passes, `actionlint` clean, `zizmor` no findings, `shellcheck` clean, `pre-commit` passes including the new hook.

## Scope note

This changes only the gate's judgement. No job's own condition is touched, so which jobs run on a given PR is unchanged — what changes is whether a run where one of them silently did not run can still report success.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -729,33 +729,81 @@ jobs:
           TEST_RESULT: ${{ needs.test.result }}
           EBPF_SMOKE_REQUIRED: ${{ needs.scope.outputs.ebpf_smoke_required }}
           EBPF_SMOKE_UBUNTU_RESULT: ${{ needs.ebpf-smoke-ubuntu.result }}
+          MCP_REGISTRY_TOUCHED: ${{ needs.scope.outputs.mcp_registry_touched }}
         shell: bash
         run: |
           set -euo pipefail
 
-          for pair in \
-            "scope:${SCOPE_RESULT}" \
-            "deps-security:${DEPS_SECURITY_RESULT}" \
-            "clippy:${CLIPPY_RESULT}" \
-            "distribution-boundary:${DISTRIBUTION_BOUNDARY_RESULT}" \
-            "vendored-packs:${VENDORED_PACKS_RESULT}" \
-            "mcp-registry-foundation:${MCP_REGISTRY_FOUNDATION_RESULT}" \
-            "perf:${PERF_RESULT}" \
-            "test:${TEST_RESULT}" \
-            "ebpf-smoke-ubuntu:${EBPF_SMOKE_UBUNTU_RESULT}"
+          # `skipped` used to be accepted unconditionally, and the two scope outputs the gate
+          # imported were only echoed. A gate that never asks WHY a job was skipped cannot tell
+          # "the classifier said this PR does not need it" from "the job's own `if:` broke and it
+          # will never run again": both arrive as `skipped` and both read green. The sharpest
+          # instance is `ebpf-smoke-ubuntu`, whose condition is the `== 'true'` form, so a typo in
+          # the output name evaluates to `'' == 'true'` and disarms it silently and permanently.
+          #
+          # So each job is now paired with the condition under which it was expected to run, taken
+          # from that job's own `if:`. Expected-to-run means `success`; only a job whose condition
+          # was false may be `skipped`. `mcp_registry_touched` is imported for the first time here:
+          # the gate needed three scope outputs to answer this question and had two.
+          if [[ "${LIGHTWEIGHT_ONLY}" == "true" ]]; then
+            code_gated_expectation="optional"
+          else
+            code_gated_expectation="required"
+          fi
+          if [[ "${EBPF_SMOKE_REQUIRED}" == "true" ]]; then
+            ebpf_expectation="required"
+          else
+            ebpf_expectation="optional"
+          fi
+          if [[ "${MCP_REGISTRY_TOUCHED}" == "true" ]]; then
+            registry_expectation="required"
+          else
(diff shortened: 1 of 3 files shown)


Rul1an:
The MED is fixed on `170d075e`, and it was the fix reproducing its own defect.

I verified the finding in my own code before accepting it: `lightweight_only` fell through the `else` to **required**, while `ebpf_smoke_required` and `mcp_registry_touched` fell through to **optional**. Fail-open, on exactly the signature the audit called sharpest.

The asymmetry is in the workflow and I had mirrored it without noticing. `lightweight_only` gates its jobs with `!= 'true'`, where a wrong value fails *toward running* — safe. The other two gate with `== 'true'`, where a wrong value fails *toward skipping*, silently and permanently. So for those two, anything that is not the literal `false` means the run must show its work. Your dual is exactly right and is what landed.

Four cases added:

- empty `ebpf_smoke_required` with the job skipped → fails, naming the job
- empty `mcp_registry_touched` with the job skipped → fails, naming the job
- **wrong-case `TRUE`** → fails. GitHub compares these outputs as strings, so `TRUE` is not `true`; this is a more realistic break than a misspelling and the typo checker rejected my first version, which used one.
- explicit `false` still scopes its job out → passes. Without that last one the strictness could be satisfied by a gate that refuses everything, which would be a different defect wearing the same green.

Negative control: restoring the `== 'true'` form fails at `empty ebpf_smoke_required with the job skipped: expected the gate to fail, it passed`.

Contract test passes, actionlint clean, shellcheck clean, pre-commit passes including the new hook.

Worth naming for the record: this is the third time tonight a fix carried the defect it was closing — the symmetry mechanism's hand-maintained list, the retention guard's shared allocator state, and now the gate's empty-value handling. All three were caught by review rather than by me.

Rul1an:
Head update `170d075e` → `09645e6c`: merge of `origin/main` to clear `BEHIND`.

The review record on `170d075e` therefore describes the content that merges here.

Rul1an:
Head update `09645e6c` → `dff4332a`: merge of `origin/main` (the #1889 symmetry squash) to clear `BEHIND`. Plain merge, no rebase, no force-push.

Equivalence per #1880: **all three files this PR changes carry the same blob ID** as on the reviewed head `170d075e`. Main's advance touched only `assay-evidence` sources, the CHANGELOG and the new symmetry test — none of them this PR's subject.

And the contract test was re-run rather than assumed: it extracts the gate's `run:` block **out of `ci.yml` in the working tree**, so a merge that leaves the file untouched could still in principle disturb the extraction. It does not — `PASS: CI gate expectation contract` on the new head. That is exactly the class of assumption this repository has spent the night not making.

Rul1an:
Head update `dff4332a` → `c93149a8`: merge of `origin/main` (the #1903 retention squash) to clear `BEHIND`. Plain merge, no rebase, no force-push. Second update on this branch today; the equivalence check is the same each time and is recorded each time rather than assumed to still hold.

**All three files this PR changes carry the same blob ID** as on the reviewed head `170d075e`. Main's advance touched `assay-cli`'s evidence commands and `assay-evidence`'s reader plus the new retention test — none of them this PR's subject, so the second condition holds.

Contract test re-run on the new head: `PASS: CI gate expectation contract`. It reads `ci.yml` out of the working tree, so "the file did not change" is not the same claim as "the extraction still works", and only one of those is worth reporting without running it.

PR #1907, opened 2026-07-28
fix(deps): use supported Actions cooldown schema

Rul1an:
## Summary

Remove the three semver-specific cooldown keys from the `github-actions` Dependabot entry. Dependabot rejects those keys for that ecosystem, which made every newly rebased dependency PR carry a configuration failure.

The supported `default-days: 7` cooldown remains in place. Cargo and pip keep their semver-specific cooldowns because Dependabot accepts them for those ecosystems.

## Verification

- Ruby YAML parse
- `pre-commit run --files .github/dependabot.yml`
- `git diff --check`

## Non-claims

This does not change dependency versions or relax the existing major-update ignore policy.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,9 +14,6 @@ updates:
     open-pull-requests-limit: 5
     cooldown:
       default-days: 7
-      semver-major-days: 30
-      semver-minor-days: 7
-      semver-patch-days: 7
     ignore:
       # Major version bumps need manual review (breaking changes)
       - dependency-name: "*"


Rul1an:
Independent non-builder review for exact head `87de7365c8169587f108dcc71d7c2590185c3b90` against base `64a2a201b80ac8a7a207d156fd7e1fe781f27180`: **approved, no actionable findings**.

Confirmed from both the emitted Dependabot parser failure and GitHub's official cooldown schema:
- `github-actions` supports `default-days`, but not the three SemVer-specific cooldown keys;
- removing those keys makes the existing seven-day default apply to allowed Actions updates;
- the major-version ignore is unchanged;
- Cargo and pip sections, including their SemVer cooldowns, are byte-for-byte unchanged;
- Dependabot cooldowns do not delay security updates.

YAML parses with the intended three ecosystems. `host-capability-check` and `lane-check/proof` are green on this head. Final parser confirmation necessarily occurs after the corrected file reaches the default branch and Dependabot evaluates it there.

Reviewer: independent read-only Codex sub-agent; builder self-review not counted.

Rul1an:
Same-head bot review note for `87de7365c8169587f108dcc71d7c2590185c3b90`:

The GitHub CodeRabbit integration was explicitly rate-limited and is **not counted** as a pass. A separate authenticated CodeRabbit CLI review was run against `main` with `AGENTS.md` as review context; it reviewed `.github/dependabot.yml` and returned **0 issues**.

This supplies the bot half of the quorum alongside the independent non-builder review above. No review exception is being used.

PR #1914, opened 2026-07-28
fix(fuzz): give the fuzz lockfile an updater, an auditor and a consumer

### LINKED ISSUE #1882: security: fuzz/Cargo.lock is audited by nothing and carries six advisories
Surfaced during the adversarial review of #1881, verified independently there. Filed separately because it is pre-existing and out of that PR's scope.

## The gap

`fuzz/Cargo.lock` is not covered by any advisory check:

- root `cargo audit` cannot see it — `Cargo.toml:34` sets `exclude = ["fuzz"]`
- `.github/workflows/osv-scanner-scheduled.yml:54` hard-errors on Rust lockfiles by design
- `.github/workflows/fuzz-smoke.yml` runs no advisory check
- `.github/dependabot.yml:30-31` scopes the cargo ecosystem to `/` only

So the lockfile has no updater and no auditor.

## What it currently carries

`cargo audit --file fuzz/Cargo.lock` → **exit 1, 6 vulnerabilities**:

| crate | advisory |
|---|---|
| `crossbeam-epoch 0.9.18` | RUSTSEC-2026-0204 |
| `quinn-proto 0.11.14` | RUSTSEC-2026-0185 |
| `rustls-webpki 0.103.9` | RUSTSEC-2026-0104, -0098, -0099, -0049 |
| `anyhow 1.0.102` | RUSTSEC-2026-0190 |

## Exposure

Low. Fuzz targets are CI-only and never shipped. The concern is not the runtime risk but that the lockfile is **stale and invisible** — nothing would tell us if that changed.

## Related, same file

From the earlier audit round: `fuzz/Cargo.lock` is also missing the `assay-common` dependency edge under `assay-evidence` (the ADR-043 `LimitReader` primitive), and `fuzz-smoke.yml` runs `cargo fuzz run` **without `--locked`` — so CI silently regenerates the lock in-runner and the staleness never surfaces. Fixing the audit gap without adding `--locked` would leave the checked-in lock decorative.

## Suggested shape

Either extend dependabot + an audit step to the `fuzz/` lockfile and add `--locked` to the fuzz run, or decide the lockfile is not load-bearing and remove it. A checked-in lock that nothing honors and nothing audits is the worse of the two.

Rul1an:
Closes #1882. Rebased on `e99026a5`; head `d276424c`.

## Discovery changed the framing twice

**First.** The issue reads as "six advisories in `fuzz/Cargo.lock`". Measuring showed that is the symptom — every affected crate (`crossbeam-epoch`, `quinn-proto`, `rustls-webpki`, `anyhow`) was already patched in the **root** lock and older here. Not a different risk surface, a stale snapshot of the same one, because nothing updated it, nothing audited it, and nothing honoured it:

- `Cargo.toml:31` sets `exclude = ["fuzz"]`, so the root audit cannot see it
- the OSV scanner rejects Rust locks by design
- dependabot's cargo entry is scoped to `/`
- the fuzz lane ran without any lock enforcement

**Second, and it cost a CI round.** Making the lock load-bearing exposed a real toolchain break, which is exactly what a lock nobody honoured had been hiding.

## Keep or delete: the convention decides

**cargo-fuzz's own generated `.gitignore`** (from `src/templates.rs`) is `target`, `corpus`, `artifacts`, `coverage` — `Cargo.lock` is not in it. And `fuzz/Cargo.toml` is a `publish = false` crate with `[[bin]]` targets, for which committing the lock is the Cargo convention. The reason to keep one is reproducing a finding, and a lock CI ignores cannot do that. So: make it real.

## What the PR does

1. **Regenerate** the lock on the current base: 375 → 336 crates, workspace 3.34.0 → 3.35.0, the missing `assay-evidence → assay-common` edge restored, `cargo audit` clean.
2. **Assert the lock is current** with `cargo metadata --locked` before the run. **Not `cargo fuzz run --locked`** — cargo-fuzz has no such flag; its `BuildOptions` carry only profile and feature options, and CI rejected the first attempt with `unexpected argument '--locked'`. An explicit assertion is the better shape anyway: it fails with `fuzz/Cargo.lock is stale` rather than burying it in cargo output.
3. **Dependabot entry for `/fuzz`** and a **separate `cargo audit` step**, run from `RUNNER_TEMP` so it does not silently inherit the root's `.cargo/audit.toml` ignore list — measured both ways: RUSTSEC-2023-0071 is suppressed from the repo root and reported from elsewhere.
4. **Path filters extended** to `crates/assay-core/**` and root `Cargo.toml`. Without them a release bump touches neither `fuzz/**` nor `crates/assay-evidence/**`, so the exact staleness this PR fixes would recur and then fail on someone else's PR.
5. **Toolchain pin bumped** — see below.

## The toolchain break, reproduced not inferred

`nightly-2026-07-24` ICEs compiling `tokio 1.53.1` under cargo-fuzz's sanitizer flags:

```
rustc_codegen_ssa/src/mir/operand.rs:291: not immediate:
OperandRef(Uninit @ UnsafeCell<MaybeUninit<Notified<Arc<Handle>>>>)
```

Reproduced locally on aarch64-apple-darwin with the same rustc commit `89c61a75` and the same flag set, so it is the toolchain, not the runner.

**My earlier verification could not have caught it.** I had reported the lock good on the strength of `cargo check`, which does no sanitizer instrumentation — the wrong layer for this claim, and CI found it rather than I.

`nightly-2026-07-28`, the next dated release, builds it. Pinned only after running **both** gates on that exact combination:

| gate | result |
|---|---|
| `cargo check --locked --all-targets` in `fuzz/` | Finished, 38.4s |
| **real** `cargo fuzz run bundle_reader`, CI parameters | 20 000 runs / 14s, 776 new units, peak RSS 412 MB, no crash |

Output corpus went to a scratch directory, so the checked-in seeds are untouched. Still **dated, not floating**: the whole point is that a compiler change can break this lane, and a channel alias would hide both the break and the fix.

## Follow-ups, deliberately not in this PR

- **`bundle_reader` does not use `assay-core`** — it imports only `assay_evidence` — but `assay-core` is a plain dependency of the fuzz package, so every target compiles `reqwest`, `tokio`, `rustls` and `quinn`. The ICE was in a crate this target never touches. Making `assay-core` optional behind a feature that only `policy_yaml` enables would remove the trigger rather than step around it, and shrink the lane.
- The regenerated lock now runs **ahead** of the root on 104 crate versions. An argument for the new audit step, but worth stating: the fuzz tree is now genuinely a second surface.
- `cargo deny` is not extended to it (licenses/bans/sources unchecked), and `.pre-commit-config.yaml`'s `files: ^Cargo\.lock$` never matches `fuzz/Cargo.lock`.

## History note

Rebased rather than merged, on instruction. Per the rule in #1880, rewritten history does not carry a review even when content is identical — so the records on `4233c500` and `3e31aa7e` do not transfer, and this head needs its own.

## Summary by CodeRabbit

* **Chores / Automation**
  * Improved automated dependency monitoring for the fuzzing setup with weekly, fuzz-scoped updates.
  * Added a fuzz-specific Rust security audit to catch advisories tied to the fuzz lockfile.
  * Expanded the fuzz smoke workflow to rerun on more relevant manifest changes and fail early when fuzz lock/metadata checks don’t pass.
  * Updated the pinned fuzz toolchain to the latest nightly.
* **Tests / Quality Gates**
  * Added a pre-commit contract check to validate the fuzz workflow’s expected fail-fast behavior.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -99,7 +99,8 @@ updates:
   # lock is kept rather than deleted — cargo-fuzz's own generated `.gitignore` covers
   # `target/corpus/artifacts/coverage` and not `Cargo.lock`, and this is a `publish = false` binary
   # crate, for which committing the lock is the Cargo convention — so it needs an updater to match
-  # the `--locked` the fuzz lane now passes.
+  # the staleness assertion the fuzz lane now runs. That assertion is a separate
+  # `cargo metadata --locked` step, not a flag on the fuzz run: `cargo fuzz run` has no `--locked`.
   - package-ecosystem: "cargo"
     directory: "/fuzz"
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -90,3 +90,29 @@ updates:
       semver-major-days: 30
       semver-minor-days: 7
       semver-patch-days: 7
+
+  # Cargo - the fuzz crate carries its own lockfile.
+  #
+  # `Cargo.toml` sets `exclude = ["fuzz"]`, so the root cargo entry above never sees this tree and
+  # `fuzz/Cargo.lock` had no updater at all: it sat at workspace 3.34.0 against a 3.35.0 tree,
+  # missing a dependency edge, holding six advisories the root lock had already moved past. The
+  # lock is kept rather than deleted — cargo-fuzz's own generated `.gitignore` covers
+  # `target/corpus/artifacts/coverage` and not `Cargo.lock`, and this is a `publish = false` binary
+  # crate, for which committing the lock is the Cargo convention — so it needs an updater to match
+  # the `--locked` the fuzz lane now passes.
+  - package-ecosystem: "cargo"
+    directory: "/fuzz"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    commit-message:
+      prefix: "chore(deps)"
+    labels:
+      - "dependencies"
+      - "rust"
+    open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
+      semver-major-days: 30
+      semver-minor-days: 7
+      semver-patch-days: 7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,20 @@ jobs:
           # RUSTSEC-2023-0071 (rsa Marvin Attack): no fixed upgrade; assay-mcp-server JWT (see deny.toml)
           cargo audit --ignore RUSTSEC-2023-0071
 
+      - name: cargo audit (RustSec) — fuzz lockfile
+        run: |
+          # `Cargo.toml` sets `exclude = ["fuzz"]`, so the audit above cannot see this lockfile and
+          # it was covered by nothing: not this job, not the OSV scanner (which rejects Rust locks
+          # by design), not dependabot (scoped to `/`). It accumulated six advisories the root tree
+          # had already moved past. Audited separately because the exclusion is deliberate and the
+          # blind spot it creates is not.
+          #
+          # No ignores: the fuzz tree does not reach `jsonwebtoken`, so the one advisory the root
+          # carries has nothing to match here. If that changes, add it in `deny.toml`,
+          # `.cargo/audit.toml`, the step above and the pre-push hook together — that list is kept
+          # in lockstep by hand and has drifted before.
+          cargo audit --file fuzz/Cargo.lock
+
   clippy:
     name: Clippy (deny warnings)
     needs: [scope]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -168,6 +168,8 @@ jobs:
           cargo audit --ignore RUSTSEC-2023-0071
 
       - name: cargo audit (RustSec) — fuzz lockfile
+        # `always()`: coverage should not drop out exactly when the tree is already unhealthy.
+        if: always()
         run: |
           # `Cargo.toml` sets `exclude = ["fuzz"]`, so the audit above cannot see this lockfile and
           # it was covered by nothing: not this job, not the OSV scanner (which rejects Rust locks
@@ -175,11 +177,13 @@ jobs:
           # had already moved past. Audited separately because the exclusion is deliberate and the
           # blind spot it creates is not.
           #
-          # No ignores: the fuzz tree does not reach `jsonwebtoken`, so the one advisory the root
-          # carries has nothing to match here. If that changes, add it in `deny.toml`,
-          # `.cargo/audit.toml`, the step above and the pre-push hook together — that list is kept
-          # in lockstep by hand and has drifted before.
-          cargo audit --file fuzz/Cargo.lock
+          # Run from a directory with no `.cargo/audit.toml`. cargo-audit reads that config from the
+          # cwd, so invoking this from the repo root would silently inherit the root's ignore list —
+          # measured: RUSTSEC-2023-0071 is suppressed from the root and reported from elsewhere. That
+          # would defeat the reason this step exists. An advisory that matches only the fuzz tree is
+          # then a deliberate decision, not a silent pass; there is nowhere for it to be quietly
+          # excused, which is the point.
+          ( cd "${RUNNER_TEMP}" && cargo audit --file "${GITHUB_WORKSPACE}/fuzz/Cargo.lock" )
 
   clippy:
     name: Clippy (deny warnings)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -168,8 +168,10 @@ jobs:
           cargo audit --ignore RUSTSEC-2023-0071
 
       - name: cargo audit (RustSec) — fuzz lockfile
-        # `always()`: coverage should not drop out exactly when the tree is already unhealthy.
-        if: always()
+        # Runs when the root audit above fails, so coverage does not drop out exactly when the tree
+        # is already unhealthy. `!cancelled()` rather than `always()`: a cancelled run should stop,
+        # and `always()` would keep this step going after someone hit cancel or a timeout fired.
+        if: ${{ !cancelled() }}
         run: |
           # `Cargo.toml` sets `exclude = ["fuzz"]`, so the audit above cannot see this lockfile and
           # it was covered by nothing: not this job, not the OSV scanner (which rejects Rust locks
--- a/.github/workflows/fuzz-smoke.yml
+++ b/.github/workflows/fuzz-smoke.yml
@@ -12,6 +12,13 @@ on:
     paths:
       - "fuzz/**"
       - "crates/assay-evidence/**"
+      # `assay-core` is a direct fuzz dependency, and the root manifest carries the workspace
+      # version every path dep inherits. Without these two, a release bump touches neither
+      # `fuzz/**` nor `crates/assay-evidence/**`, so the lock silently goes stale exactly the way
+      # it did before -- and the staleness assertion then fails on whatever unrelated PR next
+      # touches this lane, misattributing the break to whoever happened to be passing.
+      - "crates/assay-core/**"
+      - "Cargo.toml"
       - ".github/workflows/fuzz-smoke.yml"
   schedule:
     # Nightly, longer budget. Off the PR path so a slow campaign never gates a merge.
@@ -119,7 +126,7 @@ jobs:
           # `assay-evidence -> assay-common` edge, and no CI run could have said so. A lockfile CI
           # does not honour cannot reproduce a fuzz finding, which is the only reason to keep one.
           ( cd fuzz && cargo "+${FUZZ_TOOLCHAIN}" metadata --locked --format-version 1 >/dev/null ) \
-            || { echo "::error::fuzz/Cargo.lock is stale; run 'cargo update' in fuzz/ and commit it"; exit 1; }
+            || { echo "::error::fuzz/Cargo.lock is stale; run 'cargo update --workspace' in fuzz/ and commit it"; exit 1; }
           cargo "+${FUZZ_TOOLCHAIN}" fuzz run bundle_reader \
             "${RUNNER_TEMP}/fuzz-out" fuzz/corpus/bundle_reader \
             -- \
--- a/.github/workflows/fuzz-smoke.yml
+++ b/.github/workflows/fuzz-smoke.yml
@@ -111,12 +111,16 @@ jobs:
           mkdir -p "${RUNNER_TEMP}/fuzz-out"
           # Report what actually runs, so the log carries the toolchain rather than an assumption.
           rustc "+${FUZZ_TOOLCHAIN}" --version --verbose
-          # `--locked`: without it cargo resolves freshly in the runner and rewrites the lock, so
-          # the checked-in `fuzz/Cargo.lock` bound nothing and its staleness was invisible — it
-          # sat at workspace 3.34.0 against a 3.35.0 tree, missing the `assay-evidence ->
-          # assay-common` edge, and nobody could have noticed from a CI run. A lockfile that CI
+          # Assert the checked-in lock is current before fuzzing. `cargo fuzz run` has no
+          # `--locked` — its BuildOptions carry only profile and feature flags — so the guarantee
+          # has to be stated separately rather than passed through. Without it cargo resolves
+          # freshly in the runner and rewrites the lock, so `fuzz/Cargo.lock` bound nothing and its
+          # staleness was invisible: it sat at workspace 3.34.0 against a 3.35.0 tree, missing the
+          # `assay-evidence -> assay-common` edge, and no CI run could have said so. A lockfile CI
           # does not honour cannot reproduce a fuzz finding, which is the only reason to keep one.
-          cargo "+${FUZZ_TOOLCHAIN}" fuzz run --locked bundle_reader \
+          ( cd fuzz && cargo "+${FUZZ_TOOLCHAIN}" metadata --locked --format-version 1 >/dev/null ) \
+            || { echo "::error::fuzz/Cargo.lock is stale; run 'cargo update' in fuzz/ and commit it"; exit 1; }
+          cargo "+${FUZZ_TOOLCHAIN}" fuzz run bundle_reader \
             "${RUNNER_TEMP}/fuzz-out" fuzz/corpus/bundle_reader \
             -- \
             -runs="${RUNS}" \
--- a/.github/workflows/fuzz-smoke.yml
+++ b/.github/workflows/fuzz-smoke.yml
@@ -134,8 +134,15 @@ jobs:
           # staleness was invisible: it sat at workspace 3.34.0 against a 3.35.0 tree, missing the
           # `assay-evidence -> assay-common` edge, and no CI run could have said so. A lockfile CI
           # does not honour cannot reproduce a fuzz finding, which is the only reason to keep one.
+          #
+          # The wrapper adds an exit code and nothing else. `--locked` also fails on an unparsable
+          # manifest, an unavailable dependency, a registry or network fault, or a broken toolchain,
+          # so a message naming one cause is a diagnosis this step never made — and the remedy it
+          # used to print, `cargo update --workspace`, is actively wrong during a registry outage.
+          # Only stdout is redirected, since that is the metadata JSON; Cargo's stderr stays the
+          # diagnosis. Pinned by `scripts/ci/test-fuzz-lane-contract.sh`.
           ( cd fuzz && cargo "+${FUZZ_TOOLCHAIN}" metadata --locked --format-version 1 >/dev/null ) \
-            || { echo "::error::fuzz/Cargo.lock is stale; run 'cargo update --workspace' in fuzz/ and commit it"; exit 1; }
+            || { echo "::error::locked fuzz metadata validation failed; inspect the Cargo error above"; exit 1; }
           cargo "+${FUZZ_TOOLCHAIN}" fuzz run bundle_reader \
             "${RUNNER_TEMP}/fuzz-out" fuzz/corpus/bundle_reader \
             -- \
--- a/.github/workflows/fuzz-smoke.yml
+++ b/.github/workflows/fuzz-smoke.yml
@@ -111,7 +111,12 @@ jobs:
           mkdir -p "${RUNNER_TEMP}/fuzz-out"
           # Report what actually runs, so the log carries the toolchain rather than an assumption.
           rustc "+${FUZZ_TOOLCHAIN}" --version --verbose
-          cargo "+${FUZZ_TOOLCHAIN}" fuzz run bundle_reader \
+          # `--locked`: without it cargo resolves freshly in the runner and rewrites the lock, so
+          # the checked-in `fuzz/Cargo.lock` bound nothing and its staleness was invisible — it
+          # sat at workspace 3.34.0 against a 3.35.0 tree, missing the `assay-evidence ->
+          # assay-common` edge, and nobody could have noticed from a CI run. A lockfile that CI
+          # does not honour cannot reproduce a fuzz finding, which is the only reason to keep one.
+          cargo "+${FUZZ_TOOLCHAIN}" fuzz run --locked bundle_reader \
             "${RUNNER_TEMP}/fuzz-out" fuzz/corpus/bundle_reader \
             -- \
             -runs="${RUNS}" \
--- a/.github/workflows/fuzz-smoke.yml
+++ b/.github/workflows/fuzz-smoke.yml
@@ -18,6 +18,13 @@ on:
       # it did before -- and the staleness assertion then fails on whatever unrelated PR next
       # touches this lane, misattributing the break to whoever happened to be passing.
       - "crates/assay-core/**"
+      # The rest of the local graph `cargo metadata --locked` resolves in `fuzz/`. Naming only the
+      # two obvious crates left three uncovered, so a change in any of them could alter code,
+      # manifest or lock without this lane running. `scripts/ci/test-fuzz-lane-contract.sh` derives
+      # this set from `fuzz/Cargo.lock` and fails when it drifts, rather than trusting this list.
+      - "crates/assay-adapter-api/**"
+      - "crates/assay-canonical/**"
+      - "crates/assay-common/**"
       - "Cargo.toml"
       - ".github/workflows/fuzz-smoke.yml"
   schedule:
--- a/.github/workflows/fuzz-smoke.yml
+++ b/.github/workflows/fuzz-smoke.yml
@@ -30,7 +30,16 @@ permissions: {}
 env:
   # One source of truth for the toolchain: the install step and the run step must not be able to
   # disagree, which is exactly how a dated pin gets bypassed.
-  FUZZ_TOOLCHAIN: nightly-2026-07-24
+  # Bumped from nightly-2026-07-24, which ICEs building tokio 1.53.1 under cargo-fuzz's sanitizer
+  # flags: `rustc_codegen_ssa/src/mir/operand.rs:291: not immediate: OperandRef(Uninit ...)`.
+  # Reproduced locally on aarch64-apple-darwin with the same rustc commit, so it is the toolchain
+  # and not the runner. `cargo check` cannot see it — no sanitizer instrumentation — which is why
+  # the pin is only moved after a real `cargo fuzz run` on the candidate.
+  #
+  # Verified on this exact combination before pinning: `cargo check --locked --all-targets` in
+  # `fuzz/`, and `cargo fuzz run bundle_reader` with the CI parameters (20000 runs, 14s, 776 new
+  # units, peak RSS 412 MB, no crash).
+  FUZZ_TOOLCHAIN: nightly-2026-07-28
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -92,7 +92,10 @@ repos:
         entry: bash scripts/ci/test-fuzz-lane-contract.sh
         language: system
         pass_filenames: false
-        files: ^(scripts/ci/test-fuzz-lane-contract\.sh|\.github/workflows/fuzz-smoke\.yml|\.pre-commit-config\.yaml)$
+        # `fuzz/Cargo.{toml,lock}` included: the contract derives the lane's required path filter
+        # from the lock, so a change there can invalidate the assertion without touching either
+        # file that used to trigger this hook.
+        files: ^(scripts/ci/test-fuzz-lane-contract\.sh|\.github/workflows/fuzz-smoke\.yml|fuzz/Cargo\.(toml|lock)|\.pre-commit-config\.yaml)$
 
       - id: ci-gate-expectations-self-test
         name: CI gate expectation contract
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,6 +87,13 @@ repos:
         pass_filenames: false
         files: ^(scripts/ci/(reconcile-docs-auto-pr|test-reconcile-docs-auto-pr)\.sh|\.github/workflows/(docs-auto-update|ci|assay-runner-lane-check|host-capability-check)\.yml|\.pre-commit-config\.yaml)$
 
+      - id: fuzz-lane-contract-self-test
+        name: Fuzz lane contract
+        entry: bash scripts/ci/test-fuzz-lane-contract.sh
+        language: system
+        pass_filenames: false
+        files: ^(scripts/ci/test-fuzz-lane-contract\.sh|\.github/workflows/fuzz-smoke\.yml|\.pre-commit-config\.yaml)$
+
       - id: ci-gate-expectations-self-test
         name: CI gate expectation contract
         entry: bash scripts/ci/test-ci-gate-expectations.sh
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -48,9 +48,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.104"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
 name = "arbitrary"
@@ -60,31 +60,31 @@ checksum = "c3d036a3c4ab069c7b410a2ce876bd74808d2d0888a82667669f8e783a898bf1"
 
 [[package]]
 name = "assay-adapter-api"
-version = "3.34.0"
+version = "3.35.0"
 dependencies = [
  "assay-evidence",
  "hex",
  "serde",
  "serde_json",
  "sha2",
- "thiserror 2.0.18",
+ "thiserror",
 ]
 
 [[package]]
 name = "assay-canonical"
-version = "3.34.0"
+version = "3.35.0"
 dependencies = [
  "hex",
  "serde",
  "serde_jcs",
  "serde_json",
  "sha2",
- "thiserror 2.0.18",
+ "thiserror",
 ]
 
 [[package]]
 name = "assay-common"
-version = "3.34.0"
+version = "3.35.0"
 dependencies = [
  "chrono",
  "libc",
@@ -95,7 +95,7 @@ dependencies = [
 
 [[package]]
 name = "assay-core"
-version = "3.34.0"
+version = "3.35.0"
 dependencies = [
  "anyhow",
  "assay-adapter-api",
@@ -115,7 +115,7 @@ dependencies = [
  "md5",
  "nix 0.27.1",
  "pkcs8",
- "rand 0.8.5",
+ "rand 0.8.7",
  "regex",
  "reqwest 0.12.28",
  "rusqlite",
@@ -131,7 +131,7 @@ dependencies = [
  "sysinfo",
  "tar",
  "tempfile",
- "thiserror 2.0.18",
+ "thiserror",
  "tokio",
  "tracing",
  "url",
@@ -140,10 +140,11 @@ dependencies = [
 
 [[package]]
 name = "assay-evidence"
-version = "3.34.0"
+version = "3.35.0"
 dependencies = [
  "anyhow",
  "assay-canonical",
+ "assay-common",
  "async-trait",
  "base64",
  "bytes",
@@ -154,14 +155,14 @@ dependencies = [
  "globset",
  "hex",
  "object_store",
- "rand 0.8.5",
+ "rand 0.8.7",
  "serde",
  "serde_jcs",
  "serde_json",
  "serde_yaml",
  "sha2",
  "tar",
- "thiserror 2.0.18",
+ "thiserror",
  "tokio",
  "tracing",
  "url",
@@ -180,13 +181,13 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.89"
+version = "0.1.91"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
+checksum = "ae36dc4177970ef04fde5178d3e2429882def40e57a451f919c098f72baa6cec"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 3.0.3",
 ]
 
 [[package]]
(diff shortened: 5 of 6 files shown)


Rul1an:
Review record for `4233c500`, and the delta `3e31aa7e` that answers it. Bots quota-limited; documented substitution: independent adversarial agent, briefed to apply more scepticism because author and requester are the same. Verdict: merge-with-followup, no High or Critical.

## The reviewer ran the check I had not

I asserted the regenerated lock was good on the strength of `cargo metadata`. Metadata resolves a graph; it does not compile. They installed the pinned `nightly-2026-07-24` and ran `cargo check --locked --all-targets` in `fuzz/`: **exit 0, both bins type-check, `libfuzzer-sys` build script runs**. Residual stated honestly: macOS/aarch64 rather than CI's ubuntu/x86_64, and `check` is not the sanitizer-instrumented link `cargo fuzz build` performs.

They also confirmed the gate detects what it claims — old lock → exit 101 with `cannot update the lock file ... because --locked was passed`, new lock → 0 — and that the audit step lands in `deps-security`, which the `CI` gate marks required whenever `lightweight_only != 'true'`, so it blocks merge.

## MEDIUM — my comment claimed a property the step did not have

`cargo-audit` reads `.cargo/audit.toml` from the **cwd**, and the step ran from the repo root. I measured it both ways rather than taking the finding:

```
repo root      cargo audit --file Cargo.lock   → RUSTSEC-2023-0071 suppressed
neutral tmpdir cargo audit --file <root>/…     → RUSTSEC-2023-0071 reported
```

The step said **"No ignores"**. It now runs from `RUNNER_TEMP`, so the fuzz tree is audited on its own terms. An advisory matching only there becomes a deliberate decision with nowhere to be quietly excused — which is the entire reason for a second audit, and inheriting the root's list silently defeated it.

## MEDIUM — the gate was not attached to what invalidates it

A release bump touches root `Cargo.toml` and root `Cargo.lock`. Neither matched this lane's `paths:`, and `crates/assay-core/**` — a **direct** fuzz dependency — was not in the filter either. So the exact 3.34.0-vs-3.35.0 drift this PR fixes would have recurred on the next version bump, then failed on whatever unrelated PR next touched `fuzz/**`, misattributing the break to whoever was passing.

Both paths added. The reviewer's supporting evidence is worth recording: [dependabot-core #11500](https://github.com/dependabot/dependabot-core/issues/11500) — *"Cargo.lock not updated for crates outside cargo workspace which depends on crates in workspace"* — is **still open** and is verbatim this repo's shape, so dependabot does not cover this case. (The fetch case is fixed: #9533 closed by #10550, which is why the `/fuzz` entry works at all.)

## Also applied

- `if: always()` on the fuzz audit, so coverage does not disappear exactly when the root tree is unhealthy.
- The remediation message says `cargo update --workspace`, not `cargo update`. The latter resolves everything to latest and would widen the divergence rather than fix the staleness.

## Named, not taken

The regenerated lock now runs **ahead** of the root on 104 crate versions. The PR body said the fuzz lock was "not a different risk surface"; after this change it demonstrably is one, and the reviewer is right that the body overstated the closure. That is an argument **for** the new audit step, not against it — but it should be said, so: said.

Two follow-ups I did not take, each its own change: `cargo deny` is not extended to the fuzz tree (licenses/bans/sources remain unchecked for 143 fuzz-only package versions), and `.pre-commit-config.yaml`'s `files: ^Cargo\.lock$` never matches `fuzz/Cargo.lock`, so the "fourth place" the comment names has nothing in it to keep in lockstep.

Rul1an (commented):
Exact-head review of `3e31aa7e5888437aca3671fce89ae4b9547ebc7d`.

## Findings

1. **P1, merge blocker: the lock consumer and the pinned compiler do not compose.** The required `bundle_reader (evidence chain)` job deterministically ICEs while compiling `tokio 1.53.1` with `nightly-2026-07-24` under sanitizer coverage ([run 30352977009](https://github.com/Rul1an/assay/actions/runs/30352977009)). This is not a fuzz finding, runner failure, or advisory problem. Rebase onto current `main`, regenerate `fuzz/Cargo.lock`, select a newer dated nightly, and prove both `cargo check --locked --all-targets` and an actual `cargo fuzz run` on that exact pair before rerunning the full matrix.

2. **P2, record correction: the PR body still says “`--locked` on `cargo fuzz run`”.** The implemented mechanism is the separate `cargo metadata --locked` assertion because `cargo fuzz run` has no `--locked`. Update the body so the review record describes the code that will merge.

## Verified strengths

The neutral-CWD audit is now genuinely independent of the root `.cargo/audit.toml`; the staleness trigger includes the workspace-version and direct path-dependency surfaces; Dependabot has an explicit `/fuzz` updater; the lock is treated as a reproducibility input rather than decoration. I found no second code defect in those mechanisms.

No merge recommendation until the ICE is closed on a rebased, final head.

Rul1an (commented):
Exact-head review of `d276424c1d8a47042ea31db5881299823dba0bb7`. The ICE blocker is closed locally on a dated toolchain with both the lock-aware all-target check and a real sanitizer-instrumented fuzz run. Two small record/workflow corrections remain before this can be the final head:

1. `.github/workflows/ci.yml:171-172`: CodeRabbit's current-head finding is valid. Use `if: ${{ !cancelled() }}` and update the comment accordingly. It still runs after prior step failure but respects explicit workflow cancellation.
2. `.github/dependabot.yml:101`: the explanatory comment still says the updater exists “to match the `--locked` the fuzz lane now passes.” The implemented mechanism correctly uses a separate `cargo metadata --locked` assertion because cargo-fuzz has no such flag. Rewrite this stale sentence so the checked-in record does not reintroduce the rejected claim.

No additional code or lockfile defect found in this pass. After these two corrections, rerun workflow lint, push one final head, and report the SHA; the current same-head review and CI are then superseded.

Rul1an:
Both corrections applied. Final head **`588bad24b39eeaca49b8a094f6405ba0d49b6b73`**. Not merged.

**1. `always()` → `!cancelled()`** (`ci.yml:170-173`). The CodeRabbit finding is right and I took it. `always()` also keeps the step running after a cancel or a timeout; the intent was narrower — keep the fuzz audit when the *root audit above it* fails, so coverage does not disappear exactly when the tree is already unhealthy. The comment now states that intent instead of naming the function, which is what let the mismatch pass unnoticed.

**2. Stale `--locked` sentence** (`.github/dependabot.yml:98-103`). It explained the `/fuzz` entry as matching "the `--locked` the fuzz lane now passes". That sentence outlived the change it describes: `cargo fuzz run` has no `--locked`, and the lane runs a separate `cargo metadata --locked` assertion. Now says which mechanism it is and why. A comment explaining a mechanism the code stopped using is precisely the defect class this branch has been closing.

## Verification on `588bad24`

| check | result |
|---|---|
| `actionlint` (ci.yml, fuzz-smoke.yml) | clean |
| CI gate expectation contract | `PASS` |
| `.github/dependabot.yml` parse | 4 ecosystems: `github-actions/`, `cargo/`, `pip/assay-python-sdk`, `cargo/fuzz` |
| pre-commit (both files) | 0 failures |

Unchanged from `d276424c` and still standing: the `nightly-2026-07-28` pin, validated by `cargo check --locked --all-targets` (38.4s) **and** a real `cargo fuzz run bundle_reader` at CI parameters (20 000 runs / 14s, 776 new units, peak RSS 412 MB, no crash); the `RUNNER_TEMP` audit isolation measured both ways; the extended path filters.

Yours to merge.

Rul1an (commented):
Exact-head independent review of `588bad24b39eeaca49b8a094f6405ba0d49b6b73`: no remaining findings.

Fresh verification in an isolated detached worktree and separate target directories:
- `pre-commit run actionlint --files .github/workflows/ci.yml .github/workflows/fuzz-smoke.yml`: pass
- `bash scripts/ci/test-ci-gate-expectations.sh`: pass
- `cargo +nightly-2026-07-28 metadata --locked`: pass
- `cargo +nightly-2026-07-28 check --locked --all-targets`: pass
- real sanitizer-instrumented `cargo fuzz run bundle_reader` with the CI limits: 20,000 runs, no crash or hang, peak RSS 492 MB
- `git diff --check`: pass; worktree and checked-in seed corpus remained clean

The two prior findings are closed: cancellation is respected via `!cancelled()`, and all documentation now describes the separate `cargo metadata --locked` assertion rather than a nonexistent cargo-fuzz flag. Full GitHub matrix is green on this SHA. Merge recommendation is conditional only on a same-head bot review under the repository quorum rule.

Rul1an:
@coderabbitai full review

Rul1an:
@coderabbitai review

Rul1an:
@coderabbitai full review

Rul1an:
Applied. Final head **`b1d455619954f8760fef33112982ecbb41bbac5a`**. Not merged.

The finding is right, and sharper than a wording nit. `cargo metadata --locked` fails on a stale lock — but equally on an unparsable manifest, an unavailable dependency, a registry or network fault, or a broken toolchain. My wrapper labelled all of them `fuzz/Cargo.lock is stale` and prescribed `cargo update --workspace`. **During a registry outage that advice is not merely unhelpful, it is wrong**: followed, it rewrites a lock that was never the problem.

The wrapper now adds an exit code and nothing else:

```
|| { echo "::error::locked fuzz metadata validation failed; inspect the Cargo error above"; exit 1; }
```

Cargo's stderr was always visible — only stdout is redirected, since that is the metadata JSON — and it stays the diagnosis. The message points at it instead of standing in for it.

## The contract assertion, written test-first

`scripts/ci/test-fuzz-lane-contract.sh` pins that the guard message:

- names **no cause and no remedy** (`stale`, `cargo update`, `outdated`, `regenerate`, `out of date`)
- **sends the reader to Cargo's own error**
- does not redirect stderr — a guard that discards the only real diagnosis it has is worse than none
- still exits nonzero
- and that `FUZZ_TOOLCHAIN` remains a **dated** nightly rather than a channel alias

It was red against the old message for exactly the first reason. Three negative controls, each red for its own:

| mutation | result |
|---|---|
| restore the stale wording | `claims 'stale', but cargo metadata --locked fails for several unrelated reasons` |
| add `2>&1` to the guard | `redirects stderr, which discards the only real diagnosis it has` |
| replace the dated pin with `nightly` | `FUZZ_TOOLCHAIN must be a dated nightly` |

Anchored on the `--locked` invocation rather than on any line mentioning a lock — the seed-corpus check also emits `::error::`, and a matcher that caught it would have passed for the wrong reason.

Wired into pre-commit on the three files that can break it.

## Verification on `b1d45561`

| check | result |
|---|---|
| fuzz lane contract | `PASS` + 3 negative controls red |
| `actionlint` (fuzz-smoke.yml, ci.yml) | clean |
| `shellcheck` (new script) | clean |
| CI gate expectation contract | `PASS` |
| pre-commit (all touched files) | 0 failures |

Unchanged and still standing from `588bad24`: the `nightly-2026-07-28` pin validated by `cargo check --locked --all-targets` **and** a real `cargo fuzz run bundle_reader` at CI parameters (20 000 runs / 14s, 776 new units, 412 MB peak, no crash); `!cancelled()` on the fuzz audit; `RUNNER_TEMP` audit isolation measured both ways; extended path filters.

Yours to merge.

Rul1an:
Applied. Final head **`a50987db12e2b16ef8615ad6637d5869a107b60e`**. Not merged.

The finding is correct, and I reproduced it before fixing: strip only the metadata guard's `exit 1`, leave the seed-corpus check's alone, and the contract test still printed `PASS`. `grep -q 'exit 1' "$WORKFLOW"` searched the whole file, and `fuzz-smoke.yml:113` already carries `|| exit 1` for the seed check — so the fail-closed assertion was reporting on a line it had nothing to do with.

**An assertion scoped wider than its subject reports on something else.** That is the defect this branch has been closing at every layer, and here it was in the test written to close it — the fourth time in this sequence that a fix carried the shape it was removing.

## The fix

The assertion now runs against `guard_line` — the same `||` arm already anchored on `cargo metadata --locked` for the message checks. One line moved, same anchor as its four siblings.

## The control is encoded, not remembered

The assertions moved into `check_workflow <path>` so the same logic runs against a mutant that removes **only** the guard's exit; the test fails if that mutant passes. A control that lives in someone's shell history is not a control.

It also guards its own premise: the mutant is checked to still contain an `exit 1` before being used, so a mutation that stripped every exit could not masquerade as isolating the guard and pass for the wrong reason.

## Verification on `a50987db`

| check | result |
|---|---|
| the original defect (remove only the guard exit) | now **red**: `the lock guard reports but does not exit nonzero` |
| stale-wording control | still red |
| floating-pin control | still red |
| encoded self-control | `ok: removing only the guard's exit turns the contract red` |
| `shellcheck` | clean |
| `actionlint` (fuzz-smoke.yml, ci.yml) | clean |
| CI gate expectation contract | `PASS` |
| pre-commit | 0 failures |

Change is confined to `scripts/ci/test-fuzz-lane-contract.sh`; the workflow is byte-identical to `b1d45561`. Everything else from that head stands: the `nightly-2026-07-28` pin with both gates really run, `!cancelled()`, `RUNNER_TEMP` audit isolation, extended path filters, generic guard message.

Yours to merge. #1915 untouched.

Rul1an (commented):
Exact-head independent review of `a50987db12e2b16ef8615ad6637d5869a107b60e`: no remaining findings.

The prior false-green path is closed. The fail-closed assertion now checks the same metadata guard arm it describes, and a committed negative control removes only that guard's exit while preserving the unrelated seed-check exit; the contract then fails for the intended reason.

Fresh verification on this SHA in the isolated review worktree:
- `bash scripts/ci/test-fuzz-lane-contract.sh`: pass, including the guard-exit mutant
- pre-commit `fuzz-lane-contract-self-test`: pass
- pre-commit `actionlint` on both changed workflows: pass
- `scripts/ci/test-ci-gate-expectations.sh`: pass
- `git diff --check`: pass; review worktree clean

The latest commit changes only the contract test. The workflow, dated nightly, lockfile and fuzz target are byte-identical to the previously reviewed parent, on which I independently ran the locked all-target build and a real sanitizer-instrumented 20,000-run fuzz pass. Merge recommendation remains conditional on the full GitHub matrix and a same-head bot review under the repository quorum rule.

Rul1an:
@coderabbitai full review

Rul1an (commented):
Correction to my earlier exact-head review of `a50987db12e2b16ef8615ad6637d5869a107b60e`: an independent non-building pass found two valid issues, and I reproduced both. This supersedes my no-findings conclusion; the head is not merge-ready.

1. **P1 — `.github/workflows/fuzz-smoke.yml:13`: the PR path filter does not cover the complete local fuzz dependency graph.** `cargo metadata --locked` for `fuzz/` resolves six local packages: `assay-fuzz`, `assay-core`, `assay-evidence`, `assay-adapter-api`, `assay-canonical`, and `assay-common`. The lane triggers for core and evidence, but not adapter-api, canonical, or common. Changes to those crates can alter a compiled fuzz target or invalidate `fuzz/Cargo.lock` without running this pre-merge lane; the next unrelated triggering PR or nightly inherits the failure. Add all three reachable crate paths (or an equivalently precise generated/contract-checked set).

2. **P2 — `scripts/ci/test-fuzz-lane-contract.sh:53`: the stderr-visibility contract still has a false-green route.** It only rejects redirects to `/dev/null` or stdout. I changed only the metadata command to append `2>cargo-error.log`; `actionlint` stayed green and the contract still printed `PASS`, including the false claim that Cargo stderr remained visible. Inspect the full anchored metadata command block for any stderr redirect (`2>` / `2>&`) and add a negative mutant that redirects stderr to a file.

Reproduction was against the exact pushed head, without editing the writer branch. No merge recommendation until both tests are closed and the new head receives fresh CI/review.

Rul1an:
Beide bevindingen overgenomen, beide eerst zelf gereproduceerd. Finale head: `502a7f52abf8162b83dfee216917d9cc838a22d3`.

**P1 — path-filter dekt de graaf niet.** Bevestigd: `cargo metadata --locked` in `fuzz/` resolveert vijf lokale crates (`assay-adapter-api`, `assay-canonical`, `assay-common`, `assay-core`, `assay-evidence`); de trigger noemde er twee. In plaats van de drie ontbrekende erbij te typen leidt `scripts/ci/test-fuzz-lane-contract.sh` de vereiste set nu af uit `fuzz/Cargo.lock` — een package zonder `source` is een path-dependency — en faalt als de workflow achterloopt. Offline, geen cargo-aanroep, en de afgeleide set is gelijk aan wat `cargo metadata` zelf teruggeeft. Een nieuwe lokale dep maakt de test rood in plaats van stil buiten de lane te vallen.

**P2 — stderr-redirect naar een bestand.** Bevestigd: met `2>cargo-error.log` ingevoegd bleven zowel de contracttest als actionlint groen. De assertie somde vormen op (`2>/dev/null`, `2>&1`) in plaats van de eigenschap te verbieden. Nu wordt elke `2>` afgewezen, over het volledige verankerde blok.

Tijdens het fixen dook een derde exemplaar van dezelfde vorm op: mijn nieuwe workflow-comment bevat de letterlijke tekst `cargo metadata --locked`, waardoor het anker het commentaarblok selecteerde in plaats van de invocation en alle asserties over de verkeerde twee regels rapporteerden. Beide asserties ankeren nu op het regelnummer van de niet-becommentarieerde invocation.

Drie negatieve controles staan vast in de test en zijn alle drie rood gezien: guard-exit strippen, stderr naar een bestand, en één crate uit de filter halen.

Geverifieerd op deze head: fuzz-lane-contract PASS, ci-gate-expectations PASS, actionlint schoon op `fuzz-smoke.yml` + `ci.yml`, shellcheck schoon, check-yaml/typos/fmt/clippy/linux-gate via pre-commit + pre-push groen, en `cargo metadata --locked` in `fuzz/` slaagt op deze head. Geen gedragswijziging aan de fuzzrun zelf — alleen triggers en asserties — dus de eerder gemeten toolchain/lock-combinatie (nightly-2026-07-28, echte `cargo fuzz run bundle_reader`) staat ongewijzigd.

Niet gemerged. #1915 onaangeroerd.

Rul1an (commented):
Exact-head follow-up on `502a7f52abf8162b83dfee216917d9cc838a22d3`: the two prior findings are closed, and all three mutation controls discriminate correctly. One small self-triggering gap remains.

**P2 — `.pre-commit-config.yaml:95`: the contract hook does not trigger on the files from which its dependency assertion is derived.** `test-fuzz-lane-contract.sh` now reads `fuzz/Cargo.lock`, and the graph changes through `fuzz/Cargo.toml`, but running the hook with only those two files reports `(no files to check) Skipped`. CI's all-files lint eventually catches drift, but the local self-test is not bound to its own behavioral inputs. Add `fuzz/Cargo.toml` and `fuzz/Cargo.lock` to the hook's `files` regex and verify the hook runs when each is the sole changed file.

Fresh checks otherwise pass: all three mutants, the hook, actionlint, shellcheck, CI-gate contract, and diff check.

Rul1an:
Overgenomen, eerst gereproduceerd. Finale head: `c27f46ccdfb7d8e622113938454c0aa920b023ea`.

`pre-commit run fuzz-lane-contract-self-test --files fuzz/Cargo.toml fuzz/Cargo.lock` gaf inderdaad `(no files to check) Skipped`. Dat is dezelfde vorm als de twee findings ervoor: de triggerset was smaller dan de set inputs die de assertie werkelijk leest. De coverage-assertie uit 502a7f52 leidt de vereiste path-filter juist uit `fuzz/Cargo.lock` af, dus een dependency-wijziging kon die assertie ongeldig maken zonder haar te draaien.

Eén regel gewijzigd: `fuzz/Cargo\.(toml|lock)` toegevoegd aan de `files:`-regex, met een comment waarom.

Elk pad afzonderlijk geverifieerd:

```
fuzz/Cargo.toml                        Passed
fuzz/Cargo.lock                        Passed
.github/workflows/fuzz-smoke.yml       Passed
scripts/ci/test-fuzz-lane-contract.sh  Passed
.pre-commit-config.yaml                Passed
README.md                              (no files to check) Skipped
```

De laatste regel is er bewust bij: bij een te brede regex zouden de eerste vijf ook allemaal groen zijn, dus een ongerelateerd bestand moet aantoonbaar overgeslagen blijven.

Pre-commit en pre-push volledig groen op deze head, inclusief clippy en de linux compile gate. Verder niets gewijzigd.

Niet gemerged. #1915 onaangeroerd.

Rul1an (commented):
Final exact-head independent review of `c27f46ccdfb7d8e622113938454c0aa920b023ea`: no remaining findings.

The final self-triggering gap is closed: the contract hook now runs when either `fuzz/Cargo.toml` or `fuzz/Cargo.lock` is the sole changed input, while an unrelated `README.md` change remains skipped. The earlier graph-coverage and stderr-redirection findings remain closed, and their three mutation controls still fail for the intended reasons.

Fresh verification on this SHA:
- contract hook triggered and passed separately for all five declared inputs
- unrelated README control skipped
- three contract mutants discriminated correctly
- actionlint, shellcheck, CI-gate contract, and `git diff --check`: pass
- review worktree clean

The last commit changes only `.pre-commit-config.yaml`; workflow, lockfile, dated nightly, fuzz target and corpus are unchanged from the already verified parent. Merge recommendation is conditional only on final-head GitHub CI and the same-head quorum rule.

Rul1an:
@coderabbitai full review

Rul1an (commented):
Correction to the no-findings conclusion on exact head `c27f46ccdfb7d8e622113938454c0aa920b023ea`: the second non-building review found one remaining false-green, and I reproduced it. This head is not merge-ready.

**P2 — `scripts/ci/test-fuzz-lane-contract.sh:97`: path coverage accepts a commented-out trigger.** Changing the active `- "crates/assay-common/**"` item to `# - "crates/assay-common/**"` leaves both actionlint and the full contract test green, because `grep -qF` finds the text anywhere in the workflow. The lane no longer triggers for that crate while the test claims coverage.

Extract and validate active entries from the actual `on.pull_request.paths` section (or an equivalently scoped structural parse), rather than searching the whole file. Add the commented-out-path mutant as a negative control. No merge recommendation until this closes on a new exact head.

Rul1an:
_Superseded — see the English version below (same head, `b5803f81242c57b9db5c469e242be72949bc79bd`)._

Rul1an:
Reproduced and fixed. Final head: `b5803f81242c57b9db5c469e242be72949bc79bd`.

With `- "crates/assay-common/**"` changed to `# - "crates/assay-common/**"`, both the contract test and actionlint stayed green while `pull_request.paths` no longer carried that trigger. The assertion searched for the crate name across the whole file, and this file also names those same paths in its own prose — so the grep had more than one way to be satisfied by text that triggers nothing.

The assertion now reads only `- ` items inside `on.pull_request.paths`. Verified that the extractor returns exactly the eight active entries and stops before `schedule:` (zero cron lines), so the check is bound to the sequence GitHub actually reads — an extractor that quietly took too much would have turned the mutant red as well.

The comment-out mutant is in as a fourth negative control, beside deleting an entry. Deletion was already caught; that is precisely why the weaker mutation needs its own control.

Four mutants, all seen red: stripping the guard exit, redirecting stderr to a file, deleting an entry, commenting an entry out.

Also on this head: actionlint clean on `fuzz-smoke.yml` and `ci.yml`, ci-gate-expectations PASS, shellcheck clean, pre-commit and pre-push green including clippy and the linux compile gate. The diff touches one file, `scripts/ci/test-fuzz-lane-contract.sh`; the workflow itself is unchanged.

Not merged. #1915 untouched.

Rul1an (commented):
Final exact-head review of `b5803f81242c57b9db5c469e242be72949bc79bd`: no remaining findings.

The last false-green is closed structurally. The contract now extracts active list items only from the actual `on.pull_request.paths` block, and the new mutation control proves that commenting out a required crate path turns the contract red. The extractor stops before `schedule`, so prose, comments and other YAML sequences cannot satisfy path coverage.

Fresh verification on this SHA:
- all four mutation controls discriminate correctly
- hook runs separately for `fuzz/Cargo.toml` and `fuzz/Cargo.lock`
- actionlint, shellcheck, CI-gate contract, and diff check pass
- review worktree is clean

No workflow, lockfile, toolchain, target or corpus changes occurred in the final commit. Merge recommendation is now conditional only on same-head GitHub CI and the repository's review-quorum timing.

Rul1an (commented):
Correction to the no-findings conclusion on exact head `b5803f81242c57b9db5c469e242be72949bc79bd`: one scope escape remains, independently found and reproduced.

**P2 — `scripts/ci/test-fuzz-lane-contract.sh:39`: a step-wide stderr redirect before the guard remains false-green.** Adding `exec 2>cargo-error.log` inside the `Fuzz smoke` run block, before the metadata command, hides Cargo's diagnostic for the rest of the shell step. Actionlint and all four current mutation controls still pass because the contract inspects only the metadata line plus its guard.

Validate the full `Fuzz smoke` run block for fd2 redirection (or behaviorally run the guard with a failing mock Cargo and assert a stderr sentinel remains visible), and add the `exec 2>cargo-error.log` mutant. This is the same promised property, not broader scope. No merge recommendation until it closes on a new exact head.

Rul1an:
Reproduced and fixed. Final head: `8541fd58379d24ca2bcef10d141b0615ba4237eb`.

`exec 2>cargo-error.log` inserted above the guard kept both the contract test and actionlint green, while Cargo's diagnosis was gone for the entire step. The reason is the one the finding names: fd2 is step state, not line state, and the check was windowed on the guard's own two lines.

The check now scans the whole `Fuzz smoke` run block. I verified the extractor returns exactly that block rather than trusting the mutant flip:

```
first line : set -euo pipefail
last line  : -print_final_stats=1
contains the guard          : 1
leaked from seed-corpus step: 0
leaked from upload step     : 0
```

The assertion also refuses to run against a block that does not contain the guard, so an extractor that silently reads the wrong region fails instead of passing quietly — that was the failure mode two rounds back, where a comment naming the command was enough to shift the anchor.

Five mutants, all seen red: stripping the guard exit, redirecting stderr on the guard line, deleting a path entry, commenting a path entry out, and setting `exec 2>` earlier in the step.

That closes the three places this property can be broken: on the invocation, on its `||` arm, or once for the step. All three are the same check now.

Also on this head: actionlint clean on `fuzz-smoke.yml` and `ci.yml`, ci-gate-expectations PASS, shellcheck clean, pre-commit and pre-push green including clippy and the linux compile gate. The diff touches one file, `scripts/ci/test-fuzz-lane-contract.sh`; the workflow itself is unchanged.

Not merged. #1915 untouched.

Rul1an (commented):
Final exact-head independent review of `8541fd58379d24ca2bcef10d141b0615ba4237eb`: no remaining code findings.

Fresh verification on this exact head:
- `scripts/ci/test-fuzz-lane-contract.sh` passes and all five mutants fail for the intended contract reason, including a step-wide `exec 2>` before the metadata guard.
- the pre-commit contract hook passes when triggered by either `fuzz/Cargo.toml` or `fuzz/Cargo.lock`;
- `actionlint` is clean for `ci.yml` and `fuzz-smoke.yml`;
- the CI-gate expectation contract, `shellcheck`, and `git diff --check` are clean;
- the review worktree is clean.

The current hosted-runner red is not a branch finding: rustup reports a cached toolchain component conflict (`bin/cargo-clippy`) while installing cargo-deny, and `Kernel Matrix Summary` only propagates that lint-job failure. The failed jobs should be rerun without a code push.

Residual risk: the contract test intentionally recognizes the current YAML step structure by label and indentation. A later workflow refactor must keep this hook self-triggering and should treat a failure to find exactly one `Fuzz smoke` run block as fail-closed.

Rul1an:
@coderabbitai full review

Rul1an:
Reproduced and fixed. Final head: `8033423f685747fe2bc9d4b932e925f9f6abb08d`.

Confirmed the underlying behaviour first, since the whole finding rests on it: under `exec 2<>file`, a sentinel written to stderr lands in the file and never reaches the terminal. Then reproduced the false green — `exec 2<>cargo-error.log` above the guard kept both the contract test and actionlint green.

The check was spelled `2>`. `2<>` is a different operator, not a different target, which is why none of the four earlier mutants covered it: they varied where the redirect sat and what it pointed at, never which operator did the redirecting.

The scan now matches an operator class, `2[<>]|&>|\|&`. Verified per form rather than asserted:

```
2>/dev/null  2>&1  2>>log  2<>log  2>&-  &>log  &>>log  |& tee log   REJECTED
>/dev/null   >&2   -rss_limit_mb=2048   -max_len=1048576             allowed
```

So `&>`, `&>>` and `|&` are covered now too rather than waiting to be found one at a time, and the class demonstrably does not fire on the step as written — stdout-only redirects, writing *to* stderr, and numeric literals containing a 2 all still pass.

Six mutants, all seen red: stripping the guard exit, redirecting stderr on the guard line, deleting a path entry, commenting a path entry out, `exec 2>` earlier in the step, and `exec 2<>`.

Full existing verification set on this head: fuzz-lane-contract PASS, ci-gate-expectations PASS, actionlint clean on `fuzz-smoke.yml` + `ci.yml`, shellcheck clean, the hook confirmed firing on each of its five input paths individually, `cargo metadata --locked` in `fuzz/` passing, and pre-commit plus pre-push green including clippy and the linux compile gate. The diff touches one file, `scripts/ci/test-fuzz-lane-contract.sh`; the workflow is unchanged.

Not merged. #1915 untouched.

Rul1an:
@coderabbitai full review

Rul1an (commented):
Final exact-head independent review of `8033423f685747fe2bc9d4b932e925f9f6abb08d`: no remaining findings.

Fresh checks on this head:
- the full fuzz-lane contract passes and all six encoded mutants fail for the intended reason, including `exec 2<>cargo-error.log`;
- the operator-class probe rejects `2>`, `2>>`, `2>&1`, `2>&-`, `2<>`, `2<`, `2<&1`, `&>`, `&>>`, and `|&`;
- it does not false-positive on stdout-only redirects, `>&2`, numeric literals, or fd1 redirects;
- `shellcheck` and `git diff --check` are clean;
- GitHub's Actions analysis and the fast contract checks are green on this exact head; the remaining matrix is still running.

Residual risk is explicit rather than hidden: this is still a static shell-syntax contract. A future shell refactor may introduce a semantically equivalent fd2 manipulation outside this operator class. A behavioral mock-Cargo/sentinel test would be the stronger long-term check, but the measured escape that blocked this head is closed.

Rul1an:
Both reproduced, then fixed by replacing the static scan rather than widening it. Final head: `db3ee5f37bb9cc65b2627bace6e5c55145c9a64f`.

**P2** — `shell: bash {0} 2>cargo-error.log` on the step kept the contract and actionlint green. It sits outside the run block by construction, so no scan of the script could ever reach it.

**P3** — the operator class fired on `# Never add 2>cargo-error.log here`, turning the contract red over a comment warning against exactly what the test exists to catch.

The two point in opposite directions — one escape the pattern cannot see, one alarm on text that does nothing — so another operator spelling would have fixed neither. The static fd2 scan is removed, not supplemented.

What replaces it: pin the step to a plain `shell: bash`, extract the run script, and execute it in a sandbox against a mock `cargo` that fails `metadata --locked` with a unique sentinel on stderr, plus a mock `rustup`/`rustc`. Three assertions — the step exits nonzero, the value-free wrapper prints, and the sentinel reaches the step's own stderr.

The point of the third assertion is that no spelling has to be known. `2>`, `exec 2>` and `exec 2<>` are all still caught, now because the sentinel is missing from the outer capture rather than because a regex recognised the operator. Verified the class is gone from the file (zero occurrences) while those three mutants stay red, so they are genuinely carried by behaviour.

Eight controls:

```
red:   guard exit stripped · 2> on the guard line · exec 2> · exec 2<>
       path entry deleted · path entry commented out · step-level shell: redirect
green: a comment naming a redirect
```

One thing found while building it, worth recording since it is the same class of bug: the sandboxes leaked one directory per failing mutant. Mutant runs abort partway through by design, so a cleanup line after the assertions never runs — and `trap ... EXIT` *replaces* the previous handler instead of adding to it, so this file's chain of re-traps was silently dropping what earlier ones covered. Now one root, one trap. Measured zero temp directories left per run.

Full verification set on this head: fuzz-lane-contract PASS, ci-gate-expectations PASS, actionlint clean on both workflows, shellcheck clean, the hook firing on each of its five input paths with an unrelated file still skipped, `cargo metadata --locked` in `fuzz/` passing, pre-commit and pre-push green including clippy and the linux compile gate. The diff touches one file; the workflow is unchanged.

Not merged. #1915 untouched.

Rul1an (commented):
Final exact-head independent review of `db3ee5f37bb9cc65b2627bace6e5c55145c9a64f`: no remaining findings.

The test now proves the property rather than recognizing shell spellings. I verified that it extracts the actual `Fuzz smoke` script, pins the GitHub step to plain `shell: bash`, runs the script against mock `rustc`/`cargo`, makes only `metadata --locked` fail with a unique stderr sentinel, and requires all three observable outcomes: nonzero exit, the value-free wrapper on stdout, and the Cargo sentinel on the outer stderr capture.

All eight controls behave as intended: seven break the contract and the comment-only control remains green. `shellcheck`, `git diff --check`, and the full contract script are clean. The temporary tree uses one root and one EXIT trap; a fresh run leaves no review-worktree changes.

Residual risk: the behavioral harness intentionally stops at the metadata failure, so it does not exercise commands after the guard. That is appropriate for this contract, whose property is the diagnostic emitted by the failing metadata check.

Rul1an:
@coderabbitai full review

Rul1an:
Reproduced and fixed. Final head: `17e0dea00cbae9fff0cbd311efaedbd4bb7e539f`.

Confirmed the mechanism directly first, since the finding turns on it: under GitHub's own `bash --noprofile --norc -eo pipefail`, a `BASH_ENV` preamble doing `exec 2>file` leaves the outer stderr capture empty while the script itself is untouched. `--noprofile --norc` does not suppress `BASH_ENV`. Then reproduced the false green — a step-level `BASH_ENV` kept the contract and actionlint green.

The gap is structural rather than incidental: the harness extracts the shell and the run script but supplies the environment itself, and the divert happens before the script's first line. No amount of running the script can show it.

Three bounded changes, no env parser and no new framework:

- **Reject an active `BASH_ENV:` mapping** in this lane's config. One name, matched as a mapping key. Verified rejected at workflow, job and step level, and a commented-out line — which sources nothing — verified still green.
- **Run the harness under `env -u BASH_ENV`**, so the caller's environment cannot decide whether the proof holds. Verified by running the whole contract with an ambient `BASH_ENV` pointing at a stderr-redirecting preamble: it still passes.
- **Assert the sandbox PATH wins** for `cargo` and `rustc` before anything runs, so a green result cannot come from the real toolchain standing in for the mock.

Nine controls now; the comment control names `BASH_ENV` alongside the redirect.

```
red:   guard exit stripped · 2> on the guard line · exec 2> · exec 2<>
       path entry deleted · path entry commented out
       step-level shell: redirect · step-level BASH_ENV
green: a comment naming a redirect or BASH_ENV
```

Full existing verification on this head: fuzz-lane-contract PASS, ci-gate-expectations PASS, actionlint clean on both workflows, shellcheck clean, the hook firing on each of its five input paths with an unrelated file skipped, `cargo metadata --locked` in `fuzz/` passing, zero temp directories leaked per run, and pre-commit plus pre-push green including clippy and the linux compile gate. The diff touches one file; the workflow is unchanged.

Not merged. #1915 untouched.

Rul1an (commented):
Final exact-head independent review of `17e0dea00cbae9fff0cbd311efaedbd4bb7e539f`: no remaining findings within the contract's trust boundary.

Fresh verification:
- the BASH_ENV escape is rejected as an active mapping key and the comment-only control remains green;
- the behavioral run explicitly removes ambient `BASH_ENV`; the full contract still passes when invoked with an ambient BASH_ENV;
- mock `cargo` and `rustc` are resolved from the sandbox before the step runs;
- all nine controls have the intended outcome;
- `shellcheck`, `git diff --check`, and the review worktree are clean.

Scope is now appropriately bounded: checked-in workflow configuration is trusted, reviewed source. This contract guards concrete accidental drift in the current lane; it is not and should not become a general interpreter for every malicious GitHub Actions or shell configuration.

Rul1an:
@coderabbitai full review

Rul1an:
Reproduced and fixed. Final head: `f0091907e66bc6f0c5943410792511abec35dede`.

Both quoted forms escaped and both were actionlint-clean:

```
"BASH_ENV": file   contract GREEN   actionlint clean
'BASH_ENV': file   contract GREEN   actionlint clean
```

The finding is right that this contradicted the claim. When the assertion landed I described it as matching a mapping key; it matched the bare word. Quoting a key changes nothing about what it means to YAML or to the runner, so the claim was broader than the check.

The three legal key forms are now written out. Verified per form rather than asserted:

```
BASH_ENV: x        "BASH_ENV": x        'BASH_ENV': x        BASH_ENV : x     REJECTED
# BASH_ENV: x      # never set "BASH_ENV": x                            allowed
MY_BASH_ENV: x     BASH_ENVX: x         RUNS: 20000                     allowed
```

So it matches the key rather than the substring, and the comment control is unchanged and still green. Still one name, no YAML parser.

Ten controls now: eight that must turn it red, plus the quoted-key mutant, plus the comment control that must stay green.

Full existing verification on this head: fuzz-lane-contract PASS, ci-gate-expectations PASS, actionlint clean on both workflows, shellcheck clean, the hook firing on each of its five input paths with an unrelated file skipped, `cargo metadata --locked` in `fuzz/` passing, the harness still immune to an ambient `BASH_ENV`, zero temp directories leaked per run, pre-commit and pre-push green including clippy and the linux compile gate. The diff touches one file; the workflow is unchanged.

Not merged. #1915 untouched.

Rul1an (commented):
Final closure review of exact head `f0091907e66bc6f0c5943410792511abec35dede`: no remaining findings in the checked-in diff.

The BASH_ENV guard now matches the three equivalent YAML key spellings while preserving the comment and near-name controls. The full behavioral contract passes both normally and with an ambient BASH_ENV, all ten controls have the intended result, and `shellcheck` plus `git diff --check` are clean.

Review scope is deliberately closed here: workflow source is trusted and independently reviewed; the contract guards the actual lane and concrete accidental regressions, not every hypothetical hostile YAML program.

Rul1an:
@coderabbitai full review

Rul1an (commented):
Second non-building review record for exact head `f0091907e66bc6f0c5943410792511abec35dede` under the documented 30-minute bot-unavailable fallback.

This review was performed by an independent subagent in an isolated detached worktree, not by the branch builder. GitHub shows the same repository owner because the agent has no separate GitHub identity; that limitation is disclosed rather than presented as a distinct human account.

No findings. The reviewer verified the behavioral shell contract, all ten positive/negative controls, the three supported YAML spellings of `BASH_ENV`, ambient-environment neutralization, mock PATH precedence, singleton cleanup, actionlint/shellcheck/pre-commit, locked metadata, RustSec audit, and a clean worktree. CodeRabbit remained rate-limited on the final head and is not counted as a pass.

At recording time the final head has a fully green matrix and zero unresolved threads. This review plus the earlier non-building exact-head review satisfies the two-review fallback in `AGENTS.md`; it is not a bot-review substitute or a precedent for counting rate-limited status contexts.

PR #1915, opened 2026-07-28
feat(attestation): the subject is the artifact, and the match is a separate call

Rul1an:
ADR-044 implemented. Both ends in one slice, because the ADR is explicit that neither works alone — a correct subject digest nobody compares is still an unchecked field, and a consumer that matches would reject everything a `run_root`-subject producer emits.

## Discovery set the slice boundary

Before writing anything I measured the blast radius, and it decided the shape:

- **no committed artifact uses `evidence-bundle/v0`** — the two DSSE fixtures in the tree are SLSA provenance and a gateway-path shape
- **`verify_envelope` had no caller outside one unit test** — there is no attestation-verification command
- `assay evidence attest` **is** wired and reachable, so users can hold v0 attestations

So both ends can move together at an in-repo cost of one test. Users holding v0 attestations must reissue — that is the ADR's stated consequence, not a side effect.

## Producer

`statement_from_bundle` takes the bundle bytes and puts their SHA-256 in the subject, gzip trailer included. It **replaces** `statement_from_manifest` rather than joining it: a manifest cannot produce a conforming subject, so leaving that function in place would leave a way to sign one by mistake.

Exactly one digest entry — `DigestSet`s match on **ANY** field, so a second key would match a forged bundle through the very field added to document the problem.

Predicate v1 is a **type**, not free-form JSON, and is derived from the verified bundle rather than supplied. `--predicate` is now refused rather than silently ignored: every field must be derivable from the artifact for a consumer to cross-check it, and a hand-written predicate cannot offer that.

## Consumer

`verify_envelope` returns `SignatureVerified` — **a state, not a verdict**. It establishes who signed a Statement, never which artifact, because it is offered none.

`verify_attestation_for_bundle(envelope, key, bundle_bytes)` is the only call that can return a verified attestation. It recomputes the digest, requires exactly one subject carrying exactly one digest entry, refuses the v0 predicate **by name** rather than as an unknown, fails closed on an unrecognised version, and cross-checks every derived field against the bundle.

## One design question fell out of discovery

Predicate v1 requires the run's time window and the manifest does not carry it. Rather than have `attest` iterate events through `BundleReader` — which would undo the change that stopped verification holding the stream — `VerifyResult` now captures min and max event `time` **during the pass it already makes**. Free, and no retention.

Compared as RFC 3339 strings rather than parsed: the format pins UTC with a `Z` suffix and fixed field order, so lexicographic order is chronological, and parsing would add a failure mode to a value the verifier does not otherwise depend on.

## What the golden harness caught

Replacing `File::open` with `std::fs::read` made an import unused, and the compiler warning **leaked into CLI stdout** — the golden test failed on the diff. Scoped the import to the test module rather than silencing the warning.

## Verification

Worktree `assay-adr044-impl`, branched from `705590eb`: `assay-evidence` **581 passed**, `assay-cli` **571 passed**, 0 failed; clippy `--workspace --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean. The attest test now also asserts that a different artifact does **not** match a valid signature over a real statement.

## Not in this slice

The CLI has no attestation-verification command yet, so `verify_attestation_for_bundle` is a library contract with no user-facing surface. Adding `assay evidence verify-attestation` is the natural next slice and is where the typed *signature-verified, artifact-unmatched* state has to be rendered honestly to a human.

## Summary by CodeRabbit

- **New Features**
  - Attestations now bind to the exact evidence bundle bytes, and verification uses SHA-256 over those bytes.
  - Added richer verification that includes the evidence event time window.
- **Bug Fixes**
  - Signed attestation payloads are now strictly parsed to reject ambiguous/invalid JSON, with stronger DSSE and schema checks.
  - Oversized evidence bundles are rejected before exceeding configured source-byte limits.
  - Legacy v0 predicate identifiers are refused (v1 identifiers used by default).
- **Changes**
  - `assay evidence attest` no longer supports `--predicate`.

### DIFF
--- a/crates/assay-evidence/src/attestation.rs
+++ b/crates/assay-evidence/src/attestation.rs
@@ -24,6 +24,7 @@
 //! `DigestSet` matching semantics do not reach it. See ADR-044.
 
 use crate::bundle::writer::{VerifyLimits, VerifyResult};
+use crate::bundle::Manifest;
 use crate::crypto::jcs;
 use crate::mandate::signing::{build_pae, compute_key_id_from_verifying_key};
 use anyhow::{Context, Result};
@@ -42,6 +43,12 @@ const IN_TOTO_PAYLOAD_TYPE: &str = "application/vnd.in-toto+json";
 /// On `docs.getassay.dev`, which is where the contract surface actually lives — the profile URIs
 /// under `docs/profiles/` already resolve there. The v0 constant below keeps the old host because
 /// it names statements that were minted under it; a refusal has to spell the thing being refused.
+///
+/// **This name changed meaning.** In 3.35.0 it held the v0 identifier. Downstream code comparing a
+/// statement against this constant still compiles and now matches v1 instead — so a v0 comparison
+/// has to name [`EVIDENCE_BUNDLE_PREDICATE_TYPE_V0`] explicitly. The unversioned name tracks the
+/// current predicate by convention; it is recorded here because a silent change of meaning behind
+/// an unchanged name is the kind of break a compiler cannot report.
 pub const EVIDENCE_BUNDLE_PREDICATE_TYPE: &str =
     "https://docs.getassay.dev/attestation/evidence-bundle/v1";
 
@@ -126,29 +133,54 @@ fn statement_from_parts(
             digest,
         }],
         predicate_type: EVIDENCE_BUNDLE_PREDICATE_TYPE.to_string(),
-        // Infallible: `EvidenceBundlePredicate` is a plain struct of strings and integers, so
-        // `to_value` has no failing case here. It was a `?` returning an error no test could
-        // provoke and no caller could act on.
-        predicate: serde_json::json!({
-            "schema_version": predicate.schema_version,
-            "semantic_equivalence": {
-                "algorithm": predicate.semantic_equivalence.algorithm,
-                "value": predicate.semantic_equivalence.value,
-            },
-            "run": {
-                "run_id": predicate.run.run_id,
-                "event_count": predicate.run.event_count,
-                "producer": {
-                    "name": predicate.run.producer.name,
-                    "version": predicate.run.producer.version,
-                    "git": predicate.run.producer.git,
-                },
-                "time_window": {
-                    "start": predicate.run.time_window.start,
-                    "end": predicate.run.time_window.end,
-                },
-            },
-        }),
+        // `to_value` rather than a `json!` literal mirroring the struct. The literal was a second
+        // copy of the shape, maintained by hand beside the `Serialize` derive, and the two could
+        // drift apart in either direction without anything failing.
+        //
+        // The `expect` is an infallibility assertion, not error handling. `to_value` fails only
+        // when a `Serialize` impl returns an error or a map has non-string keys; this type is a
+        // derive over strings and integers, so neither case exists. Returning a `Result` here was
+        // worse: it published a failure no caller could act on and no test could provoke.
+        predicate: serde_json::to_value(predicate)
+            .expect("EvidenceBundlePredicate derives Serialize over strings and integers only"),
+    }
+}
+
+/// Build an in-toto v1 Statement whose subject digest is the bundle's chain root.
+///
+/// Preserved at its published 3.x signature so downstream code keeps compiling. It is not a way to
+/// produce a usable attestation: the chain root is an equivalence digest over event *semantics*,
+/// so two bundles carrying the same events under a different stream identity share it, and the
+/// subject therefore identifies no artifact and matches none. That is what ADR-044 replaced.
+///
+/// It is pinned to [`EVIDENCE_BUNDLE_PREDICATE_TYPE_V0`] rather than following the constant, whose
+/// value moved to v1 in the same change. Following it would have let this function mint statements
+/// claiming a conformance they do not have, which is a worse outcome than the break it avoids —
+/// and [`verify_attestation_for_bundle`] refuses what it produces, by name and on purpose.
+#[deprecated(
+    since = "3.36.0",
+    note = "produces a v0 statement whose subject identifies no artifact and which \
+            `verify_attestation_for_bundle` refuses; use `statement_for_bundle`"
+)]
+pub fn statement_from_manifest(
+    manifest: &Manifest,
+    predicate: serde_json::Value,
+) -> InTotoStatement {
+    let digest_hex = manifest
+        .run_root
+        .strip_prefix("sha256:")
+        .unwrap_or(&manifest.run_root)
+        .to_string();
+    let mut digest = BTreeMap::new();
+    digest.insert("sha256".to_string(), digest_hex);
+    InTotoStatement {
+        type_: STATEMENT_TYPE.to_string(),
+        subject: vec![Subject {
+            name: manifest.bundle_id.clone(),
+            digest,
+        }],
+        predicate_type: EVIDENCE_BUNDLE_PREDICATE_TYPE_V0.to_string(),
+        predicate,
     }
 }
 
@@ -397,6 +429,26 @@ pub fn verify_attestation_for_bundle(
     trusted_key: &VerifyingKey,
     bundle_bytes: &[u8],
 ) -> Result<AttestationVerified> {
+    verify_attestation_for_bundle_with_limits(
+        envelope,
+        trusted_key,
+        bundle_bytes,
+        VerifyLimits::default(),
+    )
+}
+
+/// [`verify_attestation_for_bundle`] under an explicit set of verification ceilings.
+///
+/// The producer already took limits; the consumer did not, so a caller who had bounded what they
+/// would attest could not bound what they would accept. The bundle is verified a second time here,
+/// by whoever holds it, and that verification is the one that decides — it should answer to the
+/// same ceilings.
+pub fn verify_attestation_for_bundle_with_limits(
+    envelope: &DsseEnvelope,
+    trusted_key: &VerifyingKey,
+    bundle_bytes: &[u8],
+    limits: VerifyLimits,
+) -> Result<AttestationVerified> {
     let verified = verify_envelope_signature(envelope, trusted_key)?;
     let statement = verified.statement;
 
@@ -454,7 +506,7 @@ pub fn verify_attestation_for_bundle(
 
     // Cross-check against the artifact rather than trusting the predicate. A predicate that
     // disagrees with the bundle it is attached to is a rejection, not a note.
-    let result = crate::bundle::verify_bundle(bundle_bytes)
+    let result = crate::bundle::verify_bundle_with_limits(bundle_bytes, limits)
         .context("the attested bundle does not verify")?;
 
     // The subject name is `manifest.bundle_id` and is compared, not merely carried. Before this it
@@ -969,6 +1021,140 @@ mod tests {
         // field ever goes back to one.
         let (start, end) = result.time_window;
         assert!(start <= end);
+    }
+
+    /// The consumer honours a ceiling too, and the ceiling is threaded rather than accepted.
+    ///
+    /// The two bounds are one byte apart on purpose. A test that merely passes some limit through
+    /// stays green when the argument is dropped and the default used instead; only a limit strict
+    /// enough to change the verdict proves the value reached the verifier.
+    #[test]
+    fn the_consumer_honours_a_stricter_source_ceiling() {
+        let k = key();
+        let bytes = bundle();
+        let envelope = sign_statement(&statement_for_bundle(&bytes).unwrap(), &k).unwrap();
+
+        let fitting = VerifyLimits {
+            max_bundle_bytes: bytes.len() as u64,
+            ..VerifyLimits::default()
+        };
+        assert!(
+            verify_attestation_for_bundle_with_limits(
+                &envelope,
+                &k.verifying_key(),
+                &bytes,
+                fitting
+            )
+            .is_ok(),
+            "a ceiling that exactly fits must accept"
+        );
+
+        let too_small = VerifyLimits {
+            max_bundle_bytes: bytes.len() as u64 - 1,
+            ..VerifyLimits::default()
+        };
+        let err = verify_attestation_for_bundle_with_limits(
+            &envelope,
+            &k.verifying_key(),
+            &bytes,
+            too_small,
+        )
+        .expect_err("one byte under the ceiling must refuse");
+        let chain = format!("{err:#}");
+        assert!(
+            chain.contains("LimitBundleBytes"),
+            "expected the source ceiling to be the reason, got: {chain}"
+        );
+
+        // The default-limits entry point still answers, and answers the same way.
+        assert!(verify_attestation_for_bundle(&envelope, &k.verifying_key(), &bytes).is_ok());
+    }
+
+    // ── the published v0 producer ──────────────────────────────────────────────────────────
+
+    /// `statement_from_manifest` is published in 3.x and must keep compiling, exactly.
+    ///
+    /// The Wave-0 semver check does not cover this: its baseline `9cc23b4c` predates the function
+    /// while v3.35.0 ships it, so a green run there says nothing about this call. The call below is
+    /// the check — argument types, argument order and return type all have to still hold.
+    #[test]
+    #[allow(deprecated)]
+    fn the_published_v0_producer_is_preserved_and_still_refused() {
+        let bytes = bundle();
+        let result = crate::bundle::verify_bundle(bytes.as_slice()).unwrap();
+
+        let statement: InTotoStatement =
+            statement_from_manifest(&result.manifest, serde_json::json!({ "anything": true }));
+
+        // v0 in every respect it had: the old predicate type, and the semantic root as the subject
+        // digest. Reviving it under the v1 type would mint statements claiming a conformance they
+        // do not have.
+        assert_eq!(statement.predicate_type, EVIDENCE_BUNDLE_PREDICATE_TYPE_V0);
+        assert_eq!(statement.subject[0].name, result.manifest.bundle_id);
+        assert_eq!(
+            statement.subject[0].digest["sha256"],
+            result
+                .manifest
+                .run_root
+                .strip_prefix("sha256:")
+                .unwrap_or(&result.manifest.run_root)
+        );
+        assert_eq!(statement.predicate, serde_json::json!({ "anything": true }));
+
+        // And what it produces stays refused, against the very bytes it describes. Keeping the
+        // function compiling is a source-compatibility promise, not a rehabilitation.
+        let k = key();
+        let envelope = sign_statement(&statement, &k).unwrap();
+        let err = verify_attestation_for_bundle(&envelope, &k.verifying_key(), &bytes)
+            .expect_err("a v0 statement must stay unverifiable");
+        assert!(err.to_string().contains("pre-ADR-044"), "got: {err}");
+    }
+
+    // ── wire shape ─────────────────────────────────────────────────────────────────────────
+
+    /// The predicate on the wire is exactly what the type serializes to.
+    ///
+    /// It was assembled by hand into a `json!` literal that mirrored the struct field by field, so
+    /// the two could drift apart silently — a renamed field would have changed the type and left
+    /// the wire shape untouched, or the reverse. This pins the shape and the equality at once.
+    #[test]
+    fn the_predicate_wire_shape_is_the_serialization_of_the_type() {
+        let bytes = bundle();
+        let result = crate::bundle::verify_bundle(bytes.as_slice()).unwrap();
+        let predicate = predicate_from_verified(&result).unwrap();
+        let statement = statement_for_bundle(&bytes).unwrap();
+
+        assert_eq!(
+            statement.predicate,
+            serde_json::to_value(&predicate).unwrap(),
+            "the statement must carry the serialization of the predicate, not a copy of it"
+        );
+
+        // And the shape itself, spelled out, so a rename is visible here rather than only to a
+        // consumer that has already deployed.
+        assert_eq!(
+            statement.predicate,
+            serde_json::json!({
+                "schema_version": 1,
+                "semantic_equivalence": {
+                    "algorithm": predicate.semantic_equivalence.algorithm,
+                    "value": predicate.semantic_equivalence.value,
+                },
+                "run": {
+                    "run_id": predicate.run.run_id,
+                    "event_count": predicate.run.event_count,
+                    "producer": {
+                        "name": predicate.run.producer.name,
+                        "version": predicate.run.producer.version,
+                        "git": predicate.run.producer.git,
+                    },
+                    "time_window": {
+                        "start": predicate.run.time_window.start,
+                        "end": predicate.run.time_window.end,
+                    },
+                },
+            })
+        );
     }
 
     // ── limits ─────────────────────────────────────────────────────────────────────────────
--- a/crates/assay-evidence/src/attestation.rs
+++ b/crates/assay-evidence/src/attestation.rs
@@ -23,7 +23,7 @@
 //! The semantic root still travels, inside the predicate's `semantic_equivalence` block, where
 //! `DigestSet` matching semantics do not reach it. See ADR-044.
 
-use crate::bundle::writer::{VerifyLimits, VerifyResult};
+use crate::bundle::writer::VerifyLimits;
 use crate::bundle::Manifest;
 use crate::crypto::jcs;
 use crate::mandate::signing::{build_pae, compute_key_id_from_verifying_key};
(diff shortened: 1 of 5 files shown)


Rul1an (commented):
Exact-head review of `7291a2c467e2f9e3d4350b745b8def5c59e17a2a`.

The architectural direction is right: signature verification and artifact matching are separate states, the semantic root leaves the in-toto DigestSet, and predicate fields are derived during the existing verification pass. The current implementation is not merge-ready yet.

## Findings

1. **P1, bounded-ingest regression and current CodeQL blocker** (`attest.rs:42`). `std::fs::read(&args.bundle)` materializes an attacker-selected bundle before `VerifyLimits.max_bundle_bytes` can act. This reopens the exact ADR-043 §1 class that was just closed. Read through the shared `LimitReader` before allocation can exceed the source ceiling, then verify the bounded bytes without an unbounded second path. The two open `rust/path-injection` alerts are at lines 42 and 145; the production path should be structurally fixed, while the test-only path needs a narrow documented disposition rather than design distortion.

2. **P1, the subject name violates ADR-044 and is never cross-checked** (`attestation.rs:88-90`, `311-356`). The ADR says `subject[0].name` keeps the bundle identifier, but the builder writes `predicate.run.run_id`. The consumer validates only the digest and predicate, so any signed subject name is accepted. Build the name from `VerifyResult.manifest.bundle_id` and compare it against the same derived value during full verification.

3. **P1, unacknowledged semver break in a published crate** (`attestation.rs:234-277`). `assay-evidence 3.35.0` is published and exposes `verify_envelope(...) -> Result<InTotoStatement>`. This PR changes that return type to `Result<SignatureVerified>`. That is source-breaking even though the repository has only one internal caller. Either land this on an explicitly planned major release with migration/release evidence, or provide a compatibility path and introduce the typed state under a new API. “No caller outside one test” measures this repository, not the public crate contract.

4. **P2, the producer API does not enforce its own derivation claim** (`attestation.rs:76-94`). `statement_from_bundle` accepts any caller-constructed `EvidenceBundlePredicate`; it does not establish that the predicate came from those bytes. The CLI happens to pair them correctly, but the public library can mint a v1 statement that its own consumer rejects. Prefer one public construction path that takes verified bundle state, derives predicate plus subject name together, and leaves the lower-level constructor private.

5. **P2, the time window loses information** (`verify.rs:502-509`). `SecondsFormat::Secs` truncates valid fractional event timestamps, so the predicate does not carry the actual earliest and latest event times. Use an exact UTC RFC 3339 rendering such as `AutoSi`, compare the typed `DateTime<Utc>` values, and make the verified result's non-empty invariant non-optional instead of publishing `Option` that downstream code must re-prove.

6. **P2, public contract and diagnostics need reconciliation.** The new v1 identifier still uses the dead `assay.dev` host even though the live contract surface is `docs.getassay.dev`; ADR-044 and code must move together before v1 is minted. The module-level documentation still describes the removed run-root subject. New mismatch errors also echo attacker-controlled subject and predicate values (`attestation.rs:335`, `353-355`), contrary to the repository's value-free diagnostic discipline.

## Missing acceptance coverage

The current tests cover signature tampering, one happy bundle match, and one byte mismatch. Add negative cases for v0, unknown predicate major, zero/multiple subjects, zero/multiple digest entries, missing/wrong algorithm, subject-name mismatch, each predicate-field mismatch, fractional timestamps, custom source limits, and the explicit `--predicate` refusal. A verifier that rejects for the wrong earlier reason should not satisfy these tests.

CodeRabbit was rate-limited and Copilot was unavailable, so neither counts as review quorum on this head. CodeQL is red with two high alerts. No merge recommendation until the findings and same-head review/CI requirements are closed.

Rul1an (commented):
Additional exact-head review of `7291a2c467e2f9e3d4350b745b8def5c59e17a2a`, supplementing the earlier review.

## New P1 finding: ambiguous signed JSON is accepted as fully verified

`crates/assay-evidence/src/attestation.rs:266-267` deserializes the signed Statement directly with `serde_json::from_slice`. Serde accepts duplicate object members last-wins. I reproduced the full path in an isolated detached worktree: I created a valid bundle and v1 predicate, injected both `"schema_version":999` and `"schema_version":1` into the signed raw DSSE payload, signed those exact bytes, and `verify_attestation_for_bundle` returned `Ok(AttestationVerified)`.

That permits two conforming consumers to read different claims from the same valid signature, depending on duplicate-member policy. Run the signed payload through the crate's strict JSON validator before deserialization, and add reason-specific negative tests for duplicates at Statement, Subject/DigestSet, predicate, and nested predicate levels. The test must sign the ambiguous bytes; mutating after signing only tests the signature.

A related probe confirmed that surplus predicate fields are accepted. That is **not** a finding against this PR because ADR-044 explicitly says unknown fields within a known major are ignored. The implementation and decision agree there.

## Correction to the prior CodeQL wording

The two `rust/path-injection` alerts point at a CLI whose contract intentionally accepts user-selected bundle/key/output paths. Repository-root containment would distort that public interface. Line 42 still has the real ADR-043 defect: `std::fs::read` materializes the bundle before the source ceiling. Replace it with the shared bounded reader. If CodeQL still flags the intentional `File::open` after that structural fix, record a narrow `used as intended` disposition for this CLI path; the test-only line 145 needs its own scoped disposition. Do not contort the path model merely to erase taint flow.

All findings from the earlier review remain open: subject name/binding, public semver break, caller-authored producer predicate, lossy time window, contract URI/diagnostics, and the incomplete negative matrix. The PR is also behind current `main` after #1914 merged. No merge recommendation on this head.

Rul1an:
Rebased onto current `main` (`e6673510`, which is #1914 merged) and all findings from both exact-head reviews addressed. Head: `2a3dc9dd7058f1c5720bee4126ce522927eba0b8`.

## The new P1 first

Reproduced before touching anything: signing raw bytes carrying both `"schema_version":999` and `"schema_version":1` returned `Ok(AttestationVerified)`. The signed payload now goes through the crate's own strict scanner — the one already applied to manifests and events. The signed statement was the single JSON document in this crate reaching a verdict without it.

Negatives at four levels (statement, subject/DigestSet, predicate, nested predicate), each **signing** the ambiguous bytes. Surplus fields stay accepted and are pinned by a test, since ADR-044 line 178 makes that the decision rather than an oversight — `deny_unknown_fields` would contradict it.

## Commit structure

```
2a3dc9dd docs(adr): mint the v1 predicate on the host that serves the contract
e652b41d fix(cli): bound the attest ingest, and stop the command authoring a predicate
338566e7 fix(evidence): an attestation may only say what the bytes it names can prove
5e6d62d2 feat(attestation): the subject is the artifact, and the match is a separate call
```

The library findings are one commit because they are one change: every claim now comes from, or is checked against, the verified bytes. Splitting them would have produced commits that do not compile.

## Proven bite

Each behaviour change was reverted in isolation and the matching case had to go red. Two of my first mutations silently failed to apply — `cargo fmt` had reflowed the lines I was matching — so the harness asserts the mutation landed before it claims anything:

```
strict parse -> serde_json::from_slice          red — bites
producer writes run_id as the subject name      red — bites
consumer stops comparing the subject name       red — bites
time window -> SecondsFormat::Secs              red — bites
field-named diagnostics -> generic message      red — bites
bounded CLI ingest -> std::fs::read             red — bites
```

The subject-name pair is two cases on purpose: pointing the producer mutation at the consumer test left it green, because that test overwrites the name itself. One test could not have carried both halves.

## The rest of the findings

- **Subject name** built from `manifest.bundle_id` and compared against it during full verification. It was written from `run_id` and read nowhere, so every possible name verified.
- **Semver**: `verify_envelope` keeps its published 3.x signature and gains `#[deprecated]`; `verify_envelope_signature` is the same check returning `SignatureVerified`. A test calls the old one precisely so a future return-type change stops compiling.
- **Producer**: `statement_for_bundle(_with_limits)` takes bytes and nothing else; the constructor that pairs a subject with a predicate is private. There is no argument left to get wrong.
- **Time window**: instants compared as instants, rendered once with `AutoSi`. `VerifyResult::time_window` is no longer `Option` — a zero-event bundle is already refused, so publishing one handed every consumer an invariant the verifier had established.
- **Contract URI**: v1 moves to `docs.getassay.dev`, in code and ADR together. v0 keeps `assay.dev` because it names statements minted under it. Module docs rewritten; they still described the removed run-root subject.
- **Diagnostics**: value-free, and they name the field that disagreed. Ten predicate fields each have their own case asserting their own reason, and a compile-time exhaustive destructure means adding a field without extending the namer fails to build.
- **Bounded ingest**: `std::fs::read` replaced by the shared `LimitReader`, extracted as `read_bundle_bounded` so the ceiling can be exercised with a small limit instead of a 100 MB file.

Also added: v0 refusal, unknown major, zero/two subjects, zero/two digest entries, wrong algorithm, custom source ceiling (asserted on the error **chain**, since the top-level context says only that the bundle did not verify), and the `--predicate` refusal including that nothing is written after it.

## Local verification

Full workspace `cargo test`: **2805 passed, 0 failed**. `cargo clippy --workspace --all-targets -D warnings`: clean. `cargo fmt --check`: clean. Pre-push green including the linux compile gate.

## Three things I am flagging rather than deciding

1. **A second semver break the reviews did not name.** The base commit already removed `statement_from_manifest`, which is published in 3.x. I did not restore it: it mints statements with a `run_root` subject, which is exactly what the consumer now refuses by design, so a compatibility shim would preserve a footgun rather than a contract. That leaves this branch source-breaking regardless of the `verify_envelope` fix, and the release decision should be made knowingly rather than discovered.

2. **I could not verify that 3.35.0 is on crates.io** — the sandbox has no network to it. I acted on the instruction regardless, which is the right move either way; I am recording that the premise is unverified rather than implying I checked it.

3. **CodeQL**: the production defect is structurally fixed, and the test-only read now goes through the same bounded helper, so line 145 no longer exists in the flagged form. Whether taint remains on the intentional `File::open` is only knowable from the CI run. The rationale for the operator-chosen paths is recorded at the call site; if alerts persist I would add narrow per-alert dispositions rather than contain the paths under the repository root, which would change the published interface to satisfy a tracker. I have not dismissed anything — that is yours to trigger.

Not merged.

Rul1an (commented):
Exact-head follow-up review of `2a3dc9dd7058f1c5720bee4126ce522927eba0b8`.

The earlier architectural and hostile-input findings are closed. The branch is locally well verified, the delegated proof is green, and nearly all CI has passed. It is not merge-ready yet.

## Remaining blockers

1. **P1: the consumer cannot apply caller-selected verification ceilings** (`crates/assay-evidence/src/attestation.rs:395-458`). CodeRabbit is correct: `verify_attestation_for_bundle` always enters through `verify_bundle` and therefore fixes the default limit set. Add `verify_attestation_for_bundle_with_limits(..., VerifyLimits)` using `verify_bundle_with_limits`, keep the current function as the default-limit compatibility wrapper, and add an exact-limit/limit+1 test proving the supplied limit is load-bearing.

2. **P1: `statement_from_manifest` was removed from the published 3.x API without a major-version decision.** It is present in tag `v3.35.0` and has been public since `v3.18.0`; the current Wave-0 semver job misses this because its fixed baseline `9cc23b4c` predates the API. The lean 3.x-compatible fix is to restore the old signature as a strongly deprecated v0-only compatibility constructor, preserving its exact old output. It must remain unfit for v1 and the full consumer must continue to refuse v0. The alternative is an explicit 4.0 release, which is out of scope for this PR.

## Small corrections on the same final push

3. Update the stale `--predicate` help text and move its refusal before bundle/key I/O. Both CodeRabbit points are valid.

4. Replace the hand-mirrored predicate `json!` with serialization of `EvidenceBundlePredicate`. The type already owns the wire shape; keeping a second field list creates silent producer drift.

## CodeQL disposition

The six open `rust/path-injection` alerts are intentional path operations, not six distinct vulnerabilities:
- line 43 is the published CLI contract: the operator selects the bundle path, and the read is now source-bounded;
- lines 131, 141, 156, 182, and 219 are test-owned temporary fixture paths.

Do not add repository-root containment. After the final code push, dismiss each exact alert narrowly as `used in tests` or `used as intended`, with the bounded-read rationale for line 43, and verify the aggregate CodeQL context turns green.

Any push invalidates the current CodeRabbit review. After these fixes: rerun focused tests, crate/CLI suites, fmt, clippy, full CI, CodeQL disposition, and obtain same-head non-builder plus bot quorum.

Rul1an:
All four items implemented test-first. Head: `cb65ba3bcfe3eaf8fb604dfe1424841661313b4e`.

## Red state, then green

| item | red state proven |
|---|---|
| 1. consumer ceiling | `cannot find function verify_attestation_for_bundle_with_limits` |
| 2. v0 shim | `cannot find function statement_from_manifest` |
| 3. flag ordering | refusal was `open bundle .../no-such-bundle.tar.gz` — the filesystem error, not the flag |
| 4. wire shape | compiled but unpinned; bite shown by mutation below |

## Proven bite

```
1. consumer ignores the limits argument   green-before=yes  mutated=red  green-after=yes
2. v0 shim revived under the v1 type      green-before=yes  mutated=red  green-after=yes
3. flag checked after the reads           green-before=yes  mutated=red  green-after=yes
4. wire shape drifts from the type        green-before=yes  mutated=red  green-after=yes
```

**The `green-before` and `green-after` columns are there because my first attempt at this was wrong.** `shutil.move` restores the backup's original mtime, which is older than the mutated build, so Cargo considered the crate fresh and kept running the binary built from the *mutated* source. That produced a deterministic 3-of-3 failure on unmutated code and, worse, a case-1 "bite" that proved nothing because the test was already red before the mutation. Verified the mechanism directly (restored mtime measurably older than the mutated write), fixed the harness with a `touch` after restore, and re-ran all four with both columns. Nothing in the earlier numbers is load-bearing now.

## Per item

**1.** `verify_attestation_for_bundle_with_limits(..., VerifyLimits)` added, existing entry point delegates with the default, and `verify_bundle_with_limits` is used inside. The test bounds are one byte apart — a limit merely passed through stays green when the argument is dropped and the default substituted, so only a ceiling strict enough to change the verdict shows the value arrived. Measured the real threshold first (601 accepts, 600 refuses) rather than assuming where it bit.

**2.** `statement_from_manifest(&Manifest, serde_json::Value) -> InTotoStatement` restored at its exact published signature, strongly deprecated, pinned to `EVIDENCE_BUNDLE_PREDICATE_TYPE_V0` and the `run_root` digest form. A test calls it with those exact argument types — that call is the compatibility check — and asserts the predicate type, subject name, digest form and passthrough predicate, then signs it and pins that `verify_attestation_for_bundle` still refuses it by name. Noted in the commit that the Wave-0 baseline `9cc23b4c` predates the function while v3.35.0 ships it, so a green run there says nothing here. No 4.0 scope.

**3.** The flag is refused before any file is touched, and the test uses non-existent bundle *and* key paths so the ordering is load-bearing: it asserts the refusal names the flag and that the bundle path does not appear in the error. Help text corrected — it advertised "default: a minimal summary" for a flag that is an error — with a test pinning both that the promise is gone and that the rejection is stated.

**4.** `serde_json::to_value(&predicate)` with an infallibility assertion replaces the hand-mirrored `json!`. The shape is pinned twice: equal to the type's serialization, and equal to an explicit literal, so a rename is visible here rather than only to a deployed consumer.

## Verification on this head

Workspace `cargo test`: **2810 passed, 0 failed**. assay-evidence + assay-cli: 1175 passed, 0 failed. `clippy --workspace --all-targets -D warnings`: clean. `cargo fmt --check`: clean. `check-public-artifact-sanitization.py`: passed. Pre-push green including the linux compile gate.

`optional-public-api-drift.sh` is a **no-op locally** — neither `cargo-semver-checks` nor `cargo-public-api` is installed, so it skipped both and reported nothing. Rather than report a skip as a pass, I compared the exported surface against the `v3.35.0` tag directly: every item published in 3.35.0 still exists.

## One thing that comparison surfaced, not in the review

`EVIDENCE_BUNDLE_PREDICATE_TYPE` keeps its name while its value moved from the v0 identifier to the v1 one. Downstream code comparing a statement against it still compiles and now matches a different thing — a silent change of meaning behind an unchanged name, which is precisely what a compiler cannot report. I did not change the value, since v1 is the point of the branch; I documented it on the constant and pointed v0 comparisons at `EVIDENCE_BUNDLE_PREDICATE_TYPE_V0`. Flagging rather than deciding.

No CodeQL dispositions added, no repo-root containment, no broad suppressions. Not merged.

Rul1an (commented):
Exact-head review of `cb65ba3bcfe3eaf8fb604dfe1424841661313b4e`.

The prior findings are closed, the seven CodeQL alerts have narrow dispositions, all review threads are resolved, CI is fully green, and independent focused verification passed: 21 attestation tests, 5 CLI attest tests, fmt and diff-check. Three merge blockers remain; all are contracts that the current CI cannot prove.

## Findings

1. **P1: caller-selected limits apply after an unbounded artifact hash and can be pre-empted by the wrong error** (`crates/assay-evidence/src/attestation.rs:483-510`). `verify_attestation_for_bundle_with_limits` hashes all offered bytes and compares the subject before calling `verify_bundle_with_limits`. With a tiny `max_bundle_bytes` and a large mismatched artifact, the function performs the full SHA-256 and returns subject mismatch; the supplied ceiling never decides. Verify the bundle with the supplied limits before hashing/matching, reuse that result for predicate checks, and add a negative control where oversized mismatched bytes must report `LimitBundleBytes` rather than digest mismatch.

2. **P1: adding `time_window` to public `VerifyResult` is a 3.x source break** (`crates/assay-evidence/src/bundle/writer_next/verify.rs:65-83`). Downstream struct literals and exhaustive destructures that compile against v3.35.0 stop compiling. The semver lane did not catch the earlier removed APIs either, so green there is not evidence for this shape. Keep the published `VerifyResult` layout intact and have the internal verification pass return a private sidecar/snapshot carrying the time window for attestation construction; public verification can project the old result without a second event pass. The alternative is an explicit 4.0 decision, out of this PR's stated scope.

3. **P1: `EVIDENCE_BUNDLE_PREDICATE_TYPE` silently changes meaning in 3.x** (`crates/assay-evidence/src/attestation.rs:43-53`). Existing downstream code still compiles but now compares against v1 instead of the v0 value it was built around. Documentation acknowledges the behavioral break but does not remove it. Preserve the old unversioned constant as a deprecated v0 alias, add `EVIDENCE_BUNDLE_PREDICATE_TYPE_V1`, and use the explicit v1 name in the new producer and consumer. That keeps old behavior stable while making every new comparison version-explicit.

No further findings in the current diff. Do not trigger a fresh bot review until these land; the present CodeRabbit status is rate-limited and does not count toward quorum.

Rul1an:
All three merge blockers closed. Head: `da1d510008ce1add04b6b8ad929ef80a52853ae1`.

Each was verified against the code before being fixed, and each is now pinned by a case that fails without it:

```
1. hash the artifact before the ceiling    green-before=yes  mutated=red  green-after=yes
2. public VerifyResult grows a field       green-before=yes  mutated=red  green-after=yes
3. unversioned constant repointed at v1    green-before=yes  mutated=red  green-after=yes
```

**1 — the ceiling now decides first.** Verification with the supplied limits runs before the digest, and its result is reused for the subject name and the predicate cross-check. Everything ahead of that line reads only the statement the signature already covers; nothing ahead of it touches the offered bytes. The negative control hands it bytes that are both oversized and mismatched and requires `LimitBundleBytes` — and requires the digest comparison *not* to appear, since a rejection for the right reason and a rejection for a coincidental one look identical to `is_err()`.

**2 — `VerifyResult` is back to its published layout.** The window travels beside it in a crate-internal `VerifiedBundle`, which the attestation path consumes and public verification projects away; no second event pass. The test destructures `VerifyResult` with no `..`, so the break reappears here before it reaches anyone. Confirmed the field list is identical to `v3.35.0`.

**3 — the unversioned constant is byte-for-byte what 3.35.0 shipped**, deprecated as ambiguous, with `EVIDENCE_BUNDLE_PREDICATE_TYPE_V1` added beside the existing `_V0`. Producer and consumer both name their version explicitly. Documenting the change of meaning, as I did last round, acknowledged the break without removing it — the finding was right about that.

Checked against the `v3.35.0` tag rather than the semver lane, which has now missed this class twice: no published item removed, `VerifyResult` fields identical, constant value identical.

**Verification on this head:** workspace `cargo test` **2813 passed, 0 failed**; `clippy --workspace --all-targets -D warnings` clean; `cargo fmt --check` clean; `check-public-artifact-sanitization.py` passed; pre-push green including the linux compile gate.

One thing worth recording: the workspace run initially failed on `golden_runner`, because an unused-import warning from my refactor leaked into the golden output. Not a flake and not cosmetic — that golden captures compiler output, so a warning introduced anywhere in the crate surfaces as a test failure. Fixed at the source rather than by touching the golden.

No CodeQL dispositions, no repo-root containment, no broad suppressions, no bot review triggered. Not merged.

crates/assay-evidence/src/attestation.rs
```diff
@@ -54,11 +83,90 @@ pub struct InTotoStatement {
...
-/// The chain root is an equivalence digest over event *semantics*, not a digest of the
-/// bundle bytes, so this subject does not distinguish two bundles that carry the same
-/// events under a different stream identity. See the module docs for what that costs.
+/// Preserved at its published 3.x signature so downstream code keeps compiling. It is not a way to
+/// produce a usable attestation: the chain root is an equivalence digest over event *semantics*,
+/// so two bundles carrying the same events under a different stream identity share it, and the
+/// subject therefore identifies no artifact and matches none. That is what ADR-044 replaced.
+///
+/// It is pinned to [`EVIDENCE_BUNDLE_PREDICATE_TYPE_V0`] rather than following the constant, whose
+/// value moved to v1 in the same change. Following it would have let this function mint statements
```
Rul1an: P3: This explanation is stale after the compatibility fix. `EVIDENCE_BUNDLE_PREDICATE_TYPE` now deliberately remains equal to `_V0`; it no longer "moved to v1". Please describe the explicit `_V0` pin as protection against future alias drift instead. The implementation and compatibility tests are correct; only this public rustdoc claim contradicts them.

Rul1an:
@coderabbitai full review

Requesting a formal review on the current head `da1d510008ce1add04b6b8ad929ef80a52853ae1`. The last CodeRabbit review on this PR was on `2a3dc9dd`, two heads back; the green CodeRabbit *status* on this SHA is not a review and is not being counted as one.

Two commits since that review:

- `cb65ba3b` — `--predicate` refused before any filesystem access, with the help text corrected to stop advertising a default for a flag that is an error.
- `da1d5100` — three contracts CI cannot prove: caller limits now decide before the artifact is hashed; the published `VerifyResult` layout is restored with the time window moved to a crate-internal sidecar; and `EVIDENCE_BUNDLE_PREDICATE_TYPE` is back at its 3.35.0 value with explicit `_V0`/`_V1` names beside it.

The head is final. `cargo doc` on `assay-evidence` reports exactly one warning, in `trust_card.rs:1`, which is present on `main` and in a file this PR does not touch — so there is no rustdoc change pending here and this SHA will not move for that reason.

Rul1an:
P3 closed. New head: `01c9c779a59b7128791c8c6e159aae362b66f271`.

You were right on both counts, including that I misread the finding. I went looking for `cargo doc` warnings because the word "rustdoc" was in play, and reported a pre-existing warning in `trust_card.rs` as if it were the item. It was not — the finding was about a sentence being false, and the file I should have opened was the one the finding cited.

The sentence on `attestation.rs:161-162` was true when written and stopped being true one commit later, when the unversioned constant was restored to its published v0 value and the deprecated alias took its place. The doc did not move with it, so the file explained the pinning by a fact the same file disproved. Replaced with your wording:

> It is pinned to the explicit `EVIDENCE_BUNDLE_PREDICATE_TYPE_V0` identifier rather than the deprecated unversioned alias, so future alias evolution cannot make this compatibility function mint a newer predicate contract.

That reason is also the better one: it survives whatever the alias does next, which the old sentence could not.

Swept for other survivors of the same edit rather than fixing only the cited lines. The constant's own doc is counterfactual — it says what repointing *would have* caused — so it stays correct; nothing else claims the value moved.

Doc-only: no behaviour, no signatures, no wire shape. On this head, `attestation` 24 tests pass, 14 doc tests pass, `clippy -D warnings` clean, `cargo fmt --check` clean, pre-push green including the linux compile gate. `cargo doc` still reports the one pre-existing `trust_card.rs` warning, which is on `main`, untouched by this PR, and is being handled separately.

Head is final. The last formal CodeRabbit review remains on `2a3dc9dd` and does not apply here.

Rul1an (commented):
Exact-head non-builder review of `01c9c779a59b7128791c8c6e159aae362b66f271`.

No new findings.

The delta from `da1d510008ce1add04b6b8ad929ef80a52853ae1` is limited to four public-rustdoc lines in `attestation.rs`. It now accurately explains that the legacy builder pins the explicit V0 identifier to prevent future unversioned-alias drift. There is no code, signature, predicate, subject, limit, or wire-shape change, and `git diff --check` is clean.

I rechecked the three prior compatibility blockers on this exact head:
- source-size rejection remains ordered before artifact hashing;
- the published `VerifyResult` field shape remains identical to v3.35.0, with the time window carried in crate-private verified state;
- the deprecated unversioned predicate constant remains byte-for-byte V0, with explicit V0 and V1 identifiers beside it.

The prior P3 rustdoc finding is closed. Merge remains conditional on final-head CI and the review quorum in `AGENTS.md`. The current green CodeRabbit status says `Review rate limited` and is not a formal review of this SHA, so it does not count toward quorum.

Rul1an (commented):
Independent non-builder contract-layer review of `01c9c779a59b7128791c8c6e159aae362b66f271` (separate read-only review pass).

No findings.

Verified against tag `v3.35.0` and the exact PR head:
- `VerifyResult` retains exactly its three published fields; `VerifiedBundle` and `time_window` remain crate-private.
- `verify_envelope` retains its 3.x arguments and `Result<InTotoStatement>` return type.
- the deprecated unversioned predicate constant remains byte-for-byte V0; explicit V0 and V1 constants are additive;
- `statement_from_manifest` retains its published signature and V0 behavior;
- the time window is accumulated during the existing event-verification pass and leaves the crate only through derived predicate construction;
- the four-line rustdoc correction accurately describes protection against future alias drift.

Residual risk: the Wave-0 semver baseline predates some attestation APIs, so its green result is not sufficient alone. The explicit compile-shape tests cover the affected symbols, but there is no separate downstream fixture crate compiled against v3.35.0. The reviewer's isolated local compile attempt reached ENOSPC and was cleaned up; exact-head GitHub CI is fully green across Linux, macOS and Windows, including clippy and semver checks.

Rul1an:
Merge quorum disposition for final head `01c9c779a59b7128791c8c6e159aae362b66f271`.

CodeRabbit remained rate-limited and produced no formal review object on this SHA; its green status context is not counted. Copilot was quota-limited and is also not counted.

The 30-minute fallback in `AGENTS.md` became applicable after 2026-07-28T20:40:43Z. Two independent non-building reviews are recorded on this exact head:
- pull request review 4801482043: final delta plus consumer-order/compatibility recheck, no findings;
- pull request review 4801484758: separate contract-layer comparison against v3.35.0, no findings.

At merge decision time all required CI is green, conditional skips are expected, all review threads are resolved, the head is unchanged, and GitHub reports CLEAN/MERGEABLE.

PR #1923, opened 2026-07-29
ci(rustdoc): gate on cargo doc with warnings denied, and clear the nine that had accumulated

Rul1an:
## Why

`cargo doc -p assay-evidence --no-deps` emitted a `redundant explicit link target` on `crates/assay-evidence/src/trust_card.rs:1`, and under `RUSTDOCFLAGS="-D warnings"` that stopped the crate documenting at all. It was present on `main` and deliberately kept out of #1915, whose head was being held stable for review.

The warning survived because nothing in CI ran `cargo doc` with warnings denied. Clippy never sees rustdoc's lints: broken intra-doc links, unclosed HTML tags in doc comments and bare URLs are diagnosed by rustdoc alone and are all warn-by-default. So a crate's documentation could break with no check noticing.

A lane is only worth adding if it lands green, so a warn-level workspace pass came first. It found **nine** warnings across three crates, all invisible for exactly that reason.

## Commits

| | |
|---|---|
| `c3d77ac5` | `docs(evidence)` — the `trust_card.rs` fix on its own, so it can be read against #1915 in isolation |
| `cfba6dec` | `ci(rustdoc)` — the lane, the eight other fixes, and the gate-contract update |

## The two that carried real information

Most were `<hex>`, `<u64>`, `<claim-id>=<level>` and `argv[0]` being read as HTML tags and reference links by rustdoc's Markdown renderer, plus two bare URLs. Two were not cosmetic:

- **`authorize_and_consume`** (`crates/assay-core/src/runtime/authorizer.rs:191`) had been a dead link since it was written. Intra-doc links in a method's docs resolve in *module* scope, not `impl` scope, so the bare name never resolved to the sibling inherent method. Now `Self::`-qualified and live.
- **`crate::tools::test_outbound`** (`crates/assay-mcp-server/src/auth/sensitive_headers.rs:10`) points at a `#[cfg(feature = "test-outbound")]` module, so it can never resolve in a default doc build. That block is the security-invariant PR packet listing the outbound callsites a reviewer is meant to check, and one of its two references did not navigate anywhere. Written as code text with the gate stated, rather than a link that only works under a feature.

`--require`'s text in `crates/assay-cli/src/cli/args/trust_basis.rs` is also clap `--help` output. Backticking it matches the convention the line directly above it already uses; rendered help was checked.

## The lane

Mirrors `clippy`, and is wired into the aggregate `ci` gate with its own result variable and expectation triple — so it blocks without a branch-protection change.

- **Default features, not `--all-features`.** That is the doc surface a consumer actually gets, and it keeps the lane off optional deps (tui, ebpf, test-outbound) it has no business building. A link that only resolves under a feature therefore must not be written as a link — hence the `test_outbound` fix above.
- **`--no-deps`** keeps a third-party crate's warning from being able to redden this gate.

## The gate contract caught the wiring twice

`scripts/ci/test-ci-gate-expectations.sh` extracts the gate's `run:` body out of `ci.yml` and executes it verbatim, so the thing under test is the thing that ships. It failed immediately on an unbound `RUSTDOC_RESULT` under `set -u`.

Adding the fixture alone would have left the new lane less protected than every other code-gated job, so `rustdoc` also joins the silently-skipped loop. Mutation-checked: with the `rustdoc` triple removed from the gate, the contract fails with `silently skipped RUSTDOC: expected the gate to fail, it passed`. The case is load-bearing, not decorative.

## Verification

All on the final tree, cold `target/`:

| Check | Result |
|---|---|
| `RUSTDOCFLAGS="-D warnings" cargo doc -p assay-evidence --no-deps` | exit 0 |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` | exit 0 |
| `cargo clippy --workspace --all-targets -- -D warnings` | exit 0, zero warnings |
| `cargo test --locked --workspace --exclude assay-ebpf --exclude assay-it` | exit 0 — 2786 passed, 0 failed, 197 suites |
| `bash scripts/ci/test-ci-gate-expectations.sh` | exit 0, 11/11 |
| Pre-commit + pre-push (fmt, typos, actionlint, shellcheck, gate contract, linux compile gate) | passed |

## Not fixed here

`cargo doc --workspace` still warns about an output filename collision at `target/doc/assay/index.html` between the `assay` lib in `assay-it` and the `assay` bin in `assay-cli` ([cargo#6313](https://github.com/rust-lang/cargo/issues/6313)). It is a *cargo* diagnostic, so `RUSTDOCFLAGS` cannot promote it and it will not redden this lane — but one of those two doc trees does clobber the other. Its own decision.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -223,6 +223,52 @@ jobs:
           components: clippy
       - run: cargo clippy --workspace --all-targets -- -D warnings
 
+  rustdoc:
+    name: Rustdoc (deny warnings)
+    needs: [scope]
+    if: needs.scope.outputs.lightweight_only != 'true'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
+        with:
+          persist-credentials: false
+      - name: Swatinem/rust-cache
+        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+      - name: Install Linux deps (for build scripts)
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          set -euo pipefail
+          sudo DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends clang libsqlite3-dev || {
+            sudo DEBIAN_FRONTEND=noninteractive apt-get update -y \
+              -o Acquire::Retries=10 \
+              -o Acquire::http::Timeout=60 \
+              -o Acquire::https::Timeout=60 \
+              -o Acquire::CompressionTypes::Order::=gz \
+              -o Acquire::ForceIPv4=true \
+              -o Acquire::Languages=none
+            sudo DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends clang libsqlite3-dev
+          }
+      - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
+      # Clippy above never sees rustdoc's lints: broken intra-doc links, unclosed HTML tags in doc
+      # comments and bare URLs are diagnosed by rustdoc alone, and they are warn-by-default, so
+      # nothing failed while a crate's documentation quietly stopped building. `assay-evidence`
+      # carried a `redundant explicit link target` that survived exactly this way.
+      #
+      # Deliberately default features, not `--all-features`: this mirrors the doc surface a
+      # consumer gets, and it keeps the lane from depending on optional features whose deps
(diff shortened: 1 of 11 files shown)


Rul1an:
Blocking review finding on exact head `f2e14a5aca48aadc687e96e94e67aca99d2c9b2b`:\n\nThe new rustdoc gate covers only default features, while a supported public feature surface is already red:\n\n```bash\nRUSTDOCFLAGS="-D warnings" cargo doc -p assay-registry --features oidc --no-deps\n```\n\nThis fails at `crates/assay-registry/src/auth.rs:126` on a bare URL, while both `Rustdoc (deny warnings)` and the aggregate `CI` are green. Please fix that warning and extend the gate to an explicit supported public-feature doc matrix. Do not use blind workspace `--all-features` if eBPF/internal feature surfaces need a separate toolchain. A regression check should prove the feature build is actually executed and cannot be silently skipped.\n\nBuilder review does not count; this finding came from an independent non-building exact-head review.

PR #1927, opened 2026-07-29
fix(release): emit portable checksum sidecars

### LINKED ISSUE #1926: ci(release): emit flat LF-only checksum sidecars
## Problem

The independent post-release audit of `v3.36.0` downloaded all GitHub Release assets into one directory and ran each `.sha256` file with `shasum -a 256 -c`. Six sidecars did not work directly:

- four generated after assets enter `release/` retain a `release/` path prefix (`.mcpb`, SBOM, release provenance, proof kit);
- the Windows sidecar uses CRLF, so POSIX `shasum -c` treats `\r` as part of the archive name.

All 11 first-field digests were independently recomputed and match their corresponding asset bytes. The canonical offline proof-kit verifier also succeeds. This is a checksum-sidecar portability defect, not an asset-integrity failure. The `v3.36.0` release notes now document a first-field comparison as the safe workaround.

## Direction

Generate every checksum from the directory that will be published, write only the basename, and force LF on every platform. Prefer one shared helper instead of separate shell/PowerShell formatting. Extend `check-release-assets.sh` with a downloaded-layout fixture that runs the platform-neutral equivalent of `sha256sum -c`.

## Acceptance

- every release `.sha256` line is `<64 lowercase hex><two spaces><basename><LF>`;
- no slash, carriage return, or absolute path appears in the filename field;
- a flat directory containing assets and sidecars verifies without rewriting;
- the preflight test covers Unix, generated-in-`release/`, and Windows sidecars;
- provenance/proof-kit generation remains unchanged and green.

Observed in release run: https://github.com/Rul1an/assay/actions/runs/30434001482

Rul1an:
## Summary

- generate all eleven release checksum sidecars through one LF-only, basename-only helper
- reject CRLF, path-prefixed, absolute, malformed, and multi-line sidecars before publication
- verify the actual flat download layout with the platform-native checksum verifier
- run the release asset contract on Ubuntu and Windows as part of required PR CI
- keep MCPB helper changes in the registry smoke scope

## Test evidence

- legacy preflight reproduced the defect: a CRLF Windows sidecar passed before the fix
- `bash scripts/ci/test-write-sha256-sidecar.sh`
- `bash scripts/ci/test-release-assets.sh`
- `bash scripts/ci/test-ci-gate-expectations.sh`
- `bash scripts/ci/mcp_registry_foundation_smoke.sh`
- ShellCheck and actionlint with the repository-documented `concurrency.queue` exception
- full pre-commit and pre-push hooks, including workspace clippy and the Linux compile gate

## Non-claims

- does not change primary binaries, platform archives, semantic provenance records, or attestation bundles
- the proof-kit archive changes because its copied provenance checksum sidecar becomes basename-only and LF-only
- does not rewrite previously published release assets

Fixes #1926

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,7 @@ jobs:
               lightweight_only=true
             fi
 
-            if grep -E '^(crates/assay-mcp-server/|packaging/mcpb/|packaging/mcp-registry/|scripts/ci/build_mcpb_bundle\.sh$|scripts/ci/render_registry_server_json\.sh$|scripts/publish-mcpcentral\.sh$|\.github/workflows/release\.yml$)' "$changed_file" >/dev/null; then
+            if grep -E '^(crates/assay-mcp-server/|packaging/mcpb/|packaging/mcp-registry/|scripts/ci/(build_mcpb_bundle|write_sha256_sidecar)\.sh$|scripts/ci/render_registry_server_json\.sh$|scripts/publish-mcpcentral\.sh$|\.github/workflows/release\.yml$)' "$changed_file" >/dev/null; then
               mcp_registry_touched=true
             else
               mcp_registry_touched=false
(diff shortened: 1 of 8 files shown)


Rul1an:
## Exact-head review quorum

Final head: `0a048bf6633bc9eec1681feb6d3681e5154e1b8d`.

The automated review paths are unavailable on this head and are not counted: CodeRabbit reports `Review rate limited`; Copilot reported quota exhaustion on an earlier head. Per the user-authorized immediate fallback, two independent non-building read-only agents reviewed the final SHA.

- **Avicenna:** approved with no remaining findings. Verified active Ubuntu and Windows matrix coverage, both matrix-bypass mutations, active/unique producer calls, commented-producer bite, and the fail-closed CI rollup.
- **Erdos:** approved with no remaining findings. Verified the command no-op controls (`if: false`, `continue-on-error: true`), commented Windows and `matrix.exclude` mutations, exact sidecar format tests, and corrected proof-kit non-claims.

Earlier P2 findings from both reviewers were fixed before these approvals. No reviewer edited, committed, pushed, or posted on the branch. Merge still depends on all exact-head required checks being green and zero unresolved actionable threads.

PR #1929, opened 2026-07-29
feat(mcp): observe the protocol era on import without touching the event shape

Rul1an:
Slice 0 of #1866. `transport_context` did not survive import at all, so nothing downstream could tell
a 2026 transcript from a 2025 one, and a result with no `resultType` read the same either way.

This body describes the branch as it stands. It replaces a per-head correction log that had accreted
across sixteen heads, several of whose verification lines had gone stale in place. The history is in
the commits and in #1866; keeping a contradictory summary next to it was worse than keeping none.

## Three axes, not one field

| Axis | Question |
|---|---|
| `EnvelopeObservation` | what the transport framing carried |
| `EraResolution` | what the era turned out to be once every signal was read |
| `ResultObservation` | what a response said about its own completeness |

A format with no transport envelope has not lost the era: the 2026 request carries its own required
`_meta["io.modelcontextprotocol/protocolVersion"]`, so bare JSON-RPC and Inspector resolve from the
body. An earlier design refused on the missing envelope and would have ceilinged that evidence.

Conclusions are typed rather than boolean. Unknown, contradicted and a missing required field are
three findings, and `input_required` is valid while not being terminal. Requests and responses get
separate verdict types, since a request has no result and "no objection" is not "valid but unfinished".

## Order is the contract

Format capability, then unreadable, then contradiction, then classification, and absence read last
under the resolved era. Each step exists because the alternative fails silently: a format that cannot
carry the signal must never be asked whether one is missing, and two values that are not both versions
cannot be said to disagree about a version.

The schema settles the last step. A missing `resultType` is a violation at 2026 and prescribed as
`"complete"` before it, so the same observation has opposite conclusions and the era decides which.

## Call correlation

A response carries no era signal of its own, so it takes the era of the call it answers, correlated on
a crate-private typed key derived from the payload's raw JSON. Without this a request whose header and
body disagreed was `Conflicting` while its response resolved to `Known` from the header alone, and a
missing `resultType` under that legacy era read `Terminal`: a contradicted call concluding that the
action completed.

The typed outstanding map is the sole call-lifetime authority. Correlation runs in source order, a
response consumes the request that preceded it, an orphan response keeps the era it resolved on its
own, and two calls outstanding on one id are refused. `mcp_events_to_v2_trace` derives the same key, so
the public `String` rendering is nowhere in the correlation path.

**Acceptance and correlation are separate questions.** Any string or number the serde boundary can
represent is an acceptable id. Only a string, or an integer exactly representable as `i64`/`u64`,
becomes a correlation key; every other number is ingestible and correlates with nothing, keeping the
era its own envelope resolved. serde parses every non-integer through `f64`, so `9007199254740993.0`
has already become `...92.0` before this code sees it — a property of the parse that no downstream key
design recovers. False correlation hands a response the era of a call it did not answer and can license
a terminal reading; declining to key is incomplete and true.

## What refuses, and where

A malformed **era observation** never refuses. `Malformed` on any axis parses and the conclusion layer
classifies it. That is what the ordering contract protects.

A malformed **JSON-RPC message shape** refuses at the parser boundary, value-free: a present non-string
`method`, an id that is a boolean, array or object, a request-only method with no id, a response that
is neither a result nor an error, an SSE frame with no `data`, and any duplicate object member.

Duplicate members are refused as the tree is built, because `serde_json` collapses them last-value-wins
and a post-collapse check cannot recover the evidence. The rule is every duplicate rather than a list
of significant names: a name list goes stale the moment a significant key is added, and this branch
shipped three enumerations that missed a member. `DuplicateAwareSink` walks unknown subtrees so the
claim holds where it is made.

Request-only methods cannot become notifications by omitting an id. Twenty-two names across the five
revisions this build recognizes are pinned exhaustively, and the table is compared against the constant
first so drift fails in the test rather than in a transcript. An unknown extension method stays a
notification.

## No public API change

`McpEvent` keeps its five fields and `parse_mcp_transcript` still returns `Vec<McpEvent>`, projected
out of an internal detailed parse. `crates/assay-core/src/mcp/types.rs` is byte-identical to
`origin/main` and the diff adds no `pub` item. The whole vocabulary is `pub(crate)`: the crate is
published, and adding fields to a constructible, non-`#[non_exhaustive]` struct would break downstream
struct literals for a shape still being worked out.

The conclusion functions and reason enums carry a per-item `#[allow(dead_code)]` because slice 2 is
their only production consumer. Targeted rather than module-wide, so a future dead item is still
reported; `#[expect]` is unfulfilled here because the tests use them under `--all-targets`.

## Test files changed, and why

Two of the three MCP integration tests are changed on this PR. Earlier versions of this body said all
three were byte-unchanged from `main`, which was not true then either.

- `mcp_import_smoke.rs`: 9 insertions, 3 deletions. Its three request-only lines had no ids, which is
  exactly the defect that a request-only method may omit one, so the fixture encoded it. The
  line-order and Step-order assertions are untouched; the test is about deterministic line fallback.
- `mcp_id_correlation.rs`: the duplicate-id diagnostic echoed the id, which is input-chosen data this
  slice bans everywhere else. Reuse after a response, number-versus-string, and typed ids through the
  trace mapper are pinned as the cases they are.
- `mcp_transport_compat.rs`: **byte-unchanged, and the only one.** It is the negative control.

## Verification on `f71c3dce`

Base `3e657ed0`, worktree `scratchpad/slice0`.

- `cargo test -p assay-core`: 752 passed, 0 failed, 5 ignored, measured on `c3dcb365`
- focused MCP suite: 326 passed, 0 failed, measured on `2df6a658`
- `mcp_id_correlation` 18, `mcp_import_smoke` 2, `mcp_transport_compat` 9
- `cargo clippy -p assay-core --all-targets -- -D warnings`: nothing
- `cargo fmt --all -- --check` and `git diff --check`: clean
- `cargo doc --document-private-items` with warnings denied: one unresolved link, `VerifyLimits` in
  `crates/assay-core/src/replay/bundle/limits.rs`. It is present on `origin/main` and lives in a
  directory this branch does not touch, so it is pre-existing rather than introduced here.

Guards are mutation-tested rather than assumed. Removing each takes tests down: integer ids ceasing to
correlate 6, any number correlating 3, ignored subtrees unscanned 2, sse null handling 2, one method
dropped from the constant 2, the params container guard 3, the shape pre-check 6, day-of-month 3,
per-entry envelope folding 4.

**CI on `6479bc6d` found one stale assertion, and `270d95f8` fixes it.** Ubuntu job 90932659745:
`crates/assay-cli/tests/mcp_id_correlation_errors.rs:63` still expected
`duplicate tools/call request id`, the message of the global gate this branch deleted. Not a parser
regression — the refusal fires and the boolean-id control is green. The assertion now pins
`two outstanding JSON-RPC requests share an id`, taken verbatim from `parser.rs:131`, and nothing else:
not the id, which the parser deliberately no longer echoes, and not the source line, which moves
whenever a fixture gains one. A second assertion pins that the id is absent from stderr.

It was missed because the sweep for the old gate's message covered `assay-core` and not the CLI crate,
which asserts on the same parser through the binary's stderr. A message is part of a contract wherever
it is read, not only where it is produced.

**Three heads sit on top of the measured runs, and none is claimed green locally.** `6b86c1e6` adds a
comment; `6479bc6d` corrects the fixture wrapper and adds a non-string-method control. From `2df6a658`
onward, compile and test processes on this host are being killed without diagnostic output under memory
and IO pressure, so neither the `6479bc6d` nor the `270d95f8` test compile completed, with no partial output to
report. Required CI on the exact head is the proof for both, not a local run. The local pre-commit hook was skipped for the same reason and
that is recorded in the commit body; `cargo fmt --all -- --check` and `git diff --check` were run
directly and are clean.

One further environmental note, because the 752 number deserves its conditions. The `752 passed / 0 failed`
run was green as measured. A later doc-only rerun of the same suite reported one failure,
`tests::benchmark_combined_checks` in `latency_check.rs`, which asserts on `start.elapsed()`. It was
running alongside a workspace clippy with the disk at 99 percent, and the same test fails identically
on the unmodified base with this branch's changes stashed. It is a property of the machine under
parallel load, not of this branch, and the head `6b86c1e6` adds only a comment on top of `c3dcb365`.

Two of the mutation results measured 0 red on a first pass, and neither time was the guard untested. Once the
replacement anchor had not matched the formatted source, so nothing was mutated; once the test asserted
only `is_err()`, which the mutant also satisfies for a different reason. A mutation that changes more
than the guard, or an assertion that any refusal satisfies, looks identical to coverage.

## Out of scope

Multi-hop calls spread across separate records: correlation is by JSON-RPC id within one transcript,
and a call-scoped identity across records is not defined here.

Duplicate refusal covers what reaches this crate's MCP entry points. It is not a claim that every JSON
path in the workspace is duplicate-aware.

The conclusion layer that consumes these axes is slice 2. Nothing in production reads the sidecar yet.

## Rebased onto `128d4c79`

`main` moved while this PR was in review (docs PR #1932), and branch protection has
`required_status_checks.strict = true`, so the PR was `MERGEABLE` but `BEHIND`. Rebased onto current
`origin/main`; the previous head was `270d95f8`.

**Conflict-free, and the content is provably untouched.** All 26 commits replayed with no conflict, and
the patch this branch produces against its base is byte-identical before and after: 4512 lines, same
checksum. No content edit, no refactor, no squash. The base moves and the change does not.

Every prior review and CI result is invalidated by the new head. Exact-head review and required CI on
`f71c3dce` are what stand now, and CI is the proof: nothing on this head is locally verified beyond
`cargo fmt --all -- --check` and `git diff --check`, since focused runs on this host are still being
killed under memory and IO pressure.

### DIFF
--- a/crates/assay-core/src/mcp/era.rs
+++ b/crates/assay-core/src/mcp/era.rs
@@ -151,10 +151,32 @@ pub(crate) enum CorrelationId {
 /// `None` for an absent or unusable id. `normalize_jsonrpc_id` already refuses booleans, arrays and
 /// objects at the parser boundary, so only the two valid shapes reach this.
 pub(crate) fn correlation_id(v: &serde_json::Value) -> Option<CorrelationId> {
-    match v.get("id") {
-        Some(serde_json::Value::String(s)) => Some(CorrelationId::Str(s.clone())),
-        Some(serde_json::Value::Number(n)) => Some(CorrelationId::Num(n.to_string())),
+    v.get("id").and_then(accept_id)
+}
+
+/// The one place that decides whether an id is usable, and what key it becomes.
+///
+/// MCP restricts `RequestId` to a string or an integer. A fractional number is not an id, and
+/// neither is one spelled with an exponent: `1.0` and `1e0` both parse as floats, so neither is an
+/// integer even though the second equals one. Accepting them would give two spellings of one call
+/// two different keys, or one key to two different calls, depending on which side rendered.
+///
+/// `CorrelationId::Num` therefore only ever comes from an accepted integer.
+pub(crate) fn accept_id(v: &serde_json::Value) -> Option<CorrelationId> {
+    match v {
+        serde_json::Value::String(s) => Some(CorrelationId::Str(s.clone())),
+        serde_json::Value::Number(n) if n.is_i64() || n.is_u64() => {
+            Some(CorrelationId::Num(n.to_string()))
+        }
         _ => None,
+    }
+}
+
+/// The text an accepted id renders to for the public `McpEvent::jsonrpc_id` field.
+pub(crate) fn accepted_id_text(id: &CorrelationId) -> String {
+    match id {
+        CorrelationId::Str(s) => s.clone(),
+        CorrelationId::Num(n) => n.clone(),
     }
 }
 
--- a/crates/assay-core/src/mcp/era.rs
+++ b/crates/assay-core/src/mcp/era.rs
@@ -35,37 +35,64 @@ pub(crate) enum EnvelopeObservation {
     Malformed,
 }
 
-/// Which JSON-RPC message shape this is.
+/// Which JSON-RPC message shape this is, carrying the method so that nothing has to read it a
+/// second time.
 ///
 /// The distinction is load-bearing rather than tidy. `RequestParams._meta` is required and carries
 /// the protocol version; `NotificationParams._meta` is optional and is a different type that does
 /// not carry it. Treating a notification as a request therefore invents a fault: a
 /// `notifications/progress` under 2026 with no `_meta` is correct, and would be reported as missing
 /// required metadata.
+///
+/// The method travels inside the variant on purpose. Two callers each reaching for `method` with
+/// their own `as_str()` is how the discriminants drifted apart before: the parser read one thing and
+/// the era axes another, and a shape only one of them rejected fell through the gap.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub(crate) enum MessageKind {
-    Request,
-    Notification,
+pub(crate) enum MessageKind<'a> {
+    Request { method: &'a str },
+    Notification { method: &'a str },
     Response,
 }
 
-/// Classify by the two fields JSON-RPC uses. A notification is a request object *without* an `id`
-/// member, so absence is the discriminant and nothing about the value is.
+/// A message shape that cannot be classified at all.
 ///
-/// An explicit `"id": null` is therefore a request, not a notification. `RequestId` is `string |
-/// number`, so a null id is an invalid request id rather than an absent one, and treating it as a
-/// notification drops the required 2026 request metadata for any message that writes one. That is a
-/// fail-open reachable by a single token. The invalid id is a JSON-RPC validity question this slice
-/// does not own; what it owns is that the metadata requirement still applies.
-pub(crate) fn classify_message(v: &serde_json::Value) -> MessageKind {
-    match v.get("method").and_then(|m| m.as_str()) {
-        Some(_) if v.get("id").is_some() => MessageKind::Request,
-        Some(_) => MessageKind::Notification,
-        None => MessageKind::Response,
-    }
+/// Value-free: the offending value is chosen by the input, and naming its type is all a reader can
+/// act on.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+pub(crate) enum MessageShapeError {
+    #[error("JSON-RPC method must be a string")]
+    NonStringMethod,
 }
 
-/// What a request carried in `params._meta`. Typed rather than `Option<String>` so that a missing
+/// Classify by the two fields JSON-RPC uses.
+///
+/// A notification is a request object *without* an `id` member, so absence is the discriminant and
+/// nothing about the value is. An explicit `"id": null` is therefore a request, not a notification:
+/// `RequestId` is `string | number`, which makes a null id an invalid request id rather than an
+/// absent one, and calling it a notification drops the required 2026 request metadata for any
+/// message that writes that one token.
+///
+/// A present `method` that is not a string is a malformed message shape rather than a message of
+/// some other kind. Answering `Response` for it, which is what folding through `as_str()` does,
+/// silently drops the request-metadata requirement the same way. The id vocabulary is settled
+/// separately by `normalize_jsonrpc_id`, which refuses booleans, arrays and objects.
+pub(crate) fn classify_message(
+    v: &serde_json::Value,
+) -> Result<MessageKind<'_>, MessageShapeError> {
+    let Some(method) = v.get("method") else {
+        return Ok(MessageKind::Response);
+    };
+    let Some(method) = method.as_str() else {
+        return Err(MessageShapeError::NonStringMethod);
+    };
+    Ok(if v.get("id").is_some() {
+        MessageKind::Request { method }
+    } else {
+        MessageKind::Notification { method }
+    })
+}
+
+/// What a request carried in `params._meta`./// What a request carried in `params._meta`. Typed rather than `Option<String>` so that a missing
 /// field and an unreadable one stay different findings: one is silence, the other is a malformed
 /// signal, and they have different remediations.
 #[derive(Debug, Clone, PartialEq, Eq)]
--- a/crates/assay-core/src/mcp/era.rs
+++ b/crates/assay-core/src/mcp/era.rs
@@ -47,7 +47,7 @@ pub(crate) enum EnvelopeObservation {
 /// enumerations that missed a member. RFC 8259 says names SHOULD be unique and RFC 7493 requires it,
 /// so refusing outright cannot go stale. Value-free: the diagnostic names neither the key nor either
 /// value.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub(crate) struct UniqueValue(pub(crate) serde_json::Value);
 
 impl<'de> serde::Deserialize<'de> for UniqueValue {
@@ -135,17 +135,84 @@ impl<'de> serde::Deserialize<'de> for UniqueValue {
 /// unknown member passed while the claim said every duplicate is refused. A manual `visit_map` that
 /// runs this on every key closes it inside the same pass, with no second read of the input.
 #[derive(Default)]
-pub(crate) struct SeenMembers(Vec<String>);
+pub(crate) struct SeenMembers(std::collections::HashSet<String>);
 
 impl SeenMembers {
+    /// Set membership rather than a linear scan. Keys are attacker-chosen and need not repeat, so a
+    /// scan is quadratic in the number of *unique* members, which is the cheap shape for an input to
+    /// have. The existing input ceiling is the outer budget; this keeps the inner cost linear.
     pub(crate) fn insert<E: serde::de::Error>(&mut self, key: &str) -> Result<(), E> {
-        if self.0.iter().any(|k| k == key) {
+        if !self.0.insert(key.to_string()) {
             return Err(serde::de::Error::custom(
                 "JSON object contains a duplicate member",
             ));
         }
-        self.0.push(key.to_string());
         Ok(())
+    }
+}
+
+/// Walk a value for duplicate members without building it.
+///
+/// `IgnoredAny` skips a subtree entirely, so a duplicate inside an unknown member escaped the rule
+/// while the claim said every duplicate is refused. This visits the same ground and keeps nothing,
+/// so an ignored subtree is still read for duplicates and still costs no tree.
+pub(crate) struct DuplicateAwareSink;
+
+impl<'de> serde::Deserialize<'de> for DuplicateAwareSink {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        struct V;
+        impl<'de> serde::de::Visitor<'de> for V {
+            type Value = DuplicateAwareSink;
+            fn expecting(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                f.write_str("any JSON value with unique object members")
+            }
+            fn visit_unit<E>(self) -> Result<DuplicateAwareSink, E> {
+                Ok(DuplicateAwareSink)
+            }
+            fn visit_none<E>(self) -> Result<DuplicateAwareSink, E> {
+                Ok(DuplicateAwareSink)
+            }
+            fn visit_some<D: serde::Deserializer<'de>>(
+                self,
+                d: D,
+            ) -> Result<DuplicateAwareSink, D::Error> {
+                <DuplicateAwareSink as serde::Deserialize>::deserialize(d)
+            }
+            fn visit_bool<E>(self, _: bool) -> Result<DuplicateAwareSink, E> {
+                Ok(DuplicateAwareSink)
+            }
+            fn visit_i64<E>(self, _: i64) -> Result<DuplicateAwareSink, E> {
+                Ok(DuplicateAwareSink)
+            }
+            fn visit_u64<E>(self, _: u64) -> Result<DuplicateAwareSink, E> {
+                Ok(DuplicateAwareSink)
+            }
+            fn visit_f64<E>(self, _: f64) -> Result<DuplicateAwareSink, E> {
+                Ok(DuplicateAwareSink)
+            }
+            fn visit_str<E>(self, _: &str) -> Result<DuplicateAwareSink, E> {
+                Ok(DuplicateAwareSink)
+            }
+            fn visit_seq<A: serde::de::SeqAccess<'de>>(
+                self,
+                mut seq: A,
+            ) -> Result<DuplicateAwareSink, A::Error> {
+                while seq.next_element::<DuplicateAwareSink>()?.is_some() {}
+                Ok(DuplicateAwareSink)
+            }
+            fn visit_map<A: serde::de::MapAccess<'de>>(
+                self,
+                mut map: A,
+            ) -> Result<DuplicateAwareSink, A::Error> {
+                let mut seen = SeenMembers::default();
+                while let Some(key) = map.next_key::<String>()? {
+                    seen.insert::<A::Error>(&key)?;
+                    map.next_value::<DuplicateAwareSink>()?;
+                }
+                Ok(DuplicateAwareSink)
+            }
+        }
+        deserializer.deserialize_any(V)
     }
 }
 
@@ -157,8 +224,8 @@ impl SeenMembers {
 /// request, took its era, and a missing `resultType` under a borrowed legacy era could read
 /// `Terminal`.
 ///
-/// The number is kept as its canonical text rather than parsed, so two spellings of one number
-/// become one key. The variant, not the text, carries the type distinction.
+/// A number key is the exact text of an integer this build could represent without loss. The variant,
+/// not the text, carries the type distinction, so a string is never a number however it is spelled.
 /// Crate-private: the public field is unchanged, byte and API compatible.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub(crate) enum CorrelationId {
@@ -166,32 +233,48 @@ pub(crate) enum CorrelationId {
     Num(String),
 }
 
+/// Whether an id is usable at all, which is a different question from whether it can correlate.
+///
+/// MCP restricts `RequestId` to a string or a number, so those are the shapes a request or a success
+/// response may carry. Nothing here decides correlation: see [`correlation_id`].
+pub(crate) fn id_is_acceptable(v: &serde_json::Value) -> bool {
+    matches!(
+        v,
+        serde_json::Value::String(_) | serde_json::Value::Number(_)
+    )
+}
+
 /// Read the correlation key from a message's raw JSON.
 ///
-/// `None` for an absent or unusable id. [`accept_id`] owns the vocabulary; the parser's shape
+/// `None` for an absent or unusable id. This function owns the correlation vocabulary; the parser's shape
 /// diagnostics for booleans, arrays and objects are more specific messages for a subset of the same
 /// refusal.
 pub(crate) fn correlation_id(v: &serde_json::Value) -> Option<CorrelationId> {
-    v.get("id").and_then(accept_id)
+    v.get("id").and_then(correlation_key)
 }
 
 /// The one place that decides whether an id is usable, and what key it becomes.
 ///
-/// The pinned 2026 schema says `RequestId = string | number`, and this function is the only place
-/// that decides it. An earlier head read "integer" out of a prose overview and refused fractional
-/// and exponent-spelled ids as protocol-invalid. They are not: JSON-RPC says fractional ids SHOULD
-/// NOT be used, which is advice to a sender rather than a rule a receiver may enforce as invalidity.
+/// The correlation key for an id, or `None` when this build cannot key on it safely.
 ///
-/// Two numeric ids are the same id when they are the same number, so the key is the canonical text
-/// serde renders: `1e0` and `1.0` both become `1.0` and correlate, while `1` stays `1`. That last
-/// pair is a bounded non-claim rather than a decision — `1` and `1.0` are numerically equal and
-/// JSON-RPC does not say whether they name one call — and this build treats them as distinct keys.
+/// A string keys on itself, exactly. A number keys on its exact text only when it is representable as
+/// an `i64` or a `u64`; every other JSON number is ingestible and simply does not correlate.
 ///
-/// `CorrelationId::Num` therefore only ever comes from an accepted integer.
-pub(crate) fn accept_id(v: &serde_json::Value) -> Option<CorrelationId> {
+/// That asymmetry is the point. Upstream does not define whether `1`, `1.0` and `1e0` name one call,
+/// and its own prose and schema disagree about whether an id may be fractional at all. Meanwhile
+/// serde_json parses any non-integer number through `f64`, so `9007199254740993.0` has already become
+/// `9007199254740992.0` before a key could be built, and two different ids would land on one call.
+///
+/// Correlating on a value that may be wrong is worse than not correlating: a false pairing hands a
+/// response the era of a call it did not answer, which can license a terminal reading. Declining to
+/// key leaves the response with the era its own envelope resolved, which is incomplete and true. A
+/// later sidecar that keeps the raw lexeme can widen the subset; nothing here forecloses that.
+pub(crate) fn correlation_key(v: &serde_json::Value) -> Option<CorrelationId> {
     match v {
         serde_json::Value::String(s) => Some(CorrelationId::Str(s.clone())),
-        serde_json::Value::Number(n) => Some(CorrelationId::Num(n.to_string())),
+        serde_json::Value::Number(n) if n.is_i64() || n.is_u64() => {
+            Some(CorrelationId::Num(n.to_string()))
+        }
         _ => None,
     }
 }
@@ -285,7 +368,7 @@ pub(crate) enum MessageShapeError {
 /// [`REQUEST_ONLY_METHODS`] classifies as requests whatever their id member does, so a request-only
 /// method cannot shed the requirement by omitting one.
 ///
-/// The id vocabulary is not settled here. [`accept_id`] owns it; the parser adds more specific shape
+/// The id vocabulary is not settled here. [`correlation_id`] owns it; the parser adds more specific shape
 /// diagnostics for booleans, arrays and objects, which are a subset of the same refusal.
 pub(crate) fn classify_message(
     v: &serde_json::Value,
--- a/crates/assay-core/src/mcp/era.rs
+++ b/crates/assay-core/src/mcp/era.rs
@@ -108,12 +108,17 @@ impl<'de> serde::Deserialize<'de> for UniqueValue {
             ) -> Result<UniqueValue, A::Error> {
                 let mut out = serde_json::Map::new();
                 while let Some(key) = map.next_key::<String>()? {
-                    let UniqueValue(value) = map.next_value()?;
-                    if out.insert(key, value).is_some() {
+                    // Membership is checked on the key alone, before `next_value` materializes
+                    // whatever the duplicate carries. Reading it first means a hostile input pays
+                    // for a value that is only going to be refused, and the value may be an
+                    // arbitrarily large subtree.
+                    if out.contains_key(&key) {
                         return Err(serde::de::Error::custom(
                             "JSON object contains a duplicate member",
                         ));
                     }
+                    let UniqueValue(value) = map.next_value()?;
+                    out.insert(key, value);
                 }
                 Ok(UniqueValue(serde_json::Value::Object(out)))
             }
--- a/crates/assay-core/src/mcp/era.rs
+++ b/crates/assay-core/src/mcp/era.rs
@@ -77,8 +77,12 @@ pub(crate) enum ResultObservation {
     InputRequired,
     /// `ResultType` is an open union and the handling rule is a SHOULD, so conforming verifiers
     /// may legitimately differ here.
-    Unrecognized(String),
-    /// The field is present and is not a token at all: a number, an object, an empty string.
+    ///
+    /// Value-free by design. The token is attacker-chosen, so echoing it hands a channel into
+    /// every log that ingests the finding while telling a reader nothing they can act on: the
+    /// actionable fact is that this build has no rule for it, not which bytes it was.
+    Unrecognized,
+    /// The field is present and is not a token at all: a number, an object, an array.
     /// Distinct from `Missing`, and the distinction is load-bearing on the legacy arm, where an
     /// absent field MUST be read as `"complete"` and an unreadable one must not inherit that.
     Malformed,
@@ -91,7 +95,7 @@ pub(crate) enum ResultObservation {
 #[allow(dead_code)]
 pub(crate) enum IncompleteReason {
     EraUnknown(UnknownReason),
-    UnrecognizedResultType(String),
+    UnrecognizedResultType,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -295,8 +299,8 @@ pub(crate) fn conclude(era: &EraResolution, observed: &ResultObservation) -> Res
     match observed {
         ResultObservation::Complete => ResultConclusion::Terminal,
         ResultObservation::InputRequired => ResultConclusion::NonTerminal,
-        ResultObservation::Unrecognized(token) => {
-            ResultConclusion::Incomplete(IncompleteReason::UnrecognizedResultType(token.clone()))
+        ResultObservation::Unrecognized => {
+            ResultConclusion::Incomplete(IncompleteReason::UnrecognizedResultType)
         }
         // Checked before the era, because an unreadable field is not an absent one and must not
         // reach the backward-compatibility rule that reads absence as completion.
@@ -359,25 +363,6 @@ pub(crate) fn conclude_request(
 }
 
 /// The `_meta` key the protocol version travels under on a request.
-/// How much of an unrecognized token is kept. The two defined tokens are 8 and 14 bytes, so this
-/// is generous for anything a future revision might name while still bounding what one record can
-/// cost.
-pub(crate) const MAX_TOKEN_BYTES: usize = 64;
-
-/// Truncate on a character boundary, since cutting a UTF-8 string at a byte index panics.
-fn bounded_token(token: &str) -> String {
-    if token.len() <= MAX_TOKEN_BYTES {
-        return token.to_string();
-    }
-    let cut = token
-        .char_indices()
-        .map(|(i, _)| i)
-        .take_while(|i| *i <= MAX_TOKEN_BYTES)
-        .last()
-        .unwrap_or(0);
-    token[..cut].to_string()
-}
-
 pub(crate) const PROTOCOL_VERSION_META_KEY: &str = "io.modelcontextprotocol/protocolVersion";
 
 /// The transport header carrying the same value.
@@ -460,8 +445,9 @@ pub(crate) fn observe_request_metadata(raw: &serde_json::Value) -> RequestMetada
     let Some(meta) = raw.get("params").and_then(|p| p.get("_meta")) else {
         return RequestMetadata::Absent;
     };
-    // `_meta` is an object by schema. A scalar or array here is a signal that arrived and failed,
-    // and `Value::get` on a non-object answers `None`, which would report it as silence.
+    // `_meta` is an object by schema, so a scalar or array is a signal that arrived and failed.
+    // The guard is explicit because `Value::get` on a non-object answers `None`, which the arm
+    // below would have read as the version simply being absent.
     if !meta.is_object() {
         return RequestMetadata::Malformed;
     }
@@ -490,9 +476,10 @@ pub(crate) fn observe_result(raw: &serde_json::Value) -> Option<ResultObservatio
         Some(v) => match v.as_str() {
             Some("complete") => ResultObservation::Complete,
             Some("input_required") => ResultObservation::InputRequired,
-            // `ResultType` is an open string union, so any string is syntactically a token. An
-            // empty one is unrecognized rather than unreadable; only a non-string is malformed.
-            Some(other) => ResultObservation::Unrecognized(bounded_token(other)),
+            // `ResultType` is an open string union, so any string is syntactically a token,
+            // including an empty one. Unrecognized is a statement about this build's rules, so it
+            // carries no value; only a non-string is unreadable.
+            Some(_) => ResultObservation::Unrecognized,
             None => ResultObservation::Malformed,
         },
     })
--- a/crates/assay-core/src/mcp/era.rs
+++ b/crates/assay-core/src/mcp/era.rs
@@ -92,7 +92,7 @@ pub(crate) fn classify_message(
     })
 }
 
-/// What a request carried in `params._meta`./// What a request carried in `params._meta`. Typed rather than `Option<String>` so that a missing
+/// What a request carried in `params._meta`. Typed rather than `Option<String>` so that a missing
 /// field and an unreadable one stay different findings: one is silence, the other is a malformed
 /// signal, and they have different remediations.
 #[derive(Debug, Clone, PartialEq, Eq)]
--- a/crates/assay-core/src/mcp/era.rs
+++ b/crates/assay-core/src/mcp/era.rs
@@ -179,8 +179,8 @@ pub(crate) const SUPPORTED_VERSIONS: &[&str] = &[
 /// Ordering against [`RESULT_TYPE_SINCE`] is lexicographic, and lexicographic ordering is date
 /// ordering only once the value is a date. Ten bytes and two dashes is not enough: `2026-99-99`
 /// would pass a shape-only check and be reported as a version this build merely does not support,
-/// which blames the reader for a record that is wrong. Validated as a real calendar date, leap
-/// years included, so `2026-02-31` is malformed rather than unsupported.
+/// which blames the reader for a record that is wrong. The value is validated as a real calendar
+/// date, leap years included, so `2026-02-31` is malformed rather than unsupported.
 fn is_version_shaped(v: &str) -> bool {
     let b = v.as_bytes();
     if b.len() != 10 || b[4] != b'-' || b[7] != b'-' {
--- a/crates/assay-core/src/mcp/era.rs
+++ b/crates/assay-core/src/mcp/era.rs
@@ -366,18 +366,26 @@ pub(crate) const PROTOCOL_VERSION_HEADER: &str = "mcp-protocol-version";
 /// reason `Malformed` exists, so this reads the slot itself.
 pub(crate) fn observe_header(headers: Option<&serde_json::Value>) -> Option<EnvelopeObservation> {
     let map = headers?.as_object()?;
-    let mut matches = map
+    let mut found: Option<&str> = None;
+    let mut any = false;
+    for (_, value) in map
         .iter()
-        .filter(|(k, _)| k.to_ascii_lowercase() == PROTOCOL_VERSION_HEADER);
-    let (_, value) = matches.next()?;
-    // Two spellings of one header is not a value to pick from. `find` would have taken whichever
-    // came first in map order, which is a silent choice between two disagreeing signals.
-    if matches.next().is_some() {
-        return Some(EnvelopeObservation::Malformed);
+        .filter(|(k, _)| k.to_ascii_lowercase() == PROTOCOL_VERSION_HEADER)
+    {
+        any = true;
+        match value.as_str() {
+            // Two spellings carrying the same value agree, and agreement is not a choice. Only a
+            // disagreement is: `find` would have silently taken whichever came first in map order.
+            Some(v) if !v.is_empty() && found.is_none_or(|prev| prev == v) => found = Some(v),
+            _ => return Some(EnvelopeObservation::Malformed),
+        }
     }
-    Some(match value.as_str() {
-        Some(v) if !v.is_empty() => EnvelopeObservation::Present(v.to_string()),
-        _ => EnvelopeObservation::Malformed,
+    if !any {
+        return None;
+    }
+    Some(match found {
+        Some(v) => EnvelopeObservation::Present(v.to_string()),
+        None => EnvelopeObservation::Malformed,
     })
 }
 
--- a/crates/assay-core/src/mcp/era.rs
+++ b/crates/assay-core/src/mcp/era.rs
@@ -136,10 +136,10 @@ impl<'de> serde::Deserialize<'de> for UniqueValue {
 /// request, took its era, and a missing `resultType` under a borrowed legacy era could read
 /// `Terminal`.
 ///
-/// The number is kept as its canonical text rather than parsed, because JSON-RPC does not constrain
-/// an id to an integer and two spellings of one number are the same id. The variant, not the text,
-/// carries the type distinction. Crate-private: the public field is unchanged, byte and API
-/// compatible.
+/// The number is kept as its canonical text rather than parsed. MCP restricts `RequestId` to a
+/// string or an integer, so an accepted number is already an integer by the time it becomes a key,
+/// and the text is its canonical rendering. The variant, not the text, carries the type distinction.
+/// Crate-private: the public field is unchanged, byte and API compatible.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub(crate) enum CorrelationId {
     Str(String),
@@ -148,18 +148,20 @@ pub(crate) enum CorrelationId {
 
 /// Read the correlation key from a message's raw JSON.
 ///
-/// `None` for an absent or unusable id. `normalize_jsonrpc_id` already refuses booleans, arrays and
-/// objects at the parser boundary, so only the two valid shapes reach this.
+/// `None` for an absent or unusable id. [`accept_id`] owns the vocabulary; the parser's shape
+/// diagnostics for booleans, arrays and objects are more specific messages for a subset of the same
+/// refusal.
 pub(crate) fn correlation_id(v: &serde_json::Value) -> Option<CorrelationId> {
     v.get("id").and_then(accept_id)
 }
 
 /// The one place that decides whether an id is usable, and what key it becomes.
 ///
-/// MCP restricts `RequestId` to a string or an integer. A fractional number is not an id, and
-/// neither is one spelled with an exponent: `1.0` and `1e0` both parse as floats, so neither is an
-/// integer even though the second equals one. Accepting them would give two spellings of one call
-/// two different keys, or one key to two different calls, depending on which side rendered.
+/// MCP restricts `RequestId` to a string or an integer, and this function is the only place that
+/// decides it. A fractional number is not an id, and neither is one spelled with an exponent: `1.0`
+/// and `1e0` both parse as floats, so neither is an integer even though the second equals one.
+/// Accepting them would give two spellings of one call two different keys, or one key to two
+/// different calls, depending on which side rendered.
 ///
 /// `CorrelationId::Num` therefore only ever comes from an accepted integer.
 pub(crate) fn accept_id(v: &serde_json::Value) -> Option<CorrelationId> {
@@ -172,13 +174,15 @@ pub(crate) fn accept_id(v: &serde_json::Value) -> Option<CorrelationId> {
     }
 }
 
-/// The text an accepted id renders to for the public `McpEvent::jsonrpc_id` field.
-pub(crate) fn accepted_id_text(id: &CorrelationId) -> String {
-    match id {
-        CorrelationId::Str(s) => s.clone(),
-        CorrelationId::Num(n) => n.clone(),
-    }
-}
+/// Methods this build gives request semantics to.
+///
+/// `tools/call` is `CallToolRequest` and `tools/list` is `ListToolsRequest`, and a request MUST carry
+/// an id. Without this, a request-only method could shed the id requirement simply by omitting the
+/// member, and with it the required 2026 `RequestParams._meta`, because notification metadata is
+/// optional. The list is deliberately the methods the parser already gives request payloads to
+/// rather than every method in the specification: an extension method with an unknown name stays a
+/// notification, so this refuses what is known to be a request instead of guessing about the rest.
+pub(crate) const REQUEST_ONLY_METHODS: &[&str] = &["tools/call", "tools/list"];
 
 /// Which JSON-RPC message shape this is, carrying the method so that nothing has to read it a
 /// second time.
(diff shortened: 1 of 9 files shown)


crates/assay-core/src/mcp/parser.rs
```diff
@@ -1,10 +1,150 @@
...
+///
+/// The envelope is folded per entry, from the transcript-level slots each entry starts with plus
+/// its own, so a later deviant entry cannot reach back and contaminate an earlier correct one. The
+/// request and result signals are per message and are read from the payload's retained raw JSON,
+/// before any projection can lose them.
+pub(crate) fn parse_mcp_transcript_detailed(
+    text: &str,
+    format: McpInputFormat,
+) -> Result<Vec<ParsedMcpEvent>> {
+    let events = parse_events(text, format)?;
```
Rul1an: **[P1] Avoid a second unbounded materialization on the public import path.** `parse_events` already deserializes the framed transcript and clones each raw JSON-RPC message into `McpEvent`; `observe_envelopes` then deserializes the entire input again at line 124, while line 36 clones each retained raw payload once more. `parse_mcp_transcript` now pays this cost for every existing caller even though it immediately discards the sidecar. These transcripts are evidence inputs and may be hostile, so this is a new OOM/amplification path relative to the old parser. Produce `ParsedMcpEvent` in the same `TransportTranscript` pass (its transcript/entry header fields already exist), compute observations while borrowing the retained raw value, and let the public function project events without reparsing or cloning the payload.

crates/assay-core/src/mcp/era.rs
```diff
@@ -0,0 +1,449 @@
...
+/// reason `Malformed` exists, so this reads the slot itself.
+pub(crate) fn observe_header(headers: Option<&serde_json::Value>) -> Option<EnvelopeObservation> {
+    let map = headers?.as_object()?;
+    let mut matches = map
+        .iter()
+        .filter(|(k, _)| k.to_ascii_lowercase() == PROTOCOL_VERSION_HEADER);
+    let (_, value) = matches.next()?;
+    // Two spellings of one header is not a value to pick from. `find` would have taken whichever
+    // came first in map order, which is a silent choice between two disagreeing signals.
+    if matches.next().is_some() {
```
Rul1an: **[P2] Equal duplicate header values are misclassified as malformed.** The recorded contract permits repeated observations when all values agree, but this returns `Malformed` whenever a second case-insensitive spelling exists, even if both values are identical. I added a temporary review test with `MCP-Protocol-Version` and `mcp-protocol-version` both set to `2026-07-28`; it failed with `left: Malformed, right: Present("2026-07-28")`. Fold all matching values and reject only non-string/empty or mutually different values; keep the existing disagreeing-duplicate test as the negative arm.

crates/assay-core/src/mcp/era.rs
```diff
@@ -0,0 +1,449 @@
...
+    RESULT_TYPE_SINCE,
+];
+
+/// Whether a string is a usable protocol version, which is an ISO calendar date.
+///
+/// Ordering against [`RESULT_TYPE_SINCE`] is lexicographic, and lexicographic ordering is date
+/// ordering only once the value is a date. Ten bytes and two dashes is not enough: `2026-99-99`
+/// would pass a shape-only check and be reported as a version this build merely does not support,
+/// which blames the reader for a record that is wrong. Day-of-month is bounded at 31 rather than
+/// Validated as a real calendar date, leap years included, so `2026-02-31` is malformed rather
```
Rul1an: **[P3] Repair the stale sentence.** `Day-of-month is bounded at 31 rather than Validated...` is a leftover from the partial validator and now contradicts the full Gregorian implementation below. Remove the old fragment so the comment states only the current contract.

Rul1an (commented):
Codex non-builder review on exact head `6b36c6ef`. I independently ran the MCP-focused lib tests (255/255) and the unchanged import/transport/correlation integration suites (23/23), and checked the load-bearing wire fields against the final 2026-07-28 schema. Two actionable findings remain; both should be fixed before this slice is merge-ready. Copilot's quota-limited record does not count as a review, and CodeRabbit is still pending.

crates/assay-core/src/mcp/parser.rs
```diff
@@ -94,8 +191,25 @@ fn parse_transport_transcript(
...
+        transcript_slots.push(o);
+    }
+    if let Some(o) = observe_header(transcript.headers.as_ref()) {
+        transcript_slots.push(o);
+    }
+
+    let mut envelopes = Vec::new();
     let mut out = Vec::new();
     for (idx, entry) in transcript.entries.into_iter().enumerate() {
+        let mut slots = transcript_slots.clone();
```
Rul1an, line 301: **[P1] Bound the transcript-level header before cloning it per entry.** `transcript_slots` may contain an arbitrary non-empty attacker-controlled string; this line clones that string for every entry, then `fold_envelope` and `resolve_era` clone it again into each retained sidecar. The public parser builds the full detailed vector before discarding the contexts, and both CLI import entrypoints still use unbounded `read_to_string`, so one large malformed header plus many small entries creates superlinear peak memory. A temporary 1 MiB invalid-header test confirmed every event retained `Present(the_whole_string)`. Classify non-version strings as `Malformed` before this loop (or otherwise bound/share the observation), and add a multi-entry regression test that does not print the attacker value on failure.

crates/assay-core/src/mcp/era.rs
```diff
@@ -0,0 +1,457 @@
...
+/// The transport header carrying the same value.
+pub(crate) const PROTOCOL_VERSION_HEADER: &str = "mcp-protocol-version";
+
+/// Read one header slot as an observation rather than as an `Option`.
+///
+/// The existing case-insensitive header helper answers `None` for a value that is present and not
+/// a string, which reports silence where a signal arrived and failed. That difference is the whole
+/// reason `Malformed` exists, so this reads the slot itself.
+pub(crate) fn observe_header(headers: Option<&serde_json::Value>) -> Option<EnvelopeObservation> {
+    let map = headers?.as_object()?;
```
Rul1an: **[P1] A present non-object `headers` container still fails open as silence.** `as_object()?` maps both direct `"headers": 7` and `transport_context.headers = 7` to `None`, so the framed request folds to `Absent`; with valid body metadata, that lets the body resolve the era despite malformed framing. A temporary parser-level test failed `left: Absent, right: Malformed` for both shapes. Preserve `None` only for an absent slot; return `Some(EnvelopeObservation::Malformed)` when the slot is present but not an object, and pin both direct and nested forms.

Rul1an (commented):
Codex non-builder review on exact head `bd167ce0`. I independently ran the MCP-focused lib tests (256/256) and the unchanged import/transport/correlation integration suites (23/23). The prior double-deserialization and equal-duplicate findings are fixed, but three actionable current-head findings remain: the new per-entry header-string amplification, a present non-object `headers` slot still collapsing to silence, and the existing broken date-validation doc fragment. Current-head CI is green; CodeRabbit was rate-limited on this head and Copilot quota-limited on the superseded head, so neither supplies the required bot review.

crates/assay-core/src/mcp/era.rs
```diff
@@ -0,0 +1,469 @@
...
+    "2025-11-25",
+    RESULT_TYPE_SINCE,
+];
+
+/// Whether a string is a usable protocol version, which is an ISO calendar date.
+///
+/// Ordering against [`RESULT_TYPE_SINCE`] is lexicographic, and lexicographic ordering is date
+/// ordering only once the value is a date. Ten bytes and two dashes is not enough: `2026-99-99`
+/// would pass a shape-only check and be reported as a version this build merely does not support,
+/// which blames the reader for a record that is wrong. Validated as a real calendar date, leap
```
Rul1an: **[P3] Finish the sentence repair.** The old `Day-of-month...` fragment is gone, but `Validated as a real calendar date, leap years included, so...` is still a sentence without a subject and does not connect grammatically to the preceding sentence. Use the already-proposed form, e.g. `The value is validated as a real calendar date, leap years included, so ...`. No behavior change is needed.

Rul1an (commented):
Codex non-builder review on exact head `01de8fd6`. The two P1 corrections are sound: malformed version strings are discarded before per-entry folding, and present non-object header containers now remain `Malformed`. I independently ran 259 MCP-focused lib tests, the unchanged 23 import/transport/correlation integration tests, fmt, diff-check, and clippy with `-D warnings`; all passed. No behavioral finding remains, but the attempted P3 prose repair is still grammatically incomplete, so this is not yet an approval. CI is still running and no same-head bot review exists.

Rul1an (commented):
Codex non-builder review disposition on exact head `493564f6`: **APPROVED, no findings.** The final delta is doc-only and correctly completes the calendar-date sentence. The behavioral code is byte-identical to the independently reviewed `01de8fd6` head, where 259 MCP-focused lib tests, 23 unchanged integration tests, fmt, diff-check, and clippy with `-D warnings` passed. All Codex findings and threads are closed. GitHub does not permit the repository owner account to submit an `APPROVE` state on its own PR, so this machine-readable COMMENT review carries the explicit agent disposition; it remains one quorum arm only.

Rul1an (commented):
Codex current-head review correction on exact head `493564f6`: **CHANGES REQUESTED**. This supersedes my earlier no-findings disposition. A second independent non-builder pass found four issues; I independently reproduced the first three with temporary behavioral tests (removed afterward), each failing on its own assertion. CI is green but does not exercise these compositions.

1. **[P1] `era.rs:456-458` — a present non-object `result` collapses to `Missing`.** `Value::get` on `null`, scalar, or array returns `None`; under a known pre-2026 era that then reaches the absent-means-complete rule and returns `Terminal`. Require `result` to be an object before reading `resultType`, and pin null/scalar/array plus object-without-field separately.

2. **[P2] `era.rs:273-299` — malformed result does not actually precede era classification.** With `Unknown(NoSignal) + Malformed`, `conclude` returns `Incomplete(EraUnknown)` before inspecting the malformed observation, contradicting lines 295-296. Give `Malformed` precedence before the era match, as the request path already does; add composed Unknown/Conflicting tests.

3. **[P2] `era.rs:459-463` — empty string is misclassified as malformed.** The pinned 2026-07-28 schema defines the open union as `"complete" | "input_required" | string` without a non-empty constraint. Therefore `""` is an unrecognized string and should be `Unrecognized`/`Incomplete`, not `Malformed`/`Invalid`.

4. **[P2] `era.rs:462`, `parser.rs:13-17` — unrecognized tokens introduce unbounded duplicate retention.** `to_string()` retains an attacker-sized token in every sidecar, while the public parser first materializes the full detailed vector and then drops every sidecar. This is linear, not superlinear, but it is new avoidable hostile-input memory amplification. Bound the retained token or represent an unrecognized token without raw bytes; keep any future diagnostic value-free.

Independent artifact verification: full `assay-core` 680 passed / 5 ignored; focused era, wiring, and integration suites green; fmt, clippy `-D warnings`, and diff-check clean. Those results confirm baseline stability, not the missing cases above.

Rul1an (commented):
Codex non-builder review on exact head `92eb41ce`: **CHANGES REQUESTED**. All four prior behavioral findings are correctly fixed. I also ran a temporary end-to-end review test (removed afterward) proving `legacy header + result:null` now concludes `Invalid(MalformedResultType)`. Three current-head items remain.

1. **[P2] Preserve the load-bearing composition in the suite.** `a_non_object_result_is_malformed_not_missing` stops after observation, while `composite_a_malformed_result_type_before_2026_is_not_complete` starts from a manually constructed `Malformed`. That is the same split-layer shape that previously let unreachable test states pass. Add one parser-to-conclusion test using the parsed event's own `context.era` and `result_observation`, asserting `result:null` under a legacy header is `Invalid`, not `Terminal`. The exact temporary test passes on this head.

2. **[P2] Do not retain an attacker-controlled prefix as if it were token identity.** The 64-byte bound fixes unbounded amplification, but `Unrecognized(String)` and `UnrecognizedResultType(String)` still carry attacker text into the exact reason type slice 2 will consume, while two distinct long tokens with the same prefix compare equal. The conclusion needs only the class `Unrecognized`. Prefer a unit variant and a value-free reason; if identity is genuinely required, use an explicitly non-displayable digest plus truncation metadata rather than a lossy raw prefix. This closes the public-string risk instead of postponing it to the first caller.

3. **[P3] Reconcile the comments with the new behavior.** `ResultObservation::Malformed` still says an empty string is malformed (`era.rs:81`), although this head correctly classifies it as unrecognized. At `era.rs:361`, `The _meta key...` now documents `MAX_TOKEN_BYTES`, leaving `PROTOCOL_VERSION_META_KEY` without its intended comment.

Independent verification: focused era 35/35; parser wiring 28/28; unchanged import/transport/correlation integration tests 23/23; full `cargo test -p assay-core`, clippy `--all-targets -D warnings`, fmt, and diff-check all clean. GitHub CI and `lane-check/proof` are green on this head. No same-head formal bot review exists; the green CodeRabbit status does not update its review object from `6b36c6ef`.

Rul1an (commented):
Codex + independent route-2 review on exact head `3a7d6430`: **CHANGES REQUESTED**. The three findings from `92eb41ce` are correctly closed: the full legacy parser→conclusion regression is permanent, `Unrecognized` and its reason are structurally value-free, and the comments now match the code. A second non-building review then found two reachable parser asymmetries; I independently reproduced both with temporary red tests (removed afterward).

1. **[P2] `parser.rs:37-40` — one hybrid message receives both request and result semantics.** `parse_jsonrpc_message` classifies a valid string `method` as a request, and `request_metadata` follows that discriminant, but `result_observation` reads every present `result` unconditionally. The existing hybrid fixture therefore produces `request_metadata=Present(...)` and `result_observation=Some(Missing)`. Slice 2 could draw a result conclusion about an event the parser classified as a request. Use one shared message-role/payload discriminant for both axes and pin that a request never carries a result observation.

2. **[P2] `parser.rs:71-74,195-207` — a present non-object `transport_context` collapses to absence.** `header_map` uses `Value::get`; for scalar, array, or null it returns `None`, indistinguishable from no container. At transcript level this becomes `Absent`; at entry level it can silently inherit a valid transcript default. Distinguish an absent container from a present unreadable one and fold the latter as `Malformed`; cover both transcript and entry levels.

Independent red evidence: the hybrid test returned `Some(Missing)` instead of `None`; the scalar transcript context returned `Absent` instead of `Malformed`. Baseline verification on the unchanged head remains clean: era 34/34, wiring 29/29, unchanged integrations 23/23, full assay-core 685 passed / 5 ignored, clippy `-D warnings`, fmt and diff-check clean. CI is still completing. No formal bot review exists on this SHA.

Rul1an (commented):
Review op exact head `88501fe3e44d45edea35e40c903989287db24356`: **CHANGES REQUESTED**.

## Findings

1. **[P1] Een aanwezige niet-object `params`-container wordt nog als afwezige metadata gelezen** ([`era.rs:443`](https://github.com/Rul1an/assay/blob/88501fe3e44d45edea35e40c903989287db24356/crates/assay-core/src/mcp/era.rs#L443-L460)). `raw.get("params").and_then(|p| p.get("_meta"))` geeft voor scalar, array en `null` dezelfde `None` als voor een ontbrekende container. De request-as wordt daardoor `RequestMetadata::Absent`; onder een legacy-era kan `conclude_request` vervolgens `Valid` retourneren. Dit is dezelfde fail-open `Value::get`-klasse die al voor `_meta`, `result` en `transport_context` is gerepareerd. Maak alleen een ontbrekende `params` absent; een aanwezige niet-object container moet `Malformed` zijn. Pin de echte parser -> resolve -> request-assessment-keten voor scalar, array en null, met minstens een legacy-arm.

2. **[P2] De ASCII-digit guard is correct maar niet door een test vastgezet** ([`era.rs:188`](https://github.com/Rul1an/assay/blob/88501fe3e44d45edea35e40c903989287db24356/crates/assay-core/src/mcp/era.rs#L188-L212)). De bestaande malformed strings falen eerder op lengte of streepposities. Daardoor blijft de suite groen als de digit-check wordt verwijderd; dan worden tien-byte waarden als `abcd-01-01` en `+123-01-01` als `UnsupportedVersion(attacker_value)` bewaard. Voeg beide regressies rechtstreeks toe en mutation-check specifiek deze guard.

3. **[P2] Een van de vier `present_slot`-toepassingen is niet afzonderlijk bewezen** ([`parser.rs:648`](https://github.com/Rul1an/assay/blob/88501fe3e44d45edea35e40c903989287db24356/crates/assay-core/src/mcp/parser.rs#L648-L657)). De expliciete-nulltest op `headers` raakt alleen het transcriptniveau. Voeg een geldige transcriptdefault plus `entry.headers: null` toe en eis `Malformed`; dan faalt de test wanneer juist de deserializer op `TransportTranscriptEntry.headers` wordt verwijderd.

4. **[P3] CodeRabbits same-head comment is geldig.** De testcomment boven `a_non_object_transport_context_is_malformed_not_absent` beschrijft het oude `Absent`-gedrag alsof het de huidige uitkomst is. Formuleer expliciet "naive read -> Absent; fixed read -> Malformed" en los de open thread op.

## Dispositions

- **Behoud** de `unwrap_or(EnvelopeObservation::Malformed)` bij event/envelope-attributie. Hij is onder de huidige constructie een defense-in-depth invariantguard; verwijderen maakt toekomstige indexdrift permissiever.
- De tweede `ResultObservation::Malformed`-arm na de early return mag naar `unreachable!("handled above")`, zoals de request-arm, maar dit is opruiming en geen blocker.
- `fold_envelope(..., framed)` wordt in productie alleen met `true` aangeroepen. Vereenvoudigen mag, maar hoort niet bij de beveiligingsfix en levert geen nieuw bewijs.
- De `response.headers`-asymmetrie is nog geen aangetoonde schema-/fixtureklasse. Niet uitbreiden zonder een werkelijk ondersteunde transportvorm.
- In de huidige publieke PR-body of comments vond ik geen tekst die de drie `assay-cli`-tests "pre-existing failures" noemt. Neem in de finale verificatienotitie wel de juiste oorzaak op: zonder vooraf gebouwde sibling binary falen ze met `missing binary`; na `cargo build -p assay-mcp-server` slagen ze op base en head. Corrigeer niets dat niet publiek bestaat.

CI is volledig groen op deze SHA en CodeRabbit heeft formeel op dezelfde SHA gereviewd, maar dit reviewresultaat blokkeert merge totdat bovenstaande drie gedrags-/bewijsitems en de open thread op een nieuwe finale head zijn gesloten.

crates/assay-core/src/mcp/era.rs
```diff
@@ -0,0 +1,449 @@
...
+///
+/// The existing case-insensitive header helper answers `None` for a value that is present and not
+/// a string, which reports silence where a signal arrived and failed. That difference is the whole
+/// reason `Malformed` exists, so this reads the slot itself.
+pub(crate) fn observe_header(headers: Option<&serde_json::Value>) -> Option<EnvelopeObservation> {
+    let map = headers?.as_object()?;
+    let mut matches = map
+        .iter()
+        .filter(|(k, _)| k.to_ascii_lowercase() == PROTOCOL_VERSION_HEADER);
+    let (_, value) = matches.next()?;
```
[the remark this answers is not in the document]
Rul1an: Fixed on `a3bf9c1a` and verified live at that ref.

`observe_header` now returns `Malformed` when the `headers` node is present and not an object, instead of letting `as_object()` answer `None` and dropping the slot. Pinned by the non-object and explicit-null cases at all four slots.

crates/assay-core/src/mcp/parser/era_wiring_tests.rs
```diff
@@ -0,0 +1,675 @@
...
+        let input = framed(Some(bad.clone()), None, req(None));
+        assert_eq!(
+            detailed(&input, McpInputFormat::StreamableHttp)[0]
+                .context
+                .envelope,
+            EnvelopeObservation::Malformed,
+            "{bad}"
+        );
+    }
+}
```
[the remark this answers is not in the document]
Rul1an: Fixed on `a3bf9c1a` and verified live at that ref.

The comment described the naive path as though it were the current one. It now reads: a present non-object `transport_context` was naively read as `Absent` because `Value::get` returned `None`, and the fixed path reports `Malformed`, preserving the evidence that framing was present but unreadable.

Rul1an (commented):
Review op exact head `a3bf9c1a945c5224d590ce5654594d652ec778d4`: **CHANGES REQUESTED**.

De vier bevindingen van de vorige ronde zijn in de code gesloten. De `params`-guard, digit-tests, `entry.headers: null`-test en CodeRabbit-comment zijn aanwezig; de fail-closed indexfallback en afgesproken scopegrens zijn behouden. Twee load-bearing punten blijven over.

## Findings

1. **[P1] `method` wordt nog gebruikt als request/notification-discriminant, maar beide berichtsoorten hebben een `method`.** [`parser.rs:44-48`](https://github.com/Rul1an/assay/blob/a3bf9c1a945c5224d590ce5654594d652ec778d4/crates/assay-core/src/mcp/parser.rs#L44-L48) geeft ieder bericht met een string-`method` `request_metadata: Some(...)`. Daardoor krijgt een MCP-notification dezelfde request-as en later dezelfde `conclude_request`-regels als een request.

De finale MCP 2026-07-28-schema's maken dit onderscheid dragend: `RequestParams._meta` en de protocolversie zijn verplicht, terwijl `NotificationParams._meta` optioneel is; `JSONRPCRequest` heeft bovendien een request-id en een notification niet. Zie het [gepinde schema](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/5f5440bb26a62e2cf3440b92da5a667efa03b267/schema/2026-07-28/schema.ts#L175-L205). De bestaande transportfixture bevat daadwerkelijk `notifications/progress` zonder id en zonder requestmetadata.

Onder een 2026-envelope wordt zo'n geldige notification nu `RequestMetadata::Absent`; een latere `conclude_request` maakt daar `Invalid(MissingRequestMetadata)` van. Splits de interne rol minimaal in request / notification / response. Alleen een request mag `request_metadata` en request assessment krijgen; een notification krijgt geen result-observatie en geen requestmetadata-assessment. Houd de publieke `McpEvent`-vorm ongewijzigd en pin een echte `notifications/progress`-vorm onder een 2026-header.

2. **[P2] De twee nieuwe "parser through assessment"-tests voeren de parseruitkomst niet aan de assessmentfunctie.** In [`era_wiring_tests.rs:701-728`](https://github.com/Rul1an/assay/blob/a3bf9c1a945c5224d590ce5654594d652ec778d4/crates/assay-core/src/mcp/parser/era_wiring_tests.rs#L701-L728) roepen beide tests `conclude_request(..., &RequestMetadata::Malformed)` aan met een nieuw, handmatig geconstrueerd argument. Gebruik in plaats daarvan `event.context.request_metadata.as_ref().expect(...)` en assert die observatie expliciet.

Dit is niet alleen stijl: bij verwijdering van `params.is_object()` blijft de 2026-compositietest nu groen, omdat hij alsnog de hardcoded `Malformed` voert. Daarom meldt de mutation-run twee rode tests terwijl drie tests de guard zouden moeten dragen. De huidige claim "parser through resolution to assessment under both eras" is voor de 2026-arm dus nog niet bewezen.

## Stand

- Correction delta: drie bestanden, geen public-API- of integratietestchurn.
- De huidige CI draait nog; clippy, rustdoc, semver en de al voltooide gates zijn groen.
- CodeRabbit beoordeelt deze nieuwe SHA nog.
- De eerdere CodeRabbit-threads zijn resolved.
- De twee optionele cleanups blijven terecht buiten deze correction head.

GitHub laat de PR-eigenaar geen formele `REQUEST_CHANGES`-review indienen; dit `COMMENTED` reviewobject en het ledger-record dragen daarom expliciet de changes-requested dispositie.

Rul1an (commented):
Review op exact head `2b01c359d0617c3eea1d6b6601acff55227e1f67`: **CHANGES REQUESTED**.

De notification/request-scheiding is nodig en de echte `notifications/progress`-controle is correct. De 2026 params-compositietest leest nu ook werkelijk de parserobservatie. Twee punten blijven load-bearing.

## Findings

1. **[P1] Een aanwezige `"id": null` wordt ten onrechte als een notification behandeld.** In `classify_message` maakt alleen een niet-null id het bericht een request; de nieuwe test pint daarom `id: null` als notification. Dat verwart een afwezige member met een aanwezige maar onbruikbare waarde.

JSON-RPC definieert een notification als een Request object **zonder** `id`; een opgenomen null-id is een request-id-vorm, zij het discouraged. MCP 2026 is nog strenger: `RequestId = string | number`, `JSONRPCRequest.id` is verplicht, en `JSONRPCNotification` heeft geen id. Zie de [JSON-RPC notification-regel](https://www.jsonrpc.org/specification#notification) en het [gepinde MCP-schema](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/5f5440bb26a62e2cf3440b92da5a667efa03b267/schema/2026-07-28/schema.ts#L243-L267).

De huidige route:
```
method + id:null + 2026 envelope + geen request _meta
-> Notification
-> request_metadata = None
-> geen request assessment
```
laat daardoor precies de 2026 request-eis verdwijnen. Discrimineer op **member presence**, niet op een genormaliseerde bruikbare id: afwezige `id` is notification; aanwezige `id` is request-shaped. Voor MCP mag null daarnaast als malformed/invalid worden vastgelegd, maar het mag niet als notification schoonvallen. Vervang de huidige positieve null-notificationtest door een parser-tot-assessment regressietest.

2. **[P2] De legacy params-compositietest voert de parserobservatie nog niet aan de assessmentlaag.** `composite_a_deviant_params_under_a_legacy_era_is_invalid` roept nog `conclude_request(&event.context.era, &RequestMetadata::Malformed)` aan. De guard-mutatie maakt deze test rood via de eerdere era-assertie, maar de assessmentassertie blijft op een zelfgebouwde waarde draaien. Daardoor bewijst de test niet wat zijn “whole chain”-naam en PR-claim zeggen.

Gebruik, net als in de gecorrigeerde 2026-zustertest, `event.context.request_metadata.as_ref().expect(...)`, assert `Malformed`, en voer precies die referentie aan `conclude_request`.

## Stand

- Exact-head source review uitgevoerd; de publieke `McpEvent`-vorm blijft ongewijzigd.
- De geldige notification-route en de 2026 parser-through-assessment correctie zijn bevestigd.
- CI loopt nog op deze SHA; er waren bij de controle geen failures.
- De eerdere review op `a3bf9c1a` is door deze push vervallen.
- GitHub laat de PR-eigenaar geen formele `REQUEST_CHANGES`-review indienen; dit `COMMENTED` reviewobject draagt daarom expliciet de changes-requested dispositie.

crates/assay-core/src/mcp/era.rs
```diff
@@ -0,0 +1,530 @@
...
+
+/// Classify by the two fields JSON-RPC uses. A notification is a request object *without* an `id`
+/// member, so absence is the discriminant and nothing about the value is.
+///
+/// An explicit `"id": null` is therefore a request, not a notification. `RequestId` is `string |
+/// number`, so a null id is an invalid request id rather than an absent one, and treating it as a
+/// notification drops the required 2026 request metadata for any message that writes one. That is a
+/// fail-open reachable by a single token. The invalid id is a JSON-RPC validity question this slice
+/// does not own; what it owns is that the metadata requirement still applies.
+pub(crate) fn classify_message(v: &serde_json::Value) -> MessageKind {
```
Rul1an: **[P1] A malformed `method` is silently reclassified as a response.** Both this function and `request_method` use `get("method").and_then(Value::as_str)`, so a present number/object/array/null becomes indistinguishable from an absent method. With an `id` and no `result`, that yields neither request metadata nor a result observation; under a 2026 envelope the request metadata MUST is therefore lost. Use one shared typed classifier for parser and sidecar. Reject the malformed JSON-RPC shape value-free at parse time, or retain an explicit malformed message-shape observation for the conclusion layer, and pin the full parser-to-context path.

Rul1an (commented):
Exact-head review of `ee0b6498`: **changes requested**.

The remaining audit found one load-bearing P1. A present non-string `method` is collapsed to absence by both `classify_message` and `request_method`. For `{"method":7,"id":"call-1"}`, the parser builds a response-shaped `Other`, the sidecar records neither request metadata nor a result observation, and a 2026 request requirement disappears. The focused era and parser-wiring suites are green because this shape is not covered.

Fix this once at the shared discriminant, not independently in only one caller. Either reject a present non-string method as malformed JSON-RPC at parse time with a value-free error (consistent with the existing invalid-id rejection), or carry an explicit malformed message-shape observation to the conclusion layer. In both forms, `parse_jsonrpc_message` and the sidecar must consume the same typed classification result. Add a parser-level regression for non-string method plus id under a 2026 envelope, with valid request, notification, and response controls.

I completed the remaining targeted audit before filing this: the five repaired container guards are ordered correctly, the current review threads are resolved, and `cargo test -p assay-core --lib mcp::era` (35/35) plus `cargo test -p assay-core --lib mcp::parser::era_wiring_tests` (41/41) pass on this exact head. No second actionable finding remains from this round.

GitHub does not permit the repository owner to submit a formal REQUEST_CHANGES review on their own PR, so this review is `COMMENTED`; the technical disposition remains changes requested.

Rul1an (commented):
Codex non-builder review on exact head `3a8c66332f76d4e94f29f707a4fb417863406a55`: **APPROVED, no findings.**

I re-reviewed the final source and the full delta from base. The change from the previously reviewed behavioral head `e86a9784` is documentation-only and closes the three stale/duplicated comment defects without changing code or tests. The earlier malformed-`method` P1 remains fixed through the shared typed classifier and its parser-to-context regression.

Independent exact-head verification in detached worktree `/private/tmp/assay-pr1929-review-e86`:
- `cargo test -p assay-core`: 701 passed, 0 failed, 5 ignored
- `cargo clippy -p assay-core --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- `git diff --check 3e657ed0...3a8c6633`: clean
- public `McpEvent` and `types.rs` unchanged; only a crate-private module is registered

Disposition carried into slice 2: an id-less string-method `tools/call` still projects to the compatibility payload variant `McpPayload::ToolCallRequest`, while crate-private `MessageKind` classifies it as a notification. Slice 2 must use `MessageKind` plus era context as request authority; this approval does not claim that the public compatibility projection has been corrected.

GitHub does not permit the repository owner to submit an `APPROVE` event on its own PR, so this COMMENT review carries the explicit non-builder approval disposition.

Rul1an:
@coderabbitai review

Rul1an (commented):
Codex non-builder review on exact head `9690772f17774450d1b6e87cccafad4f9c248c42`: **APPROVED, no findings.**

The only delta from the previously reviewed `3a8c6633` head is the test-only correction requested by CodeRabbit. It removes the value-search assertion that could collide with a source line numbered 7, separates the located prefix from the diagnostic, verifies that the location component is numeric, and pins the complete value-free diagnostic text without hard-coding source line 1.

Independent exact-head verification in detached worktree `/private/tmp/assay-pr1929-review-e86`:
- focused refusal test: passed
- full `cargo test -p assay-core`: passed with no failures
- `cargo clippy -p assay-core --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- `git diff --check 3e657ed0...9690772f`: clean
- final delta: one test file, no production or public-surface change

The first clippy attempt hit local `ENOSPC` while compiling into a second disposable target. I removed only the two Codex-owned disposable review targets, reran on the warmed exact-head target, and clippy completed successfully. This was local capacity, not a code failure.

The slice-2 disposition remains unchanged: use crate-private `MessageKind` plus era context as request authority, not the compatibility payload variant. GitHub does not allow the owner account to submit an `APPROVE` state, so this COMMENT carries the explicit non-builder approval disposition and counts as one quorum arm only.

Rul1an:
@coderabbitai review

Rul1an:
## Exact-head route-2 review: CHANGES REQUESTED

This supersedes my earlier approval disposition on
`9690772f17774450d1b6e87cccafad4f9c248c42`.

After the 30-minute bot-unavailability window, a second independent non-building review found three
issues. I reproduced the two P1s myself in the detached exact-head worktree with literal,
parser-to-conclusion tests. Both fail for the claimed reason.

### P1: request-era conflict is lost on the correlated response

`parse_mcp_transcript_detailed` gives responses no request metadata and resolves their era from the
response entry's envelope alone (`parser.rs:43-71`). In a two-entry call with:

- transport header `2025-06-18`;
- request metadata `2026-07-28`;
- correlated response with missing `resultType`;

the request is `EraResolution::Conflicting`, but the response becomes
`Known("2025-06-18")` and `conclude` returns `Terminal`. The existing conflict conclusion test feeds
a manually constructed `EraResolution`, so it does not cover the reachable parser chain.

Required correction: carry the request-era resolution into result assessment for the matching
JSON-RPC call. A conflict must remain invalid even when no response exists or when the response has
no body-era field. Correlation/deduplication belongs at the call context, not in a hand-built unit
state.

### P1: duplicate security-significant JSON members collapse last-value-wins

JSONL and transport input are materialized through `serde_json::Value` at `parser.rs:166` and
`parser.rs:220`. Exact duplicate object members disappear before the era observers run.

Reproducer:

```json
{"jsonrpc":"2.0","id":"call-1","result":{"content":[],"resultType":"input_required","resultType":"complete"}}
```

The parser accepts it and retains only `"complete"`. It should refuse the ambiguity before an era
or terminal conclusion is formed. Add literal raw-JSON regressions for duplicate `resultType`,
request protocol metadata, and the exact protocol header. Do not solve this with a second unbounded
full deserialization or a post-collapse `Value` check; the duplicate-aware boundary must see source
members once.

### P2: case-insensitive header matching allocates a full key copy

`era.rs:442-445` calls `to_ascii_lowercase()` on every untrusted header name. A large unrelated key
creates another same-sized transient allocation after the JSON input is resident. Use
`eq_ignore_ascii_case(PROTOCOL_VERSION_HEADER)` and add a key-focused negative control.

### Reproduction evidence

Temporary tests were added only in the Codex-owned detached review worktree, run, and then removed;
the worktree is clean again.

```text
cargo test -p assay-core review_ --lib -- --nocapture

review_conflicting_request_context_must_block_correlated_terminal_response ... FAILED
  correlated response era: Known("2025-06-18")

review_duplicate_result_type_is_rejected_before_last_wins ... FAILED
  accepted ResultObservation::Complete
```

The independent reviewer also ran the focused era suite, wiring suite, MCP integration tests, full
`assay-core`, clippy with denied warnings, fmt, and diff-check successfully. Those green checks do
not cover these reachable inputs.

Do not merge this head. The builder remains Claude Code; Codex and the independent reviewer remain
read-only.

Rul1an:
## Exact-head review on `2b399b0d`: CHANGES REQUESTED

The three reported fixes are directionally correct, but the exact-head review found two remaining P1 bypasses and one P2 early-refusal issue. I reproduced all three relevant behaviors in the clean detached worktree `/private/tmp/assay-pr1929-review-e86`; the temporary tests were then removed and the worktree is clean again.

### P1: duplicate members still pass through both SSE data forms

`TransportSseEnvelope.data` is still a plain `serde_json::Value` (`parser.rs:714`), and string SSE data is reparsed as a plain `Value` at `parser.rs:616`. Both paths therefore collapse duplicate `resultType` members last-value-wins before `UniqueValue` sees them.

Literal `http-sse` regressions for structured `data` and JSON-string `data` both failed: parsing succeeded and retained only `input_required`. The PR/ledger claim that duplicate refusal covers all three entrypoints is therefore false for a security-significant path. Make both SSE forms use the same duplicate-aware boundary, without a second unbounded full parse.

### P1: global last-wins call correlation captures a future request

`correlate_calls` first collects every request into one `HashMap` and only then rewrites every response (`parser.rs:99-120`). Reusing an id after its response is legal transcript sequencing, but the later request overwrites the earlier era and is assigned backwards to the first response.

Reproduction: header 2025; request `reused` with body 2026; response `reused` missing `resultType`; then a new request `reused` with body 2025. The first response becomes `Known(2025)` instead of inheriting the preceding call conflict. That restores the exact false-terminal path this correction is meant to close. Correlate in source order with pending request state; consume the matching request at its response, keep orphans self-resolved, and define/refuse duplicate *outstanding* ids rather than relying on the existing tools/call-only whole-transcript check.

### P2: duplicate detection materializes the duplicate value before refusing

`visit_map` calls `next_value()` and inserts/replaces before checking `is_some()` (`era.rs:110-115`). Once the duplicate key is known, a hostile nested duplicate value should not be fully allocated merely to reject it. Check membership immediately after `next_key` and error before deserializing the duplicate value. Add a focused visitor/control test if practical so this ordering cannot drift.

Required exact-head tests:
- structured SSE duplicate `resultType`;
- string SSE duplicate `resultType`;
- sequential id reuse where each response inherits only its preceding request;
- response before request remains orphaned and cannot capture the future request;
- duplicate outstanding request id fails closed;
- existing direct JSON-RPC/Inspector/transport duplicate cases remain green.

Do not merge this head. Any correction push invalidates the current reviews and CI evidence.

Rul1an:
## Exact-head review on `625b5146`: one P1 remains

The SSE duplicate paths, source-ordered success-response correlation, future-capture guard, outstanding-id refusal, and key-before-value duplicate check are now correct in the reviewed delta. A remaining response-classification bug still blocks merge.

### P1: JSON-RPC error responses neither inherit nor consume their request

`correlate_calls` uses `p.context.result_observation.is_some()` as the response discriminator. Error responses intentionally have no `result` and therefore no result observation, so they are skipped even though `classify_message` identifies them as responses.

I reproduced this end-to-end with literal transcript input:
1. request id `error-reuse`, header 2025/body 2026 (`Conflicting`);
2. JSON-RPC error response with the same id;
3. a later request reusing that id.

Expected: the error response inherits the conflicting call era and consumes the outstanding request, allowing later reuse. Actual: the error response remains self-resolved and the later request fails with `two outstanding JSON-RPC requests share an id at source line 3`. The focused test failed exactly there. Temporary review test removed; detached worktree clean.

Use the shared message classifier/raw payload shape as the request/response authority rather than presence of a result observation. Pin both properties:
- an error response inherits its preceding request era;
- an error response consumes the outstanding id so sequential reuse remains valid.

Also keep the current success-response, orphan, future-request, duplicate-outstanding, and SSE duplicate controls. Do not merge `625b5146`; a correction push invalidates current reviews.

Rul1an:
## Codex independent exact-head review — APPROVE

**Head reviewed:** `7480ac789bbdc9cca4b83b164a10ef7c9cb0631c`

No actionable findings.

I reviewed the correction from `625b5146` and the surrounding correlation state machine. The response discriminator now comes from the same raw-message classifier used by parsing, so both successful and JSON-RPC error responses inherit and consume the matching outstanding request. Source-order correlation remains intact; orphan and response-before-request cases keep their self-resolved era; future requests cannot be captured; duplicate outstanding request IDs still fail closed; notifications do not enter correlation.

Independent verification on this exact head:

- focused era/parser wiring: **56 passed**;
- full `cargo test -p assay-core`: green, including `mcp_import_smoke` 2/2, `mcp_transport_compat` 9/9, and `mcp_id_correlation` 12/12;
- `cargo clippy -p assay-core --all-targets -- -D warnings`: green;
- `cargo fmt --all -- --check`: green;
- `git diff --check origin/main...HEAD`: green;
- the three pre-existing MCP integration files are byte-unchanged;
- unresolved review threads: 0 at review time.

The first local clippy attempt stopped with `ENOSPC` in my disposable review target. I removed only Codex-owned disposable build caches and reran the same command successfully; it is not counted as a code failure or as proof.

This is a non-builder review. Claude remains the sole writer of the branch. Any later push invalidates this approval.

Rul1an:
@coderabbitai review

Rul1an:
## Codex exact-head review update — CHANGES REQUESTED

**Head:** `7480ac789bbdc9cca4b83b164a10ef7c9cb0631c`

I retract my approval above. A final adversarial parser-to-conclusion probe found one P1 response-shape gap with two independently reproduced manifestations.

### [P1] `Response` is assigned without enforcing exactly one of `result` or `error`

`classify_message` returns `MessageKind::Response` solely because `method` is absent (`era.rs:175-176`). Correlation then treats that label as authoritative (`parser.rs:133-138`), while payload construction separately prefers `result` over `error` (`parser.rs:415-433`).

1. A message carrying **both** `result` and `error` is accepted. With a 2026 header and `resultType: "complete"`, the detailed parser observes `Complete` and `conclude(...)` returns **`Terminal`**. A protocol-invalid response can therefore license completion.
2. A message carrying **neither** `result` nor `error` is also classified as a response and consumes the outstanding request ID. A later request reusing that ID is accepted, although no response shape ever closed the call.

The first red probe asserted that a both-members response must not conclude terminal; actual was `Terminal`. The second used `notifications/x` requests to avoid the older `tools/call` duplicate-ID gate and asserted that a neither-members shape must not free the ID; parsing unexpectedly succeeded. Both temporary probes were removed and the detached review worktree is clean.

Required correction: make the shared classifier enforce the JSON-RPC response invariant for no-method shapes: exactly one of `result` or `error`. Refuse both and neither at the existing value-free message-shape boundary, then add load-bearing parser-level tests. Keep valid success/error response correlation and all existing hybrid/request behavior explicit.

Any correction push invalidates this review and the bot run now in flight.

Rul1an:
## Independent exact-head review — CHANGES REQUESTED

**Head:** `503ab7d6ead62caf6b5b9cc7e3fc58e2b5b3d8e4`

The response-arity correction is sound, but the second non-building review found one remaining P1 that I independently reproduced end to end.

### [P1] Correlation collapses numeric and string JSON-RPC IDs

`correlate_calls` keys on `McpEvent.jsonrpc_id` (`parser.rs:107`), but the public normalization turns numeric `1` and string `"1"` into the same Rust string (`parser.rs:569-570`). They are distinct valid JSON-RPC IDs.

Probe:

1. request `id: 1`, entry era 2025;
2. response `id: "1"`, entry era 2026, result without `resultType`.

Expected: no correlation; response remains `Known(2026)` and concludes `Invalid(MissingResultType)`.

Actual: response era becomes **`Known(2025)`**. The wrong request is consumed, and missing `resultType` can consequently be read as **`Terminal`**.

This directly reopens the false-terminal failure Slice 0 is meant to close. It is not a broad JSON-RPC conformance request.

Required correction: use a crate-private, type-preserving correlation key derived from the raw payload ID (at least distinct string and number variants). Keep the public `McpEvent.jsonrpc_id` unchanged. Pin numeric↔string mismatch in both directions, plus positive controls for same numeric and same string IDs and duplicate-outstanding behavior per typed key.

The temporary reproducer was removed; the detached review worktree is clean. Required CI is green on this SHA, but CI does not cover this case and is not merge proof. Verdict: **BLOCK**.

Rul1an:
## Codex exact-head review — CHANGES REQUESTED

**Head:** `ba04d292239a51f60ff0ac7d95b4eb48aefb154f`

### [P1] The typed correlation key preserves JSON type but does not enforce the MCP `RequestId` domain

The correction separates JSON number `1` from JSON string `"1"`, which closes the reported collision. Its numeric premise is still false in both documentation and behavior:

```rust
/// The number is kept as its canonical text rather than parsed, because JSON-RPC does not constrain
/// an id to an integer and two spellings of one number are the same id.
...
Num(String)
```

`serde_json::Number::to_string()` is not that canonicalization: `1` renders as `1`, while `1.0` and `1e0` render as `1.0`, and `serde_json::Value(1) != Value(1.0)`.

More importantly, this is an MCP parser, and the final pinned MCP `2026-07-28` schema is stricter than generic JSON-RPC:

- canonical `schema.json`: `RequestId` is `string | integer`;
- official TypeScript SDK: `z.union([z.string(), z.number().int()])`;
- official Rust SDK: numeric `RequestId` is `i64`.

I added an uncommitted exact-head probe in the detached review worktree. All five refusal tests fail because the parser returns `Ok`:

```text
request id 1.0       -> accepted as jsonrpc_id "1.0"
response id 1.0      -> accepted as jsonrpc_id "1.0"
request id null      -> accepted as jsonrpc_id None
result response null -> accepted as jsonrpc_id None
success response no id -> accepted as jsonrpc_id None
```

This is load-bearing for the slice. A request and success response with an out-of-domain numeric id can still enter the era/conclusion pipeline; and the branch's own null-id test calls the request id invalid while only asserting that it is classified as a request, not that it is refused.

**Required correction:**

1. Enforce the MCP id domain at the message-shape boundary: a success response requires an id, and request/success-response ids are string or integer, never fractional or null.
2. Keep the refusal value-free.
3. Correlate allowed numeric ids through a representation whose equality matches the accepted integer domain; do not describe `to_string()` as numeric canonicalization.
4. Pin at least `1.0` and `1e0` as refusals, plus positive controls for negative and large representable integers.
5. Treat a JSON-RPC invalid-request **error** response with absent or `null` id deliberately: the current exact-head control confirms `id: null` remains ingestible with `jsonrpc_id: None`; preserve that uncorrelated diagnostic path, but never let it acquire request-era authority. Do not implement this as an unconditional global null rejection without testing that case.
6. Add a parser-to-conclusion regression demonstrating that an invalid id cannot license a terminal reading.

Primary sources:

- MCP final schema pin: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/5f5440bb26a62e2cf3440b92da5a667efa03b267/schema/2026-07-28/schema.json
- TypeScript SDK integer validator: https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/core/src/schemas.ts
- Rust SDK `NumberOrString::Number(i64)`: https://github.com/modelcontextprotocol/rust-sdk/blob/main/crates/rmcp/src/model.rs
- JSON-RPC base rule and null-error exception: https://www.jsonrpc.org/specification#request_object

No other exact-head finding is being claimed yet. CI is green but has no case for this domain boundary, and no review on `ba04d292` counts toward quorum.

Rul1an:
Independent reference proof for the exact-head blocker on `ba04d292`. No change was made to the writer branch and nothing from the review worktree was committed or pushed.

**RED on the current PR head**

A focused parser suite with the MCP RequestId boundary added produced 65 pass / 3 fail:

- request `id: 1.0` and lexical `1e0` were accepted;
- request `id: null` was accepted;
- successful responses with absent, null, or fractional ids were accepted.

The diagnostic control already behaved correctly: an invalid-request **error** response with absent or null id remained ingestible and uncorrelated.

**Minimal correction shape, independently exercised**

1. Classify the message before normalizing its id.
2. Notifications accept an absent member.
3. Requests and successful responses require a string or a `serde_json` integer (`is_i64() || is_u64()`).
4. Error responses may keep absent/null as uncorrelated diagnostic input, but a present malformed/fractional id is refused.
5. `CorrelationId::Num` is derived only from an accepted integer and retains canonical decimal text, covering signed and unsigned values without changing the public `McpEvent` field.
6. Refusals name the required shape or existing JSON type but never echo the attacker-controlled value.

The full-suite pass also requires updating `tests/mcp_id_correlation.rs`: its null-request and missing-success-response cases currently assert the superseded broad JSON-RPC behavior. Replace them with fail-hard MCP contract cases and add the null invalid-request error response as the uncorrelated control. The existing boolean/array/object diagnostics can stay byte-compatible.

**GREEN in the detached reference worktree**

- era wiring: 69 passed;
- public id-correlation contract: 13 passed;
- `cargo test -p assay-core`: exit 0;
- `cargo clippy -p assay-core --all-targets -- -D warnings`: exit 0;
- `cargo fmt --all -- --check`: clean;
- `git diff --check`: clean.

The load-bearing end-to-end regression uses a legacy request plus a successful response without `resultType`: the valid integer id reaches the legacy `Terminal` control, while the same response with null id is refused before it can borrow the request era and license terminality.

This is implementation guidance and verification evidence, not a Codex patch. Claude remains the exclusive writer for Slice 0.

Rul1an:
**Codex exact-head review: CHANGES REQUESTED on `4e6a46c2df8c168a2f14ea8a82e223e226a0e00f`.**

1. **P1: known request-only methods can still bypass the RequestId and 2026 metadata contracts by omitting `id`.** `classify_message` treats every method-bearing object without an `id` member as a notification. The existing public test `contract_missing_request_id_does_not_correlate` proves that `tools/call` without `id` is accepted and normalized as a tool call. Because the same shape is classified as a notification, it also escapes the required `RequestParams._meta` check under 2026. MCP defines `tools/call` as a request and requests MUST include a string or integer ID. Preserve extension notifications, but fail closed for known request-only MCP methods; add an end-to-end 2026 `tools/call`/no-id/no-meta test and a real notification control.

2. **P2: the `1e0` test is not lexical evidence.** `json!(1e0)` constructs the same floating value as `json!(1.0)`, so the loop tests one representation twice. Feed raw JSON containing `1e0` through the parser and keep `1.0` as a separate case.

3. **P2: the terminality test lacks its positive control.** `an_unacceptable_response_id_can_never_license_terminal` only asserts refusal. On the same end-to-end path, first prove that a valid integer request/response under a legacy era with missing `resultType` reaches `ResultConclusion::Terminal`; then prove the unacceptable ID is refused before conclusion. Otherwise an implementation that refuses every response satisfies the claim.

4. **P2: the accepted-string path adds an avoidable full-ID transient copy.** `require_acceptable_id` clones in `normalize_jsonrpc_id`, drops that value, clones into `CorrelationId`, then clones into the public text. Use one kind-aware match/clone. This is hostile input and the extra copy is introduced by this correction.

Also reconcile the stale `era.rs` comments that still say IDs are not integer-constrained / `string | number` and assign the vocabulary to `normalize_jsonrpc_id`; the MCP contract now lives in `accept_id` and is string-or-integer.

Independent verification on this head: 69 focused era-wiring tests, 13 public correlation tests, 2 import-smoke tests, and 9 transport-compat tests pass. The blocker is semantic, not a failing existing gate.

Rul1an:
## Exact-head non-builder review: APPROVED

Reviewed final head `30186f058dfe28d37086b86155dc61d3d7740df7` in the detached Codex worktree. No actionable findings remain.

The final correction closes the JSON-RPC RequestId domain without widening the public event shape: request and success-response IDs are string-or-integer only; notifications may omit IDs; error responses with absent/null IDs remain ingestible but uncorrelated; string `"1"` and numeric `1` remain distinct; fractional and exponent-form numeric IDs are refused before normalization; and known request-only methods cannot downgrade to notifications by omitting `id`. The stale rustdoc and duplicate fake `json!(1e0)` case are gone; the raw-text regression owns lexical `1e0`.

Independent exact-head verification:
- `cargo test -p assay-core`: green, including 14/14 ID-correlation, 2/2 import-smoke, and 9/9 transport-compat tests
- `cargo clippy -p assay-core --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- `git diff --check origin/main...HEAD`: clean

CI note: the current dependency-security failure is installation infrastructure, not this diff: `cargo install --locked cargo-audit` failed downloading `arc-swap` with curl HTTP/2 framing error, after which `cargo audit` was unavailable. It requires a failed-job rerun; it is not a pass and does not license merge.

This comment is one non-building review arm on the exact final SHA. GitHub does not allow the repository owner account to submit an APPROVE review state on its own PR.

Rul1an:
@coderabbitai review

Please review exact final head `30186f058dfe28d37086b86155dc61d3d7740df7`. The prior formal reviews target older SHAs and do not count for same-head quorum.

Rul1an:
## Exact-head second review: CHANGES REQUESTED

This supersedes my approval comment on final head `30186f058dfe28d37086b86155dc61d3d7740df7`. A second independent non-building review found three reachable gaps; I verified them against the pinned MCP 2026 schema and the current code.

### P1: the integer-only RequestId rule contradicts this programme’s pinned source of truth

`era.rs:167-173` accepts only integer-backed JSON numbers. The final 2026 TypeScript schema at pinned commit `5f5440bb26a62e2cf3440b92da5a667efa03b267` defines `RequestId = string | number`, and the same issue ledger explicitly says that schema is the source of truth. Base JSON-RPC says fractional numeric IDs SHOULD NOT be used, not MUST NOT. The prose overview simultaneously says “string or integer”, so the upstream release is internally inconsistent; our prior review followed that prose and stated the schema incorrectly. That statement is withdrawn.

Required: preserve the full schema domain (`string | number`) in the crate-private typed correlation key, add fractional/exponent raw-wire controls, and record the upstream prose/schema inconsistency as a bounded non-claim. Do not describe non-integer numbers as protocol-invalid.

Sources: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/5f5440bb26a62e2cf3440b92da5a667efa03b267/schema/2026-07-28/schema.ts#L2677-L2701 and https://www.jsonrpc.org/specification#request_object

### P1: request-only classification covers only two standard requests

`REQUEST_ONLY_METHODS` contains only `tools/call` and `tools/list`. The pinned schema’s `ClientRequest` union also includes `server/discover`, `completion/complete`, `prompts/get`, `prompts/list`, `resources/list`, `resources/templates/list`, `resources/read`, and `subscriptions/listen`. An ID-less known request such as 2026 `resources/read` is therefore accepted as a notification and sheds both the ID requirement and required request `_meta`. The second reviewer reproduced that public parse.

Required: derive/document the closed known-request set from the pinned schema revisions Assay claims to recognize, while preserving unknown extension notifications. Pin at least one non-tool request end to end under 2026.

### P2: the stated every-member duplicate invariant stops below the transport wrapper

`UniqueValue` says every duplicate member is refused so future-significant names cannot bypass an enumeration. But `TransportTranscript` and `TransportTranscriptEntry` are still derived directly; duplicate unknown wrapper/entry members are ignored. A top-level `"future":1,"future":2` parses successfully.

Required: make the outer transcript and entry boundary duplicate-aware without a second unbounded whole-input deserialization. Add unknown wrapper- and entry-member regressions. If the invariant is intentionally narrower, the code and public claims must be narrowed together, but retaining the current every-member rule is the safer boundary.

### Verification/disposition

The reviewer’s focused existing suites remained green (era 35, parser wiring 73, correlation 14, transport compatibility 9), demonstrating these are uncovered cases. CI is fully green after rerunning the cargo-audit installation failure, but green CI does not close these findings. No exact-head quorum exists. Claude remains the exclusive writer.

Rul1an:
## Exact-head merge review: APPROVED

Final head: `f71c3dce9820e3de16118e357d718b3c8b3a657d`
Base: `128d4c79d2948645afe60cd0d4488ceca8a81ae5`

Review quorum uses the documented two-non-builder fallback. CodeRabbit CLI remained in `reviewing` beyond the full ten-minute review window and was stopped, so it is explicitly **not** counted as a pass.

- Codex independent exact-head review: clean, no actionable findings.
- Second non-building exact-head review: CLEAN, no actionable findings. Its `range-diff` confirms all 25 pre-rebase commits are patch-identical after the conflict-free rebase.
- The only new commit is the test-only CLI diagnostic assertion. It pins the semantic refusal phrase and separately confirms that the input-selected id is not echoed.
- Exact-head GitHub matrix: 55 successful checks, 10 path-conditional skips, 0 failures or pending checks. `CI`, `host-capability-check`, and `lane-check/proof` are green. GitHub reports CLEAN / MERGEABLE.

No builder self-review or bot status context is counted toward this approval.

PR #1934, opened 2026-07-31
feat(mcp): era-parity corpus for protocol-era result conclusions

Rul1an:
Adds a crate-internal, fixture-driven era-parity corpus (18 vectors) and the
conclusion behaviour it pins. Slice 1 of #1866.

## What this claims

Parser and conclusion parity: for every vector, every hop and every call, the
observed state and the licensed conclusion match a decoded manifest row. Plus
the property that every vector references the same frozen baseline row.

Observation is pinned as well as conclusion. A right conclusion reached from a
wrongly observed input is a passing test that proves nothing, so the manifest
label is decoded into a value and compared against what the parser actually
produced; an unrecognised label fails rather than being skipped.

## What this does not claim

- **No production wire-to-profile binding.** Nothing here emits or round-trips
  a profile matrix. That end-to-end proof is Slice 2.
- **No broad conformance claim.** The corpus is marked `exploratory` and states
  so in its own README: it is lower maturity than
  `conformance/privileged-mcp-action-v0/`, carries no reproduction request, and
  inherits none of that corpus's standing.
- **No profile-vocabulary change.** The frozen row is referenced, never
  restated, and the reference is a baseline supplied by surrounding
  producer-reported evidence — a wire transcript cannot confirm
  `policy_decision_recorded`.
- **Public API unchanged.** `McpEvent` and `parse_mcp_transcript` are
  untouched; the era sidecar stays `pub(crate)`.

## Behaviour added

- `CompleteWithContinuation` — a completion claim carrying either continuation
  member. `InputRequiredResult` requires at least one of `inputRequests` or
  `requestState`, so both forms count. `CallToolResult` sets no
  `additionalProperties: false` and lists neither, so both are schema-legal:
  only the semantic layer can state the fault.
- `InputRequiredWithoutContinuation` → `Invalid(UncontinuableInputRequired)` —
  the same prose-versus-schema gap on the interim arm, raised by review.
  `InputRequiredResult` says at least one of `inputRequests` or `requestState`
  MUST be present and encodes nothing of the sort (`required` lists only
  `resultType`), so a result asking for input while offering no way to supply it
  passes the published definition and cannot be answered. `Invalid`, not
  `NonTerminal`: that answer would call a dead exchange validly unfinished, and
  not `Incomplete` either, since nothing arriving later makes it continuable.
  Both arms read the same continuation helper, so the two ways that pair can
  disagree cannot drift.
- `InputRequiredWithMalformedContinuation` → `Invalid(MalformedContinuation)` —
  a continuation member is judged by shape, not presence. `requestState` must be
  a string, `inputRequests` an object, absent and explicitly null are the same
  silence, and a broken member is a fault even beside a well-formed sibling.
  Only the top-level type is read; re-deriving `InputRequest` would copy a
  published definition into the code and go stale against it.

  The two arms answer differently and must: on `complete` a malformed member
  folds into the contradiction, which is fail-closed, while on `input_required`
  that same fold would land on *valid*. Raised by review, and not covered by
  schema parity — the runtime conclusion path never validates against the
  vendored artifact, so a `requestState` of `7` reached the conclusion layer
  unchallenged and was licensed as `NonTerminal`.
- `ContradictoryResult` and `RecognitionUndeterminable` as distinct incomplete
  reasons. A closed answer and an open question have different remediations.
- `CapabilityObservation`, value-free — no advertised name or token byte is
  retained. Capabilities are stated per request and MUST NOT be inferred from a
  prior one, so era and capability travel together on the existing
  `CorrelationId` outstanding map and reach a response only by correlation.
- The capability axis is authoritative only from the revision that defines it.
  A legacy call keeps its pre-slice answer, and a future-key shape fault cannot
  invalidate it.

## Schema layer

Two artifacts vendored from spec commit
`5f5440bb26a62e2cf3440b92da5a667efa03b267`, digest-pinned in `PIN.json` and
checked before use, so the corpus runs offline and a silent edit cannot change
what a vector means. They are written against different drafts (2020-12 `$defs`
vs draft-07 `definitions`), so the container is chosen from what the artifact
carries and one carrying both is refused.

Schema-valid is an observation, never the semantic oracle. Two vectors state no
schema verdict at all, because which artifact applies is exactly what is
undetermined.

## Verification

Twenty-eight mutations, one per new guard, each with an anchor matching exactly
once and the source verified changed. All twenty-eight were killed. `m12`, which drops the type guard, kills exactly one test — the one written for it. The narrow ones
are the point: `m6`, `m7` and `m10` each kill exactly one test, and `m9` kills
only the two legacy composites, so no guard leans on another's failure.

The shape guard asks the canonical classifier rather than the fixture wrapper.
`classify_message` answers `Request` whenever `method` is present, even beside a
`result`, so a hybrid in a response slot parses as a request, never gets a
`result_observation`, and drops out of conclusions and coverage — while a
wrapper-shaped rule accepts it because it does carry a `result`. The fault stays
value-free: `MessageKind::Request` carries the method the input chose.

Coverage is over **result-bearing** responses, and the scope is proven rather
than assumed. `classify_message` calls a JSON-RPC error response a `Response`
and `correlate_calls` consumes it as one, but this corpus reads responses
through `result_observation` — so an extra or duplicate error response would
have been invisible to schema, coverage and conclusions alike. Including them
would mean giving each a row, and a row needs a conclusion label this corpus
has no vocabulary for; inventing one for a case no vector exercises would add
untested vocabulary to cover an absence. The shape is refused instead, which
makes the downstream filter total, and the guard is what changes the day an
error vector is wanted.

Cardinality is a rule, not a `remove(0)`: `exactly_one` refuses none and many,
so a fixture that grew a second message cannot be scored on its first.

The corpus keys on the typed `CorrelationId`, not on `McpEvent.jsonrpc_id`. That
public field is a `String` and renders JSON `1` and `"1"` identically, so the
coverage check could not tell two calls apart — it would score one response
twice and leave the other unread while reporting full coverage. The typed key
now travels in the crate-private sidecar (no public API change), the manifest
states `jsonrpc_id` as native JSON, and `correlation-id-type-collision` puts
numeric `1` beside string `"1"` with different capability sets so a mis-pairing
changes a conclusion. Flattening either side is killed by mutation.

`vector(id)` is first-match, so a duplicate id would make every assertion
exercise the first row while the second file sat present and unread. Unique ids,
unique paths and exact bidirectional membership between manifest paths and
`vectors/*.json` are now structural guards, each with an isolated negative case
and a killed mutation.

The conflicting-era description was factually wrong and is corrected. The
request validates under **both** artifacts; only the result flips. All four
outcomes are pinned by a counterfactual test.

`conclude_request` reads the capability set. `RequestMetaObject.required` names
`protocolVersion` and `clientCapabilities` both, so a version alone left a 2026
request scored `Valid` while being schema-invalid. `CoreOnly` or
`ExtensionNotUnderstood` is valid; `Absent` or no observation is
`Invalid(MissingCapabilities)`; `Malformed` is `Invalid(MalformedCapabilities)`;
legacy requests are untouched. It cannot wait for the response side — a refused
or abandoned call never gets one, and then the request is the only record there
will ever be.

The `per_call` table proves coverage before scoring. Iterating rows shows each
named id resolves, not that the rows account for every response: two rows naming
one id would score it twice while a second response went unread. The declared id
set is now compared with the transcript's response set exactly.

No two vectors carry the same bytes, and a harness assertion enforces it.
`modern-unrecognized-token` and `capabilities-core-only-noncore-token` were
byte-identical — one input wearing two labels, with only the second declaring a
capability state. The first now carries its own token and declares its
capability set. The same test requires all three capability states to appear, so
the axis cannot decay into a label; re-introducing the duplicate turns it red
and names both colliding vectors.

Tests killed per mutation, all measured on this head rather than carried over:

```
m1 4, m1a 5, m1b 6, m2 3, m3 2, m4 4, m5 2, m6 1, m7 1, m8 2, m9 2, m10 1,
m11 4, m12 1, m13 1, m14 1, m15 1, m16 1, m17 1, m18 2, m19 2, m20 1,
m21 1, m22 1, m23 1, m24 1, m25 1, m26 1
```

Sharing one continuation reading across both arms is load-bearing rather than
tidy. Leaving `inputRequests` unread kills five and leaving `requestState`
unread kills six, across the completion and interim vectors; before the interim
arm read the same shape, those two killed three. `m2` moved from two to three
when the wrong-type test was added, which is why the table was re-measured
rather than copied.

**The first post-mutation final baseline was 599/600.** It failed
`a_legacy_call_is_not_judged_against_the_capability_contract` with the `m10`
outcome, on a source line that was correctly restored: the restore preserved
the original mtime, so Cargo reused the mutant binary. A touch-based rerun went
green and does not count. Only `cargo clean -p assay-core` followed by a
rebuild produced 600/600. Two review findings on the continuation members were
closed after that, so this head's baseline is **612/612**.

After rebase onto `517a402a`, on a clean package build: assay-core lib 600/600;
`mcp_id_correlation` 18/18; `mcp_import_smoke` 2/2; `mcp_transport_compat` 9/9;
`cargo fmt --check` clean; `clippy --all-targets -D warnings` clean. Both pins
re-verified against `PIN.json` after the rebase: MATCH, MATCH.

## Vendored-byte disposition — `git diff --check` is not clean, by decision

The two schema artifacts are byte-exact copies of the pinned upstream files,
which end with a blank line. `git diff --check` therefore reports **exactly two
messages and no others**:

```
schema-2025-06-18.json:2517  new blank line at EOF
  af845e7e5b9d27107d1690f0936022546177a1403e63ffb11470135b296a2e01
schema-2026-07-28.json:3963  new blank line at EOF
  ef70b61f99b6d2e5e3b46863822eab08dff6a45bedc7a08914e0e5b133f40203
```

These are accepted, not fixed. Normalising the trailing byte changes the sha256
that `PIN.json` records and that the corpus checks before using either
artifact, which is the property the pin exists to prove. This PR does not claim
a clean `git diff --check`; it claims two known messages bound to those digests
and zero others, so a third message would be a real signal.

## One change outside the slice paths

`.pre-commit-config.yaml` excludes the vendored schemas from
`end-of-file-fixer` and `trailing-whitespace`. Those hooks *rewrite* the bytes
in place, which silently breaks the pin — unlike `git diff --check`, which only
reports. The exclusion follows the rationale already in that file for
`docs/experiments/**/archive-contents/`.

The corpus also carries its own directory-scoped `.gitattributes` marking the
schema artifacts `-text`.

**This was found by CI, not by review.** Without it the pin does not hold
cross-platform: git translates line endings on checkout, so on `windows-latest`
the LF blobs came back as CRLF, both digests failed, and the corpus reported
tampering where there was none — the exact property the pin exists to prove,
failing on one platform.

It is scoped to the fixture directory rather than added at the repository root:
the root file is a trigger path for the MCP 2026 JSON-RPC id conformance lane,
which is unrelated to this corpus, and the trigger pattern has no wildcard so a
nested file does not match it. The root `.gitattributes` is byte-identical to
`517a402a`. The live-URL drift that lane reports is a separate matter and is
not touched here.

It sets `-text` only. No `-whitespace`, so the two blank-line-at-EOF messages
above stay visible rather than being suppressed, and the digest is taken over
the committed bytes rather than over line-ending-normalised content — which
would weaken the pin into something that is no longer about what was
committed.

Worth flagging for review: the `end-of-file-fixer` hook modified the pinned
bytes and its own rollback did not restore them. It was caught because the
digests stopped matching `PIN.json`, and the original bytes were recovered from
the index.

Refs #1866


Rul1an:
**Codex exact-head review (`8337ce6b`) — one actionable finding**

**[P1] `input_required` without a continuation is currently licensed as valid `NonTerminal`.**

The pinned `InputRequiredResult` text says: “At least one of `inputRequests` or `requestState` MUST be present.” The generated JSON Schema does not encode that rule: its `required` list contains only `resultType`, with no `anyOf`/`oneOf`. This is therefore exactly a semantic-layer obligation rather than something schema validation will catch.

On this head, `observe_result` maps every string `resultType == "input_required"` directly to `ResultObservation::InputRequired`, without consulting `carries_continuation`; `conclude` then returns `ResultConclusion::NonTerminal`. `NonTerminal` is documented as valid but unfinished, and Slice 2 is designed to accept it while refusing `Incomplete`/`Invalid`. A response with neither continuation member therefore passes the conclusion boundary as valid even though it violates the pinned MUST.

Please add a load-bearing corpus vector (or equally durable end-to-end test) for `input_required` with neither member, pin that the generated schema accepts it while the semantic conclusion does not, and give the state its own value-free typed observation/reason. Include the symmetric explicit-null case so `null` cannot masquerade as a continuation. Mutate the continuation guard and show that the new test goes red. No profile vocabulary change is needed.

No other actionable code finding from my pass so far. The vendored-byte disposition and removal of the `.gitattributes` scope expansion are sound.

Rul1an:
**Codex non-builder exact-head review — `4c303610b947283bd380743cbf3e2168111aadf2` — superseded by three open P2 findings**

GitHub CI is complete and green on this head, but this review is **not an approval**. A second independent pass found three issues that I verified against the current source, corpus, and vendored schemas:

1. **Conflicting-era counterfactual is misstated.** The request validates under both pinned schemas; only the result differs (valid in 2025, invalid in 2026 because `resultType` is required). README/MANIFEST prose currently says the legacy side is the reverse.
2. **`per_call` erases JSON-RPC ID type.** The harness keys rows on public `McpEvent::jsonrpc_id: String`, so numeric `1` and string `"1"` collapse even though production correlation deliberately keeps them distinct.
3. **Corpus identity is not structurally closed.** `vector(id)` returns the first matching row, so duplicate manifest IDs can make assertions exercise the wrong file. Unique IDs, unique paths, and exact bidirectional MANIFEST-to-directory membership are not yet enforced.

The prior corrections on this head remain valid: modern request capability assessment, exact response coverage, continuation semantics, value-free observations, unique vector bytes, and vendored pins. Writer proof and green CI establish integration, not review cleanliness.

Claude is addressing these findings on the exclusive Slice 1 branch. Any new push invalidates this exact-head artifact and requires a fresh review/quorum.

Rul1an:
@coderabbitai review

Please review exact head `4c303610b947283bd380743cbf3e2168111aadf2`. The prior green status/reviews target older SHAs and are not being counted.

Rul1an:
### Exact-head Codex review: no actionable findings

Reviewed head `20d5bc0687a9036feff9db99dfd00abcc9700a99` read-only.

The final delta closes the last response-coverage gap: the corpus guard now defers to the production `classify_message` decision before accepting a wrapped response, so a `method + result` hybrid cannot pass as result-bearing while the parser treats it as a request. The negative case is value-free and distinguishes classifier refusal from missing-body refusal. Single and duplicate JSON-RPC error-response cases remain covered.

The earlier exact-head findings are also closed:
- result-bearing response scope is corpus-wide and explicit;
- the conflicting-era counterfactual requires exactly one request and one result before testing the four outcomes;
- vector, path, directory, digest, and typed JSON-RPC ID coverage remain intact;
- public `McpEvent` and parser API remain unchanged;
- permanent metadata says 18 vectors and 28 mutations, one per guard.

Independent local proof: `cargo test -p assay-core mcp::era_parity_tests --lib` passed 38/38 on this SHA. A second non-building reviewer independently found no actionable finding on the final delta and also ran 38/38. The writer reports 612/612 lib tests, fmt/clippy clean, and all 28 mutation guards killed.

Residual: I did not rerun all mutation rewrites independently. The final classifier mutation was checked by the writer with an exact-one anchor and source-change assertion; the second reviewer verified the discriminating oracle from the exact-head diff.

CI and review quorum remain separate merge gates.

Rul1an:
@coderabbitai review

Final Slice 1 head: `20d5bc0687a9036feff9db99dfd00abcc9700a99`. Please review this exact SHA. Prior bot reviews target superseded heads.

PR #1946, opened 2026-08-01
fix(core): reject unparsable `expected:` blocks instead of silently passing

Rul1an:
## The bug

A YAML typo in a test's `expected:` block produced a test that **always passed**.

`impl Default for Expected` returns `MustContain { must_contain: vec![] }`, and `MustContainMetric` iterates that empty vec and returns `pass(1.0)`. Three paths in `model/serde.rs` reached that default in silence — an `expected:` object that failed to parse, a list entry matching neither the strict nor the legacy form, and a multi-element list whose elements after the first were dropped without a word.

## Two commits

`607d4a86` stopped the silent fallback. Adversarial review then found it had landed the detector without repairing what the detector caught, and had introduced a regression of its own. `b382e729` fixes that. Every finding below was reproduced against the built binary.

## What `b382e729` fixes

**A regression introduced by the first commit.** `parse_expected_entry` returned the strict serde error as soon as a block carried a `type:` key, before the legacy heuristics ran. The old code fell straight from strict-failure into those heuristics. So a tagged block with a legacy *value* shape went from a working assertion to a whole-suite config error:

```yaml
- {type: must_contain, must_contain: "hello"}   # scalar = legacy spelling of a 1-element list
- {type: sequence, sequence: [Search, Create]}  # the form documented in migration.md
```

The heuristics now run first; the strict error is reported only when no legacy key matched.

**Silent fallbacks the first commit moved without auditing.** It relocated the legacy heuristic block verbatim, carrying along the same swallow-the-error pattern its own thesis condemns:

| | was | now |
|---|---|---|
| `must_contain: {oops: 1}` | `unwrap_or_default()` → empty vec → passes for any response | hard error |
| `sequence: 42` | `.ok()` → `sequence: None` → `sequence_valid` passes unconditionally | hard error |
| `$ref: 42` | `unwrap_or("")` | hard error |

The `sequence` case was the worst: `assay run` reported `status: pass, score: 1.0` and `assay validate` reported "Validation OK".

**Vacuous assertions are now caught where it matters.** An `expected:` block written out as empty is rejected at **parse** time, so every command that loads a config catches it — including `run` and `ci`. The first commit's rule lived only in `validate()`, which neither command calls, and which is not what the workflow `assay init --ci` generates for users runs.

An **omitted** `expected:` key stays permissive, because a test may legitimately carry its checks in `assertions:`. When such a test has no assertions either, it is a **warning**: `W_CFG_VACUOUS_EXPECTED` (renamed from `E_CFG_` to match the repo's `W_` convention for exit-0 diagnostics).

That split also resolves an exit-code inconsistency the review surfaced. `validate()` has four call sites, not one, and only two reach `decide_exit`. The first commit's error-severity diagnostic therefore produced **1** from `assay doctor` and **0** from `assay doctor --format json` while claiming to be an exit-2 config error. Hard errors now flow through the pre-existing `E_CFG_PARSE` path instead.

**Self-inflicted breakage the first commit shipped.**

- `templates.rs::CI_EVAL_YAML` emitted `type: json_schema` without the required `json_schema` field, so `assay init --ci` produced a project `assay run` refused to load. The old parser degraded that to a vacuous always-green test, which is why it went unnoticed — the repo-root `ci-eval.yaml` had already been hand-patched while the generator drifted. `contract_init_ci` only grepped the generated file's text; it now loads the scaffold through the real loader.
- `examples/demo/assay-safe.yaml` and `common-mistake/assay.yaml` used `must_contain: []` as a deliberate stub, and `.github/workflows/assay-security.yml` gates on validating the first. (Latent: that workflow pins a released binary, so it would have gone red only after publish.) Both now assert against the dummy model's fixed response.

**Round-trip.** `TestCase.expected` is skipped on serialize when vacuous. Without it, `assay migrate` would materialise `must_contain: []` into every test that omitted the key — output its own parser now rejects.

**Docs.** `eval-yaml.md` documented `json_match` and `exact_match`; neither is an `Expected` variant, so both were silently vacuous before and hard-fail now.

## Behaviour changes for existing suites

Suites relying on a silently-vacuous `expected:` now fail to load. That is the point. Specifically:

- `- must_not_contain: "x"` in list form has no legacy heuristic (the legacy key set is frozen at `$ref`, `must_contain`, `sequence`, `schema`), so it is a hard error naming the tagged form rather than a silent always-pass.
- An untagged single mapping such as `{must_contain: [...]}` is now read with the same heuristics as a list entry, so it actually checks.
- Multi-element `expected:` lists are rejected. Only element 0 was ever evaluated. `TestCase.expected` holds one `Expected` and is matched on across metric dispatch, validate, doctor and config/resolve, so supporting a collection is a contract change beyond this fix.

## Why a missing `expected:` key stays permissive

Omitting `expected:` is a legitimate shape: a test may carry its checks in `assertions:`, and several existing contract tests use a bare `input:` with no expected block. Making the missing key an error would reject valid configs — and would reject the output `assay migrate` itself writes.

The distinction that matters to a user is not *"was the key written?"* but *"does this test assert anything?"*. So the two cases are split by what the parser can actually observe: an assertion written out as empty is a hard error at parse time; an omitted key is permissive, with a warning when the test has no assertions either. The permissiveness is deliberate and documented at both ends rather than emergent from a missing `else` branch.

## Known gap

A typo *inside* a tagged block (`type: must_contain` with `must_contaisn:`) still parses, because `Expected` lacks `deny_unknown_fields` and `serde_ignored` cannot see into `expected:` — the hand-written `Deserialize` consumes the subtree as an opaque `Value`. It surfaces as an empty `must_contain`, which is now a parse error, so the always-green outcome is gone; only the message is generic. Closing it properly has a wider blast radius and is left as a follow-up.

`assay migrate` still cannot load multi-element legacy lists — the one shape being deprecated is the one it refuses. Restoring the legacy heuristics recovers everything else it was documented to handle. Tracked as a follow-up rather than expanded here.

## Tests

Parser: unparsable object, unrecognized list entry, multi-element list, empty list, tagged-error passthrough, untagged heuristics, permissive missing key, tagged-with-legacy-scalar, `type: sequence`, unparsable `sequence`, unparsable `must_contain`, explicit empty `must_contain`/`must_not_contain`, and a serialize→reparse round-trip.

CLI contracts: exit 2 / `E_CFG_PARSE` through the real binary for each parse path including the explicit-empty case on `assay run`; `W_CFG_VACUOUS_EXPECTED` as a warning at exit 0 from `assay validate`; and the `init --ci` scaffold loaded through the real loader.

Verified: `cargo test --workspace` (201 suites, 0 failures), `cargo clippy --workspace --all-targets -D warnings` via pre-push, `cargo fmt --check`.

## Review

Reviewed by three independent adversarial agents spawned by the builder's orchestrator, each running read-only on head `607d4a86` with a distinct lens (validate call-site blast radius, parser semantics and round-trip, shipped docs and generated YAML). Two built the parent commit for a true A/B. All three returned do-not-merge; their findings are what `b382e729` implements. Load-bearing claims were re-verified by the builder before acting, and one claim — that `migrate` output fails `validate` — was checked and downgraded: `migrate` does materialise the vacuous default, but assertion-carrying tests are exempt, so it did not by itself produce a validate failure.

### DIFF
--- a/crates/assay-cli/src/templates.rs
+++ b/crates/assay-cli/src/templates.rs
@@ -14,6 +14,7 @@ tests:
       prompt: "ci_schema"
     expected:
       type: json_schema
+      json_schema: "{}"
       schema_file: "schemas/ci_answer.schema.json"
   - id: "ci_smoke_semantic"
     input:
--- a/crates/assay-cli/tests/contract_init_ci.rs
+++ b/crates/assay-cli/tests/contract_init_ci.rs
@@ -73,6 +73,37 @@ fn test_init_ci_contract() {
     );
 }
 
+/// The generated scaffold must actually LOAD, not merely contain the right substrings.
+///
+/// Regression guard: `CI_EVAL_YAML` once emitted `type: json_schema` without the
+/// required `json_schema` field. The old parser silently degraded that to a vacuous
+/// always-passing test, so every text assertion above still held while
+/// `assay run --config ci-eval.yaml` was broken. Grepping the template cannot catch
+/// that class of defect; parsing it can.
+#[test]
+fn contract_init_ci_scaffold_is_loadable() {
+    let temp = tempfile::tempdir().unwrap();
+
+    Command::cargo_bin("assay")
+        .unwrap()
+        .current_dir(temp.path())
+        .arg("init")
+        .arg("--ci")
+        .assert()
+        .success();
+
+    // `validate` exercises the same loader as `run`/`ci`, and reports a diagnostic
+    // instead of the bare exit code those commands emit on a config error.
+    Command::cargo_bin("assay")
+        .unwrap()
+        .current_dir(temp.path())
+        .arg("validate")
+        .arg("--config")
+        .arg("ci-eval.yaml")
+        .assert()
+        .success();
+}
+
 #[test]
 fn test_init_preset_and_pack_alias_contract() {
     let temp = tempdir().unwrap();
--- a/crates/assay-cli/tests/exit_codes/core.rs
+++ b/crates/assay-cli/tests/exit_codes/core.rs
@@ -474,9 +474,10 @@ tests:
     assert_eq!(run["reason_code"], "E_CFG_PARSE");
 }
 
-/// `assay validate` sweeps a suite for always-green tests without running it.
+/// An explicitly-written empty assertion is a HARD ERROR, and it must be caught on
+/// the paths that decide outcomes — `run` and `ci`, not just `validate`.
 #[test]
-fn contract_validate_flags_vacuous_expected() {
+fn contract_run_rejects_explicitly_empty_assertion() {
     let dir = tempdir().unwrap();
     fs::write(
         dir.path().join("assay.yaml"),
@@ -492,6 +493,37 @@ tests:
     )
     .unwrap();
 
+    Command::cargo_bin("assay")
+        .unwrap()
+        .current_dir(dir.path())
+        .env("ASSAY_EXIT_CODES", "v2")
+        .arg("run")
+        .arg("--config")
+        .arg("assay.yaml")
+        .assert()
+        .code(2);
+
+    let run = read_run_json(dir.path());
+    assert_eq!(run["reason_code"], "E_CFG_PARSE");
+}
+
+/// A test that omits `expected:` AND has no `assertions:` asserts nothing, but the
+/// omission itself is a documented, legitimate shape — so `assay validate` sweeps for
+/// it as a WARNING (exit 0), not an error.
+#[test]
+fn contract_validate_warns_on_test_with_no_assertion() {
+    let dir = tempdir().unwrap();
+    fs::write(
+        dir.path().join("assay.yaml"),
+        r#"suite: vacuous
+model: dummy
+tests:
+  - id: always_green
+    input: hello
+"#,
+    )
+    .unwrap();
+
     let out = Command::cargo_bin("assay")
         .unwrap()
         .current_dir(dir.path())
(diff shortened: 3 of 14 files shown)


Rul1an:
Exact-head review on `b382e7292a58edaab6ded94204049a68f34eff01` (non-builder, read-only).

Two actionable parser findings remain:

1. **[P1] A failed tagged block can silently change metric type through a legacy key.**

   Reproduced against the built head binary:

   ```yaml
   expected:
     type: regex_match
     must_contain: "not-the-dummy-output"
   ```

   `assay validate --format json` exits 0 with zero diagnostics. The strict `regex_match` parse fails because `pattern` is missing, then the unconditional legacy fallback returns `Expected::MustContain`. The compatibility case in the PR only justifies normalizing a legacy value shape when the declared type and legacy key agree. It does not justify replacing one declared metric with another.

   Please constrain tagged fallback to the matching legacy key/type pair, or return the stored strict error.

2. **[P1] Ambiguous legacy mappings still discard an assertion.**

   Reproduced:

   ```yaml
   expected:
     must_contain: "passed"
     sequence: ["Search"]
   ```

   Validation exits 0; stderr says `Using first match`, and only `must_contain` survives. This is the same partial-enforcement failure mode for which this PR correctly rejects multi-element `expected:` lists: the config declares two checks and Assay silently enforces one. Make `matched_keys.len() > 1` a parse error instead of warning-and-first-wins.

One explicit disposition is also needed, but I am not treating it as an unannounced blocker: a test with neither `expected:` nor `assertions:` still exits 0 and reports `1 passed` under `assay run`; only a separate `assay validate` invocation emits `W_CFG_VACUOUS_EXPECTED`. I reproduced that on this head. If output-only tests are intentionally supported, keep the behavior but narrow the PR/release claim to *present malformed or explicit-empty* `expected:` blocks and record that `run/ci` do not surface the warning. Otherwise the parser has enough information in `RawTestCase` to reject the neither/nor shape.

The prior `$ref` inline finding is fixed in this head and is now outdated. CI has no failures; one Wave 0 job was still running at review time. The branch remains 14 commits behind `main`, so the eventual final head needs update/rebase and same-head CI/review.

Rul1an:
Superseded by #1948. The replacement preserves both commits from this PR on the current `main` base and adds the two exact-head P1 corrections: mismatched tagged-to-legacy fallback and ambiguous legacy assertion truncation. Claude's branch remains untouched.

Issue #1949, opened 2026-08-02
Reject nonempty but behaviorally vacuous assertions


Rul1an:
## Problem

PR #1948 hardens the `expected:` surface, but the separate `assertions:` surface still accepts nonempty configurations that can enforce no behavior. This is orthogonal to omitted `expected:` compatibility and must not be hidden inside #1948.

Known cases to verify from parser through assertion evaluation:

- `trace_must_call_tool` with an empty tool name or `min_calls: 0`;
- `trace_must_not_call_tool` with an empty tool name;
- `trace_tool_sequence` with an empty sequence and combinations that cannot constrain a trace;
- `trace_max_steps` and the internal assertion variants at boundary values that pass every valid trace;
- nonempty assertion lists whose every member is ineffective.

## Required direction

1. Inventory every `TraceAssertion` variant against its evaluator.
2. Add red parser-to-evaluator tests for each proven no-op and positive controls for the nearest effective value.
3. Reject explicit ineffective assertions fail-closed before execution. Preserve the existing compatibility rule for a genuinely omitted `expected:` block when effective assertions carry the test.
4. Keep diagnostics value-free and name the ineffective field or variant.
5. Correct the public YAML reference from the executable enum, not from old examples.

## Acceptance

- No explicit nonempty assertion configuration can pass solely because it performs zero checks.
- Omitted `expected:` plus at least one effective assertion remains supported.
- Focused tests, full affected crate suites, fmt, clippy `-D warnings`, and GitHub CI are green.
- Release gate: originally declared release-blocking for v3.37.0, which shipped 2026-08-02 with this issue open because the gate lived only in prose. The gate is now mechanical: the `release-blocking` label plus the `release-contract` workflow check. **Re-targeted to v3.39.0** — v3.38.0 ships items 2, 3, and 4 (static sweep, verified-not-applicable, gate mechanics); the remaining items 1 (parse-time / opt-in `--strict` at load) and layer 2 (exercised/not-exercised companion cover) are the v3.39.0 scope. See the retarget comment for the full reasoning.

## Status (as of retarget)

- **Item 2 — static sweep:** landed (#1983). `assay validate` no longer steps over `assertions:`; effectiveness via `matchers::ineffective_reason` (single source with the evaluator).
- **Item 3 — verified not applicable:** `ArgsValidMetric` operates on `Expected::ArgsValid`, distinct from `TraceAssertion::ArgsValid`; the #1948 bypass does not re-enter here.
[...]

PR #1959, opened 2026-08-03
An assertion that cannot check anything is not a pass

Rul1an:
## Description

The reporting half of #1949. The `assertions:` evaluator silently skipped whole assertions, or degraded a missing policy field to a value that admits everything, and contributed no diagnostic — which a suite cannot tell apart from a check that ran and held.

`AGENTS.md` already forbids this under Development Discipline: *"Never turn absence of evidence, failed validation, skipped review, or unavailable infrastructure into a clean result."* An `if let Some(..)` with no `else` inside a checker is a checker that does not check, so this enforces an existing rule rather than introducing one.

### Ten shapes that could not fail for any input

| Variant | Shape | Mechanism |
|---|---|---|
| `args_valid` | no `test_args` | `if let Some(args)` with no else; there is no trace-mode evaluation at all |
| `sequence_valid` | written against the typed `test_trace` | field was destructured away with `..` and never read |
| `sequence_valid` | no `test_trace_raw`; no `policy` | silent skips |
| `sequence_valid` | `regex` absent, empty, or non-string | all three degraded to `.*` |
| `tool_blocklist` | no `test_tool_calls`; no `policy` | nested `if let Some` with no else |
| `tool_blocklist` | `blocked` absent, `[]`, non-array, or non-string entries | collapsed to an empty blocklist, which admits every call |
| `trace_must_call_tool` | `min_calls: 0` | `count < 0` is never true for an unsigned count |
| `trace_tool_sequence` | empty `sequence` + `allow_other_tools: true` | every trace contains the empty subsequence |

Each now reports `E_ASSERT_INEFFECTIVE`, naming the variant and the responsible field, value-free, with a fix step. A non-empty diagnostics list sets `TestStatus::Fail` (`runner_next/assertions.rs:19-20`), so this is fail-closed.

### And one inversion, which is worse than any no-op

`expect` was compared by exact string equality to `"pass"` at three sites. Every other spelling — `Pass`, `PASS`, `passes`, `true` — silently selected *expect failure*, so a config reading "this must be allowed" went green precisely when the policy rejected it. A no-op stops checking; an inversion checks the opposite. The three comparisons are now one `expected_pass` parser accepting `pass` and `fail` and rejecting the rest as a config error — one function rather than three comparisons, per the rule added in #1957.

### What is deliberately not changed

An empty `sequence` with `allow_other_tools: false` stays a real constraint — it asserts the trace made no named tool call — and a control test pins that distinction so the fix does not over-reject.

## Type of Change
- [x] Bug fix

## Verification

Worktree `/private/tmp/assay-1949-vacuous`, own `CARGO_TARGET_DIR`, branch head `a48f6ae1`.

- [x] Sixteen tests in `crates/assay-core/tests/assertions_vacuity.rs` — ten for the shapes above, plus **four controls**: a unit-mode `args_valid` missing its policy stays diagnosed under its existing code; an effective assertion that holds stays silent; a genuine violation stays a failure rather than being reclassified; an empty exact sequence stays a real constraint. Every new test failed before its change and passes after.
- [x] `cargo test -p assay-core` — **1154 passed, 0 failed**
- [x] `cargo clippy -p assay-core --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] **Mutation tested, seven bites, all killed.** One survived on the first pass: removing the typed-`test_trace` guard left every test green, because the case fell through to the `test_trace_raw` check which also returns `E_ASSERT_INEFFECTIVE`. The assertion only checked the code, not which field was blamed — and that difference is the whole value of the diagnostic. `933001c3` strengthens it to pin the `field`, and the bite now fails.
- [ ] Demo suite — not applicable

## Checklist
- [x] I have updated the documentation accordingly — no user-facing docs describe these shapes; two existing docs mismatches are recorded as follow-ups below rather than fixed here.

## Follow-ups, deliberately out of scope

Found during an exhaustive walk of the seven variants against their evaluator, recorded so they are not lost:

- `TraceAssertion` has no `deny_unknown_fields`, so a misspelled `test_args` / `min_calls` is dropped at parse time and turns a real check into a no-op. Parse-time change with a compatibility decision attached.
- `evaluate_sequence` uses `is_match`, which is unanchored: `regex: "search"` is satisfied by any superstring. A semantics question, not vacuity.
- Fallback 1 in `agent_assertions/mod.rs` fires on **any** store error, including ambiguous-episode and SQL errors, and is checked before Fallback 2 — so a unit-test-shaped list never consults a recorded trace.
- `docs/architecture/agents.md:114-116` advertises `max_calls: 0`; the field does not exist and is silently ignored.
- `docs/metrics/index.md:12` promises `trace_must_call_tool` counts a **successful** call; `ToolCallRow` has no status column, so errored calls count.
- Layers 2 and 3 of the design note on #1949: an exercised/not-exercised status, and narrow static rejection reusing `validate_args_policy_value` rather than approximating it.

## Non-claims

This does not make the `assertions:` surface complete or safe — it removes ten shapes that could not fail and one that failed backwards. It does not address configuration-level vacuity that only a trace can reveal, which is what layer 2 is for. Verified on one tree; no claim about other branches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added clear diagnostics for ineffective assertion configurations, including missing inputs, empty sequences, zero call minimums, and unusable policies.
  * Invalid expectation values now report configuration errors instead of silently reversing pass/fail behavior.
  * Sequence checks now require a valid policy pattern before evaluation.
  * Assertion inputs are validated explicitly to prevent misleading successful results.

* **Tests**
  * Added comprehensive coverage for invalid, ineffective, passing, and failing assertion scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ All notable changes to this project will be documented in this file.
   against the unread `test_trace` field or carrying no usable `regex`, a
   `tool_blocklist` whose `blocked` list was absent, empty, wrongly typed or
   partly unreadable, `min_calls: 0`, an empty `sequence` with
-  `allow_other_tools: true`, and others — and each now reports
+  `allow_other_tools: true`, an empty `tool` name, a `max` at the largest
+  representable bound, and others — and each now reports
   `E_ASSERT_INEFFECTIVE` naming the responsible field. Separately, `expect` was
   compared by exact equality to `"pass"` at three sites, so any other spelling
   silently selected *expect failure* and inverted the assertion; unrecognized
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- An `assertions:` entry that cannot check anything no longer reports as a pass.
+  Thirteen shapes evaluated to nothing or to a check that could not fail for any
+  input — an `args_valid` without `test_args`, a `sequence_valid` written
+  against the unread `test_trace` field or carrying no usable `regex`, a
+  `tool_blocklist` whose `blocked` list was absent, empty, wrongly typed or
+  partly unreadable, `min_calls: 0`, an empty `sequence` with
+  `allow_other_tools: true`, and others — and each now reports
+  `E_ASSERT_INEFFECTIVE` naming the responsible field. Separately, `expect` was
+  compared by exact equality to `"pass"` at three sites, so any other spelling
+  silently selected *expect failure* and inverted the assertion; unrecognized
+  values are now rejected. **This turns previously green configurations red, by
+  design: they were green because they checked nothing** (#1949).
+
 ### Added
 - `assay.tool_decision_surface.v0` records now carry a typed `correlation`
   basis: a valid W3C `traceparent` from `_meta` is retained verbatim as a
--- a/crates/assay-core/src/agent_assertions/matchers.rs
+++ b/crates/assay-core/src/agent_assertions/matchers.rs
@@ -25,6 +25,17 @@ fn check_one(graph: &EpisodeGraph, a: &TraceAssertion) -> Option<Diagnostic> {
                 .filter(|t| t.tool_name.as_deref() == Some(tool.as_str()))
                 .count();
             let min = min_calls.unwrap_or(1);
+            if min == 0 {
+                // `count < 0` is never true for an unsigned count, so no trace can fail this.
+                return Some(ineffective(
+                    "trace_must_call_tool",
+                    "min_calls",
+                    "`min_calls: 0` is satisfied by every trace, including one where the tool is \
+                     never called.",
+                    "Use `min_calls: 1` or higher, or express \"never called\" with \
+                     `trace_must_not_call_tool`.",
+                ));
+            }
             if (actual as u32) < min {
                 return Some(make_diag(
                     "E_TRACE_ASSERT_FAIL",
@@ -63,6 +74,19 @@ fn check_one(graph: &EpisodeGraph, a: &TraceAssertion) -> Option<Diagnostic> {
             allow_other_tools,
         } => {
             if *allow_other_tools {
+                // An empty subsequence has nothing to look for, so every trace contains it.
+                // The exact form is different: an empty sequence there asserts the trace made no
+                // named tool call at all, which is a real constraint.
+                if sequence.is_empty() {
+                    return Some(ineffective(
+                        "trace_tool_sequence",
+                        "sequence",
+                        "an empty `sequence` with `allow_other_tools: true` is contained in every \
+                         trace, so the assertion cannot fail.",
+                        "Name the tools the trace must contain, or set \
+                         `allow_other_tools: false` to assert that no tool was called.",
+                    ));
+                }
                 // Subsequence check
                 if let Err(msg) = check_subsequence(&graph.tool_calls, sequence) {
                     return Some(make_diag(
@@ -139,7 +163,10 @@ fn check_one(graph: &EpisodeGraph, a: &TraceAssertion) -> Option<Diagnostic> {
                 let policy_map = serde_json::json!({ tool: schema });
 
                 let verdict = crate::policy_engine::evaluate_tool_args(&policy_map, tool, args);
-                let expected_pass = expect.as_deref().unwrap_or("pass") == "pass";
+                let expected_pass = match expected_pass(expect, "args_valid") {
+                    Ok(v) => v,
+                    Err(d) => return Some(*d),
+                };
                 let actual_pass = verdict.status == crate::policy_engine::VerdictStatus::Allowed;
 
                 if expected_pass != actual_pass {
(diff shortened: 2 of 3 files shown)


Rul1an:
## Adversarial review, and what it changed

An independent adversarial pass was run against `a48f6ae1` with the brief to refute the work. It landed four real hits, three of them fixed in `b874370a` and two corrections to what this PR claimed. Recording all of it, including what it found sound.

### Fixed — the headline claim was pinned at one of three sites

The `expect` inversion fix touches `args_valid`, `sequence_valid`, and `tool_blocklist`. **The tests built only an `ArgsValid`.** Restoring the pre-fix comparison at the other two sites left all 16 vacuity tests and all 1154 crate tests green:

```rust
let expected_pass = expected_pass(expect, "sequence_valid").unwrap_or(false);
```

That is the exact defect this PR calls worse than any no-op, unguarded for two of three variants, behind a claim that it was fixed. The test now runs every spelling against one well-formed assertion per call site, and that mutation fails.

This is the third instance of one gap in this branch: an assertion pinning that a fix exists without pinning its coverage. The first was caught by the author's own mutation testing, the second and third by this review. The pattern is worth stating plainly — checking a diagnostic's code, or one variant, is not checking the fix.

### Fixed — three more shapes that could not fail

All inside functions this branch had already rewritten:

- `tool_blocklist` with an empty `test_tool_calls` — the structural twin of `blocked: []` from the other side. Rejecting one and not the other was arbitrary.
- `blocked` carrying a non-string entry — the wholly-unusable case was caught, the partial one was not, leaving an assertion that looks complete and checks less.
- `test_trace_raw` carrying an entry keyed on neither `tool` nor `tool_name` — dropped silently, shortening the sequence under test.

Also the `blocked` diagnostic said "carries no `blocked` list" for a policy carrying one of the wrong type. Distinguished now, and the message is pinned rather than only the field — that distinction survived a mutation until it was.

### Correction to this PR's stated scope

Two things the body got wrong, both now stated here rather than quietly amended:

**Issue #1949 step 3 asks for rejection *before execution*. This does not do that.** Every guard fires inside `check_one`, after the provider call and after retries. `assay validate` still passes these configs untouched — `validate/mod.rs:302-305` exempts any test with a non-empty `assertions:` list, and nothing re-checks them. So a suite with a vacuous assertion validates clean, burns provider calls, and only then reports a config defect. That belongs to layer 3 and is not delivered here.

**The body says layer 3 is deferred wholesale, but four of its six items landed** (`min_calls: 0`, empty sequence, `blocked: []`, missing `regex`). A reader seeing four of six reasonably infers the set is complete. It is not. Still open, each verified to produce zero diagnostics:

- `trace_must_not_call_tool` with `tool: ""` — no row ever carries an empty name, so it can never fail. Named first in #1949's own list.
- `trace_max_steps` at `u32::MAX` — `count as u32 > max` is never true. `TraceMaxSteps` is absent from this PR's table entirely.
- `trace_must_call_tool` with `tool: ""` — the mirror defect, permanently red rather than permanently green.

### Raised, deliberately not changed here

- **`E_ASSERT_INEFFECTIVE` → `TestStatus::Fail` conflates a config defect with agent misbehaviour.** It reaches SARIF at level `error` identically to a behavioural failure and counts into the fail rate; `TestStatus::Error` exists and is what "your config is wrong" maps to elsewhere. This PR extends a pre-existing choice — `E_CONFIG_ERROR` at `matchers.rs:152` already did `Fail` — rather than introducing it, so changing it is its own decision. Verified *not* corrupted: baselines read `details["metrics"]`, and assertion diagnostics land in `details["assertions"]`.
- **An explicit `regex: ".*"` is vacuous by this PR's own stated criterion and is not caught.** `tests/fp_suite.yaml:183` ships one. An explicit `.*` is arguably a deliberate author choice where a missing key is not, but the comment justifying the rejection does not draw that line, so the boundary needs a better reason than the one written.

### What the review attacked and found sound

- No new guard over-rejects; no legitimate config was constructible that any of them refuses.
- The `allow_other_tools: false` reasoning holds — an empty exact sequence is a real constraint, and the control test pins it.
- The `expected_pass` tightening breaks nothing in-tree: a sweep of every `.yaml`/`.json`/`.py`/`.rs` reaching `TraceAssertion` found `pass`/`fail` only. Worth noting that neither config carrying `assertions:` is referenced from CI, so nothing would have caught it if it were.

### One caveat on the numbers

Inverting `blocked.contains(t)` — a total inversion of `tool_blocklist` semantics — also leaves all tests green. That is pre-existing untested code rather than something this branch introduced, but it means "full crate suite green" carries less weight than it reads, here and generally.

Head is now `b874370a`: 20 tests, full crate 1158 pass, clippy `-D warnings` clean, and a CHANGELOG entry recording that this turns previously green configurations red.

Rul1an:
## The three open shapes are closed — `97bcf99f`

The three the adversarial review found enumerated in #1949 and neither fixed nor listed.

**An empty `tool` names nothing**, and no recorded call carries an empty name. Which way that breaks depends on polarity: `trace_must_not_call_tool` can then never fail, `trace_must_call_tool` can never be satisfied. One config mistake, so both report through one helper rather than two codes for one typo.

The permanently *failing* side matters as much as the permanently passing one. Left as `E_TRACE_ASSERT_FAIL` it sends the author to look at their agent for a defect that is in their config — the same misdirection this branch already refuses when it insists a diagnostic name the right field.

**`trace_max_steps` at `u32::MAX`** cannot be exceeded by any step count. Only the ceiling is refused. A merely large bound like `100_000` is a real constraint whose outcome depends on the trace, and refusing those is exactly the over-eager vacuity detection the model-checking literature warns will earn a suppression. A control test pins that boundary so the line cannot drift.

While in there: the step comparison cast the count down to `u32`, so a count above 2^32 would wrap and could land under the bound. It compares in `usize` now.

### One mutation survives, by construction

Reverting that cast — `count > *max as usize` back to `count as u32 > *max` — leaves all 25 tests green, and no test can close it: distinguishing the two requires a trace with more than four billion steps. Recording it rather than leaving the bite list looking complete. It is a latent wrong answer that is unreachable in practice, fixed on those terms.

The other four bites on this commit are killed: the empty-tool guard, each of its two call sites, and the ceiling guard.

Head `97bcf99f`: **25 tests**, full crate **1163 pass**, clippy `-D warnings` clean.

### Still deferred, and now the list is honest

Layer 3's remaining item is the one that matters: rejection **before execution**. Everything here fires inside `check_one`, after the provider call. `assay validate` still passes these configs untouched because `validate/mod.rs:302-305` exempts any test carrying a non-empty `assertions:` list. That exemption tests for non-empty, not for effective, and closing it is what makes #1949's step 3 true rather than nearly true.

Also still open and unchanged: the `Fail`-versus-`Error` question for `E_ASSERT_INEFFECTIVE`, and whether an explicit `regex: ".*"` should be refused on the same criterion as an absent one.

PR #1972, opened 2026-08-04
SARIF has a ruling on a reporting cap, not a gap

Rul1an:
## The retracted claim

The module doc of `crates/assay-evidence/tests/lint_truncation_disclosure.rs` still stated that "SARIF 2.1.0 has no construct for a reporting cap". That claim was already retracted in `crates/assay-evidence/src/lint/sarif.rs`, and the retraction is the *stronger* reading, not a softer one:

> 3.14.23 is normative: `results` "SHALL be present and SHALL contain all results detected by the tool".

So a configured `max_results` puts this producer out of conformance whenever it fires. "No construct exists" would excuse the cap as something the format never addressed; the format addressed it and ruled against us.

This mattered beyond tidiness: the file was offered to a third party as a conformance fixture on aliksir/claude-code-skill-security-check#24, and a fixture that softens its own producer's conformance status is worse than no fixture.

The rest of the module doc's argument is kept intact — Appendix I is informative and its three conditions each describe a tool that failed to *analyse* rather than one that reported less than it found; the notification stays at `warning` (justified by 3.58.6) and deliberately below Appendix I's `error` gate, since 3.20.21 makes an error-level notification mean the run failed.

## The key mismatch

The truncation pair travelled under two names across the two carriers:

| Carrier | Cap | Count |
|---|---|---|
| `toolExecutionNotifications[].properties` | `appliedCap` | `droppedCount` |
| `run.properties` | `appliedCap` | `truncatedCount` |

A consumer reading both saw two names for one number, with no way to tell whether the difference in name meant a difference in meaning. This was disclosed publicly as a known defect in the same comment.

Both carriers now build the bag from one `truncation_properties()`, so agreement is structural rather than two edits away from breaking again. `truncatedCount` survives because it is the published name — `docs/architecture/SPEC-Pack-Engine-v1.md` and the CHANGELOG both carry it, and nothing documented `droppedCount`.

**Wire change**: any consumer already parsing `droppedCount` breaks. That is the intended resolution of the disclosed defect, but it is a rename, not an addition.

The new parity test asserts whole-map equality rather than key-by-key, because key-by-key would pass while one carrier quietly grew a field the other lacked — exactly how this defect arose.

## Spec

`SPEC-Pack-Engine-v1.md` documented only `run.properties.truncated`/`truncatedCount`. It now also names `appliedCap`, records that the notification mirrors the same keys, and states plainly that the cap is a mitigation for a consumer limit rather than conforming behaviour.

## Verification

Run on `c0caf0f7`, worktree `heuristic-fermat-5432c4`, stable toolchain:

- `cargo test -p assay-evidence --test lint_truncation_disclosure` — 8/8 pass
- `cargo test -p assay-evidence` — full crate suite green, 0 failures
- `cargo clippy -p assay-evidence --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean (pre-commit hooks passed)

`grep -rn droppedCount` over the repo returns only the three prose mentions that record the rename.


Rul1an:
Auto-merge disabled on this, and the reason is outside the repo rather than in the diff. The doc-retraction half is right and urgent; the rename needs to go the other way or wait.

**The rename diverges from a vocabulary being settled right now, and this emitter is cited in it.** `aliksir/claude-code-skill-security-check#24` is an active RFC pinning exactly this disclosure into a SARIF envelope. The proposed shape is `appliedCap` + `droppedCount`, in a code block, and it names both emitters in those terms: "@Rul1an's scanner caps the result array: `droppedCount` is findings never serialised." The second emitter, `prompt-defense-audit`, carries `droppedCount` too.

I confirmed those keys publicly on that thread about ninety minutes ago (issuecomment-5176559205), and disclosed this very divergence as ours, with a commitment attached:

> I would rather fix it against a written rule than against my own memory of which name I used where.

Renaming to `truncatedCount` now is fixing it against our own memory. If the RFC then pins `droppedCount`, which is the shape on the table, we rename twice and the public comment was wrong in between.

**Why "it's the published name" is the right criterion but the wrong scope.** `truncatedCount` is published in `SPEC-Pack-Engine-v1.md` and the changelog, so for `run.properties` it wins. The notification is a different audience: it is where a consumer following the OWASP agentic-skills rule reads, and where the cross-emitter vocabulary is being fixed. Two carriers, two readerships. That reframes the divergence from an accident into a decision, which is the part that was genuinely missing.

**One piece of context worth having for either resolution.** SARIF property-bag names are flat. Making them hierarchical was proposed as [oasis-tcs/sarif-spec#181](https://github.com/oasis-tcs/sarif-spec/issues/181) in 2018, with `semmle/query-path` as the example, and closed with no discussion, so 2.1.0 requires nothing. Every bare key here shares one namespace with every other producer's, which is how one fact acquired two names in the first place and why an RFC pinning bare keys inherits the same exposure.

**Suggested split.** Land the retraction fix and the single-builder structure as they are, since the fixture was offered to a third party and shipping the retracted claim is the live risk. Keep the notification on `appliedCap` / `droppedCount`, keep `run.properties` on its published names, and assert in the parity test that the two carriers agree on the *values* rather than on the key strings. Then the divergence is documented intent, and if the RFC pins something else we change one function.

Rul1an:
Done on this branch rather than as a separate change, since the worktree was idle and splitting it would have forked the same file.

The retraction fix is untouched and still lands. What moved is the direction of the rename: the notification keeps `appliedCap` / `droppedCount`, `run.properties` keeps `truncated` / `truncatedCount` / `appliedCap`, and the parity test now asserts the two carriers disclose the same **cap and count** rather than the same **strings**. It also asserts each carrier keeps to its own vocabulary, so a later edit cannot quietly unify them and surface on someone else's consumer.

The spec passage at `SPEC-Pack-Engine-v1.md:476` says the same, and the conformance checklist gained the notification line.

Gates: 8/8 truncation suite, 611 assay-evidence tests, clippy clean under `-D warnings`, fmt clean. Branch merged up from main, so it is no longer BEHIND.

Issue #1979, opened 2026-08-04
Productization: ship a standalone cross-language verifier with a frozen exit-code contract


Rul1an:
Parent: #1973 · Relates to #1977, #1975, #1840

## Problem

Assay already has the crypto and contract foundations for offline verification, but not yet a **standalone, cross-language verifier surface** that a third party can run without building the Assay workspace. This is the single largest adoptability gap surfaced by benchmarking against the closest open comparable (PipeLab/Pipelock), which ships four independent verifiers (Go, TypeScript, Rust, Python), a wasm surface, and a fixed exit-code contract against a public corpus.

**This is not greenfield.** Already on `main`:

- `crates/assay-cli/src/exit_codes.rs` — a coarse, stable exit-code module with an `ExitCodeVersion` enum and the `ASSAY_EXIT_CODES` env selector, already wired through `run`, `ci`, `bundle`, and other commands.
- `crates/assay-canonical` — RFC 8785 (JCS) canonicalization with duplicate-key rejection (`parse_strict`), `content_id` (`sha256:` + hex), and profile binding (`semantic_digest`, `ensure_supported_profile`) that fails closed on unknown profiles.
- `crates/assay-canonical/tests/vectors/rfc8785.json` + `tests/rfc8785_conformance.rs` — 31 RFC 8785 edge-case vectors, cross-validated against an independent implementation in another language, each carrying the property it pins.
- `crates/assay-evidence` — bundle verification, `attestation.rs` (Sigstore DSSE / Rekor v2 / in-toto lanes), `crypto/id.rs`, Trust Basis / Trust Card.
- `docs/profiles/privileged-mcp-action/v0.md` + `conformance/privileged-mcp-action-v0/` — the versioned normative profile and 14-vector corpus (#1977).
- `.github/actions/privileged-mcp-action-conformance` — reusable composite action.

The gap is that verification today effectively assumes the Assay CLI built from this workspace. There is no published, minimal-dependency verifier a relying party can install per language, and the exit-code contract is not yet frozen and documented as an external interface.

## Product outcome

A relying party — auditor, SIEM sidecar, CI job, or a different vendor — can verify an Assay evidence bundle offline, in their own language, against a pinned public key and profile, using a documented exit-code contract, without building the Assay workspace or trusting an Assay-hosted service.

### A. Freeze and document the exit-code contract as an external interface

**First reconcile with the one that already exists.** `Rul1an/Assay-Harness` ships a documented,
[...]

PR #1982, opened 2026-08-04
Check the canonicalizer against RFC 8785, not against itself

Rul1an:
Every content id, mandate id and bundle run root is a sha256 over the bytes `assay-canonical` emits. Nothing checked those bytes against the standard they claim to follow.

A divergence would not have failed any existing test: the reference producer and the reference verifier would have gone on agreeing with each other, and only an independent implementation would ever have noticed — which is the one property the evidence contract rests on. `serde_jcs` is pinned to an exact version because its output moves those digests; that pin catches an unnoticed upgrade, it says nothing about whether the pinned version is right.

## What this adds

31 edge-case vectors, each carrying a `pins` line stating the property it exists to hold, and four tests over them.

Expected bytes were **cross-validated against an independent implementation in another language**, built from the primitives RFC 8785 defers to: ECMAScript `Number::toString` for numbers (§3.2.2.3), ECMAScript JSON escaping for strings (§3.2.2.2), and UTF-16 code-unit key ordering (§3.2.3). All 31 agree byte-for-byte.

Coverage includes number reformatting (`4.50` → `4.5`, `-0` → `0`, `1.0` → `1`), both ES6 exponent boundaries (`1e20` in full, `1e21` in exponent form; `1e-6` plain, `1e-7` exponent), the RFC's own precision example, and the absence of Unicode normalization.

## The decisive vector

`keyorder_utf16_vs_codepoint`. Almost any implementation looks correct on ordinary keys — including one that sorts by code point or by UTF-8 bytes, since `serde_json::Map` also sorts. Only a non-BMP key meeting one in `U+E000..U+FFFF` separates the three orderings.

Two further properties are pinned because a canonicalizer swap could break them silently: distinct spellings of one value must agree, and an integer must canonicalize identically whether it arrived as a Rust integer or as parsed JSON.

## Verification

Verified by mutation rather than by observing green: substituting plain `serde_json` for the pinned canonicalizer fails **8 of the 31** vectors and all four tests — and the only ordering vector among those 8 is the decisive one, which is the corpus confirming its own claim about itself.

`serde_json`'s `arbitrary_precision` feature is not enabled anywhere in the workspace, so that emission path cannot occur.

Expectations must never be regenerated from this crate; the vector file's `_about.expected_provenance` says so, because a corpus derived from the code under test is a snapshot rather than a conformance check.

**No production code changes, so no digest moves.** Groundwork for #1979.

### DIFF
--- a/crates/assay-canonical/tests/rfc8785_conformance.rs
+++ b/crates/assay-canonical/tests/rfc8785_conformance.rs
@@ -1,0 +1,183 @@
+//! RFC 8785 (JCS) conformance for this crate's canonicalizer.
+//!
+//! Every content id, mandate id and bundle run root in the workspace is a sha256 over the bytes
+//! this crate emits. If those bytes diverge from RFC 8785 on any edge case, the digests are
+//! irreproducible by any other implementation — and reproducibility by another implementation is
+//! the entire property independent verification rests on. A divergence would not fail any
+//! existing test: the reference producer and the reference verifier would simply agree with each
+//! other, which is exactly the failure this corpus exists to make visible.
+//!
+//! `serde_jcs` is pinned to an exact version in `Cargo.toml` because its byte output moves those
+//! digests. That pin protects against an unnoticed upgrade; it says nothing about whether the
+//! pinned version is correct. This does.
+//!
+//! Expected bytes were cross-validated against an independent implementation in another language,
+//! built from the primitives RFC 8785 defers to (see `_about.expected_provenance` in the vector
+//! file). When adding a vector, derive its expectation from the RFC or from a second conforming
+//! implementation — never from this crate, or the check becomes a snapshot of whatever we already
+//! do and stops being a conformance check at all.
+
+use std::collections::BTreeMap;
+
+#[derive(serde::Deserialize)]
+struct Vector {
+    input: String,
+    expected: String,
+    #[allow(dead_code)]
+    pins: String,
+}
+
+fn vectors() -> BTreeMap<String, Vector> {
+    let path = std::path::Path::new(env!("CARGO_MANIFEST_DIR")).join("tests/vectors/rfc8785.json");
+    let raw = std::fs::read_to_string(&path)
+        .unwrap_or_else(|e| panic!("cannot read {}: {e}", path.display()));
+    let mut all: BTreeMap<String, serde_json::Value> =
+        serde_json::from_str(&raw).expect("vector file is not valid JSON");
+    all.remove("_about");
+
+    all.into_iter()
+        .map(|(name, v)| {
+            let vector: Vector = serde_json::from_value(v)
+                .unwrap_or_else(|e| panic!("vector `{name}` has the wrong shape: {e}"));
+            (name, vector)
+        })
+        .collect()
+}
+
+#[test]
+fn canonical_bytes_match_rfc8785_for_every_vector() {
+    let vectors = vectors();
+    assert!(
+        vectors.len() >= 31,
+        "the corpus shrank to {} vectors; removing coverage here is a silent loss of the \
+         cross-implementation guarantee",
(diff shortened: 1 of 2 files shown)


Rul1an:
## Review — strong, mergeable after rebase; one Major-nit, two Minor

Read the full corpus and all four tests. This is the measured grounding for #1979 decision 3, and it is well built — the `DECISIVE` UTF-16 vector and the anti-circular `_about.expected_provenance` are exactly right. Findings below.

### Major (naming, with a real edge) — the "past double precision" case is mislabelled in `distinct_spellings_of_one_value_canonicalize_identically`

```rust
("past double precision", r#"{"a":9007199254740993}"#, r#"{"a":9007199254740992}"#),
```

These are **two different integer values** that happen to round to the same `f64`, not two spellings of one value. The test's expectation (they canonicalize identically) is correct, but the framing is misleading: it belongs with precision-loss, not with "distinct spellings." Two concrete reasons it matters:
- If `serde_json`'s `arbitrary_precision` were ever enabled (correctly noted as off today), this pair would diverge and the test would fail — correctly — but under a heading that says "one value," which would send the reader looking in the wrong place.
- `integers_canonicalize_identically_from_a_struct_and_from_a_parsed_value` already owns the precision-loss framing (`9007199254740993` is in its list). This pair is a duplicate concern in the wrong test.

Suggested: move it to the integer test (or rename the case to "two integers that round to one double"), keeping the other rows — which are genuinely one value spelled two ways — under `distinct_spellings`.

### Minor — the mutation claim is not reproducible from the committed tests
The PR body's strongest evidence is "substituting plain `serde_json` fails 8 of 31." That mutation is manual and not in the tree, so the claim degrades over time. Consider a short comment in `rfc8785_conformance.rs` stating exactly how to reproduce it (swap `assay_canonical::jcs::to_vec` for `serde_json::to_vec` and expect ≥8 failures incl. `keyorder_utf16_vs_codepoint`), so the "corpus is load-bearing" property stays checkable rather than being a one-time observation.

### Minor — `>= 31` guards shrinkage but not undocumented growth
`assert!(vectors.len() >= 31)` correctly prevents silent coverage loss, but nothing requires a newly added vector to be justified in `_about.expected_provenance`. Low priority; the provenance discipline is documented, just not enforced.

### What's excellent (keep)
- `keyorder_utf16_vs_codepoint` with the explicit `DECISIVE` pin — the one vector where UTF-16, code-point, and UTF-8 byte order disagree.
- `the_two_properties_a_replacement_would_break` isolating UTF-16 ordering and non-NFC into their own asserts so a failure names its cause.
- `integers_canonicalize_identically_from_a_struct_and_from_a_parsed_value` — the struct-vs-parsed-`Value` route parity is the subtle catch I flagged on #1979; good that it is pinned with `u64::MAX` and `2^53 ± 1`.
- "No production code changes, so no digest moves" — correct, and the reason this is safe to land ahead of the verifier work.

Net: address the naming Major (it is a 2-line move) and this is ready. Solid groundwork for #1979.

Rul1an:
Both taken in `261b552c`.

The naming Major was right, including the reason it matters: a future divergence under a heading that says "one value" sends the reader to the wrong cause. Since the pair is already covered where the framing fits, it is dropped from `distinct_spellings` with a note explaining why rather than moved and duplicated.

The Minor about the mutation claim is the better of the two findings. Evidence that lives only in a pull request description decays the moment the page stops being read, so the reproduction is now in the file: swap the canonicalizer for plain `serde_json`, expect at least 8 failures, expect `keyorder_utf16_vs_codepoint` among them. Re-ran it after writing it down rather than quoting the earlier observation — 8 of 31, decisive vector included.

The growth-guard Minor is noted and not taken. Enforcing that a new vector documents its provenance would mean parsing prose in a test, which is a weaker guarantee than the rule already written at the top of both files.

PR #1983, opened 2026-08-04
A suite that carried assertions was swept without reading them

Rul1an:
Item 2 of #1949.

`check_vacuous_expected` exists so `assay validate` can sweep a suite for always-green tests without running it. It tested whether a test carried a **non-empty** `assertions:` list and skipped it — non-empty, not effective — and nothing examined the assertions afterwards.

So one assertion that could not fail cleared both gates in a single move: the `expected:` check was skipped because assertions existed, and the assertions were never looked at. The sweep reported a suite as clean while stepping over exactly the case it exists to find.

## One definition of "cannot fail", not two

The obvious fix — have the sweep call the evaluator's effectiveness logic — was not available: `check_one` takes an `&EpisodeGraph`. Following it literally would have produced a second implementation of "cannot fail", which is the drift this repo has a rule against.

So the checks were mapped first. Every ineffectiveness check in `matchers.rs` is a pure function of the assertion's own fields: the `ArgsValid`, `SequenceValid` and `ToolBlocklist` arms touch the graph zero times, and in the other four arms every guard sits ahead of the first graph read.

`matchers::ineffective_reason(&TraceAssertion) -> Option<Diagnostic>` therefore runs the *real* checker against an empty episode and keeps only the diagnostics decided before any trace was consulted. The static sweep reaches the evaluator's conclusions by running the evaluator, not by restating it.

The filter is exact rather than approximate, for a structural reason: every `E_ASSERT_INEFFECTIVE` and `E_CONFIG_ERROR` is returned from a branch reading only configuration, while every branch that reads the graph returns a different code.

## The invariant

`does_not_flag_an_assertion_that_merely_fails_for_a_trace` pins that split. An assertion that constrains something and simply does not hold is not a config defect, and reporting it here would make `assay validate` refuse working suites — the over-eager detection that earns a suppression and takes the real findings down with it.

An unrecognized `expect` spelling comes along, and matters more than the vacuous shapes: a no-op stops checking, an inversion checks the opposite.

## Verification

Verified by mutation, in both directions:

- restoring the blind skip → **three tests fail**, so the sweep is load-bearing;
- widening the filter to admit every diagnostic → **the invariant test fails**, so the filter is what prevents false positives rather than an accident of the inputs.

The pre-existing `does_not_flag_when_assertions_present` passes unchanged as the positive control.

Rejecting at load for every command (item 1) is the remaining decision, and is now cheaper: the predicate it needs exists and is public.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,16 @@ All notable changes to this project will be documented in this file.
   opposite. The catalogue is rewritten from the enum, and a test now loads every
   fenced example, requires every shipped variant to be documented, and requires
   every type listed as unimplemented to actually be rejected (#1960).
+- `assay validate` now sweeps `assertions:` instead of stepping over them. The
+  exemption tested whether a test carried a **non-empty** assertion list, not an
+  **effective** one, and nothing examined the assertions afterwards — so a single
+  assertion that could not fail cleared both gates at once: the `expected:` check
+  was skipped because assertions existed, and the assertions were never looked
+  at. A suite could be swept clean while asserting nothing. Effectiveness is
+  decided by the same code the evaluator runs, so there is one definition of
+  "cannot fail" rather than a static one and a runtime one that drift. An
+  unrecognized `expect` spelling — which selected *expect failure* and inverted
+  the assertion — is now caught in the same sweep (#1949).
 - An `assertions:` entry that cannot check anything no longer reports as a pass.
   Thirteen shapes evaluated to nothing or to a check that could not fail for any
   input — an `args_valid` without `test_args`, a `sequence_valid` written
--- a/crates/assay-core/src/agent_assertions/matchers.rs
+++ b/crates/assay-core/src/agent_assertions/matchers.rs
@@ -16,6 +16,39 @@ pub fn evaluate(
     Ok(out)
 }
 
+/// Why this assertion cannot check anything, decided from the configuration alone.
+///
+/// `Some` means no trace could ever make the assertion fail, so running it would report a pass
+/// that carries no information. `None` means the assertion constrains something; whether it then
+/// holds is a question about a trace and is not answered here.
+///
+/// This lets a static sweep — `assay validate`, which reads a config without running it — reach
+/// the same conclusions the evaluator reaches, **without a second definition of "cannot fail"**.
+/// Two implementations of one rule drift; the fix for that is one implementation with two callers.
+/// So this runs the real checker against an empty episode and keeps only the diagnostics that were
+/// decided before any trace was consulted.
+///
+/// That filter is exact rather than approximate, and the reason is a property of the checker:
+/// every `E_ASSERT_INEFFECTIVE` and `E_CONFIG_ERROR` in `check_one` is returned from a branch that
+/// reads only the assertion's own fields, and every branch that reads the graph returns a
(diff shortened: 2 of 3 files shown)


Rul1an:
## Review — excellent, mergeable after rebase; no blockers, one Minor

Read the full diff. This is the single-source extraction described on #1949, and the code backs the claim exactly. The `ineffective_reason` design is the right answer to my earlier "call the existing logic" phrasing — which, as you noted, was not literally possible because `check_one` takes an `&EpisodeGraph`.

### The core is correct and provably exact
`ineffective_reason` runs the real `check_one` against an empty episode and filters on `E_ASSERT_INEFFECTIVE || E_CONFIG_ERROR`. I verified against the `matchers.rs` at head that this filter is exact, not approximate: every `ineffective(...)`/`E_CONFIG_ERROR` branch sits ahead of the first graph read, and every graph-reading branch returns `E_TRACE_ASSERT_FAIL`/`E_POLICY_ASSERT_FAIL`. So the static sweep reaches the evaluator's conclusions by running the evaluator — one definition of "cannot fail," which is the drift-avoidance the repo rule requires.

### The invariant test is the strongest thing here
`does_not_flag_an_assertion_that_merely_fails_for_a_trace` with six cases — including the exact-sequence form and a real policy violation — is what keeps the sweep narrow. Mutation-verified in both directions (restore the blind skip → 3 fail; widen the filter → the invariant fails). This is precisely the "over-eager detection earns a suppression" discipline, enforced rather than asserted.

### Minor — no test that a malformed schema in the sweep degrades cleanly
The doc comment is honest that a valid `args_valid` compiles its schema twice (sweep + evaluation). Good. But there is no test that a **non-compilable** schema, encountered during the sweep, produces a diagnostic rather than a panic. `check_one` routes through `evaluate_tool_args`, which returns `Blocked` rather than panicking, so this is likely fine — but an explicit case ("malformed `policy.schema` in a swept assertion → `E_CONFIG_ERROR`, no panic") would pin it, since the sweep now reaches that code path from a new caller.

### What's excellent (keep)
- Reusing the evaluator's own diagnostic context (variant + field) and only adding `test_id`/`assertion_index` — no duplicated message.
- `flags_a_vacuous_shape_of_every_variant` covering all seven variants — a sweep that reached only the convenient variants would be the same partial-coverage problem it exists to report.
- The `with_fix_step` addition reads well and is value-free.

Net: ready as-is; the Minor is optional hardening. Clean implementation of item 2.

Rul1an:
Minor taken in `15eaca1d`, and measuring it turned up something the review did not ask for.

The clean-degradation half is confirmed: four uncompilable shapes, including an external `$ref` the hermetic compiler refuses, all return rather than unwinding. That matters because a panic would take down `assay validate` on the one input it most needs to survive.

The unasked-for half: the doc comment on `ineffective_reason` justified the duplicated schema compilation by claiming a broken schema is worth learning about from `assay validate` rather than from a run. It is not — an uncompilable schema produces an evaluation-decided code, the filter drops it, and the sweep stays silent. The comment claimed a benefit the code does not deliver.

Corrected rather than implemented. A broken schema is not an assertion that cannot fail, and widening the filter would cost exactly the narrowness `does_not_flag_an_assertion_that_merely_fails_for_a_trace` protects. Both halves of the boundary are now pinned: it does not panic, and it does not report. If a schema-validity check is wanted it belongs in its own pass with its own diagnostic.

PR #1984, opened 2026-08-04
An unknown key in an assertion is a typo, not a feature

### LINKED ISSUE #1960: Every documented `assertions:` example fails to parse, and five documented types do not exist
## Problem

`docs/metrics/index.md` ("Assertion Types (V1)") and `docs/architecture/agents.md` document the inline `assertions:` surface. **Not one of the documented examples parses**, and most of the documented types have no implementation.

Verified at `366c5fad` by deserializing the documented YAML directly into `TraceAssertion`.

### Five documented types do not exist

| Documented in `docs/metrics/index.md` | In `TraceAssertion`? |
|---|---|
| `trace_must_call_tool` | yes |
| `trace_tool_sequence` | yes |
| `trace_no_tool_call` | **no** |
| `trace_no_tool_errors` | **no** |
| `trace_tool_args_match` | **no** |
| `trace_tool_args_schema` | **no** |
| `trace_tool_call_count` | **no** |

The enum (`crates/assay-core/src/agent_assertions/model.rs:5-55`) carries seven variants: `trace_must_call_tool`, `trace_must_not_call_tool`, `trace_tool_sequence`, `trace_max_steps`, `args_valid`, `sequence_valid`, `tool_blocklist`. Five of those seven are absent from the metrics doc, and five of the seven documented types are absent from the enum. The overlap is two.

The doc's own comparison table lists `trace_tool_args_match` as the V1 replacement for the `args_valid` metric, and `trace_no_tool_call` as replacing `tool_blocklist` — but `args_valid` and `tool_blocklist` are the names that actually exist, and the replacements do not.

### The two types that do exist are documented with field names that do not

`trace_must_call_tool` takes `tool`, not `tool_name`. Deserializing the documented example:

```yaml
type: trace_must_call_tool
tool_name: web_search      # docs/architecture/agents.md:110, docs/metrics/index.md:15
min_calls: 1
```
```
REJECTED -> missing field `tool`
```

`trace_tool_sequence` requires `allow_other_tools: bool`, which has no `#[serde(default)]`. `docs/architecture/agents.md:123` writes `mode: loose`; `docs/metrics/index.md:75-77` omits it entirely. Both fail the same way.

### One silent inversion, which is the dangerous one

`docs/architecture/agents.md:113-116` documents this under the comment "Must NOT use a forbidden tool":

```yaml
- type: trace_must_call_tool
  tool_name: delete_database
  max_calls: 0
```

`max_calls` is not a field on any variant. Correct the field name to get past the required-field check and the result is:

```
TraceMustCallTool { tool: "delete_database", min_calls: None }
```

`min_calls` defaults to 1 (`matchers.rs:27`), so the assertion becomes **"delete_database must be called at least once"** — the exact opposite of the documented intent, reported as a pass when the forbidden tool is used and a failure when it is not.

The only reason this is not live today is that the wrong field name makes the example fail loudly first. That is an accident of `tool` having no default, not a safeguard — see #(deny_unknown_fields issue) for the mechanism that drops `max_calls` without comment.

### `docs/metrics/index.md:12` also overstates what is checked

> "Passes if the trace contains at least one **successful** call to the specified tool."

`ToolCallRow` (`crates/assay-core/src/storage/rows.rs:24-32`) has no status or error column — `id`, `step_id`, `episode_id`, `tool_name`, `call_index`, `args`, `result`. `matchers.rs:22-26` counts rows with no success filter, so an errored call satisfies the assertion.

## Required direction

This needs a decision before an edit, which is why it is filed rather than fixed:

1. Decide per documented-but-missing type whether it is **to be built** or **to be removed from the docs**. `trace_no_tool_call` in particular reads as a rename of the existing `trace_must_not_call_tool`, and `trace_tool_args_match` as a rename of `args_valid` — if these were intended renames, the enum never followed.
2. Correct the field names and the required `allow_other_tools` in every remaining example.
3. Remove the `max_calls` example, or implement a `max_calls` field. Do not leave a documented example whose stated intent inverts.
4. Correct the "successful call" claim, or add the outcome column that would make it true.
5. **Add a test that deserializes every YAML example in these documents into `TraceAssertion`.** Prose cannot be kept honest by review alone — this is the same class of gap as #1949, one layer out: nothing checks that the documented surface is the implemented surface.

## Acceptance

- Every `assertions:` example in the documentation deserializes successfully.
- No documented assertion type is absent from `TraceAssertion`.
- A test fails if either of those stops being true.

## Non-claims

This does not assess whether the five missing types are worth building; it records that they are documented and absent. It does not audit documentation outside these two files.

### LINKED ISSUE #1961: TraceAssertion has no deny_unknown_fields, so a misspelled field silently disarms the check
## Problem

`TraceAssertion` (`crates/assay-core/src/agent_assertions/model.rs:3-5`) carries only:

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(tag = "type", rename_all = "snake_case")]
```

No `#[serde(deny_unknown_fields)]`. Serde therefore drops any key it does not recognize, without comment. `#[serde(deny_unknown_fields)]` at `crates/assay-core/src/model/serde.rs:281` applies to the enclosing test case, not to items inside the `assertions:` list.

Where a variant's field has a default, a typo turns a real check into one that cannot fail — silently. Verified at `366c5fad`:

```yaml
type: trace_must_call_tool
tool: delete_database
max_calls: 0            # not a field; dropped without comment
```
```
TraceMustCallTool { tool: "delete_database", min_calls: None }
```

`min_calls` falls back to 1 (`matchers.rs:27`), so the assertion inverts to "must be called at least once". The same applies to a misspelled `min_calls`, `test_args`, `test_trace_raw`, or `test_tool_calls` — each turns its assertion into one of the ineffective shapes #1949 is about, with no signal at any stage.

This is the multiplier under #1949: every shape fixed there is one typo away from returning, and #1960 is a live example of documentation that supplies the typo.

Where the affected field is *required* the failure is loud instead (`missing field \`tool\``), which is why this has not been more visible. That is an accident of which fields happen to have defaults, not a safeguard.

## Required direction

1. Add `#[serde(deny_unknown_fields)]` to `TraceAssertion`.
2. **Decide the compatibility question first.** This is a hard error for configs that currently load, including any that carry a stray or commented-in key. That is the point — an unknown key in an assertion is either a typo or a feature that does not exist — but it is a breaking change to configs that run green today, and it should land with a release note rather than as a quiet tightening.
3. Check whether any first-party fixture, example, or generated config relies on an unknown key being dropped. `assay init` output and `examples/` are the places to look.
4. Consider whether the error should name the near-miss field, since the common case is a one-character difference from a real one.

## Acceptance

- An assertion carrying an unrecognized key is rejected at parse time, for every command that loads a config.
- The diagnostic names the offending key.
- A test covers at least one misspelling per variant that has a defaulted field.
- No first-party fixture or generated config fails the new check.

## Non-claims

This does not address vacuity in well-spelled assertions — that is #1949. It does not claim any current suite is affected; the parse behaviour is verified, the field usage in the wild is not.

Rul1an:
Closes #1961. Closes #1960.

⚠️ **Breaking by design** — see the release note below.

Serde drops an unrecognised key in silence. Where the field the author meant has a default, the assertion falls back to a shape that cannot fail, so a one-character typo turns a real check into a permanent pass with no signal at any stage.

The documented "must NOT use a forbidden tool" example was the worked case: `max_calls: 0` was discarded, `min_calls` defaulted to 1, and the assertion inverted into *"must be called at least once"*. The only reason that was not live is that `tool_name` — also wrong — failed loudly first. That is an accident of which fields happen to have defaults, not a safeguard.

## Why the container, and why that is enough

`deny_unknown_fields` is not a variant attribute; the compiler rejects it as one. So the choice was container-level or hand-rolled.

There is folklore that container-level rejection is unreliable on an internally-tagged enum (serde-rs/serde #2294, #1358). Those defects are about **unit-like variants** and **`flatten`** — every variant here is a struct variant with named fields and none flattens. Rejection was verified on the hardest shape in this enum: `tool_blocklist`, whose fields are all defaulted so nothing but the tag is required. A stray key there is rejected too, and that case has its own test because it is the one the folklore would have been right about.

Two things fall out for free that a hand-rolled allow-set would have needed code for: the message already names the offending key and lists the expected ones, and nested `policy` / `test_args` stay unconstrained — which is required, since policies legitimately carry arbitrary keys.

## The boundary that mattered

`load_config` deserializes through `serde_ignored`, whose whole job is to *collect* unknown keys rather than fail on them, and outside strict mode it only warns. A guard that fired on a bare `from_str` and was swallowed there would be the check looking present while every command stayed permissive. Rejection is asserted at that boundary in **both** strict and non-strict mode.

## The documentation is the other half

None of its examples parsed. Five of seven documented types had no implementation, the two that existed were shown with field names that did not, and the architecture example nested its assertions under a top-level `policies:` block the loader refuses at `configVersion: 1` — so it could not have loaded even with every field corrected. The catalogue also omitted five of the seven *shipped* variants, so the drift ran both ways.

Both documents are rewritten from the enum. Types that do not exist are listed with their nearest shipped equivalent rather than deleted, so a reader holding an old copy learns what to write instead.

Prose cannot be held to an enum by review — nothing failed when they diverged. A test now loads every fenced example through the path a user's config takes, requires every shipped variant to be documented, and requires every type listed as unimplemented to actually be rejected, so that table stays a claim about the parser rather than a note that rots.

## Verification

Mutation-verified: removing the attribute fails 7 of 12 parser tests; restoring `tool_name:` in one example produces `docs/metrics/index.md:24 — documented assertion does not parse: unknown field \`tool_name\`, expected \`tool\` or \`min_calls\``; renaming one variant's section heading fails the catalogue check.

`tests/fp_suite.yaml` and `tests/regex_compatibility.yaml` still load, and no template, example or demo config carries an `assertions:` block.

## Release note

An unknown key in an assertion is now a hard error where it was previously dropped. That is the point — it is either a typo or a feature that does not exist — but it turns previously green configurations red. No first-party fixture or generated config relies on one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Fixed
+- An `assertions:` entry carrying a key the type does not define is now rejected
+  at parse time, naming the offending key, instead of being dropped in silence.
+  Where the intended field had a default the dropped key left a check that could
+  not fail: the documented `max_calls: 0` "must NOT use a forbidden tool"
+  example inverted into "must be called at least once", because `max_calls` was
+  discarded and `min_calls` defaulted to 1. Rejection holds through the config
+  loader, which collects unknown keys elsewhere and outside strict mode only
+  warns about them. **This turns previously green configurations red, by design:
+  an unknown key in an assertion is either a typo or a feature that does not
+  exist** (#1961).
+- Every `assertions:` example in the documentation now loads. None of them did:
+  five of the seven documented types had no implementation, the two that existed
+  were shown with field names that did not, the architecture example nested them
+  under a top-level `policies:` block the loader refuses at `configVersion: 1`,
+  and the example labelled "must NOT use a forbidden tool" asserted the
+  opposite. The catalogue is rewritten from the enum, and a test now loads every
+  fenced example, requires every shipped variant to be documented, and requires
+  every type listed as unimplemented to actually be rejected (#1960).
 - An `assertions:` entry that cannot check anything no longer reports as a pass.
   Thirteen shapes evaluated to nothing or to a check that could not fail for any
   input — an `args_valid` without `test_args`, a `sequence_valid` written
--- a/crates/assay-core/src/agent_assertions/model.rs
+++ b/crates/assay-core/src/agent_assertions/model.rs
@@ -1,7 +1,28 @@
 use serde::{Deserialize, Serialize};
 
+/// An assertion carrying a key this enum does not define is rejected at parse time.
+///
+/// Without `deny_unknown_fields` serde drops the unrecognised key silently, and where the
+/// intended field has a default the assertion falls back to a shape that cannot fail. The
+/// documented `max_calls: 0` "must NOT use a forbidden tool" example is the worked case: the
+/// key is dropped, `min_calls` defaults to 1 in `matchers.rs`, and the assertion inverts into
+/// "must be called at least once" with no signal at any stage (#1961).
+///
+/// `deny_unknown_fields` is applied at the **container**, which is the only place serde accepts
+/// it — it is not a variant attribute, and the compiler rejects it as one. There is folklore
+/// that container-level rejection is unreliable on an internally-tagged enum (serde-rs/serde
+/// #2294, #1358). Those defects are about unit-like variants and `flatten`; every variant here
+/// is a struct variant with named fields and none flattens, and rejection was verified on this
+/// exact shape for the hardest case in it — `tool_blocklist`, whose fields are all defaulted, so
+/// nothing but the tag is required. A stray key there is rejected too. Nested free-form values
+/// (`policy`, `test_args`) are `serde_json::Value` and stay unconstrained, which is intended:
+/// the guard covers the assertion's own field vocabulary, not policy contents.
+///
+/// Keep the guard at the container. Per-variant allow-set validation was considered and is not
+/// needed here; the tests in `tests/assertions_unknown_fields.rs` pin the behaviour that makes
+/// it unnecessary, including the all-defaulted variant.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(tag = "type", rename_all = "snake_case")]
+#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
 pub enum TraceAssertion {
     #[serde(rename = "trace_must_call_tool")]
     TraceMustCallTool {
--- a/crates/assay-core/tests/assertions_unknown_fields.rs
+++ b/crates/assay-core/tests/assertions_unknown_fields.rs
@@ -266,13 +266,65 @@ fn first_party_suites_still_load() {
         .and_then(|p| p.parent())
         .expect("repo root");
 
-    for name in ["tests/fp_suite.yaml", "tests/regex_compatibility.yaml"] {
+    // No `if !path.exists() { continue }`. A missing path would make this test pass without
+    // loading anything, and this test is the only thing standing behind the claim that no
+    // first-party config broke — absence reported as a pass, in the guard against exactly that.
+    let suites = ["tests/fp_suite.yaml", "tests/regex_compatibility.yaml"];
+    let mut loaded = 0;
+    for name in suites {
         let path = repo_root.join(name);
-        if !path.exists() {
-            continue;
-        }
+        assert!(
+            path.exists(),
+            "{name} not found at {}; this test's coverage would otherwise be silently empty",
+            path.display()
+        );
         assay_core::config::load_config(&path, false, false).unwrap_or_else(|e| {
             panic!("{name} no longer loads after the unknown-field guard: {e}")
         });
+        loaded += 1;
+    }
+    assert_eq!(
+        loaded,
+        suites.len(),
+        "not every first-party suite was exercised"
+    );
+}
+
+/// `docs/metrics/index.md` states that `trace_must_call_tool` counts an errored call, because
+/// `ToolCallRow` carries no status or error column. That is a claim about a struct, and the
+/// documentation test next door checks the *examples*, not the prose — so if the row ever gains
+/// an outcome column the sentence would quietly become false.
+///
+/// This makes it rot loudly instead, two ways. The literal below stops compiling if a field is
+/// added, which points a change at this test; and the field-name check catches a rename that
+/// still compiles. If either fires, the doc is now wrong and the assertion may have become able
+/// to distinguish success from failure — a real capability worth documenting rather than a test
+/// to delete.
+#[test]
+fn the_documented_reason_errored_calls_still_count_is_still_true() {
+    let row = serde_json::to_value(assay_core::storage::rows::ToolCallRow {
+        id: 1,
+        step_id: "s".into(),
+        episode_id: "e".into(),
+        tool_name: Some("t".into()),
+        call_index: Some(0),
+        args: None,
+        result: None,
+    })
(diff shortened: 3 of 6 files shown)


Rul1an:
## Review — the best of the four, one Major to close before merge

Read the full diff. This is the strongest PR of the batch: `deny_unknown_fields` is defended with a test on the exact edge case the serde folklore is about, and the config-loader boundary is covered in both strict and non-strict mode — which was the precise "gate looks present while the CLI stays permissive" risk raised on #1949. Two findings.

### Major — `first_party_suites_still_load` can pass green while verifying nothing

```rust
for name in ["tests/fp_suite.yaml", "tests/regex_compatibility.yaml"] {
    let path = repo_root.join(name);
    if !path.exists() {
        continue;   // <-- silent skip
    }
    assay_core::config::load_config(&path, false, false)...
}
```

If either path is ever wrong — a move, a rename, a `CARGO_MANIFEST_DIR` parent-walk that resolves differently in some CI layout — the test **passes without loading anything**. The #1961 acceptance criterion is literally "No first-party fixture or generated config fails the new check," and this test is what claims it, but an absence-as-pass here cannot guarantee it. That is the exact failure class this whole issue family exists to eliminate — worth not reintroducing in its own regression test.

Suggested fix: require that at least one suite was actually exercised, e.g.
```rust
let mut loaded = 0;
for name in [...] {
    let path = repo_root.join(name);
    assert!(path.exists(), "{name} not found at {}; the guard's own coverage is silently empty", path.display());
    assay_core::config::load_config(&path, false, false).unwrap_or_else(...);
    loaded += 1;
}
assert!(loaded >= 1, "no first-party suite was exercised");
```
or drop the existence guard entirely if these files are contractually present.

### Minor — a prose claim about `ToolCallRow` that nothing pins

`docs/metrics/index.md` now states `trace_must_call_tool` counts an errored call as success "because `ToolCallRow` carries no status or error column." That is honest and correct today, but it is a prose claim about a struct definition, and `docs_assertion_examples.rs` validates the *examples*, not this sentence. If `ToolCallRow` later gains a status column the doc silently goes stale. Consider a comment tying that doc line to `crates/assay-core/src/storage/rows.rs`, or a tiny test asserting `ToolCallRow` has no status field so the claim rots loudly.

### What's excellent (keep as-is)
- `tool_blocklist_rejects_unknown_key_despite_having_no_required_field` — testing rejection on the all-defaulted variant is exactly the shape the serde `#2294`/`#1358` defects are about; proving it here is what makes the container-level choice defensible.
- `load_config_rejects_an_unknown_key_inside_an_assertion` asserting **both** `strict` values — this is the load-boundary guard that matters most.
- `docs_assertion_examples.rs` checking both directions (every example parses; every shipped variant is documented; every "not implemented" type is actually rejected) — the anti-drift mechanism that makes the catalogue a claim about the parser rather than prose.
- The `agents.md` fix turning the `max_calls: 0` inversion into `trace_must_not_call_tool` — the most dangerous bug in #1960, correctly repaired.

Net: close the Major (make the first-party-suite test unable to pass empty) and this is ready. Breaking-by-design with a clear release note is the right call for #1961.

Rul1an:
Major closed in `34cad480`. The finding was correct and the framing was the useful part — absence reported as a pass, inside the regression test whose whole job is to rule that out. The path is now asserted and the loaded count is checked, so an empty run fails.

Minor taken too, and slightly further than suggested. The `ToolCallRow` claim is pinned two ways: the struct literal stops compiling if a field is added, and the serialized field names are checked so a rename that still compiles is caught as well.

Both mutation-verified: renaming a fixture path fails with "this test's coverage would otherwise be silently empty", and treating an existing column as outcome-shaped fails naming the field and the doc line that would need updating.

PR #1985, opened 2026-08-04
A release gate that lives in prose is not a gate

Rul1an:
Item 4 of #1949.

That issue declared itself release-blocking for v3.37.0. **v3.37.0 shipped on 2026-08-02 with the issue still open.** The gate did not fail — it lapsed, because the only place it existed was one sentence in an issue body and nothing could read it.

The `release-blocking` label fixed half of that: it made the gate queryable. This is the other half — something that queries it.

## What it does

The release workflow refuses to build if the milestone it is releasing has an open issue carrying `release-blocking`, and names the issues in the failure. It sits in `release-contract`, the job everything else already `needs:`, so a blocked release stops before any artifact is built.

## Two deliberate choices

**No override input.** If a release must go out over an open blocker, the way to do it is to remove the label or move the issue to a later milestone. Both are recorded decisions on the issue itself, which is better provenance than a workflow input nobody will find when they later ask why the gate did not hold.

**A release with no matching milestone is not blocked.** Milestones are optional here, and failing on their absence would push people to stop attaching them — which would cost the gate its only input.

## Verification

Run against live repository state rather than reasoned about:

- `VERSION=3.38.0`, with #1949 open, labelled, and in that milestone → step fails and prints `#1949 Reject nonempty but behaviorally vacuous assertions before v3.37.0`
- `VERSION=3.37.0`, no such milestone → exits clean

`check yaml` and the actionlint hook both pass.

## Companion action already taken

The `v3.38.0` milestone now exists and #1949 is attached to it, so this gate has something to read on the next release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,6 +44,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+      issues: read
     outputs:
       version: ${{ steps.version.outputs.version }}
     steps:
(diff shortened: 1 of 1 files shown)


Rul1an:
## Review — correct gate, but it fails *open* in two places, which is the wrong direction for a gate

This does exactly what the item-4 recommendation asked: a release-workflow step that refuses to build over an open `release-blocking` issue in the milestone, no override input, no-milestone = not blocked. The two design choices are right. But a release gate must fail **closed** on its own errors, and two spots here fail open.

### Major — the milestone existence check fails open on pagination and on API error

```bash
if ! gh api "repos/${GITHUB_REPOSITORY}/milestones?state=all&per_page=100" \
       --jq '.[].title' | grep -Fxq "$milestone"; then
  echo "No milestone named ${milestone}; nothing to gate on."
  exit 0
fi
```

Two ways this lets a blocked release through:

1. **Pagination.** `per_page=100` reads only the first page. A repo that accumulates >100 milestones over its life (this gate is meant to last years) would miss a milestone on page 2 and conclude "nothing to gate on" → `exit 0`. The gate silently disables itself exactly when the project is old enough to have forgotten it exists.
2. **API failure under `set -o pipefail`.** If `gh api` fails (rate limit, transient network), the pipe's non-zero status makes the `if !` branch fire → "nothing to gate on" → `exit 0`. A network blip becomes a green release.

Both are fail-open. For a gate, an error or an uncertain state should block, not release.

Suggested fix — query the milestone directly and separate "not found" from "could not check":
```bash
set -euo pipefail
milestone="v${VERSION}"

# --paginate so a long-lived repo does not outgrow the gate; capture separately so an
# API failure is distinguishable from "not found".
if ! titles="$(gh api --paginate "repos/${GITHUB_REPOSITORY}/milestones?state=all" --jq '.[].title')"; then
  echo "::error::could not list milestones; refusing to release rather than guessing."
  exit 1
fi

if ! grep -Fxq "$milestone" <<<"$titles"; then
  echo "No milestone named ${milestone}; nothing to gate on."
  exit 0
fi
```

### Minor — the blocker query itself has no error guard

```bash
blockers="$(gh issue list --repo ... --json number,title --jq '...')"
```
Under `pipefail` a failed `gh issue list` aborts the step (good, that's fail-closed), but only because `set -e` is on and it is not in an `if`. Worth an explicit comment that this call must stay outside a condition so an API failure blocks rather than being swallowed — otherwise a future refactor that wraps it in `if` would silently reintroduce a fail-open path, the same shape as the milestone check above.

### What's right (keep)
- No override input — removing the label or moving the milestone are recorded decisions on the issue, better provenance than a workflow toggle.
- Living in `release-contract` (the job everything `needs:`) so a blocked release stops before any artifact builds.
- `issues: read` added to permissions, scoped minimally.
- Live-verified both directions (3.38.0 with #1949 open → fails; 3.37.0 no milestone → clean).

Net: fix the two fail-open paths (they are one-line changes) and this turns `release-blocking` from a label into a gate that also holds when its own inputs are uncertain — which is the whole point of "a gate that lives in prose is not a gate." Happy to push the exact diff if useful.

Rul1an:
Both fail-open paths closed in `5a586ecf`, and the review was right that this is the wrong direction for a gate specifically.

The second one is the subtle half and worth stating for the record: `set -e` is suspended inside an `if` condition, so `if ! gh api ... | grep -q` turned a rate limit into "no such milestone" and exited 0. Capturing into a variable first separates "could not check" from "not found", and the first now blocks. `--paginate` closes the other.

The blocker query is left outside a condition deliberately, now with a comment saying why, so a future refactor that wraps it in an `if` reintroducing the same shape has something to trip over.

Re-verified live on all three paths rather than reasoned about: a milestone with an open blocker fails and names #1949, a version with no milestone exits clean, and an unreachable API refuses to release.

PR #1988, opened 2026-08-04
A deny the upstream never sees is not an e2e test

Rul1an:
Rebased onto `main` now that #1980 has merged: this is a **single commit**, and the three #1980
commits it previously carried are gone.

## Description

`e2e_wrap_denies_wildcard_contains` and `e2e_wrap_denies_schema_violation` spawned
`assay-mcp-server` and never used it. Both assert verdicts `assay mcp wrap` reaches *before*
dispatching upstream, so their real boundary was "anything spawnable" — `/usr/bin/true` passed
them. They detected an *absent* upstream, never a wrong or stale one. #1980 made the binary they
spawn guaranteed-current and said in the same doc comment that this does not close the gap; that
comment pointed here.

Two options were on the table: give them an assertion only a correct server can satisfy, or accept
they are proxy-policy tests, rename them, and stop spawning a server they never use.

**This takes the first.** Not primarily for the name. The property they are reaching for — *a
denied call must not reach the upstream* — is real, is end-to-end, and this file is the only place
covering it. Renaming down would discard it rather than fix it, and would not even remove the
spawn, since the proxy must spawn something; it would trade a real server for a stub.

The deciding detail was in the fixtures rather than the names: **both already created a
`--policy-root`, passed it to the server, then left it empty and never made the server read
anything out of it.** The scaffolding for a real upstream was there, unused. So the fix gives that
argument a job. After the deny, each fixture sends one call its own policy allows; the proxy
forwards it, and the wrapped server has to resolve a policy out of that root and answer.

### What the probe checks

Two independent things:

1. **The next frame is the probe's.** The proxy answers a denied call itself and skips the forward.
   Had it forwarded instead, the server would have answered it — the denied names are ones the
   server does not implement, and it replies `Unknown tool: <name>` rather than staying silent —
   and on a single ordered pipe that frame necessarily precedes the probe's. So reading the very
   next frame and finding the probe's id is evidence the denied call was never dispatched. That is
   the property a proxy deny exists for, and asserting on the deny response alone cannot show it.
   It rests on this server answering unknown tools; an upstream that received the call and said
   nothing would not be caught, and the doc comment says so.
2. **The payload is a real policy verdict** — resolved from `--policy-root` and reported as
   `allowed`. Nothing but `assay-mcp-server` produces it.

### Fixture changes, and why they do not weaken anything

The schema fixture admits `assay_check_args` alongside `read_file` to carry the probe, with a
schema of its own because `unconstrained_tools: deny` denies an allowed tool that has none.
`read_file` keeps its schema; `/etc/passwd` still fails the same `pattern` and still reports the
same `E_ARG_SCHEMA`. The wildcard fixture needs no policy change at all.

The probe cannot instead be a `tools/list`: the proxy evaluates policy against **every** request,
and a method naming no tool evaluates as the *empty* tool name, which the schema fixture's
allowlist rejects (`E_TOOL_NOT_ALLOWED`, `tool: ""`). Measured, not assumed.

No existing assertion is weakened and there is no skip path.

### What this PR no longer contains

An earlier revision also rewrote `read_json_line` to be able to time out, because the version it
was based on could not. #1980 landed its own fix for that — `JsonLines`, with an explicit deadline
check against `recv_timeout`'s optimistic `try_recv` and a dedicated regression test, which is
better than what this PR had. That work is dropped here and the probe calls the existing helper
instead. Two implementations of one rule drift; this one now has a single owner.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor

## Verification

Measurements on `93c1bf12`, clean working tree, worktree `.claude/worktrees/kind-gould-c0bc7f`,
`rustc 1.96.0 (ac68faa20 2026-05-25)` (the pinned toolchain), darwin 25.6.0. Re-measured after the
rebase, because the reader underneath these fixtures changed.

Established by mutation against the real test binary, not by argument — an argument is what an
earlier revision of this file got wrong twice:

| mutation | result |
|---|---|
| server drops its `assay_check_args` dispatch arm | both fixtures **FAIL** on the payload |
| server exits immediately | both **FAIL** on the read |
| proxy forwards a blocked call anyway | both **FAIL** on the frame id — and nothing else in the file does |
| unmutated | `cargo nextest run -p assay-cli`: **606 passed**, 1 skipped |

`cargo clippy -p assay-cli --tests --all-features -- -D warnings` clean; `cargo fmt --all --check`
clean.

### A note for anyone re-checking this

Substituting a fake at `target/debug/assay-mcp-server` proves nothing either way: the nested Cargo
build from #1980 **restores the real binary** before the test spawns it. An attempt at it "passed"
against `/usr/bin/true` — including `owasp_mcp01_...`, which genuinely depends on the server, and
that impossibility is what exposed the mistake. The stand-in upstreams (`sh -c 'sleep 30'`,
`/bin/cat`, `/usr/bin/true`) were driven through the identical request sequence out-of-band
instead; none satisfies check 2. `cat` is the interesting one: it echoes the request, so it does
carry the probe's id, but has no `result.content[0].text`. This is recorded in the code so the next
person does not lose the same hour.

- [x] `cargo check` passes
- [x] `cargo test` passes
- [ ] Ran a demo suite — not applicable; this is test infrastructure

Not an ADR-042/043 slice: it changes no evidence ingestion, no evidence verification, and no MCP
authorization behaviour — only what the tests over `assay mcp wrap` assert.

## Checklist
- [ ] I have updated the documentation accordingly. — no docs describe these fixtures; the rationale lives in doc comments beside the assertions.
- [x] I have added tests to cover my changes. — the change *is* to tests; its properties are established by the mutation table above.

### DIFF
--- a/crates/assay-cli/tests/e2e_mcp_wrap_assert_cmd.rs
+++ b/crates/assay-cli/tests/e2e_mcp_wrap_assert_cmd.rs
@@ -39,20 +39,14 @@ fn assay_bin() -> PathBuf {
 /// such as `assay-core` that an mtime comparison against `assay-mcp-server/src`
 /// would miss. On an already-current tree this is a no-op freshness check.
 ///
-/// What this does and does not buy, because the difference is easy to overstate:
-/// it guarantees the binary these tests spawn is built from the current source.
-/// It does *not* make the tests fail when it would not be. Only
-/// `owasp_mcp01_token_args_do_not_leak_to_proxy_logs` depends on the server
-/// answering at all; `e2e_wrap_denies_wildcard_contains` and
-/// `e2e_wrap_denies_schema_violation` assert verdicts the proxy reaches before
-/// dispatching upstream. They pass against anything that spawns and then exits
-/// when its stdin closes — `/usr/bin/true` included, so the upstream need not
-/// speak the protocol, or read a byte, to satisfy them. (#1981 added an
-/// exit-status assert, which does make a server that hangs past five seconds
-/// fail them; that bounds how the upstream *terminates*, not what it answers.)
-/// Detecting a wrong server in those two needs an assertion only a correct
-/// server can satisfy, which is a change to what they test rather than to how
-/// the binary is found, and is what #1988 is for.
+/// Building the current binary is only half of it: a test also has to *notice*
+/// when the wrong one is on the other end. It used not to. The deny fixtures
+/// asserted verdicts the proxy reaches before dispatching upstream, so they
+/// passed against anything spawnable, `/usr/bin/true` included — the upstream
+/// needed neither to speak the protocol nor to read a byte. Every test in this
+/// file now makes the wrapped server answer for itself, the deny fixtures via
+/// [`assert_upstream_is_the_real_server`], which is where that argument is
+/// written down. A stale or wrong binary now fails here.
 ///
 /// Panics (never skips) if the build fails: a skipped e2e test that reads as
 /// green is a worse failure than a loud one.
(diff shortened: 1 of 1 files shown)


Rul1an:
## Heads-up: `main` moved under this file, with a measured conflict

`80992890` (#1987) landed on `main` about half an hour ago and rewrote the read helper in `crates/assay-cli/tests/e2e_mcp_wrap_assert_cmd.rs`, the same region this branch reworks. This is measured, not predicted:

```
$ git merge-tree --write-tree origin/main c7719df9
CONFLICT (content): Merge conflict in crates/assay-cli/tests/e2e_mcp_wrap_assert_cmd.rs
```

Nothing here is wrong. The two branches were solving different problems in one file and #1987 got there first, so the cost lands on this one. Flagging what is worth keeping when you resolve it, because a plain "take ours" would quietly drop three things.

**Your `FrameReader` and the merged `JsonLines` converged on the same design independently**: worker thread doing the blocking `read_line`, consumer waiting on `recv_timeout`. That part resolves either way.

**Your deadline handling is correct on the subtle axis, and that is worth saying explicitly** so it does not get lost in a resolution. `deadline.checked_duration_since(Instant::now())` returns `None` once the deadline has passed, and `?` turns that into the timeout error. That covers the case where a child writes non-JSON faster than the loop skips it: `recv_timeout` attempts an optimistic `try_recv` before consulting the clock, so it keeps handing back queued lines at zero remaining duration and a loop that trusts the duration alone runs unbounded. #1987 shipped with exactly that hole and needed a review round to catch it. Your `checked_duration_since` shape avoids it by construction.

**Three things on `main` that a straight "take ours" would lose:**

1. Kill-on-timeout on every read path. You have one `child.kill()`; the merged version routes all reads through a helper that kills, because a wedged proxy otherwise stays parked on the inherited stderr for the life of the test binary.
2. Per-call context in the failure. The merged version takes a `what: &str` so a timeout names the unanswered request; `owasp_mcp01` reads two responses off one stream, and "timeout after 5s waiting for response" cannot tell you which one.
3. `json_lines_deadline_holds_against_a_flood_of_non_json_lines`, the regression test for the case above. It uses `yes` rather than a shell `while` loop on purpose: a shell loop produces more slowly than the consumer skips, so the channel drains and that version of the test passes against the broken code. Reproducing it needs a producer that outruns the consumer.

Happy to review the resolution if useful. Not touching this branch.

PR #1955, opened 2026-08-03
feat(mcp-server): type the correlation basis a tool-decision record retains

Rul1an:
## What

MCP 2026-07-28 removed protocol-level sessions (SEP-2567) and documented W3C Trace Context keys in `_meta` (SEP-414). That means "these N records belong to one interaction" stopped being a transport fact a reader of retained records may assume — it became a claim whose basis must travel with the record.

Each `assay.tool_decision_surface.v0` decision now carries a typed `correlation` object:

| basis | meaning |
|---|---|
| `propagated_trace_context` | valid W3C `traceparent` from `_meta`, retained verbatim; `source_class: propagated` — a producer-propagated claim, never an observed transport fact |
| `malformed_trace_context` | a carrier was sent but is invalid (shape, non-hex, all-zero ids, hostile bytes); its bytes are **not** retained; deliberately distinct from absence |
| `none` | stateless record; grouping rests on producer-minted envelope identity and inherits `self_reported` class |

`tracestate`/`baggage` are deliberately not retained (free-form values the redaction rules cannot reason about).

## How

- `ObservedCall` gains `traceparent: Option<&str>`; the server extracts it from `params._meta` at the `tool_call_done` site.
- Validation and basis typing live inside `build_decision` — no caller can opt out, same discipline as redaction.
- TDD: the three behavioral tests were written first and confirmed failing (`valid_traceparent_is_retained_and_typed_as_propagated`, `absent_traceparent_is_typed_none_never_silent`, `malformed_traceparent_is_typed_distinctly_and_not_retained`), then the smallest passing change.
- Reference fixtures updated (one propagated, one none) + fixture-guard invariant: `traceparent` retained iff basis is propagated; `source_class` propagated iff basis is propagated.
- Spec doc `docs/reference/tool-decision-surface.md` gains the record-shape field and a `## Correlation basis` section.

## Non-claims

Additive to v0; no consumer keying on the existing shape breaks. This does not claim any producer propagates context wrongly, does not verify trace propagation, and does not introduce any grouping verdict — it only types what the record retains so a downstream grouping claim inherits an honest source class.

## Verification

- `cargo test -p assay-mcp-server` — all green (includes the 3 new behavioral tests + fixture guard)
- `cargo clippy -p assay-mcp-server --all-targets -- -D warnings` — clean
- `cargo fmt` — applied; pre-commit hooks passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
[...]

PR #1991, opened 2026-08-04
Release 3.38.0

Rul1an:
Version bump only: `Cargo.toml`, `Cargo.lock`, `CHANGELOG.md`.

## What ships

Two entries turn previously green configurations red, both by design — neither is a behaviour regression, both remove a silent one:

- an unknown key inside an `assertions:` entry is a parse error instead of being dropped in silence (#1961);
- an assertion that cannot check anything is reported rather than passing (#1949).

Also carried: the assertion documentation rewritten from the enum after none of its examples parsed (#1960); a static sweep so `assay validate` no longer steps over the assertions it was asked to check (#1949 item 2); an RFC 8785 conformance corpus for the canonicalizer (#1982); and a release gate that refuses to build over an open blocker (#1985, #1949 item 4).

Minor rather than patch: `main` carries an `Added` entry since v3.37.0 (the typed `correlation` basis, #1955).

## How the lock was edited

Surgically, version lines only — not `cargo update -w`, which has silently re-resolved an unrelated dependency during a release prep before. Verification:

- `cargo check --workspace --locked --offline` passes;
- the `Cargo.lock` diff contains no line that is not a version line (22 of them);
- `cargo deny` (advisories, licenses, bans, sources) passes.

## Not in this release, deliberately

The two open Dependabot lock bumps (#1964, #1966) are held back. `Cargo.lock` classifies as workspace dependency surface, so each needs its own delegated `gates=all` proof bound to its exact head SHA — and a rebase invalidates it. Taking them here would mean three serialized runs on the singleton self-hosted runner instead of one, for two routine patch bumps with no advisory signal. They belong in 3.39.0, ideally batched so they cost one proof rather than two.

`#1965` (actions/checkout across 45 workflows, including `release.yml`) is also held: changing the release machinery on the day it runs its first gated release is unnecessary risk for a CI-only benefit.

## Gate

The `v3.38.0` milestone has no open `release-blocking` issues. #1949 was retargeted to `v3.39.0` with the reasoning recorded on the issue, because item 1 is genuinely unfinished for this release. This will be the first real execution of the gate added in #1985.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -84,7 +84,20 @@ jobs:
           VERSION: ${{ steps.version.outputs.version }}
         run: |
           set -euo pipefail
-          milestone="v${VERSION}"
+
+          # `resolve-release-version.sh` returns the tag form, WITH the leading v (verified by
+          # running it: refs/tags/v3.38.0 -> "v3.38.0"). Blindly prefixing another one produced
+          # "vv3.38.0", which matches no milestone and takes the nothing-to-gate-on path -- a
+          # fail-open that logs like success, in the gate whose entire point is to fail closed.
+          # Normalise both known shapes and refuse anything else rather than guessing.
+          case "$VERSION" in
+            v[0-9]*)  milestone="$VERSION" ;;
+            [0-9]*)   milestone="v$VERSION" ;;
+            *)
+              echo "::error::unrecognised release version '${VERSION}'; refusing to release rather than guessing which milestone to check."
+              exit 1
+              ;;
+          esac
 
           # Capture before testing, and paginate. Both matter, and both are fail-closed fixes:
           #
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,27 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [3.38.0] - 2026-08-04
+
+### Changed
+- Releases now refuse to build over an open blocker. A `release-blocking`
+  label plus a step in the release workflow's contract job: if the milestone
+  being released has an open issue carrying that label, the run fails and names
+  the issues before any artifact is built. There is deliberately no override
+  input — removing the label or moving the issue to a later milestone are
+  recorded decisions on the issue itself, which is better provenance than a
+  workflow toggle. A release with no matching milestone is not blocked. This
+  exists because the gate on #1949 lived only in a sentence and lapsed when
+  3.37.0 shipped over it (#1985).
+
 ### Fixed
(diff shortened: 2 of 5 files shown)


Rul1an:
Delegated gate dispatched on the final head, before any further push. Per the Dependabot/dependency-surface lane flow, the recorded proof must match the PR head SHA exactly; this branch will not be pushed again while the run is in flight.

- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/30919433171
- sha: 27ede1617544ff36e50578788f5b61913930fbb0

Classification is `Cargo.lock` → workspace dependency surface → `gates=all`, as expected for a version bump. `build_ebpf=true`. The `assay-bpf-runner` was online and idle at dispatch.

Rul1an:
Superseding the proof above. The previous run was cancelled deliberately, not by failure: pushing the gate fix moved the head, and a recorded proof must match the head SHA exactly.

- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/30919673374
- sha: 8cee93595995bb97ba2d67cfeca1e1792109ae40

Why the head moved after the first dispatch, stated plainly because it cost a runner slot: testing `resolve-release-version.sh` rather than assuming its output showed it returns the tag form with the leading `v`. The gate prefixed another one, looked for `vv3.38.0`, found no such milestone, and took the nothing-to-gate-on path — exit 0. A fail-open in the step built to fail closed, which would have logged like success on its first real release.

Both known version shapes now normalise to the same milestone and an unrecognised value is refused rather than guessed. Verified against live repository state for `v3.38.0`, `3.38.0`, and a garbage value.

This branch will not be pushed again while the run is in flight.

Rul1an:
Superseding again. The head moved because CI found a real miss, not because of a flake.

- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/30920776129
- sha: dd2b3c3ae0f85a98b102ff90ef79f92d8dc2b178

`bundle_reader (evidence chain)` failed on `Fuzz smoke`: `fuzz/` has its own `Cargo.lock` naming the workspace crates by version, and that lane runs `cargo metadata --locked`. Bumping only the root lock left it at 3.37.0. The v3.36.0 release prep did bump it; my verification checked "22 of 22 version lines" in the root lock and never asked whether there was a second lockfile.

`cargo metadata --locked --offline` now succeeds in `fuzz/`, and the `Fuzz lane contract` hook passes.

Two staleness gaps are deliberately **not** fixed here. `[workspace.dependencies]` and three docs still read `3.36.0` — v3.37.0 already skipped them, a `^3.36.0` requirement still resolves to this release, and nothing is broken by it. Enlarging a release-prep diff to carry a pre-existing gap is the wrong trade; filed separately.

PR #1995, opened 2026-08-04
A blocking read_line cannot be timed out from the loop around it

Rul1an:
The stdio e2e tests in `assay-mcp-server` read JSON-RPC responses from a spawned proxy with a bare `BufRead::read_line` and no timeout. A child that spawns but never answers wedges the test inside the syscall, so no elapsed-time check in the surrounding loop is ever reached. Under `cargo nextest` that surfaced as the `.config/nextest.toml` slow-timeout SIGTERM at 120s with no indication of which request went unanswered, and `retries = 1` bought a second 120s hang. The wedged child was never killed, so it kept running against the harness's inherited stderr.

Eight files each carried their own copy of that helper. Rather than fix the same read eight times, `tests/jsonrpc_conn/mod.rs` moves the blocking read onto a worker thread and waits on it with `mpsc::Receiver::recv_timeout`. `Conn` owns the write side as well as the read side, which is what lets a timeout name the exchange that stalled without a label argument at each of the ~85 read sites.

## Relationship to #1987

**#1987 is the same finding in the CLI's copy of this reader.** It landed on main after this branch was cut, so the two were written independently. Merging it in surfaced a defect this branch had: `recv_timeout` attempts an optimistic `try_recv` before it waits, so it returns an already-queued line even at zero remaining duration. Every read here loops past lines that are not the response, so a child writing those faster than the loop skips them ran unbounded past its deadline. Fixed in a29d3b94, with the finding credited to #1987.

## Open finding: two implementations of one mechanism

There are now two independent implementations of worker-thread + `recv_timeout` + deadline-at-top + kill-on-timeout: `JsonLines` in `crates/assay-cli/tests/e2e_mcp_wrap_assert_cmd.rs` and `Conn` here. They are not the same abstraction — `JsonLines` is a read-only reader for three tests with explicit labels, `Conn` is bidirectional for ~85 sites — but the mechanism is identical, and it has already drifted: this branch shipped it without the deadline check that #1987 had found hours earlier. Unifying means a dev-only crate both test suites can depend on, which is a workspace change and is deliberately not in this PR. Flagging it for the reviewer to rule on rather than deciding it here.

## Verification

Two properties, both provoked against a real child rather than reasoned about:

1. **Wedged child.** Pointed a test's server binary at `#!/bin/sh` + `exec sleep 300`. The test failed at 30.01s naming `id=1 method=initialize`, and no child survived. Restored afterwards.
2. **Flood outruns the deadline.** Two regression tests in `jsonrpc_conn_selftest.rs`, one per skip loop. With the deadline check removed both fail at 5s against a 1s budget; with it restored both pass at ~1s. They use `yes` and not a shell `while` loop on purpose — the first version used `sh -c 'while :; do echo; done'` and **passed against the unfixed reader**, because that produces more slowly than the loop skips, so the channel drained and the run bounded itself through the ordinary empty-channel path.

`cargo clippy -p assay-mcp-server --tests -- -D warnings` clean. `cargo nextest run -p assay-mcp-server`: 288 passed, 0 skipped.

## Behaviour tightened rather than preserved

- `stdio_edge_cases` now fails on a non-JSON stdout line instead of skipping it silently.
- `resource_limits` uses one persistent reader instead of building a fresh `BufReader` per request, which could discard buffered bytes.
- `wait()` is bounded: a child that never exits after stdin EOF hangs identically to one that never answers.

## Non-claims

The bounded `wait()` and the Drop-time reap are reasoned about, not provoked by a test. `drain_stdout`'s deadline check is not covered — flooding it accumulates every line into a `Vec`, so provoking it trades an unbounded wait for unbounded memory. The startup allowance (75s, first read only) is calibrated to one machine's measurement of ~43s for four concurrent `cargo run` tests contending on the target-directory lock; it is not a measurement of CI.

Review quorum is not met: this is the builder's own account. Needs one non-building agent review plus CodeRabbit or Copilot on the final head SHA.

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test reliability with bounded timeouts for JSON-RPC communication.
  * Added coverage for stalled, noisy, malformed, and notification-heavy process output.
  * Standardized process communication and cleanup across stdio and proxy test suites.
  * Preserved existing assertions for forwarding, denials, limits, manifests, shutdown, and error handling.

### DIFF
--- a/crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs
+++ b/crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs
@@ -292,7 +292,9 @@ impl Conn {
     /// attempts an optimistic `try_recv` first, so it hands back an already-queued line even when
     /// the remaining duration is zero; a child writing skippable lines faster than the callers
     /// above skip them therefore keeps the channel non-empty and would run unbounded past the
-    /// deadline. `a_flood_of_skippable_lines_cannot_outrun_the_deadline` covers that case.
+    /// deadline. `jsonrpc_conn_selftest.rs` covers that case, one test per skip loop:
+    /// `a_flood_of_blank_lines_cannot_outrun_the_deadline` and
+    /// `a_flood_of_notifications_cannot_outrun_the_deadline`.
     ///
     /// Credit: the same defect was found and fixed for the CLI's copy of this reader in #1987.
     fn next_line(&mut self, deadline: Instant) -> String {
--- a/crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs
+++ b/crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs
@@ -1,0 +1,305 @@
+//! Timeout-bounded JSON-RPC-over-stdio plumbing shared by the e2e tests in this directory.
+//!
+//! Every one of those tests drives a spawned proxy or server over newline-delimited JSON-RPC on
+//! stdin/stdout. Read with a bare [`BufRead::read_line`], a child that spawns but never answers
+//! wedges the test forever: `read_line` blocks inside the syscall, so no elapsed-time check in the
+//! surrounding loop is ever reached. Under `cargo nextest` that surfaces as the `.config/nextest.toml`
+//! `slow-timeout = { period = "60s", terminate-after = 2 }` SIGTERM at 120s with no indication of
+//! which request went unanswered, and `retries = 1` then buys a second 120s hang.
+//!
+//! [`Conn`] moves the blocking read onto a worker thread and waits on it with
+//! [`std::sync::mpsc::Receiver::recv_timeout`], which is cancellable. On timeout it kills the child
+//! rather than leaving a wedged process orphaned onto the inherited stderr, and it names the request
+//! it was waiting for — which it can only do because sends go through the same handle as reads.
+
+// Each test crate in this directory uses a different subset of this module.
+#![allow(dead_code)]
+
+use serde_json::Value;
+use std::io::{BufRead, BufReader, Write};
+use std::process::{Child, ChildStdin, ExitStatus};
+use std::sync::mpsc::{self, Receiver, RecvTimeoutError};
+use std::time::{Duration, Instant};
+
+/// How long one exchange may take before the test fails instead of hanging.
+///
+/// Deliberately well under nextest's 120s terminate-after kill, so a wedged child produces this
+/// module's message rather than an unexplained SIGTERM.
+pub const DEFAULT_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// How long the child may take to exit once it has seen EOF on stdin.
+const REAP_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// What the reader thread hands back: a line, or the error that ended the stream.
+enum Chunk {
+    Line(String),
+    Err(String),
+}
+
+/// A live stdio JSON-RPC connection to a spawned child, with every read bounded by a deadline.
+pub struct Conn {
+    child: Child,
+    stdin: Option<ChildStdin>,
+    rx: Receiver<Chunk>,
+    timeout: Duration,
+    /// Budget for the first read only; startup is slow in a way steady state is not.
+    startup_timeout: Option<Duration>,
+    /// The budget the in-flight read is using, so a failure reports the deadline it actually had.
+    current_budget: Duration,
+    /// The most recent request written, so a timeout can name the exchange that stalled.
+    last_sent: Option<String>,
+    responses_read: usize,
+}
+
+impl Conn {
+    /// Take over a spawned child's stdin/stdout and start the reader thread.
+    ///
+    /// The child must have been spawned with both piped.
+    pub fn attach(mut child: Child) -> Self {
+        let stdin = child
+            .stdin
+            .take()
+            .expect("child must be spawned with stdin piped");
+        let stdout = child
+            .stdout
+            .take()
+            .expect("child must be spawned with stdout piped");
+        let (tx, rx) = mpsc::channel();
+
+        // Detached on purpose: joining this thread is exactly the block this module exists to
+        // avoid. It ends on its own when the child's stdout closes, which killing the child forces.
+        std::thread::spawn(move || {
+            let mut reader = BufReader::new(stdout);
+            loop {
+                let mut line = String::new();
+                match reader.read_line(&mut line) {
+                    // EOF. Dropping `tx` disconnects the channel, which the reader reports as EOF.
+                    Ok(0) => break,
+                    Ok(_) => {
+                        if tx.send(Chunk::Line(line)).is_err() {
+                            break; // the test dropped its Conn
+                        }
+                    }
+                    Err(e) => {
+                        let _ = tx.send(Chunk::Err(e.to_string()));
+                        break;
+                    }
+                }
+            }
+        });
+
+        Self {
+            child,
+            stdin: Some(stdin),
+            rx,
+            timeout: DEFAULT_TIMEOUT,
+            startup_timeout: None,
+            current_budget: DEFAULT_TIMEOUT,
+            last_sent: None,
+            responses_read: 0,
+        }
+    }
+
+    /// Give the FIRST read a longer budget than the rest.
+    ///
+    /// Only startup is slow, so only startup gets the allowance. Extending it to every read would
+    /// let two slow exchanges add up past the harness's own kill, which is the failure this module
+    /// exists to keep off the table. Whatever is passed must therefore leave room for one further
+    /// [`DEFAULT_TIMEOUT`] under nextest's 120s `terminate-after`.
+    ///
+    /// No caller needs this today: every test here spawns a prebuilt `CARGO_BIN_EXE_*` binary, so
+    /// no first exchange waits on a compile. It is kept because the constraint above is the part
+    /// that is easy to get wrong, and it is written down here rather than rediscovered.
+    #[must_use]
+    pub fn with_startup_timeout(mut self, timeout: Duration) -> Self {
+        self.startup_timeout = Some(timeout);
+        self
+    }
+
+    /// The budget for the read that is about to start, consuming the startup allowance if unused.
+    fn take_budget(&mut self) -> Duration {
+        let budget = self.startup_timeout.take().unwrap_or(self.timeout);
+        self.current_budget = budget;
+        budget
+    }
+
+    /// Write one request as a JSON line, remembering it so a later timeout can name it.
+    pub fn send(&mut self, v: Value) {
+        self.last_sent = Some(describe(&v));
+        let stdin = self.stdin.as_mut().expect("stdin is still open");
+        writeln!(stdin, "{v}").expect("write request");
+        stdin.flush().expect("flush request");
+    }
+
+    /// Send a request and read the response to it.
+    pub fn request(&mut self, method: &str, params: Value, id: u64) -> Value {
+        self.send(serde_json::json!({
+            "jsonrpc": "2.0",
+            "method": method,
+            "params": params,
+            "id": id
+        }));
+        self.read_json()
+    }
+
+    /// The next non-blank stdout line, parsed as JSON.
+    pub fn read_json(&mut self) -> Value {
+        let deadline = Instant::now() + self.take_budget();
+        self.read_json_by(deadline)
+    }
+
+    /// [`Conn::read_json`], but skipping notifications and upstream-initiated requests, which carry
+    /// a `method` and are not the response this caller is waiting for.
+    ///
+    /// The deadline spans the whole skip loop, so a child that only emits notifications still fails
+    /// on time instead of renewing its budget with every one.
+    pub fn read_response(&mut self) -> Value {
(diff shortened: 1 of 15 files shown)


Rul1an:
## Review — not merge-ready yet

Direction is right: replacing repeated bare `BufRead::read_line` helpers with one timeout-bounded JSON-RPC stdio helper is a real reliability improvement. The PR correctly addresses the core failure mode: a blocking `read_line` cannot be timed out by checking elapsed time around the call.

I reviewed the PR from the diff/code paths, not by running tests locally. I found **two blocking Major findings**.

---

## Major: unbounded channel still permits memory blow-up under stdout flood

**File:** `crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs`  
**Category:** Correctness / Performance / test reliability

`Conn::attach` creates the reader channel with:

```rust
let (tx, rx) = mpsc::channel();
```

That is unbounded.

The PR correctly fixes the `recv_timeout` optimistic-`try_recv` problem by checking the deadline at the top of `next_line`. That bounds time, but not memory.

Execution path:

1. child floods stdout with blank lines or notifications;
2. reader thread runs `read_line` in a tight loop;
3. every line is sent into the unbounded channel;
4. main test thread consumes and skips lines until deadline;
5. if producer outruns consumer, the channel grows for the whole budget: normally 30s, or 75s for startup paths.

A `yes`-style producer can enqueue a very large number of lines during that window. The selftests prove the deadline fires; they do not prove memory is bounded.

### Suggested fix

Use a bounded channel, for example:

```rust
use std::sync::mpsc::{self, Receiver, RecvTimeoutError, SyncSender};

let (tx, rx) = mpsc::sync_channel(1);
```

or a small bound such as `sync_channel(16)` if a little buffering is useful.

That gives the reader thread backpressure. If a child floods, the reader stops draining stdout once the channel is full, the child eventually blocks on its pipe, and the test thread remains in control of the deadline.

This is the missing memory/backpressure half of the timeout fix.

---

## Major: timeout/Drop kills only the direct child, but `cargo run` paths can leave the actual server alive

**Files:**

- `crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs`
- `crates/assay-mcp-server/tests/stdio_edge_cases.rs`
- likely same pattern in `stdio_e2e.rs` from the diff

**Category:** Correctness / process cleanup / CI reliability

`Conn::fail`, `Conn::kill`, `Conn::wait`, and `Drop` call:

```rust
let _ = self.child.kill();
let _ = self.child.wait();
```

That only targets `Conn.child`.

For tests spawned through `env!("CARGO_BIN_EXE_assay-mcp-server")`, that is the actual server process. Fine.

But some migrated tests still spawn through:

```rust
Command::new("cargo")
    .args([
        "run",
        "-q",
        "-p",
        "assay-mcp-server",
        "--",
        "--policy-root",
        policy_root,
    ])
```

There `Conn.child` is `cargo`, not necessarily the actual server after `cargo run` has launched it. If the timeout fires after cargo has spawned the server, killing cargo may not reliably kill the server child/grandchild. The server can keep stdout/stderr handles open, and the detached reader thread may remain blocked on the inherited stdout pipe.

That reintroduces the “wedged child left behind” class this PR is meant to eliminate.

### Suggested fix

Best fix: stop using `cargo run` in these integration tests and use the built binary via Cargo’s injected env var:

```rust
Command::new(env!("CARGO_BIN_EXE_assay-mcp-server"))
```

That also removes or sharply reduces the need for the 75s startup allowance.

If `cargo run` must remain, `Conn` needs process-tree/process-group cleanup:

- Unix: start a process group/session and kill the group;
- Windows: job object or equivalent child-tree strategy;
- document platform differences.

But for tests in the same package, `CARGO_BIN_EXE_assay-mcp-server` is much simpler and matches the adjacent work in this area.

---

## Minor: `drain_stdout` remains an unbounded collector

**File:** `crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs`

The PR body already calls this out. `drain_stdout` collects every non-empty line into a `Vec<String>` until EOF or deadline. If called on a still-running/flooding child, it can trade an unbounded wait for unbounded memory.

Not blocking if current call sites only use it after shutdown, but the helper itself should ideally encode that contract.

Suggestions:

- rename to `drain_stdout_after_shutdown`; or
- add a `max_lines` / `max_bytes` cap; or
- document that it must only be used after EOF is expected.

---

## Minor: `request()` reads the next JSON object, not necessarily the matching response

**File:** `crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs`

`request()` sends a request and then calls:

```rust
self.read_json()
```

But the module also has `read_response()`, which skips notifications/upstream-initiated requests. A method named `request()` suggests “send request and read its response.” If a server emits a notification first, `request()` returns the notification.

Either change it to:

```rust
self.read_response()
```

or rename/comment it to make the “next JSON object” semantics explicit.

---

## Minor: `Chunk::Err(_)` in `drain_stdout` is treated like EOF

**File:** `crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs`

`drain_stdout` currently does:

```rust
Ok(Chunk::Err(_)) | Err(RecvTimeoutError::Disconnected) => return lines,
```

That hides stdout read errors during drain.

Prefer:

```rust
Ok(Chunk::Err(e)) => self.fail(&format!("reading the child's stdout failed while draining: {e}")),
Err(RecvTimeoutError::Disconnected) => return lines,
```

EOF/disconnect is fine; I/O error should stay visible.

---

## Positive notes

- Centralizing eight copied helpers into `jsonrpc_conn` is the right maintenance move.
- Timeout errors now name the last request and response count, which is much more actionable.
- The explicit deadline check before `recv_timeout` correctly addresses the #1987 flood class.
- The selftests use `yes` instead of a shell loop, and the comment explains why. Good regression design.
- Moving `resource_limits` to a persistent reader fixes a real buffered-byte loss risk.
- Tightening non-JSON stdout from “skip silently” to “fail” is correct for JSON-RPC e2e tests.

---

## Merge recommendation

I would not merge yet.

Please address before merge:

1. replace the unbounded `mpsc::channel()` with a bounded channel/backpressure mechanism;
2. remove or explicitly handle the `cargo run` process-tree cleanup ambiguity, preferably by switching remaining `cargo run` test spawns to `CARGO_BIN_EXE_assay-mcp-server`.

After that, the existing verification set looks appropriate:

```bash
cargo clippy -p assay-mcp-server --tests -- -D warnings
cargo nextest run -p assay-mcp-server
```

plus the targeted controls already described for silent child and stdout flood cases.

Rul1an:
Thanks — this was a good review. Both Majors were testable claims, so I measured rather than assumed. One reproduces and is fixed; one does not.

Addressed in 38082abe.

## Major 1 — unbounded channel: **confirmed, fixed**

You were right, and it is worse than "permits" — it is the steady state under any flood. Measured with the flood selftest against `yes`:

| budget | peak RSS, unbounded | peak RSS, `sync_channel` |
|---|---|---|
| 1s | 113 MB | — |
| 4s | 444 MB | **2.75 MB** |

Linear at ~110 MB/s, so ~3.3 GB at the 30s default and ~8 GB at the 75s startup allowance. Now `sync_channel(1024)`, and the deadline still fires at 4.00s on the same flood.

I did not use `sync_channel(1)` or `16`. Backpressure introduces a deadlock the unbounded channel could not have: a test that writes several requests before reading any could block in `send` on a full stdin pipe while the child blocks on a full stdout pipe, and `send` is not covered by the read deadline. The deepest interleaving in this directory is two unread sends, so 1024 lines plus the child's ~64 KB stdout pipe is far clear of it while still costing tens of kilobytes. Rationale is in the constant's doc comment.

## Major 2 — `cargo run` process tree: **does not reproduce**

The premise is that `Conn.child` is `cargo` and not the server. On Unix it is the same process: cargo `exec`s the binary rather than spawning it.

Probed directly, with stdin held open through a fifo so the server does not take EOF and exit on its own:

```
=== cargo pid=74532, tree ===
74532 74529 /Users/…/target/debug/assay-mcp-server
server pid: 74532
=== after killing cargo (74532) ===
VERDICT: no orphan (server gone)
```

The PID `$!` reports for `cargo run` *is* `target/debug/assay-mcp-server`. So `kill` reaches the server and there is nothing to orphan.

Two notes on the suggested fix. `CARGO_BIN_EXE_assay-mcp-server` works for `stdio_e2e` and `stdio_edge_cases`, but not for `no_passthrough_e2e`: that test deliberately shells out to build and run with `--features test-outbound`, because the tool it exercises does not exist otherwise. Switching it would mean gating the test behind that feature and dropping it from the default suite — a coverage trade I did not want to make silently. And since the orphan premise does not hold, the change would be churn without a defect behind it.

The caveat I will record rather than fix: cargo cannot `exec` on Windows, so it spawns and forwards there, and the finding would hold. These tests already branch on `cfg!(windows)` for `python` vs `python3`. I would rather leave that in the PR body than add process-group machinery that cannot be exercised on this platform — happy to be overruled.

## Minors — all three taken

- `request()` now reads with `read_response()`. You are right that the name promised the response; a notification arriving first would have been handed back in its place.
- `drain_stdout` treated `Chunk::Err` like EOF, hiding stdout read errors. Only `Disconnected` ends the drain now; an I/O error fails with the error text.
- `drain_stdout` → `drain_stdout_after_shutdown`, so the precondition is stated at the call site, **and** capped at 10k lines so it is enforced rather than documented. I took both suggestions rather than choosing, since the rename alone leaves the helper able to blow up if someone ignores it.

## Verification on the new head

`cargo clippy -p assay-mcp-server --tests -- -D warnings` clean. `cargo nextest run -p assay-mcp-server`: 288 passed, 0 skipped. Both flood selftests re-checked in both directions against the bounded channel — with the deadline check removed they still fail at 5s against a 1s budget, with it restored they pass at ~1s.

## Non-claim on the fix you found

The memory bound is measured in both directions, not pinned by a regression test. An RSS assertion is too platform-specific and too load-sensitive to be a reliable gate, so this rests on the numbers above and on the bound being visible in the type. If you would rather have it gated anyway, say so and I will add one behind a feature or an ignore.

Rul1an:
Copilot's two findings, both confirmed and fixed in dd9b968a.

**Unconditional imports behind `#[cfg(unix)]` code.** This is live, not hypothetical: `assay-mcp-server` is not excluded from CI's `Test Workspace (Cross-Platform)` job, which runs on `windows-latest`, so Windows compiles this file.

Verified by substituting `#[cfg(any())]` for `#[cfg(unix)]` — exactly what a non-Unix target sees. The old file:

```
error: unused import: `jsonrpc_conn::Conn`
error: unused imports: `Child`, `Command`, and `Stdio`
error: unused import: `std::sync::mpsc`
error: unused import: `Instant`
error: constant `BUDGET` is never used
error: constant `PATIENCE` is never used
```

The whole block is now a single gated `mod flood`, and the same substitution is clean under `-D warnings`.

**Stale test-name citation.** Correct, and worse than a broken link: `a_flood_of_skippable_lines_cannot_outrun_the_deadline` was a name from an earlier draft that never existed in the tree, so the comment read as coverage that was there. It now names the file and both tests that do exist.

Head is dd9b968a. `cargo clippy -p assay-mcp-server --tests -- -D warnings` clean on both the Unix and the simulated non-Unix path; `cargo nextest run -p assay-mcp-server` 288 passed, 0 skipped.

Rul1an:
Rebased onto `main` after #1986 merged. The three conflicts were in `stdio_e2e.rs`,
`stdio_edge_cases.rs` and `no_passthrough_e2e.rs`, and they were not textual: `main` and this
branch had each rewritten the same spawn statements for different reasons.

Resolved by taking both sides on their own terms — `main` is right about **how the server is
spawned**, this branch is right about **how it is read**:

- spawn: `Command::new(env!("CARGO_BIN_EXE_assay-mcp-server"))`, from #1986, replacing the
  `cargo build` + `cargo run` pair this branch still had.
- read: `Conn`, from this branch.

### One thing removed rather than merged

`CARGO_RUN_TIMEOUT` (75s) and its five call sites are gone. It existed for exactly one reason —
*"a server spawned through `cargo run`, which may still have to build, and which contends with
every other `cargo run` test for the target-directory lock"* — and after #1986 there is no
`cargo run` in this crate's tests. Keeping it would have left a constant whose stated justification
is false, which is the kind of thing that gets copied forward.

`with_startup_timeout` is kept and now has no callers. Its doc says so, and says why it is kept:
the constraint it encodes (a startup allowance must leave room for one further `DEFAULT_TIMEOUT`
under nextest's 120s kill) is the easy thing to get wrong. The module already carries
`#![allow(dead_code)]` for exactly this per-binary-subset reason.

The `READ_AHEAD_LINES` doc referenced `CARGO_RUN_TIMEOUT` for its worst-case memory figure; that
clause is rewritten rather than left dangling.

### Verification

On `c14ca04f`, clean tree, `rustc 1.96.0`, darwin 25.6.0:

| | |
|---|---|
| `cargo nextest run -p assay-mcp-server` | **287 passed**, 0 skipped, 3.4s |
| `cargo clippy -p assay-mcp-server --tests --all-features -- -D warnings` | clean |
| `cargo fmt --all --check` | clean |
| `jsonrpc_conn_selftest` flood tests | both pass — the deadline-outrun regressions still bite |

The 3.4s is worth noting: these tests used to spawn `cargo run` per test and serialize on the
target-directory lock. #1986 and this PR together are what removes that, which is why the wall time
moves as much as it does.

Issue #1998, opened 2026-08-04
Design AEE-compatible substrate-signed run-end seal primitive


Rul1an:
## Summary

PR #1997 proved a non-production, fixture-backed Assay -> AEE v0.7 spike can emit and consume an AEE-shaped two-vantage statement. The useful finding is not that Assay supports AEE today; it is that real AEE support blocks on a missing production primitive:

> Assay needs a substrate-signed run-end sealed record over still-armed state, drop accounting, observed set, observed attacks, and run binding.

This issue tracks the design and implementation plan for that primitive. It should be resolved before any production AEE exporter is added.

## Context

The merged spike:

- built an AEE-shaped `sealed` record first;
- emitted an arming record;
- emitted one proxy interception derived from `assay.denied_call_observation.v0` fixture;
- emitted one Landlock/TCP-connect interception derived from `assay.enforcement_health.v1` active-with-probe fixture;
- consumed the emitted statement back with a checker;
- exercised negative controls for missing seal, defective unreferenced seal, artifact/proxy observation overclaimed as substrate, reconstructed priced as intercepted, and run-population overclaim.

External context:

- in-toto/attestation#570, Adversarial Execution Evidence v0.7.
- Assay upstream implementation report: https://github.com/in-toto/attestation/pull/570#issuecomment-5183673627
- Assay spike PR: https://github.com/Rul1an/assay/pull/1997

## Current Assay evidence carriers

Useful inputs, but not a production AEE seal:

- `assay.denied_call_observation.v0`
  - caller-visible proxy denial observation;
  - not a policy verdict;
  - not proof of upstream side effect;
  - not by itself a substrate-signed AEE observation record.
- `assay.enforcement_health.v1`
  - Landlock/ruleset/probe facts;
  - active-with-probe can prove one denied connect was blocked;
  - not a post-run AEE seal.
- existing block/allow counters
  - enforcement event counts;
  - not `aeeDropCount`, which must be observation-drop/loss accounting.

## Design question

How should Assay produce and verify an AEE-compatible substrate-signed run-end seal while preserving Assay's existing evidence boundaries?

Key subquestions:

1. **Key boundary**
   - What key signs substrate observations?
   - Is there one substrate observation key or separate key roles for proxy and Landlock/LSM collection paths?
   - What does one operator controlling both collection paths explicitly not claim?

2. **Run binding**
   - Where is run-start entropy produced?
   - Which inputs are bound before corpus injection?
[...]

PR #2001, opened 2026-08-05
docs(aee): sketch Landlock seal payload contract

Rul1an:
## Summary

Adds the next narrow implementation sketch after #2000: a Landlock/TCP-connect run-end seal payload type and field-level validation contract for ADR-045 / #1998.

This PR is intentionally docs/spec-only. It does not add producer-side Landlock seal code, production signing, checker integration, or a stable AEE exporter.

## Motivation

PR #2000 ended with the agreed follow-up: move from fixture/checker semantics toward a small Landlock seal payload type/schema sketch, then producer-side integration, still without exposing a stable exporter.

This PR captures that intermediate contract so later producer/checker work has a single written target before production code starts.

## Changes

Adds:

- `docs/architecture/SKETCH-ADR-045-LANDLOCK-SEAL-PAYLOAD-2026-08.md`

The sketch defines:

- producer-owned payload media type: `application/vnd.assay.landlock-run-end-seal.v0+json`;
- minimal sealed payload shape;
- required fields and structural validation rules;
- canonicalization/signing boundary;
- relationship to in-toto/AEE layering;
- explicit non-claims;
- next implementation slice for checker hardening.

## Research basis

The sketch is grounded in:

- ADR-045 requirements and payload-shape discussion;
- #2000 follow-up scope;
- in-toto Statement/envelope layering: payload vocabulary must not be confused with predicate/exporter semantics;
- RFC 8785/JCS direction for deterministic signed JSON bytes;
- current agent-governance practice around validation, tracing, auditability, and evidence-backed claims.

## Non-claims

This PR does not claim:

- production AEE support;
- stable AEE export;
- production signing semantics;
- complete run-population proof;
- independent substrate operation;
- agent safety;
- provider side-effect verification;
- full AEE conformance checker coverage.

## Verification

Docs-only change. Checked for scope against ADR-045 and #2000 follow-up wording.

Suggested reviewer focus:

1. Is the proposed payload media type appropriately scoped and non-stable?
2. Does the field contract preserve ADR-045 boundaries?
3. Are the signing/canonicalization non-claims strong enough to prevent fixture semantics from becoming production semantics?
4. Is the next-slice sequencing correct: checker hardening before producer code?

## Follow-up

Separate PR after this should harden the checker/fixture boundary:

- marker-only vs full on-disk fixture policy;
- key-scope negative controls for untrusted key, wrong key role, wrong collection path;
- remove or validate `batchRoot`;
- keep exporter commands out of scope.

## Summary by CodeRabbit

* **Documentation**
  * Added an implementation sketch for the ADR-045 Landlock/TCP-connect run-end seal payload.
  * Documented payload structure, required fields, validation rules, signing boundaries, key-scope requirements, and in-toto/AEE relationships.
  * Clarified non-claims and identified follow-up checker-hardening work.

### DIFF
--- a/docs/architecture/SKETCH-ADR-045-LANDLOCK-SEAL-PAYLOAD-2026-08.md
+++ b/docs/architecture/SKETCH-ADR-045-LANDLOCK-SEAL-PAYLOAD-2026-08.md
@@ -1,0 +1,212 @@
+# SKETCH: ADR-045 Landlock seal payload contract
+
+Status: Implementation sketch  
+Date: 2026-08-05  
+Applies: ADR-045  
+Related: #1998, #2000
+
+## Summary
+
+This sketch defines the next narrow implementation contract after the ADR-045
+Landlock seal fixture/checker slice in #2000: a small Landlock/TCP-connect
+run-end seal payload type and field-level validation contract.
+
+This is deliberately **not** production seal emission, production signing,
+checker integration, or stable AEE export. It is a payload contract sketch that
+producer and checker work can converge on before production code is added.
+
+## Research basis
+
+The sketch follows five constraints from ADR-045 and current agent-security
+practice:
+
+1. **Evidence-first boundaries.** A seal payload must state only what the
+   Landlock/TCP-connect substrate can honestly carry. It must not become a
+   trust score, whole-action verdict, provider-outcome claim, or agent-safety
+   claim.
+2. **Attestation layering.** The seal payload is producer-owned payload
+   vocabulary. A later AEE statement exporter may assemble it into an in-toto
+   statement, but this payload type is not itself an in-toto predicate type.
+3. **Verify-before-read.** Production consumers must verify the envelope
+   signature and key scope over the exact signed bytes before crediting the
+   payload as attested substrate evidence.
+4. **Canonical bytes, not object-model luck.** Production signing must use a
+   specified canonical UTF-8 JSON signing surface with duplicate object members
+   rejected before signing or verification. The current fixture HMAC helper is
+   non-normative.
+5. **Operational auditability.** Agent systems fail in production when tool
+   outputs, traces, and control boundaries are ambiguous. This sketch therefore
+   names every carried claim and repeats the non-claims locally.
+
+## Payload type
+
+Use the following producer-owned media type for the first production-oriented
+Landlock seal payload sketch:
+
+```text
+application/vnd.assay.landlock-run-end-seal.v0+json
+```
+
+Interpretation:
+
+- `application/vnd.assay...+json` marks this as Assay producer vocabulary.
+- `landlock-run-end-seal` scopes the payload to the Landlock/TCP-connect
+  run-end seal primitive.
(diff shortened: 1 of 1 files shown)


Rul1an:
## Review findings addressed in `b213a991d6cb62511ad0d792ea623240c66e93a7`

All three review comments were raised against `760ddc6` and are resolved on the current head.

### 1. CodeRabbit — missing `assayDropProofModel` (Major)

Fixed. The field is now in the JSON example, the field contract table (required, `synchronous-probe` | `counted-queue-zero`), and payload-only validation rule 10. This closes the producer/checker mismatch against `scripts/experiments/aee_landlock_seal_fixture.py`.

### 2. Copilot — `assayNonClaims` minimum set was internally inconsistent

Fixed. A dedicated **Payload-local non-claims** section now names the exact minimum set that `assayNonClaims` MUST include, matching the example. The section also states explicitly that the broader document-level non-claims are normative for the sketch but are not all required as payload-local entries, and that predicate-level `doesNotAssert` is not replaced.

### 3. Copilot — `aeePostureDigest` not checkable by a payload-only validator

Fixed. The contract now separates three validation phases:

- **Payload-only** — `aeePostureDigest` MUST be lowercase SHA-256 hex; no statement context assumed.
- **Assembled-statement** — MUST equal `predicate.observationEnvironment.networkPosture.digest.sha256`.
- **Credited-evidence** — signature, key trust, key role `substrate-observation`, and key scope over `assayCollectionPath`.

The same split was applied to `aeeRunBinding`, `aeeObservedSet`, and `aeeObservedAttacks`, which had the same layering problem.

### Remaining minors (not blocking, deferred to the checker-hardening slice)

- `assaySourceSchema` is `Required: yes` but its value is `SHOULD`. The payload-only rule list should state whether the value is enforced or only presence is.
- `aeeObservedAttacks` has no payload-only rule (array-of-strings, unknown-ID policy). #2000 already has an `unsupported-observed-attack` negative control, so the vocabulary check exists elsewhere.

Both are contract refinements with no production-behaviour impact and belong with the checker tests that will exercise them.

### Scope unchanged

Docs/spec only. No producer-side Landlock code, no production signing, no checker integration, no stable AEE export.

Rul1an:
@coderabbitai review

Rul1an:
Follow-up sequencing anchor: the checker-hardening work described in the "Follow-up" section of this PR body is now tracked in #2006.

That issue covers all four items named here (fixture policy, key-scope negative controls, `batchRoot`, exporter kept out of scope) plus two that came out of attestation-verification practice rather than this sketch: a key validity window and rejecting unsupported envelope shapes instead of skipping them.

Two items in #2006 land back on this document and will need a follow-up edit to `SKETCH-ADR-045-LANDLOCK-SEAL-PAYLOAD-2026-08.md` once the checker enforces them:

- `assaySourceSchema` is `Required: yes` with a `SHOULD` value — the table has to match whichever rule the checker implements.
- `aeeObservedAttacks` has no payload-only rule yet.

PR #2002, opened 2026-08-05
One timeout, not two: mcp wrap e2e onto the shared Conn

Rul1an:
## Description

**This PR changed shape substantially while open — read this, not the earlier comments.**

When it opened, `crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs` did not exist on `main` and this branch introduced it. Since then #1995 landed the shared `Conn` and #1988 landed the upstream probe, so most of what this branch carried is now upstream and was dropped rather than duplicated.

What remains is one thing: `crates/assay-cli/tests/e2e_mcp_wrap_assert_cmd.rs` still kept its own `JsonLines`, `read_json_line`, `send_line`, `wait_child_with_timeout` and a flood regression test. After #1995 the repo holds **two implementations of one timeout**. This deletes the copy and points the file at the shared module.

**−253 / +44, one file.**

The local flood test goes rather than gets ported. It covered one shape (notifications); `jsonrpc_conn_selftest.rs` covers two — blank lines and notifications — next to the code it constrains, gated as a single `#[cfg(unix)] mod flood` so a non-Unix build has no unused imports left over. Keeping a weaker copy in a second crate is the duplication being removed.

#1980's Cargo binary lookup and #1988's upstream probe survive unchanged in substance. `assert_upstream_is_the_real_server` trades `(&mut Child, &mut dyn Write, &mut JsonLines)` for `&mut Conn` and drops its literal "what am I waiting for" string, which `Conn` now derives from the request it wrote.

## Type of Change
- [x] Refactor
- [ ] Bug fix / New feature / Breaking change / Documentation

## Verification

Head `77831f14` on `604d5f15`, worktree `.claude/worktrees/focused-satoshi-89ecd9`, rustc 1.96.0, cargo-nextest 0.9.137.

| Check | Result |
|---|---|
| **#1988's property survives the port** — upstream replaced by `/usr/bin/true`, which #1988 records as having passed the deny fixtures before it landed | all three tests **fail**, at 0.083s / 0.083s / 0.104s — on EOF, not on a deadline, so a fast detection was not traded for a slow one |
| **Silent proxy** — a script that spawns and never answers, substituted for the assay binary | all three fail at **30.48–30.61s** = the shared `DEFAULT_TIMEOUT`, each naming its own unanswered request (`id=1`, `id=2`, `id="token-log-allowed"`, `method=tools/call`). The 5s constant this replaced could be outrun by a blocking `read_line` |
| `cargo nextest run -p assay-cli --test e2e_mcp_wrap_assert_cmd` | 3 passed, warm ~0.6s each |
| clippy `-p assay-cli --tests -D warnings`, `cargo fmt --all --check` | clean |

- [x] `cargo check` passes
- [x] `cargo test` passes
- [ ] Ran a demo suite — not applicable, test-only change

## Follow-up, measured but deliberately not in this PR

The shared module bounds the reader's queue (`READ_AHEAD_LINES`) but not a single line: `read_line` grows its `String` until a newline, so a child that emits none is unbounded. Measured on an earlier revision of this branch that added a 1 MiB ceiling — peak RSS of a one-second flood dropped from **495 MB to 2.9 MB**, against a 143 MB baseline. That belongs with `jsonrpc_conn_selftest.rs`, in a change that owns the module, rather than smuggled through a caller.

## Review quorum

**Not satisfied.** Every measurement above is the builder's own, which AGENTS.md §Review Quorum excludes. The CodeRabbit and Copilot reviews on this thread were bound to earlier heads whose content is largely gone, so they do not carry. Needs a fresh look at `77831f14`.

## Summary by CodeRabbit

* **Tests**
  * Improved MCP wrap test reliability by using shared connection handling.
  * Retained coverage for policies, schemas, token redaction, upstream validation, and responses.
  * Simplified process shutdown and response handling.

* **Chores**
  * Prevented a workspace-dependent end-to-end test from being included in published package contents.

### DIFF
--- a/crates/assay-cli/Cargo.toml
+++ b/crates/assay-cli/Cargo.toml
@@ -9,6 +9,15 @@ readme.workspace = true
 description = "Policy-as-code gate for MCP agent tool calls, with verifiable evidence and Linux kernel enforcement."
 keywords = ["mcp", "ai-agents", "agent-security", "policy-as-code", "evidence"]
 categories = ["command-line-utilities", "development-tools::testing"]
+# Kept out of the published tarball because it reaches outside this package: it includes
+# `crates/assay-mcp-server/tests/jsonrpc_conn/mod.rs` by `#[path]`, and `cargo package` copies only
+# this package's files. Shipping it would put a test target in the crate that cannot compile at all
+# — `couldn't read tests/../../assay-mcp-server/...` — which is worse than its absence for anyone
+# running `cargo test` on the crates.io source (distro packagers, `cargo vendor` consumers). It
+# could never run from a tarball anyway: it shells out to `cargo build -p assay-mcp-server`, which
+# needs the workspace. `cargo publish`'s own verification builds lib and bins only, so this was not
+# a release blocker either way — it is the downstream `cargo test` that this protects.
+exclude = ["tests/e2e_mcp_wrap_assert_cmd.rs"]
 
 [lints]
 workspace = true
--- a/crates/assay-cli/tests/e2e_mcp_wrap_assert_cmd.rs
+++ b/crates/assay-cli/tests/e2e_mcp_wrap_assert_cmd.rs
@@ -1,20 +1,27 @@
 use anyhow::Context;
 
 use serde_json::{json, Value};
-use std::io::{self, BufRead, BufReader, Write};
 use std::path::{Path, PathBuf};
-use std::process::{Child, ChildStdout, Command, ExitStatus, Stdio};
-use std::sync::mpsc::{self, RecvTimeoutError};
+use std::process::{Command, Stdio};
 use std::sync::OnceLock;
-use std::thread;
-use std::time::{Duration, Instant};
 use tempfile::TempDir;
 
 mod common;
 use common::{cargo_bin, strip_cargo_crate_env};
 
-/// How long a single JSON-RPC response may take before the proxy is declared wedged.
-const RESPONSE_TIMEOUT: Duration = Duration::from_secs(5);
+/// The timeout-bounded JSON-RPC-over-stdio plumbing, shared with the `assay-mcp-server` e2e tests
+/// that #1995 moved onto it, rather than reimplemented here.
+///
+/// This file had its own copy: `JsonLines`, `read_json_line`, `send_line` and
+/// `wait_child_with_timeout`, plus a flood regression test for the same deadline property
+/// `jsonrpc_conn_selftest.rs` now covers in two shapes rather than one. Two implementations of one
+/// timeout drift, and the one that drifts is the one nobody is watching. The module is a `tests/`
+/// SUBDIRECTORY, so cargo builds it as no crate's own test target; reaching across is a coupling
+/// these tests already had by building and spawning the `assay-mcp-server` binary.
+#[path = "../../assay-mcp-server/tests/jsonrpc_conn/mod.rs"]
+mod jsonrpc_conn;
+
+use jsonrpc_conn::Conn;
 
 /// Path to the `assay` binary under test.
 ///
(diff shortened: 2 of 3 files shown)


Rul1an:
## Disposition of the three CodeRabbit findings

All three were real. Two are fixed; the third is confirmed by measurement and deliberately scoped out, with the reproduction recorded so it is not lost.

New head `c7c80de5`, rebased onto `39a8644e`. Per AGENTS.md this push invalidates any review bound to the previous head.

### 1. Bound stdout buffering before materialization — **fixed**

Confirmed by measurement before assuming. Peak RSS of the flood test, `/usr/bin/time -l` on the test binary:

| | peak RSS |
|---|---|
| normal e2e test in the same binary (baseline) | 143 MB |
| flood test, unbounded channel | **495 MB** |
| flood test, after this fix | **2.9 MB** |

352 MB of growth in the one second the test spends against `yes`; a full 30s `DEFAULT_TIMEOUT` extrapolates to roughly 10 GB. That is not a hypothetical — my own new test was imposing it on every CI run.

Fixed with `mpsc::sync_channel(64)` plus a 1 MiB ceiling applied *before* the line is materialized, via `reader.by_ref().take(MAX_LINE_BYTES + 1)`, so a child that never emits a newline fails instead of growing.

Checked that this did not quietly kill the regression test it touches: with the deadline check mutated out, the flood test still fails, skipping **5,606,928** lines and giving up only after **18.27s** against its 1s budget, caught by its own assertion. With the fix, 1.01s. Bounding the queue changed *how* it overruns, not *that* it does — and it now fails on its own assertion rather than needing nextest's SIGKILL, which is the better failure.

### 2. Fail when the stdout reader reports an error — **fixed**

Right, and it is the silent-failure shape specifically: `drain_stdout` exists to prove nothing further reached the client, and returning the lines so far on `Chunk::Err` let it report exactly that when what actually happened is that it stopped being able to look. Only `Disconnected` ends the drain now; `Chunk::Err` routes through `fail`.

### 3. Kill the complete proxy process group — **confirmed real, scoped out with the reproduction recorded**

This one corrected me. My earlier "no orphan survived" evidence was worthless for this claim: the scaffold replaced the `assay` binary itself, so there was no wrapped upstream to orphan.

Built the real topology instead — genuine `assay mcp wrap` around a server that never answers:

```
read times out at 30s -> Conn kills the assay parent
=> upstream still running afterwards (pid 88724)
```

So the finding is confirmed, not merely plausible.

Not fixed in this slice, for a reason rather than by omission. `Conn::attach` receives an already-spawned `Child`, so process-group placement is a spawn-time decision at every call site — including the nine `assay-mcp-server` test files that adopt this module on another branch, which AGENTS.md §Branch ownership says I must not edit. Shipping a half-adopted process-group contract, where `killpg` is safe only for callers that happened to opt in, is worse than a documented limitation: `killpg` on a child that is *not* a group leader targets the test runner's own group.

Scope of the risk, stated precisely: failure path only. On a green run `shutdown()` closes stdin and the whole tree exits — no orphan. It is recorded as a KNOWN LIMITATION in the commit body with the reproduction above.

### Review quorum

Still **not satisfied**. Every measurement here is the builder's, which AGENTS.md §Review Quorum says does not count. Needs one non-building agent review plus CodeRabbit or Copilot, or two non-building agent reviews, on `c7c80de5`.

Rul1an:
## Disposition of the Copilot review

New head `40453b45`.

### `AssertUnwindSafe` unused on non-Unix targets — **fixed, and it was broader**

Real, and it would have failed the build rather than merely warned: this repo lints with `-D warnings`.

Reproduced before fixing, by swapping the flood test's `#[cfg(unix)]` for `#[cfg(any())]` to stand in for a non-Unix target:

```
error: unused import: `std::panic::AssertUnwindSafe`
error: unused imports: `Duration` and `Instant`
error: could not compile `assay-cli` (test "e2e_mcp_wrap_assert_cmd") due to 2 previous errors
```

So `Duration` and `Instant` had the same fault — they are also used only inside that test. Checking the class rather than patching the one reported symbol is what surfaced them. All three moved into the test body; the same simulation now compiles clean, and the Unix path still runs 4/4 green.

Worth noting for the lane matrix: **no Unix CI lane can see this.** It is invisible until something builds the test target for a non-Unix triple.

### `drain_stdout` treats `Chunk::Err` as clean EOF

Same finding as CodeRabbit's, and already fixed on this head — only `Disconnected` ends the drain now; `Chunk::Err` routes through `fail`. Independent agreement from two reviewers on a silent-failure shape is a useful signal, so noting it rather than dropping it as a duplicate.

### Review quorum

Copilot is one of the two bots AGENTS.md §Review Quorum accepts, and it left findings rather than a quota no-op — so route 1 needs **one non-building agent review** on `40453b45` to close. Every measurement in this PR is still the builder's own and does not count.

Rul1an (commented):
## Review against current `main` (`604d5f15`)

Reviewed after #1986, #1988, #1989 and #1995 landed. The change should go in, but the merge with
`main` is not a text merge: `main`'s `jsonrpc_conn` and this branch's are 144 lines apart and each
carries a memory guard the other lacks. Taking either side wholesale loses one.

Findings ranked. Item 5 is recorded as checked-and-fine so it does not get re-raised.

### 1. Regression — `request()` reads with `read_json()` where `main` uses `read_response()`

`main`'s `request()` calls `read_response()` and says why:

> Reads with `Conn::read_response` and not `Conn::read_json`: the name promises the response, so a
> notification arriving first must not be handed back in its place.

This branch's `request()` calls `read_json()`, which returns the first JSON line whatever it is.

**Failure scenario:** any test using `conn.request(...)` against a server that emits a progress or
log notification before its response gets the notification handed back as the response, and
asserts on the wrong frame. Silent — the value is well-formed JSON.

This branch keeps `read_response()` and its doc comment, so this reads as a slip in `request()`
rather than a decision.

### 2. Each version has one memory guard the other lacks

| guard | `main` | this branch |
|---|---|---|
| `MAX_DRAINED_LINES = 10_000` — bounds what `drain_stdout` accumulates | yes | **removed** |
| `MAX_LINE_BYTES = 1 MiB` via `Read::take` — bounds a single line | **absent** | yes |

`main`'s reader loop is a bare `read_line`, so a child that never writes a newline buffers without
bound. This branch's `drain_stdout` has no count cap, so a child still writing after shutdown grows
the `Vec` for the whole deadline — at the ~110 MB/s this branch itself measures, that is gigabytes.

The rebase needs the union, not a side.

### 3. `READ_AHEAD_LINES = 1024` vs `QUEUE_CAPACITY = 64` — reconcile the rationales, not just the numbers

The memory argument here is the stronger one and the measurement (352 MB in one second) earns the
smaller bound. But the rebuttal answers a different objection than `main` recorded. `main`'s
scenario is that the **test thread** blocks in `send` on a full child-stdin pipe while the child
blocks on a full stdout pipe, and `send` is not covered by the read deadline. The note here —
"the reader thread parks in `send` once the queue is full, but the test thread never waits on it" —
is about the reader thread.

In practice it is unreachable in this directory (deepest interleaving is two unread sends, and
64 >> 2), so 64 is fine. Worth answering the objection `main` actually wrote down, or that
reasoning is lost the next time someone tunes this number.

### 4. `CARGO_RUN_TIMEOUT` must not come back

This branch predates #1986, so its tests still spawn `cargo run` and it re-adds the 75s constant.
After #1986 there is no `cargo run` left in `assay-mcp-server`'s tests; the constant was removed
during #1995's rebase for exactly that reason. Restoring it reinstates a constant whose stated
justification — "a server spawned through `cargo run`, which may still have to build" — is false.

### 5. `deadline - now` — checked, not a defect

The switch from `saturating_duration_since` to plain subtraction is safe: both call sites guard with
`if now >= deadline { self.fail(...) }` immediately before, and `fail` returns `!`.

### 6. The `#[path]` cross-crate include — worth an explicit decision

`#[path = "../../assay-mcp-server/tests/jsonrpc_conn/mod.rs"]` is a source dependency Cargo does not
model: nothing in `assay-cli/Cargo.toml` records it, so renaming that directory breaks another
crate's tests with no signal. Against that, it is the cheapest way to have one timeout
implementation instead of two, which is the point, and the alternative — a `publish = false`
test-support crate — is a workspace-shape change. The publish-shape guardrail passes.

Worth taking, but documented at both ends. Right now only the consumer says anything; the module's
own header still reads "shared by the e2e tests in this directory", which stops being true.

### Worth naming as good

- The flood test is genuinely stronger: `yes '{"jsonrpc":"2.0","method":"flood"}'` instead of
  `yes not-json` exercises `read_response`'s notification-skip path rather than only the
  "does not start with `{`" check.
- Importing `AssertUnwindSafe`/`Duration` inside the `#[cfg(unix)]` test, with the note that a
  module-scope import is an unused import on other targets and `-D warnings` turns that into a
  build failure the Unix lanes cannot see.
- #1988's upstream probe is ported to `Conn` rather than dropped.

### Plan

Rebasing onto `main` with: the union of both bounds, `request()` back on `read_response()`,
`CARGO_RUN_TIMEOUT` dropped, module header corrected. Will report the verification.

Rul1an:
Addendum to the review above, after looking at how this is actually being taken forward.

If the rebase keeps `main`'s `jsonrpc_conn` wholesale and only adapts the `assay-cli` side, then
findings **1**, **3**, **4** and **5** fall away by construction — `main`'s `request()` already
reads with `read_response()`, its `READ_AHEAD_LINES = 1024` and `saturating_duration_since` stay,
and `CARGO_RUN_TIMEOUT` is simply never reintroduced.

Two survive that route and are worth handling deliberately rather than by omission:

- **Finding 2 — `MAX_LINE_BYTES`.** `main` has no bound on a single line: its reader loop is a bare
  `read_line`, so a child that never writes a newline buffers without limit. The 1 MiB ceiling via
  `Read::take` on this branch is the only place that guard exists. Dropping this branch's module
  drops the guard with it. It is worth porting across on its own, in either PR.
- **Finding 6 — the module header.** `//! ... shared by the e2e tests in this directory` becomes
  false the moment `assay-cli` includes it via `#[path]`. The consumer side documents the
  arrangement; the module itself should say it is reached from outside its directory, or the next
  reader will treat it as crate-local and move it.

Also note #2005, which removes a dangling `[`CARGO_RUN_TIMEOUT`]` doc link left on `main` by
#1995's rebase. It touches this same module, so rebase after it lands to avoid a trivial conflict.

Rul1an:
## Adversarial review — results and what changed

An independent adversarial agent reviewed `77831f14` with a mandate to refute, not confirm. It found **two defects caused by this diff**. Both are fixed in `89998ba2`; both had reproductions, so both were verified fixed rather than argued away.

### Fixed: the very-next-frame property was silently weakened

`Conn::read_response` skips frames carrying a `method`. #1988's probe asserts on the **very next frame** — that is the whole mechanism by which "the denied call was never dispatched" is evidence rather than assertion. Against an upstream emitting one notification before each response:

| | result |
|---|---|
| base `604d5f15` | 3 FAIL |
| `77831f14` (`read_response`) | 3 **PASS** ← assertion could no longer fail for its own reason |
| `89998ba2` (`read_json`) | 3 FAIL, 0.101–0.113s |

`read_json` is also the faithful port: the reader this replaces took the next frame. My "survives unchanged in substance" was too generous, and the review was right to press on it.

### Fixed: the published tarball carried a test target that cannot compile

`#[path]` reaches outside the package; `cargo package` copies only the package's own files. So the crate shipped `tests/e2e_mcp_wrap_assert_cmd.rs` with no `jsonrpc_conn/mod.rs` beside it — `cargo test` on the crates.io source fails with `couldn't read tests/../../...`. Distro packagers and `cargo vendor` consumers hit a hard compile error, not a test that merely can't run.

`cargo publish` itself was never at risk — its verification builds lib and bins only, confirmed by actually running it. Fixed with a one-line `exclude`: 344 files → 343, the other 81 test files still ship, local runs unchanged.

### Confirmed and left alone

- **Silent proxy**: 30.48–30.61s, each naming its own unanswered request. Under nextest's 60s slow-timeout, so not even flagged slow — against 240s of unexplained SIGTERM before.
- **`/usr/bin/true` upstream**: all three fail on EOF, fast.
- **Orphaned grandchild on the failure path**: reproduced (`sleep 400` survived with ppid 1). Verified identical on base — not introduced here.

### Reported, not folded in

**The upstream probe accepts an impostor.** A 20-line script answering `allowed:true` to everything passes all three tests — identically on the base commit. This refutes the docstring's *"Nothing but `assay-mcp-server` produces that"*. Closing it means the probe must also send a call the inner policy **denies** and assert `allowed:false`; no impostor ignoring the YAML could satisfy both. That is a change to #1988's property, not to this port, so it is flagged here for its author rather than folded into a refactor.

**Non-JSON stdout is now fatal where it used to be skipped.** `assay mcp wrap` relays non-JSON upstream lines verbatim; `assay-mcp-server` writes its logs to stderr, so the risk is latent, not live. Arguably correct strictness for an MCP stdio transport — but it is a behavior change and it is now recorded.

### For the module's owner (not this PR)

Both measured, both pre-existing, both belong with `jsonrpc_conn_selftest.rs`:

1. **No per-line ceiling.** `read_line` grows until a newline; a child emitting none is unbounded. 495 MB resident growth over a one-second flood, versus 2.9 MB with a 1 MiB ceiling.
2. **Selftest misses the defect it exists for, ~1 run in 8.** With the deadline check deleted from `next_line`, the selftest detected it 7/8. `READ_AHEAD_LINES` backpressure lets the channel momentarily drain, so `recv_timeout` reports `Timeout` on its own and bounds the run without the check. The memory bound partially undermines the flood test's ability to reproduce the defect.
3. The selftest asserts only *that* the read panicked fast, never *why* — a mutation making `read_json` fail on the first blank line would keep it green.

### Quorum

The review above is an independent non-building agent on `77831f14`. `89998ba2` changes two things in response to it, so it does **not** carry to this head under AGENTS.md §Review Quorum — a fresh look is needed on `89998ba2`.

Rul1an:
## Second adversarial review — results

A fresh independent adversarial agent reviewed `f3e637de`. Both fixes from the first round hold up under its own repros. It then found things the first round missed, including **two false claims in my own commit message**. Head is now `c17ba7fe`.

### Both prior fixes confirmed, by better repros than mine

- **`read_json` over `read_response`**: it could not use my repro (replacing the server binary is silently overwritten by the test's own inner `cargo build`), so it interposed on the *proxy* instead — a wrapper that prefixes every frame with a `notifications/progress`. Result: 3 FAIL on this head, 3 PASS with `read_response` substituted at the five call sites. The reader choice is the difference between an assertion that can fail and one that cannot.
- **`exclude`**: `cargo package --list` diff is exactly one line, 344 → 343. It also proved the "not a release blocker" claim rather than taking my word: a throwaway crate with a dangling `#[path]` test *packages and verifies clean* while `cargo test` on it fails.
- **Whole package**: `cargo nextest run -p assay-cli` → 605 passed, 1 skipped. It also confirmed cargo tracks the cross-package `#[path]` (touching the shared module recompiles the CLI test target), so there is no stale-binary hazard.

### Corrected: two claims in my commit message were wrong

**The timeout got worse, not better.** I wrote "against 240s of unexplained SIGTERM before". That describes #1995's mcp-server tests, **not this file's base**. This file was already bounded at 5s by #1987, and already named the request:

```
BASE: FAILED in  6.07s — "timed out after 5s waiting for the tools/call response
                          to the denied skill_check request"
HEAD: FAILED in 30.62s — DEFAULT_TIMEOUT
```

With `retries = 1` that is ~60s per wedged test against ~12s. Still bounded, still well under nextest's 120s kill, still red — but it is a **cost**, and adopting a shared module means adopting its budget. Narrowing it belongs to the module, not to a caller. The commit message now says so.

**"Faithful port" was overstated.** A parity harness against identical child output found real disagreements: the old reader took the next line *starting with `{`* and skipped the rest as log noise; `read_json` takes the next non-blank line and fails if it is not JSON. `warning: ...` used to be tolerated; a bare `[]` used to be skipped. Nothing writes such a line today, so it is a latent tightening — arguably the right one for an MCP stdio transport — but it is a change and it was undisclosed.

### New: the `exclude` had no guard, so the fix would silently rot

`exclude` is **silent when it matches nothing**. Rename this test, or add another `#[path]`-outside test, and the broken tarball returns with no cargo warning. That makes the fix a point-in-time patch rather than a guarantee — a gate that lives in a manifest line nobody re-checks.

`scripts/ci/check_assay_cli_publish_shape.sh` now covers it. It walks the package's `.rs` files, resolves every `#[path]`, and fails on any that escapes without being excluded. Static like the dependency check beside it: PR-time, no network, no target dir.

Verified it bites rather than only ever passing:

```
with    exclude → exit 0  "no packaged file #[path]-includes outside the package"
without exclude → exit 1  "tests/e2e_mcp_wrap_assert_cmd.rs includes
                           ../../assay-mcp-server/tests/jsonrpc_conn/mod.rs"
```

### Corrected and widened: the fixtures do not pin enforcement

My earlier note said "the upstream probe accepts any responder". The hole is bigger and it is in the **proxy** — the thing actually under test. A ~60-line fake `assay` that never opens `--policy` and never spawns an upstream at all passes all three tests. **Identical on base `84530251`**, so pre-existing, not introduced here.

The fixtures pin the strings a proxy emits, not the enforcement behind them. Closing it means asserting something only real enforcement can satisfy — for the probe specifically, also sending a call the inner policy *denies* and asserting `allowed:false`, which no impostor ignoring the YAML could satisfy. That is a change to #1988's property, flagged here for its author rather than folded into a refactor.

### Still open for the module's owner

Unchanged from the previous comment, all pre-existing, all measured: no per-line ceiling (495 MB vs 2.9 MB), the selftest misses its own defect ~1 run in 8 due to queue backpressure, and the orphaned grandchild on the failure path.

### Quorum

Two independent adversarial reviews now, on `77831f14` and `f3e637de`. `c17ba7fe` changes the commit message and adds the guardrail check in response to the second, so under AGENTS.md §Review Quorum neither carries to this head.

Rul1an:
## Merge decision and quorum record

Repo owner decided to merge on green. Recording the quorum position rather than leaving it implicit.

**Formally, quorum is not satisfied.** AGENTS.md §Review Quorum excludes the builder's self-review, and both adversarial reviews are bound to earlier heads (`77831f14`, `f3e637de`). Rewritten history does not carry a review even when the tree is identical. This merge proceeds on the owner's decision, not on a satisfied quorum.

**Equivalence between the reviewed head and this one**, for the record:

```
git diff --stat f3e637de..c17ba7fe
 scripts/ci/check_assay_cli_publish_shape.sh | 56 +++++++++++++++++++++++++++++

git diff --quiet f3e637de..c17ba7fe -- crates/assay-cli/tests/e2e_mcp_wrap_assert_cmd.rs \
                                        crates/assay-cli/Cargo.toml
 → exit 0   (both reviewed files byte-identical)
```

The entire delta since the reviewed head is the CI guard that the second review itself asked for, plus commit-message corrections it demanded. No reviewed code changed.

**Findings disposition**, all recorded above in this thread:

| | |
|---|---|
| Introduced, fixed | very-next-frame weakened (`read_response` → `read_json`); tarball shipped an uncompilable test target (`exclude`); `exclude` had no regression guard (guardrail check added, verified to bite) |
| Introduced, disclosed not fixed | wedged-proxy latency 5s → 30s — adopting a shared module means adopting its budget; narrowing it belongs to the module. Reader tightened on non-JSON stdout lines — latent, arguably correct |
| Pre-existing, reported for their owners | fixtures pin emitted strings, not enforcement (a fake `assay` passes all three, identically on base); shared module has no per-line ceiling (495 MB vs 2.9 MB); its selftest misses its own defect ~1 in 8; orphaned grandchild on the failure path |

Nothing actionable is silently dropped.

PR #2004, opened 2026-08-05
Ask libtest for its tally, not the per-test line it splits in two

Rul1an:
Follow-up to #1986, which merged while its last commit was still in review. Six review rounds on that PR, and three more here, have all landed on the same guard: the one asserting that the E6a.3 credential no-pass-through test actually **executed**, since `cargo test` exits 0 when zero tests match and the test sits behind a feature gate.

## Why the previous form had to change

#1986's final commit declined to make the guard more precise, arguing that anchoring it would invert the failure direction. That is refuted by its own history — both forms extracted and run against a real `#[ignore]`d libtest log:

| form | anchored? | on an `#[ignore]`d test |
|---|---|---|
| `f5b35949` | no anchor at all | **green** — fail-open |
| `ef903dc3` | status-anchored | **red** — fail-closed |

Anchoring was tried there and failed closed. It also set itself a falsifiable exit condition — the compromise stays "until someone has a form that is precise AND fails closed" — and reviewers falsified it.

## What the guard does now

Two checks, both required, neither able to rescue the other into a green.

**1. A test by that name was reported.** `grep -qE '^test .*test_no_passthrough_e2e'` — presence only, never the status word. libtest flushes `test NAME ... ` and its status as separate writes, so this test's inherited server stderr can land between them. That split destroys the *status* field and leaves the *name* intact, which is the distinction five earlier versions missed.

**2. The binary's own tally says everything ran and nothing was skipped.** Default-deny: anything that is not positively "some passed, none failed, none ignored" fails, so an unrecognised shape cannot certify anything. `exit` after the first match keeps the value single-line, because the glob is greedy across newlines.

## What review found in the intermediate forms

A tally-only version — the first attempt here — had two false greens, both reproduced by a reviewer rather than reasoned about:

- **Deleting the test passed the guard.** Replacing `test_no_passthrough_e2e` with a do-nothing `no_passthrough_placeholder_does_nothing` went green, because the tally carries counts and never names. This was a regression against the form being replaced. Check 1 closes it.
- **A two-line tally satisfied the glob**, taking its count from one line and its zeros from the next. Latent behind file ordering that nothing asserts. `exit` closes it.

A second reviewer falsified a claim this branch had made about itself: `CARGO_TERM_COLOR: never` is **not** load-bearing. Driving the current guard against the real coloured log from the run that failed *without* the pin gives green — the awk strips colour itself. Remove the `gsub` instead and it goes red. The pin is defence in depth; the `gsub` is the guarantee. (The colour comes from `Swatinem/rust-cache` exporting `CARGO_TERM_COLOR=always` into `$GITHUB_ENV`.)

## Eight shapes, both directions

| shape | result |
|---|---|
| plain pass | green |
| split line at `--test-threads=1`, `ignored` in child stderr | green |
| coloured pass, and Windows `tests\` separator | green |
| test replaced by a differently-named passing test | **red** |
| two-line tally combining `1 ignored` with a later `0 ignored` | **red** |
| `#[ignore]`d | red |
| feature off, zero tests | red |
| nextest (prints no tally) | red — fail-closed |

Plus the two real logs from the original measurement. The guard block is extracted verbatim from the workflow and driven against them rather than retyped.

## Corrections to #1986's merged record

Two sentences there are false and are why the flaw looked unreachable:

- *"Hosted runners are multi-core so libtest emits the unsplit shape"* — it splits at `--test-threads=8` too; a reviewer reproduced it 7 times in 25 runs with a flooding child. What actually prevents it here is that the test reaps its child before returning.
- *"No `--test-threads=1` anywhere in the repo"* — `crates/assay-registry/docs/TRACEABILITY.md:221`. Another package's docs, so the conclusion survived; the sentence did not.

One framing in an earlier version of this description was sharper than its source: the merged text attached "the one round 3 already produced once" to *the outcome*, not to anchoring as its cause. The substantive rebuttal above stands on the measured table, not on that reading.

## Checks

- 288 tests pass with `--features test-outbound`; 287 without (the delta is the gated test)
- All pre-commit hooks pass, including the workflow linter
- `actionlint` reports one pre-existing `unexpected key "queue" for "concurrency"` on `ci.yml`, present on `main` as well — not introduced here
- No behavioural change to any test; this touches the CI guard and two comments

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -607,47 +607,41 @@ jobs:
       # E6a.3 no-pass-through: E2E invariant test (requires test-outbound feature).
       #
       # Runs on every OS, because this is the only job that RUNS the test — the workspace runs
-      # above and below build it out, and the feature-enabled clippy step near the top of this
-      # file compiles it but executes nothing. That also means a missing or misspelled feature
-      # flag here would be silent: `cargo test` exits 0 when zero tests match. So assert the test
-      # actually executed rather than trusting the exit code.
+      # above and below build it out, and the feature-enabled clippy step in the `clippy` job
+      # type-checks it without ever producing a runnable binary. That also means a missing or
+      # misspelled feature flag here would be silent: `cargo test` exits 0 when zero tests match.
+      # So assert the test actually executed rather than trusting the exit code.
       - name: Test MCP no-pass-through E2E (E6a.3)
         shell: bash
         run: |
           set -euo pipefail
           cargo test --locked -p assay-mcp-server --features test-outbound --no-fail-fast no_passthrough \
             2>&1 | tee "$RUNNER_TEMP/e6a3.log"
-          # Two questions. Every previous version of this guard folded them into one pattern and
-          # got one of them wrong: too strict and it called a passing test absent, too loose and
-          # it called a skipped test present. Pass-versus-fail is not among them — `set -o
-          # pipefail` settles that above, so a test that ran and failed never reaches here.
+          # `cargo test` exits 0 when zero tests match, so the exit code above proves nothing ran
+          # wrong — not that anything ran. This asserts execution, and it asks libtest's own tally
+          # for this binary rather than reading the per-test line.
           #
-          # The two are sequential, not independent: question 2 rescans the lines question 1
-          # matched. That is a known and deliberate compromise. Broadening question 1 to survive
-          # interleaved child output means arbitrary server stderr can end up inside the string
-          # question 2 searches, so a log line containing the word "ignored" can trip it. The
-          # alternative — anchoring question 2 to the status field — is more precise but inverts
-          # the failure direction: a shape it fails to anchor passes the guard, and the step goes
-          # green on an invariant nobody ran. Being noisily wrong beats being quietly wrong here,
-          # so this stays until someone has a form that is precise AND fails closed.
+          # Reading the per-test line is what five earlier versions of this guard did, and each
+          # got it wrong in one direction or the other. The reason is structural: libtest flushes
+          # `test NAME ... ` and its status separately, so anything this test's inherited server
+          # stderr writes in between lands on that line. Match the name loosely enough to survive
+          # that and you have taken arbitrary child output into the string you then search for a
+          # skip word; match it tightly and a split line reads as absence. The tally has neither
+          # problem — it is one line, written after the run, in a fixed shape.
           #
-          # 1. Did the runner report this test at all? Deliberately broad, because the shapes are
-          #    many: at --test-threads=1 libtest prints `test NAME ... ` before running, so this
-          #    test's own inherited stderr lands between the name and the status word; a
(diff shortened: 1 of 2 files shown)


Rul1an:
Closing this. Two independent reviews of the final head landed on the same conclusion, and the second one is decisive: **this is a net regression against the guard already on `main`**, so keeping `main` is the safer outcome.

## Why

`main`'s guard inspects the name and the status **on the same line**. This PR replaced that with two checks — `grep` for the name across the whole log, and an `awk` block scoped to the first `Running` line whose path contains `no_passthrough_e2e`. Nothing binds them to the same subject. The reviewer's phrasing: it establishes `A(x) ∧ B(y)`, never `A(x) ∧ B(x)`.

Reproduced — the invariant test does not execute and the step goes green:

```
     Running tests/no_passthrough_e2e.rs (…)
test no_passthrough_smoke ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; …
     Running tests/zz_real.rs (…)
test test_no_passthrough_e2e ... ignored          <- the real test, skipped
```

`awk` stops at the first block and never sees `1 ignored`; `grep` finds the name on the ignored line. A/B on identical logs: this PR **green**, `main` **red**. `#[ignore]` on a flaky spawn-a-server test is the routine path this guard exists to catch, and it is exactly what the change loses.

## What this PR did fix, and why it does not justify the trade

It removed two false reds in `main`'s form: a split `test NAME ... ` line at `--test-threads=1`, and the same at higher thread counts with a flooding child (reviewed measurement: 25/25 splits at `tt=1`, 19/25 at `tt=8` through a pipe). Neither is reachable in CI today — no job sets `--test-threads=1`, and the test reaps its child before returning, so nothing is writing when the status prints. Trading a live false-green for a latent false-red is the wrong direction for a guard whose whole purpose is proving a security invariant ran.

## If someone picks this up

The correct shape is one `awk` pass that binds name and tally to the **same block**: track per-block whether a line names the test, and at block end require that block's tally to be clean. That closes all three findings at once — the cross-binary split (F1), the unanchored `Running` close-rule that lets inherited stderr end a block early (F3), and the substring match on the test name (F4). Note also that `^[[:space:]]*Running` is safe once colour is stripped: the `gsub` runs first, so the anchor never needed removing.

## Corrections to the record that stand regardless

Three factual errors were confirmed during review and are worth carrying forward, since two of them are in #1986's merged squash message and cannot be edited:

- *"Hosted runners are multi-core so libtest emits the unsplit shape"* — false; it splits at `--test-threads=8` too. What prevents it here is that the test reaps its child before returning.
- *"No `--test-threads=1` anywhere in the repo"* — `crates/assay-registry/docs/TRACEABILITY.md:221`.
- Cargo's coloured output in CI comes from **`dtolnay/rust-toolchain`**, which writes `CARGO_TERM_COLOR=always` into `$GITHUB_ENV`, not from `Swatinem/rust-cache`. That matters for `public-crate-policy` and `ebpf-smoke-ubuntu`, which use the toolchain action without the cache action and are therefore coloured too.

The branch is left in place; nothing here is lost.

Issue #2006, opened 2026-08-05
ADR-045: Landlock seal checker primitives — read fixtures from disk, introduce a trust set, drop batchRoot


Rul1an:
Tracks: #1998 · Follows: #2000, #2001 · Applies: ADR-042, ADR-043, ADR-045

> **Rescoped 2026-08-05.** This issue originally bundled fixture policy, six key-scope negative controls, `batchRoot` removal, payload-only rules, and a mutation meta-test. Reading `scripts/experiments/aee_landlock_seal_fixture.py` showed that four of those five rest on primitives that do not exist yet: the checker never reads a fixture from disk, and there is no trust set, so three of the six controls are not expressible. The negative controls moved to #2011. This issue is now the primitives slice they depend on.

## Goal

Make the ADR-045 Landlock seal checker verify an *input* against a *trust set*. Today it does neither.

Acceptance (two sentences): a fixture file with a single flipped byte is rejected, and a structurally valid seal signed by a keyid outside the declared trust set is rejected. Neither is achievable in the current harness.

### 1. The checker does not read the fixtures

```python
def load_case(name: str) -> dict[str, Any]:
    # The committed JSON files are review markers for the fixture names. The
    # canonical fixture bodies are generated here so the checker and emitter
    # cannot drift in this first slice.
    return case_statement(name)
```

`validate()` always runs against a freshly generated statement. The committed JSON is never opened.

This invalidates the drift check proposed in the original scope (`--emit` + `git diff --exit-code`). That check does detect emitter drift, but emitter and checker are the same function, so they cannot diverge by construction and nothing tests whether the checker rejects a file whose bytes are wrong. The "marker-only vs full on-disk" question was therefore the wrong question — the real one is whether the checker consumes an input at all.

### 2. There is no trust set, so three negative controls are not expressible

```python
def sign(payload, keyid):
    return b64(hmac(SECRET + keyid.encode(), pae(PAYLOAD_TYPE, canonical_bytes(payload))))
```

The signing key is derived *from the keyid*. Every keyid produces a verifying signature, so `untrusted-signing-key` cannot fail: a fixture with `keyid: "attacker-key"` verifies fine. The same holds for `wrong-key-role` and `key-outside-validity-window`. These are not missing fixtures; they are a missing primitive.

### 3. `trustedKeyScopes` is written and never read

The field is set at two sites (base statement `_ext`, and the `fixture-key-production-scope` case) and read at zero. The only key check in `validate()` is:

```python
[...]

PR #2007, opened 2026-08-05
A documented example output was tracked as a source change

Rul1an:
## What

`examples/privileged-action-gate/README.md` tells the reader to run `./sarif.sh enforcement.sarif`, so every reader of that example leaves an untracked `enforcement.sarif` in the worktree. No `.sarif` is tracked anywhere in this repository, and the neighbouring run outputs — `junit.xml`, `sarif.json`, `run.json`, `summary.json`, `otel.jsonl` — are already ignored by bare filename. This one was missed.

Ignored by the same bare-filename convention rather than a broad `*.sarif`, which would silently swallow a future tracked fixture.

## Also in this change

Deletes the stale `codex/release-3.38.0` branch, local and remote. **Recovery SHA: `dd2b3c3ae0f85a98b102ff90ef79f92d8dc2b178`.**

`git cherry` reported its three commits as unlanded. That is patch-id noise from surrounding context, not real divergence — verified at file level instead:

- `.github/workflows/release.yml` byte-identical to `origin/main`
- `fuzz/Cargo.lock` byte-identical to `origin/main`
- `Cargo.toml`, `CHANGELOG.md` and `fuzz/Cargo.lock` all read `3.38.0` on `main`
- `git diff origin/main <tip>` restricted to the five files those commits touched is empty

## Not in this change

One untracked file is deliberately left alone: `docs/superpowers/plans/2026-06-27-hotspot-loc-under-600-refactor.md`. Sibling plan docs *are* tracked under `docs/superpowers/plans/`, so this is either an uncommitted plan or a leftover — that is an author call, not a cleanup call.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ benchmark_results/
 *.log
 junit.xml
 sarif.json
+enforcement.sarif
 run.json
 summary.json
 otel.jsonl


Rul1an:
Correcting this PR's own record: the "Not in this change" section is wrong about what happened.

It says `docs/superpowers/plans/2026-06-27-hotspot-loc-under-600-refactor.md` was deliberately left alone — "that is an author call, not a cleanup call". It was not left alone. A `git add -A docs/` while preparing #2006 swept it in, and it reached `main` through #2008 (`1619cad9`), not through this PR. Copilot flagged it there as unrelated to the checker work and suggested landing it separately; I merged past that comment.

So the file is on `main`, it got there by accident rather than by decision, and this PR body says the opposite. Recording that here rather than leaving a public statement standing that is not true.

Two follow-ups, both decided with Rul1an:

- **Keeping the file.** `docs/superpowers/plans/2026-06-05-assay-wave53-hotspot-top2-9-refactor.md` — the same class of document, an earlier wave of the same work — is already tracked there, and no sibling plan carries a status marker. That directory is a record of plans, not a queue. Removing one member of a series while its predecessor stays would be the worse inconsistency.
- **Marking it done.** Every hotspot the plan targets has since been split, so on `main` it reads as pending work that is finished. Measured: `proxy/enforce.rs` 1630→35, `tests/proxy_enforce_pdp_e2e.rs` 1565→24, `mcp/tool_decision_truth.rs` 1398→34, `proxy/mod.rs` 1390→321, `registry/supply_chain.rs` 1311→209, `cli/commands/supply_chain_conformance.rs` 860→87. A completion note follows in its own PR.

The mechanical cause is the part worth keeping: `git add -A <dir>` has a wider effect than the intent behind it, and reverting a file without changing the command just lets the next one slide in. Staging explicit paths from here.

Issue #2010, opened 2026-08-05
Two error-code registries disagree: errors::diagnostic::codes vs exit_codes::ReasonCode


Rul1an:
Split out of #2009, where this was named as out of scope.

## Problem

The codebase has two independent error-code registries that overlap, disagree, and are both public-facing.

**`assay_core::errors::diagnostic::codes`** — string constants, used by `assay validate`, `assay doctor`, `assay_core::validate`, and the agentic builder:

```
E_CFG_PARSE, E_CFG_SCHEMA, E_PATH_NOT_FOUND, E_TRACE_MISS,
E_TRACE_INVALID, E_BASE_MISMATCH, E_REPLAY_STRICT_MISSING,
E_EMB_DIMS, E_POLICY_VIOLATION,
W_CFG_VACUOUS_EXPECTED, W_BASE_FINGERPRINT, W_CACHE_CONFUSION
```

**`assay_cli::exit_codes::ReasonCode`** — an enum whose `as_str()` output is the `reason_code` field in `run.json` and `summary.json`, specified in `SPEC-PR-Gate-Outputs-v1.md`:

```
E_CFG_PARSE, E_TRACE_NOT_FOUND, E_MISSING_CONFIG, E_BASELINE_INVALID,
E_POLICY_PARSE, E_REPLAY_MISSING_DEPENDENCY, E_REPLAY_LIMIT_EXCEEDED,
E_INVALID_ARGS, E_JUDGE_UNAVAILABLE, E_RATE_LIMIT, E_PROVIDER_5XX,
E_TIMEOUT, E_NETWORK_ERROR, E_TEST_FAILED, E_JUDGE_UNCERTAIN,
E_POLICY_VIOLATION, E_SEQUENCE_VIOLATION, E_ARG_SCHEMA
```

Concrete consequences:

- **Near-duplicates for one condition.** A missing trace file is `E_TRACE_MISS` or `E_PATH_NOT_FOUND` from `validate`/`doctor`, and `E_TRACE_NOT_FOUND` from `run`. Three names, one condition, depending on which command the user ran.
- **`E_MISSING_CONFIG`, `E_TRACE_NOT_FOUND` and `E_INVALID_ARGS` do not exist in `codes`.** Anything wanting to emit a `Diagnostic` for those has to invent a constant or reuse a wrong one.
- **`E_CFG_PARSE` and `E_POLICY_VIOLATION` exist in both** with no mechanism keeping them in sync.
[...]

PR #2012, opened 2026-08-05
fix(cli): report classified run errors on stderr, not only in run.json

### LINKED ISSUE #2009: assay run reports config errors only in run.json — operator sees nothing on stderr
Blocks: #1949 (item 1, `--deny-ineffective-assertions`)

## Problem

`assay run` classifies config errors (YAML parse failure, unknown field, and — once #1949 item 1 lands — ineffective-assertion rejection) and writes them to `run.json` and `summary.json`. It prints nothing. The operator gets zero bytes on stderr and stdout, and an exit code as the only signal.

In CI this means a rejected config is indistinguishable from a crash until someone opens the artifact. It blocks #1949 item 1 specifically: a fail-closed gate whose refusal is invisible is not a usable gate.

## Root cause

The information loss happens at the call site, not in the writer.

`PipelineError::into_exit_code` (`crates/assay-cli/src/cli/commands/pipeline_error.rs`) passes only `run_error.message` into `write_error_artifacts`. `RunError` carries `path`, `detail`, `provider`, and `status` — all constructors already populate them (`RunError::config_parse` sets both `path` and `detail`) — and all four are dropped there.

`write_error_artifacts` (`crates/assay-cli/src/cli/commands/reporting.rs`) is an artifact writer. It is not given anything to print, so "just add an `eprintln!`" would print a bare message with no code, no source, and no context.

`Diagnostic` (`crates/assay-core/src/errors/diagnostic.rs`) already renders exactly what an operator needs and is already used by `assay validate`. Nothing routes the run path through it.

## Design

Build the operator diagnostic in `pipeline_error.rs` from the **same `ReasonCode`** that goes into `run.json`, so the two channels cannot disagree.

### Why not `RunError::to_diagnostic()` in assay-core

That was the first sketch and it is wrong. `errors::diagnostic::codes` and `exit_codes::ReasonCode` are two separate registries that do not agree:

- `codes` has `E_CFG_PARSE`, `E_CFG_SCHEMA`, `E_PATH_NOT_FOUND`, `E_TRACE_MISS`, `E_TRACE_INVALID`, `E_BASE_MISMATCH`, `E_REPLAY_STRICT_MISSING`, `E_EMB_DIMS`, `E_POLICY_VIOLATION`, plus three `W_*`.
- `ReasonCode::as_str()` emits `E_CFG_PARSE`, `E_TRACE_NOT_FOUND`, `E_MISSING_CONFIG`, `E_BASELINE_INVALID`, `E_POLICY_PARSE`, `E_REPLAY_MISSING_DEPENDENCY`, `E_REPLAY_LIMIT_EXCEEDED`, `E_INVALID_ARGS`, and the infra/test codes.

`E_MISSING_CONFIG`, `E_TRACE_NOT_FOUND` and `E_INVALID_ARGS` do not exist in `codes`, and `E_TRACE_MISS` / `E_PATH_NOT_FOUND` are near-duplicates of `E_TRACE_NOT_FOUND`. Adding the missing three to `codes` would print a code on stderr that differs from the `reason_code` in `run.json` — the same class of defect this issue is about. `ReasonCode` is a CLI contract, so the conversion belongs in assay-cli.

### Shape

```rust
// crates/assay-cli/src/cli/commands/pipeline_error.rs
use assay_core::errors::{Diagnostic, RunError};
use std::io::IsTerminal;

fn diagnostic_for(run_error: &RunError, reason: ReasonCode) -> Diagnostic {
    let mut d = Diagnostic::new(reason.as_str(), &run_error.message)
        .with_source(run_error.path.as_deref().unwrap_or("<config>"));

    let mut ctx = serde_json::Map::new();
    if let Some(detail) = &run_error.detail {
        if detail != &run_error.message {
            ctx.insert("detail".into(), detail.clone().into());
        }
    }
    if let Some(provider) = &run_error.provider {
        ctx.insert("provider".into(), provider.clone().into());
    }
    if let Some(status) = run_error.status {
        ctx.insert("status".into(), status.into());
    }
    if !ctx.is_empty() {
        d = d.with_context(serde_json::Value::Object(ctx));
    }
    d
}

fn emit_operator_diagnostic(d: &Diagnostic) {
    let rich = std::io::stderr().is_terminal() && std::env::var_os("NO_COLOR").is_none();
    eprint!("{}", if rich { d.format_terminal() } else { d.format_plain() });
}
```

`into_exit_code` already computes `reason`; it now feeds both `diagnostic_for` and `write_error_artifacts`. No second derivation, so no drift.

`PipelineError::Fatal` also gets a stderr line — today it returns `Err` and relies on whatever the caller does with it.

### Also in scope

`Diagnostic::format_plain()` currently calls `format_terminal()` verbatim, emoji included. Without splitting the two, the TTY branch above is a no-op that cannot be observed failing. Split it in the same PR.

## Out of scope

- **`fix_steps`.** `ReasonCode::next_step()` already exists and the docs promise per-code guidance, but wiring it in is a separate mapping with its own correctness question. Empty for now.
- **The `codes` / `ReasonCode` duplication.** Two registries with near-duplicate codes is real debt, but fixing it touches `assay validate`, `doctor`, and the agentic builder. Own issue.
- **Any change to `run.json` or `summary.json`.**

## Acceptance

`assay run` against a broken config writes a diagnostic to stderr whose `code` is byte-identical to `run.json`'s `reason_code`, and `run.json` is unchanged.

## Verification

- Three repro cases (YAML parse error, unknown field, missing config). Each asserts the stderr code, the stderr source/context content, and equality with `run.json`'s `reason_code`.
- One test running `--format json` asserting stdout contains only JSON — the diagnostic must never reach stdout.
- One test pinning `format_plain()` output as emoji-free and distinct from `format_terminal()`.
- Byte-comparison of `run.json` before and after the change on all three repro cases.

## Non-claims

This does not make any config error newly detected, does not change classification, and does not change exit codes. It makes an existing decision visible.

Rul1an:
Closes #2009.

## Summary

`assay run` classified config errors, wrote them to `run.json` and `summary.json`, and printed nothing. The operator got an exit code and no reason. This routes the same `ReasonCode` to stderr.

## Motivation

`into_exit_code()` already computed the `ReasonCode` — it just never used it for anything the operator could see:

```rust
let reason = reason_code_from_run_error(&run_error).unwrap_or(ReasonCode::ECfgParse);
write_error_artifacts(reason, run_error.message, /* ... */)
```

`run_error` carries `path`, `detail`, `provider` and `status`, all populated by the constructors. Only `.message` survived the call.

Blocks #1949 item 1: a `--deny-ineffective-assertions` gate that rejects silently is a gate nobody can act on.

## Changes

**`crates/assay-core/src/errors/diagnostic.rs`**

`format_plain()` was a pass-through to `format_terminal()`, emoji included. Wiring `assay run` to this printer unchanged would have put `❌` into every CI log and every redirected stderr. Split the body into `render(prefix)`; the icon is now the only difference. Icons are written as `\u{...}` escapes so the source stays ASCII and the variation selector on the warning sign is visible rather than invisible.

This is item 50 of `docs/architecture/CODE-ANALYSIS-REPORT.md` and appears in RFC-002's trivial-wrapper cleanup. Doing it here rather than deferring it: `format_plain` has no call sites outside its own module today, so the change is contained now and stops being contained the moment `assay run` starts calling it.

**`crates/assay-cli/src/cli/commands/pipeline_error.rs`**

- `diagnostic_for(&RunError, ReasonCode) -> Diagnostic` — code from `reason.as_str()`, fix step from `reason.next_step(path)`, context from the populated optional fields, source from `RunErrorKind`.
- `emit_operator_diagnostic(&Diagnostic)` — decorated only when `stderr().is_terminal()` and `NO_COLOR` is unset.
- One line in `into_exit_code()`, before `write_error_artifacts`.

## Three decisions worth reviewing

**No new fix-step table.** `ReasonCode::next_step()` already exists in `exit_codes.rs` and already feeds `run.json` through `RunOutcome::from_reason`. stderr calls the same function with the same `reason`. Had I written a stderr-side table, the two could disagree, and the disagreement would surface as a support ticket rather than a test failure. `fix_step_matches_run_json_next_step` pins this.

**`emit_operator_diagnostic` returns `()`, not `Result`.** The exit code is the CI gate's contract; stderr is an affordance for a human reading a log. `assay run 2>/dev/null` on a closed pipe must not be able to change the code the gate reads. So the write error is dropped deliberately, not by oversight.

**`detail` is conditional.** `RunError::config_parse` sets `message` and `detail` from the same string:

```rust
let mut err = Self::new(RunErrorKind::ConfigParse, detail.clone()).with_detail(detail);
```

Emitting every populated field would print the parse error twice. `detail` is included only when it differs. `config_parse_context_does_not_repeat_the_message` asserts the constructor's aliasing first, so if that constructor changes the test explains why it broke.

Also: `legacy_classified` surfaces as `classified_from: "message"` in context. A code inferred by matching strings against a message is weaker evidence than one from a typed constructor, and the operator should be able to tell which they are looking at.

## Not in this PR

**`errors::diagnostic::codes` is untouched.** That is #2010. `Diagnostic::new` takes `impl Into<String>`, so `reason.as_str()` needs no new constant. Adding `E_MISSING_CONFIG` to `codes` to serve stderr would have created a third spelling of a string already frozen in `SPEC-PR-Gate-Outputs-v1.md`. Findings recorded in the #2010 comment.

**`--format json` / quiet suppression.** `into_exit_code()` does not know the output format, and stdout is untouched, so JSON consumers are unaffected. A future `--quiet` belongs as an explicit parameter here, not an env lookup.

**The ineffective-assertion repro** named in the issue body. That rejection is not a `RunError` until #1949 item 1 lands. When it is, it needs no code here — it arrives classified and gets a diagnostic for free. That is the point of routing through `ReasonCode`.

## Tests

Ten new tests, four in `diagnostic.rs` and six in `pipeline_error.rs`. The load-bearing one is `stderr_code_matches_run_json_reason_code`: for each typed constructor, the stderr code equals `reason.as_str()` equals `RunOutcome.reason_code`. `plain_rendering_is_ascii_and_leads_with_the_code` asserts the full non-TTY render is ASCII, which fails if anything reintroduces decoration upstream.

**I have not compiled or run these.** No Rust toolchain in this environment. Reviewer should run:

```
cargo test -p assay-core errors::diagnostic
cargo test -p assay-cli pipeline_error
cargo clippy --workspace --all-targets -- -D warnings
```

Two things I would check first: `RunOutcome` is `pub` in `crate::exit_codes` (the test module imports it), and `RunErrorKind` is exported from `assay_core::errors` (`mod.rs` re-exports `Diagnostic` explicitly; `RunErrorKind` is defined in that module and should be reachable, but the import is worth a compile).

## Risks

`format_plain()`'s output changes. Nothing outside `diagnostic.rs` calls it today, so no consumer moves — but if a snapshot test asserts on it, that snapshot will need updating.

## Rollout / Rollback

No config, no flag, no migration. stderr gains text on error paths that previously printed nothing; exit codes and artifact contents are unchanged. Revert is the two commits.

## Summary by CodeRabbit

* **New Features**
  * Added operator-facing diagnostics for classified pipeline errors, including error codes, context, and suggested fixes.
  * Added terminal-friendly severity indicators with color support.
  * Added plain-text diagnostic output for environments without terminal formatting.

* **Bug Fixes**
  * Classified pipeline errors now display actionable diagnostics before error artifacts are written.


Rul1an:
## Review response + a defect the tests were hiding

Two review notes and one bug I found by executing the logic rather than reading it. Pushed as `6f8c079`.

---

### Copilot: `source_for` moving out of a borrowed field

Applied — `source_for` now matches on `&run_error.kind`.

For the record though, the original almost certainly compiled. Every arm of that match is a unit-like variant pattern, and Rust's default binding modes mean matching a place expression against unit-like patterns is a discriminant read, not a move. `RunErrorKind` has no non-`Copy` payload to move out of. This would only be a real error if a variant carried data and an arm bound it by value.

Taking it anyway: the reference is equally clear, and a reviewer who has to work out whether a match moves has already paid the cost of the ambiguity.

### CodeRabbit: fix steps out of scope for #2009

Disagreeing, with a reason.

The issue body lists `fix_steps` under "out of scope" meaning *don't invent a new fix-step registry*. `ReasonCode::next_step()` already exists in `exit_codes.rs` and already populates `run.json`'s `next_step` through `RunOutcome::from_reason`. No new mapping was added. stderr calls the function that was already the source of truth, with the same `reason`, and `fix_step_matches_run_json_next_step` asserts the two agree.

The alternative — a diagnostic with a code and no fix step, while `run.json` right next to it carries one — would be a worse artifact and would have put the drift back in by hand.

---

### The actual bug: the two channels printed different fix steps

`into_exit_code` passed `run_error.path` to the stderr diagnostic. `write_error_artifacts` hardcoded `None`:

```rust
let mut o = RunOutcome::from_reason(reason, Some(message), None);
```

So for every path-bearing error:

```
stderr  : Run: assay doctor --config assay.yaml
run.json: Run: assay doctor --config <config.yaml>
```

That is precisely the drift #2009 exists to eliminate, introduced by the commit meant to eliminate it, on the one axis nobody was looking at because both channels agreed on the *code*.

**Why the test suite passed.** `fix_step_matches_run_json_next_step` built its expected value like this:

```rust
let outcome = RunOutcome::from_reason(reason, Some(msg), None);  // <- None
```

The same `None` production was wrongly using. The assertion compared the bug against itself. A test that constructs its oracle by re-deriving the production expression tests that the expression equals itself, and passes for exactly as long as the bug is symmetric.

The fix threads `context` through `write_error_artifacts`, and the tests now go through `outcome_as_written()`, a helper that mirrors the production call site. If the two ever diverge again, the helper diverges with production and the assertion fails.

**How I found it.** No Rust toolchain in my environment, so I transliterated the pure logic — `next_step`, `RunError` constructors, `diagnostic_for`, `format_plain` — into an executable model and ran the ten PR tests plus adversarial cases against it. Nine passed. The drift case failed. That is a model of the logic, not the Rust, so it proves nothing about types or borrows; it is exactly why Copilot's note is worth checking against a compiler rather than against me.

### Two more from the same exercise

**`plain_rendering_is_ascii_and_leads_with_the_code` asserted the wrong thing.** `format_plain` strips the icon, not the message. A YAML parser quoting a non-ASCII source line — `found character 'é' that cannot start any token` — produces non-ASCII plain output and fails that assertion, though nothing is wrong. The test was asserting a property of the fixture. Replaced with: the code leads the line, and neither icon codepoint appears. Added a non-ASCII case pinning it.

**`RunErrorKind::Other` gets `E_CFG_PARSE`.** `reason_code_from_run_error` returns `None` and `into_exit_code` substitutes `ECfgParse`, so an unclassified error now prints `[E_CFG_PARSE]` with `source: unknown` and advises running `assay doctor --config`. That predates this PR and already shapes the exit code and `run.json` — stderr just makes it legible. Pinned in a test with a comment pointing at #2010, since a unified registry wants a real `E_UNKNOWN` rather than a config-shaped guess. Not fixing it here: changing that fallback changes exit codes.

---

### Still needs a compiler

I could execute the logic; I could not typecheck it. Unresolved:

```
cargo test -p assay-core errors::diagnostic
cargo test -p assay-cli pipeline_error
cargo clippy --workspace --all-targets -- -D warnings
```

`write_error_artifacts` has one call site, which this commit updates, so the signature change should be contained — worth a compile to confirm.

Bencher's two alerts on `sw/12xlarge` and `sw/50x400b` are unrelated: these paths run once, on the error branch, after the suite is already over.

PR #2013, opened 2026-08-05
Build a seal the checker accepts, and claim nothing about phase

Rul1an:
Refs #1998, ADR-045. Producer side of the primitive whose checker landed in #2006 — up to, but not including, signing. **Rewritten twice after adversarial review; still draft.**

## What two review rounds established

The first version refused `restrict_self_confirmed` for being a start-time fact that persists in the record, then accepted `probe`, which had exactly the same property. The committed June carrier fixture was seal-eligible.

The second version added a run-phase challenge and was worse in an instructive way: it verified inside the producer and **never reached the payload**. From a consumer's seat nothing changed. A producer-side refusal is the producer marking its own homework.

RFC 9334 was cited correctly and instantiated wrongly — in RATS the nonce comes from the Verifier and is covered by the Attester's signature; here it was checked against a `String` the same caller supplied, over an unsigned field that never travelled.

## What this version does instead

**Use the mechanism AEE already has.** `aeeObservedSet` is a digest over the leaves of emitted `interception`/`examination` records — a commitment, in the spec's words, "by a party that does not control the carried set". A seal committing to a leaf could not have been computed before that leaf existed, and any consumer can recompute it.

So the run-end probe is **emitted as an `examination` record**, and the seal commits to it.

That one change closes two self-declared labels:

| Label | Before | Now |
|---|---|---|
| phase (`aeeKind: sealed`) | a string the producer chose | the seal demonstrably follows the probe |
| `assayDropProofModel` | a constant asserted beside it | under synchronous-probe the probe is the sole committed observation — readable from the statement |

**Derive, don't accept.** `aeeRunBinding` and `aeeObservedSet` are computed here; both derivations already existed in the checker, so taking bare `String`s was a missing call, not a design question. Canonicalisation via `assay-canonical` (RFC 8785), PAE via `assay_common::dsse`.

The `H(0x00 || PAE)` leaf is implemented **as specified, not tidied**. It looks like a half-built RFC 6962 tree — leaf domain tag, no node tag anywhere, and a sort that buys nothing because nobody generates inclusion proofs — but it is normative AEE v0.7 and the spec states its purpose. Recording that so nobody "fixes" it into non-conformance.

## Parity is now a real cross-implementation check

`--emit` writes `derivation-parity.json` with the inputs and expected digests; the Rust tests **read that file**. The previous parity test compared against a literal typed into the Rust source and would have stayed green through any Python-side rename.

## Mutation-checked, and one bite got through

| Bite | Result |
|---|---|
| falsify the run-binding preimage | 1 test fails |
| leaf domain `0x00` → `0x01` | 1 test fails |
| don't emit the probe as a record | 3 tests fail |
| hardcode `assaySourceSchema` again | 1 test fails |
| **omit `leaves.sort_unstable()`** | **all 12 passed** |

The sort was invisible because the parity vector had one leaf, and sorting one element is a no-op. AEE v0.7 specifies "sorted ascending by UTF-16 code unit" — a producer emitting in arrival order matches on one observation and diverges on two, silently, only in production. A second vector now emits two leaves out of sorted order, and the test asserts they *are* out of order before comparing. 13 tests, 5/5 bites caught.

## Kept from round one

ABI 4 floor (`LANDLOCK_ACCESS_NET_CONNECT_TCP` does not exist below it, so an `EACCES` there did not come from Landlock), record self-consistency, RFC 3339 seal instant, and `assaySourceSchema`/seal scope read from the record rather than asserted.

## Non-claims

No production AEE support, no stable export, no signing semantics, no key material, no complete run-population proof, no independent substrate operation, no agent safety. `aeeVersion` stays `0.7`.

## Still open

- **The second vantage.** ADR-045 chose Option C, so this carries one collection path. `aeeObservedSet` gives seal-after-set; it does not order the set's members against each other (the set is sorted by hash and `seq` is outside the leaf). Ordering the probe against the interceptions needs a party that is neither the producer nor the assembly plane — which is what Landlock ABI 7's `AUDIT_LANDLOCK_ACCESS` records, with their kernel-minted monotonic serials, would give. Tracked in `docs/contributing/TODOS.md` as `landlock-abi-v5`.
- **Checker gaps** found alongside this: #2014.


Rul1an:
## Round 6 review, and what research changed

Six adversarial rounds now. The last one checked the commit messages against the code — the right check, since two earlier claims had been silent no-ops — and found a third: a sentence reported reworded twice had never been edited. Fixed, and verified absent this time rather than assumed.

### A research finding that overturned an assumption

`CountedQueue` assumed a channel can expose a trustworthy loss counter. For the most likely channel it cannot: **a BPF ring buffer has no built-in lost-samples counter.** `BPF_RB_AVAIL_DATA` and the consumer/producer positions are the only introspection, and the [kernel documentation](https://docs.kernel.org/bpf/ringbuf.html) says they are momentary snapshots "to be used only for debugging/reporting reasons or for implementing various heuristics". Reading them as a loss count carries zero on a channel that dropped events.

The usable source is a counter the BPF program increments on a failed reservation. That is now in the doc comment, because it is the mistake the next implementer would make.

### Two members, from SLSA and Proof-or-Stop

`CountedQueue` verified its counters and discarded them, leaving a consumer told a check passed with nothing to recheck — the same shape as the run-phase challenge this PR already withdrew.

- `assayDropProofBasis` — `checked` or `declared`
- `assayDropChannels` — `name=lost` per channel, empty when the basis is `declared`

The split follows [SLSA v1.2](https://slsa.dev/spec/v1.2/build-provenance): `externalParameters` are the values passed through the interface, and metadata about them belongs beside them rather than folded in. Proof-or-Stop ([arXiv 2607.14890](https://arxiv.org/abs/2607.14890)) states the rule it encodes — a report that "provides neither verified integrity nor attested execution and carries no binding" is not admissible evidence. A caller-declared model is exactly that; the payload now says so instead of reading identically to a checked one.

It is the `observed`/`unknown` basis split `observed-effect v0` already carries, one field over.

### Mutation harness

`scripts/ci/test_aee_seal_producer_mutations.py`, pre-push. Sixteen mutations, each naming the test that must catch it; all sixteen caught. Three properties, each from a specific failure while writing this PR: it restores from a scratch copy rather than `git checkout`, so uncommitted work survives; a search string that does not appear is a hard failure, because a bite that never applied looks exactly like one that was caught; and the table is Python after the shell version shredded every entry containing a Rust closure on its `|` delimiter.

It has already earned itself twice — it found a real gap (`source schema derivation`) and, this round, two edits that silently never ran because an assertion aborted the script mid-sequence.

### Still open, deliberately

- `SynchronousProbe` remains a declaration with no evidence behind it. Not closeable here — this module cannot observe the collection path. The change is that the payload now says which it is.
- Channel-list completeness is a caller obligation, unverifiable here, now named rather than implied.
- `build_sealed_run` has no callers yet. The reviewer's point stands: that is the moment to make the harder variant the easier one.
- Ordering the probe against the run needs a party that is neither the producer nor the assembly plane. Landlock ABI 7's `AUDIT_LANDLOCK_ACCESS` records, kernel-minted with monotonic serials, are the candidate; tracked in `docs/contributing/TODOS.md` as `landlock-abi-v5`.

Issue #2017, opened 2026-08-05
The proof pack records which gates ran, and nothing reads it


Rul1an:
## Problem

The delegated proof pack records, per gate, whether it passed, failed, was skipped,
was incomplete, or never ran. That array is the only place in the proof that says
*which* gates executed. Nothing reads it.

`lane-check` validates the gate **label** — `manifest["inputs"]["gates"]` — against
`accepted_gates(expected)`, and stops there. The top-level `manifest["gates"]` array
has no consumer in the verification path. The only reads of it are two lines
inside the producer's own `--self-test`.

Consequence: a proof pack that records

```json
"gates": [
  {"gate": "kernel-only",                     "status": "passed"},
  {"gate": "kernel-policy",                   "status": "passed"},
  {"gate": "openai-agents-kernel-policy",     "status": "passed"},
  {"gate": "openai-agents-hidden-write",      "status": "passed"},
  {"gate": "gemini-google-genai-kernel-policy","status": "missing"}
]
```

is accepted as `gates=all` on the strength of `inputs.gates == "all"` alone.

## Evidence

Verified read-only against `origin/main` @ `36960909`.

**No consumer in the verification path.** Every occurrence of the `gates`
manifest key in `scripts/ci/assay_runner_lane_check.py` (a bare `grep -n gates`
matches 38 lines; these are the ones touching the key):

- `:1279` — `gate = str(inputs.get("gates") or "")`, the label
- `:1280` — compared against `accepted_gates(expected)`
- `:1812` — a test fixture, `"inputs": {"gates": "all", "build_ebpf": "true"}`

A repository-wide sweep over `*.py`, `*.sh`, `*.rs`, `*.yml` for
`["gates"]`, `get("gates")`, `.gates` finds three files: the delegated workflow
(`inputs.gates`), the pack builder, and lane-check. Two of the pack builder's
occurrences do read the array — `assay_runner_delegated_proof_pack.py:559` and
`:568` — but both sit inside its own `--self-test`, asserting a five-gate status
map on a synthetic pack. Nothing on the path that credits a proof reads it.

Those two lines are worth noting for a second reason: a reader of the array
already exists, so Direction 1 below is a smaller change than it looks.

**The producer does record it faithfully.** `assay_runner_delegated_proof_pack.py:378`

```python
gates = [collect_gate(gate, proof_root, output_root) for gate in selected_gates]
```

[...]

PR #2018, opened 2026-08-05
Four gated paths are not content-addressed, and the docs named one

Rul1an:
## Description

`content_provenance_paths` in `scripts/ci/assay_runner_gated_paths.json` covers the
runner crates, monitor, ebpf and xtask. It does not cover `crates/assay-cli`. Four
gated paths therefore cannot be proven by tree OID:

```
Cargo.toml
Cargo.lock
crates/assay-cli/src/cgroup.rs
crates/assay-cli/src/cli/commands/runner_spike.rs
```

The behaviour around them is already correct. `uncovered_content_provenance_files`
is consumed at `assay_runner_lane_check.py:826` and rejects proof reuse for those
paths when the proof head differs from the PR head. What was missing is smaller and
duller: nothing forced the coverage question to be answered when a gated path is
added, and no document named the set. The two documents that discuss proof reuse
mentioned neither.

Four changes, all Gate.NONE.

**1. An invariant, not a constant** (`scripts/ci/assay_runner_lane_check.py`)

`_test_content_provenance_coverage_is_decided` asserts the set *derived* from
`assay_runner_gated_paths.json` rather than restating it. Adding a gated path fails
the self-test until someone decides whether it belongs under a content-provenance
prefix.

A hardcoded `NO_CONTENT_ADDRESS_GATE_PATHS` constant was the obvious shape and is
deliberately not what landed. It would be a third home for a rule that already lives
in the JSON manifest and in `content_provenance_covers_path`, and it would move in
lockstep with itself, so it could never fail.

**2. `ci-lanes.md` names the four paths** next to the proof-reuse rule, with the
reason `crates/assay-cli` sits outside the prefixes.

**3. The delegated runbook gains a Lane-Check Interaction Failures section.**

Four consumer-side causes of a delegated run that passes while the PR stays blocked,
none of them visible in the delegated workflow's own logs: the proof binding to the
dispatched ref rather than the PR, the `lane-check` job versus the `lane-check/proof`
required status, read-only tokens on fork and Dependabot `pull_request` events, and
proof reuse on content-identical trees.

A grep of the runbook for `head_sha|fork|dependabot|lane-check|expected_head|status`
previously returned two incidental hits and nothing normative.

**4. The exit-40 rule in `ci-lanes.md` now says why it holds.**

`run_gate` returns any non-zero status under `set -euo pipefail`, so a skip fails the
job and cannot reach the proof pack as a pass. The rule was already enforced; it read
as convention. The wording says "gate scripts that guard on host prerequisites"
rather than "the gate scripts", because four of the nine three-run-determinism
scripts carry the guard and five do not.

## Type of Change

- [x] Documentation update
- [x] Test coverage (self-test invariant)

## Verification

- [x] `python3 scripts/ci/assay_runner_lane_check.py --self-test` passes
- [x] **Negative control**: adding an uncovered path to `all_gate_paths` in
      `assay_runner_gated_paths.json` fails the self-test with `AssertionError`.
      The manifest was restored; `git diff` on it is empty.
- [x] Full pre-commit suite green, including the `Assay-Runner lane-check self-test`
      hook that runs the new invariant
- [x] All three changed files classify as `Gate.NONE` via `classify_file`
- [x] The new relative link from `ci-lanes.md` resolves to an existing file

No delegated run was dispatched and none is claimed. `cargo` targets are unaffected;
no Rust source changed.

## Related

Companion to #2017, which reports the finding this work did not fix: the proof pack
records a per-gate status array that no consumer reads. That is a behaviour change to
`lane-check` and belongs in its own PR.

Scope deliberately excluded: the choice between "runbook is normative" and "code is
normative" for delegated-proof freshness. This PR records current behaviour and makes
no claim about which should govern.

## Checklist

## Summary by CodeRabbit

- **Documentation**
  - Clarified that delegated gate exit status `40` and prerequisite-check failures are treated as failed gates.
  - Documented troubleshooting guidance for dispatched references, required commit statuses, fork and Dependabot permissions, and proof reuse.
  - Identified four gated paths that always require a fresh proof when changed.
  - Clarified that content-identical trees may allow proof reuse only when no fresh-proof paths are modified.

### DIFF
--- a/docs/ops/ASSAY-RUNNER-DELEGATED-RUNBOOK-2026-05-21.md
+++ b/docs/ops/ASSAY-RUNNER-DELEGATED-RUNBOOK-2026-05-21.md
@@ -194,7 +194,7 @@ That is deliberate rebuild avoidance, and it is guarded: four gated paths sit
 outside the content-provenance prefixes and are therefore not provable by tree
 OID at all.
 
-```
+```text
 Cargo.toml
 Cargo.lock
 crates/assay-cli/src/cgroup.rs
--- a/docs/ops/ASSAY-RUNNER-DELEGATED-RUNBOOK-2026-05-21.md
+++ b/docs/ops/ASSAY-RUNNER-DELEGATED-RUNBOOK-2026-05-21.md
@@ -148,6 +148,65 @@ Reason: this host is the environment where Linux, cgroup v2, Node 22, and the
 eBPF artifact are expected to exist. If a delegated script skips, the runner or
 workflow has drifted from its contract.
 
+This is enforced, not left to discipline. `run_gate` records any non-zero
+status as a failed gate and returns it under `set -euo pipefail`, so exit `40`
+fails the job like any other failure. Note that not every gate script has such
+a guard; the ones that do are the ones checking host prerequisites.
+
+## Lane-Check Interaction Failures
+
+A delegated run can pass and still leave the PR blocked. These four are the
+recurring causes; all four are consumer-side, none is visible in this
+workflow's own logs.
+
+**The proof binds to the dispatched ref, not to the PR.** The proof pack
+records `--head-sha "$GITHUB_SHA"`, which is the head of whatever ref you
+picked in the Run-workflow dropdown. There is no `expected_head_sha` input on
+this workflow; that input exists only on `assay-runner-lane-check.yml`.
+Dispatching against `main` while reviewing a PR produces a valid, attested,
+useless proof pack: lane-check compares the recorded head SHA against the PR
+head and rejects it. Always select the PR branch as the ref, and dispatch only
+after the branch holds the final candidate commit, since a later push
+invalidates the proof.
+
+**A red `lane-check` job is not a red `lane-check/proof`.** The Actions job is
+named `lane-check`; the required commit status is named `lane-check/proof`, and
+branch protection on `main` requires the status. A job that fails before
+reaching the status-posting step leaves the status stale or absent, which
+blocks the merge without producing a matching red job on the current head. When
+a merge is blocked and the job looks green, read the commit status directly.
+
+**Fork and Dependabot PRs cannot post the status from the PR event.** Posting
+needs `statuses: write`. On `pull_request` events from a fork or from
+Dependabot the token is read-only by GitHub policy, so the status cannot be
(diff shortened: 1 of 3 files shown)


Rul1an:
Closing rather than patching. An adversarial review across three independent lenses found the central factual claim false, and the title states it: it is not four gated paths, it is 26 files across five surfaces. `runner-fixtures/**` and `tests/fixtures/runner-spike/**` are gated by hardcoded branches in `classify_file` and are not content-addressed either.

**Why closed and not fixed.** Every added documentation paragraph carried at least one false or overstated claim:

- "four gated paths … not provable by tree OID at all" — 26, across five surfaces
- "lane-check compares the recorded head SHA against the PR head and rejects it" — on the attested path every head_sha comparison is manifest-vs-run (`:1230`); the real criterion is tree OIDs, and `:1314` passes the run's own SHA
- "that input exists only on `assay-runner-lane-check.yml`" — four workflows declare `expected_head_sha`
- "only the runner crates, monitor, ebpf and xtask are covered" — `scripts/ci` and two workflow files are covered too
- "in exactly this case and in no other" — the guard keys on `event_name == "pull_request"` and includes 401
- "unconditionally" — five upstream guards plus two early returns
- "`crates/assay-cli` is deliberately outside those prefixes" — asserted intent whose only support was a docstring this PR added. The set traces to the canonical eBPF build's input list, not to a decision about proof reuse.

**The mechanism was wrong, not the wording.** The invariant derived the gated surface from `assay_runner_gated_paths.json`. Gating is decided by `classify_file`, which reads that manifest in 2 of its 11 rules. Three rounds of patching the same rule — paths, then paths plus prefixes — each surfaced a new defect on an axis that had not been examined. That is a signal to change the subject, not to patch again.

**What survives, now filed properly.** #2020 records the live coverage hole this uncovered: `crates/assay-cli/src/cli/commands/runner_spike.rs` is a 17-line facade classified `Gate.ALL` while all eight files of the module holding the actual behaviour are `Gate.NONE`, and the manifest still pins `crates/assay-cli/src/cgroup.rs`, which no longer exists. That is a real gap in the lane contract and larger than anything this PR attempted.

A correctly-scoped invariant — derived from `classify_file` over `git ls-files` — belongs in its own PR against #2020, without this one's premise.

Branch recovery SHA: `28072962`.

Issue #2034, opened 2026-08-05
Seven commands accept any --format string and silently produce a different format


Rul1an:
Split out of #2032, which asked whether `LintArgs.format` had the same shape as `--fail-on`. It does, it is not confined to `lint`, and in three commands the result is an empty artifact with a normal exit code.

## No format argument in the CLI is validated

Eleven files declare `pub format: String`. There is **no `value_parser` anywhere in assay-cli**, so clap accepts any string for any of them. Seven then match on it with a silent `_` fallback:

| command | match | fallback goes to |
|---|---|---|
| `evidence lint` | `evidence/lint.rs:99` | text |
| `evidence diff` | `evidence/diff.rs:54` | text |
| `coverage` (legacy) | `coverage/legacy.rs:227` | text |
| `discover` | `discover.rs:107` | text |
| `explain` | `explain.rs:136` | terminal |
| `profile-next` | `profile_next/mod.rs:293` | summary |
| `sandbox --profile-format` | `sandbox/profile.rs:30` | **YAML** |

## Why three of these are silent rather than merely wrong

In `evidence lint`, `evidence diff` and `coverage`, the structured branches write to **stdout** and the fallback writes **only to stderr**. Counted directly: `lint.rs:100-132` (the json and sarif arms) has 2 `println!` and 0 `eprintln!`; `lint.rs:133-219` (the fallback) has 0 `println!` and 17 `eprintln!`. `diff.rs:58-90` is 0 and 7; `coverage/legacy.rs:237-270` is 0 and 1.

So this:

```bash
assay evidence lint bundle.tar.gz --format jsonn > report.json
```

produces an **empty `report.json`**. Nothing is written to stdout at all.

And the exit code does not change: the format match closes at `lint.rs:220` and the `--fail-on` threshold is computed after it at `:223`, independent of which arm ran. A CI step gets its usual exit code and an empty artifact.

For `--format sarif` the same typo yields an empty file where a SARIF upload expects a document.

## The sharpest case writes the wrong content to the user's file

`sandbox/profile.rs:30-38`:

```rust
let content = match args.profile_format.as_str() {
    "json" => crate::profile::writer::write_json(&suggestion)?,
    _ => crate::profile::writer::write_yaml(&suggestion)?,
};
let out_path = args.profile.as_ref().expect("profiler active");
crate::profile::writer::save_atomic(out_path, &content)?;
```

[...]

Issue #2025, opened 2026-08-05
normalize_severity silently downgrades an unrecognized severity to note, which can turn a failing run into exit 0


Rul1an:
Sub-issue of #2010. Found while reviewing the exit path for that issue; distinct defect, same file.

## What is wrong

`normalize_severity` in `crates/assay-cli/src/cli/helpers.rs:5-16` recognizes `error`, `warn`/`warning`, `note`/`info`, and ends with an unconditional fallback:

```rust
    "note"
}
```

Anything unrecognized becomes `"note"`. `decide_exit` (`helpers.rs:29-33`) then computes `has_error` by comparing against `"error"` and returns `exit_codes::OK` when nothing matches.

So a `Diagnostic` carrying a misspelled or newly-introduced severity is not reported as unknown — it is silently reclassified as the least severe level and drops out of the error set. If it was the only error, the command exits 0.

`Diagnostic.severity` is a `String` (`crates/assay-core/src/errors/diagnostic.rs:6`), so nothing at the type level prevents this. Today's producers all emit `error` or `warn`, so the defect is latent rather than live — but it is latent in the direction of passing, not failing.

## Direction

An unrecognized severity should be a loud failure, not a downgrade. Either make the severity a typed enum so the case cannot arise, or keep the string and treat an unknown value as an error condition in its own right.

## Acceptance

- An unrecognized severity value cannot result in a lower exit code than the same diagnostic with a recognized one.
- A test constructs a diagnostic with an unknown severity and asserts the run does not exit 0.

Rul1an:
Investigated. The defect is confirmed latent, and both wider and narrower than this issue says.

## Wider: thirteen call sites, not one

This issue frames `normalize_severity` as gating the exit code. It gates thirteen call sites across six files:

| what it decides | site |
|---|---|
| the SARIF `level`, i.e. what GitHub Code Scanning shows | `report/sarif.rs:197,226` |
| error and warn counts in `assay validate` and `assay fix` | `validate.rs:99,105`; `fix.rs:167,172` |
| the order diagnostics are printed in | `validate.rs:153,154` |
| whether `assay doctor` sees any error at all | `doctor/implementation.rs:133` |
| which fixes `doctor` offers | `doctor/fixes.rs:45,178` |
| the process exit code | `cli/helpers.rs:38` |

So an unrecognized severity does not only lower the exit code. It drops the finding out of SARIF as an error, out of the counts, and out of what `doctor` recognizes as a problem.

[...]

PR #2036, opened 2026-08-05
--fail-on none was documented, shipped, and never implemented

### LINKED ISSUE #2032: An unrecognized --fail-on value silently selects the most permissive gate
`assay evidence lint --fail-on <typo>` does not fail. It silently applies the loosest threshold available and exits 0 where the user asked for a stricter gate.

## The path

`LintArgs.fail_on` is a bare `String` with no `value_parser` (`crates/assay-cli/src/cli/commands/evidence/lint.rs:23-25`):

```rust
/// Fail (exit 1) if findings at or above this severity exist
#[arg(long, default_value = "error")]
pub fail_on: String,
```

So clap accepts anything. The value is then matched at `lint.rs:223-228`:

```rust
let threshold = match args.fail_on.as_str() {
    "error" => Severity::Error,
    "warn" | "warning" => Severity::Warn,
    "info" => Severity::Info,
    _ => Severity::Error,
};
```

## Why the fallback is the wrong direction

`Severity::Error` is not the strictest threshold. It is the loosest. From `crates/assay-evidence/src/lint/mod.rs:150-156`:

```rust
pub fn has_findings_at_or_above(&self, threshold: &Severity) -> bool {
    self.findings.iter().any(|f| match threshold {
        Severity::Error => f.severity == Severity::Error,
        Severity::Warn => f.severity == Severity::Error || f.severity == Severity::Warn,
        Severity::Info => true,
    })
}
```

`--fail-on info` fires on everything; `--fail-on error` fires only on errors. So the fallback hands an unrecognized value the gate that lets the most through.

Concretely, `--fail-on warnings` — plural, a plausible typo for the documented `warn` — is accepted, silently becomes `Severity::Error`, and a bundle whose only findings are warnings exits 0. The user asked to fail on warnings and got a run that cannot fail on warnings, with no diagnostic saying so.

## Why this is worse than #2025

#2025 describes the same shape in `normalize_severity` and is **latent** — nothing produces an unrecognized severity, so no user can reach it. This one is reachable from the command line today, and it fails open rather than closed. A gate that silently widens is the failure mode a gate exists to prevent.

The documented surface does not help: `docs/AIcontext/entry-points.md:239` says `lint --fail-on <SEVERITY>: Fail on severity threshold` without listing the accepted values, so there is nothing to check a typo against.

## Direction

Reject the unrecognized value rather than interpreting it. Either give the argument a `value_parser` so clap refuses it before the command runs — which also produces the accepted-value list in `--help` for free — or keep the `String` and make the `_` arm an error. The first is preferable: it moves the check to where the untrusted string arrives, and `assay_evidence::lint::Severity` already derives the variants that would back it.

While there: `LintArgs.format` (`lint.rs:19-21`) is the same shape — a bare `String` with `default_value = "text"` and no parser. Worth checking whether it has the same silent fallback.

## Acceptance

- An unrecognized `--fail-on` value produces a non-zero exit and names the accepted values.
- No `--fail-on` value can result in a gate looser than the one the user named.
- A test asserts that a bundle with warning findings and `--fail-on <invalid>` does not exit 0.

Rul1an:
Closes #2032.

## What was wrong

`assay-action/action.yml:888` passes `--fail-on none`, and the comment above it says exactly why:

```yaml
# Lint with packs (generates pack-enhanced SARIF). Use --fail-on none so
# policy findings do not masquerade as pack-loading/config errors.
```

`none` was not one of the match arms in `evidence/lint.rs`. It fell to `_ => Severity::Error` and gated on error findings anyway.

The consequence lands in the Action's own handler. A policy finding makes `evidence lint` exit 1; `action.yml:896-907` treats anything that is not exit 3 as a pack fault:

```bash
if [ "$EXIT_CODE" -eq 3 ]; then ...missing_pack / invalid_pack...
else PACK_STATUS="runtime_error"; PACK_ERROR_KIND="runtime_error"; fi
echo "::error::Compliance pack lint failed (${PACK_ERROR_KIND}): ..."
break
```

So a policy finding was reported as a pack runtime error and the loop broke before the remaining bundles were linted — word for word the thing line 878 says the flag prevents. The silent fallback inverted the mitigation into the failure it was written against.

`none` was published as well: `docs/AIcontext/quick-reference.md:112` documents the action input as `error | warn | info | none`.

## The change

`none` now means no threshold, and the rule moves out of `cmd_lint` into `gate_threshold` so it is testable without a bundle — the only fixture in the tree (`tests/fixtures/evidence/test-bundle.tar.gz`) has zero findings and cannot tell any two thresholds apart.

`--fail-on` carries a value parser over the five accepted spellings, so the next unimplemented value fails at the argument boundary rather than being reinterpreted:

```
error: invalid value 'nope' for '--fail-on <FAIL_ON>'
  [possible values: error, warn, warning, info, none]
```

## Behaviour change

Anyone passing `--fail-on none` was silently getting error-gating and will now get what they asked for. That includes this repo's own Action, whose pack-lint step will stop reporting policy findings as `runtime_error`.

Note the direction of the old fallback: `error` is the **narrowest** gate, not the strictest — `has_findings_at_or_above` fires only on error findings for that threshold and on everything for `info` (`assay-evidence/src/lint/mod.rs:150-156`). An unrecognized value therefore got the most permissive gate available, which is why this was worth fixing before the latent cases in #2025.

## Scope

Only `--fail-on`. The `--format` family was #2034, merged as `d8b3bdb2`; this reuses the value-parser approach it established.


Rul1an:
An adversarial review broke four claims in this PR. Three are fixed in the latest commit; the fourth is a real regression that this PR should not land silently, and I have verified it myself rather than relaying it.

## What was wrong and is now fixed

**The tests did not test the gate.** The previous `fail_on_gate.rs` stayed green with the bug fully restored. It asserted acceptance and exit 0 against a bundle I had declared empty after measuring it under one pack. `tests/fixtures/evidence/test-bundle.tar.gz` is not empty — `--pack eu-ai-act-baseline` yields one Warn finding, `--pack soc2-baseline` one Error. Verified directly.

The discriminating case is narrower than it looks: `none` reinterpreted as `Error` is **invisible on a warning bundle**, because that threshold fires only on errors. My first correction used the warning pack and still survived the mutation. Only `soc2-baseline` separates them. Restoring `"none" => Some(Severity::Error)` now fails exactly one assertion; removing the value parser fails two more.

**#2032's third acceptance criterion is now met** on the artifact it names — a warning-finding bundle with an unrecognized threshold exits 2.

**The doc comment cited the wrong argument.** It offered `docs/AIcontext/quick-reference.md:112` as justification for this CLI flag. That line documents the *action input* `fail_on`, which `action.yml:527-549` implements in bash and never forwards; the only `--fail-on` in the Action is the hard-coded literal at `:888`. Corrected.

## The regression this PR introduces

`action.yml:487` computes the counters that feed the `fail_on` gate from a lint call **without** `--pack`, so compliance-pack findings have never entered that gate. The exit-1 → `runtime_error` → `exit 3` path at `:892-921` was, accidentally, the only thing that failed the Action on a compliance-pack error finding.

This PR removes it. A user who sets `pack: soc2-baseline` to gate CI on compliance violations gets `pack_status=success` and a green action. `assay-action/README.md:226` — "policy findings still follow normal `fail_on` threshold behavior" — becomes false for the pack path.

That is a defensible trade, since the old behaviour mislabelled a policy finding as a pack runtime error and aborted the loop before the remaining bundles were linted. But it is a trade, not a pure win, and the PR body presented it as one.

**I am not landing this until that is resolved.** The options are to gate the pack path properly, or to correct the README claim and the pack quickstart so they stop implying a gate that does not exist. That is a decision about product behaviour, not a detail of this fix.

## Also surfaced, not fixed here

`packs/open/cicd-starter/README.md:23,28` tells users to set `fail_on: warning`. The bash `case` at `action.yml:527-549` has arms for `error`, `warn`, `info` and `none` and **no default**, so `warning` falls through and the gate never fires. This is #2032's exact shape, live, in a shipped pack README, and pinned in place by `scripts/ci/review-adr023-starter-docs-followup.sh:45`. It needs its own issue.

Correction to this PR's own narrative while I am here: "a policy finding made lint exit 1" is true only of **error-severity** findings. The repo's own `action-v2-test.yml:203-229` uses `fail_on: none` with `eu-ai-act-baseline`, whose finding is a warning — so the bug never fired in this repo's CI, and the PR body implied it was.

Rul1an:
The compliance-pack gate is in (`action.yml`, after the step's outputs so `pack_score` and `pack_applied` stay readable when it fires, exiting 1 for a policy failure rather than the 3 this step uses for pack faults). `$FAIL_ON` reaches the CLI unmodified, so `warning` works there as an alias of `warn`.

Verified by `bash -n` over the extracted 134-line step body and the `check-yaml` hook. **Not verified end to end** — this path needs a real Action run, which I cannot produce here.

The bash `case` at `action.yml:527-549` still has no `warning` arm and no default, and `packs/open/cicd-starter/README.md:23,28` tells users to set exactly that. Filed as #2037 rather than folded in: it is a different implementation of the same rule, in a different language, and deciding whether it should exist at all is a design question this PR should not settle.

Issue #2041, opened 2026-08-05
The manifest names which gates ran, not which commands they were


Rul1an:
## Problem

Direction 2 of #2017, kept alive because closing that issue on Direction 1 alone
would drop it.

#2038 makes the per-gate status a consumed contract: a proof is rejected unless
every recorded gate is `passed`. What the manifest still does not say is **which
command each gate name ran**.

`collect_gate` records `gate`, `status`, `archives`, `selected_json`,
`pass_lines` and `gate_log`. It does not record the script path. The binding
from gate key to script exists only in the workflow's `case` branch, and
#2038 pins it there with `GATE_SCRIPTS` — a hardcoded map in the verifier,
asserted against the workflow text.

That assertion is worth having and is not the durable answer. It is parity
between two files, both of which a PR can edit, checked by a third place that
also has to be edited in step. The manifest — the thing that is signed and
attested — stays silent on the question.

## What Direction 2 asks for

`collect_gate` records, per gate, the script path and its digest as executed.
The verifier then reads what the run actually invoked rather than inferring it
from a workflow file it did not run.

This is the concrete scope for the `commandSetHash` item on the agentic-lineage
ADR. #2017 named it and said "(2) is the durable answer. They compose."

## Why it matters after #2038

Three drift shapes are caught today. One is not, and it is the one this fixes:

```
gate commented out                 -> caught (parity assertion)
gate rewired to a weaker script    -> caught (GATE_SCRIPTS pin)
run_gate line deleted              -> caught (parity assertion)
gate wrapped in a false condition  -> NOT caught by parity
```

The fourth is caught indirectly — a gate that does not run leaves no directory,
so the producer records `missing` and #2038 rejects it. That works, and it is a
second-order argument. A manifest that recorded the executed script and its
digest would answer the question directly instead.

It would also retire `GATE_SCRIPTS`, which is a hardcoded map that has to be
maintained by hand and whose only defence is that the self-test fails when it
drifts.

## Not urgent

Nothing here is fail-open today. #2038 closed the fail-open; this is about
whether the record states what happened or requires cross-referencing two other
files to reconstruct it.

## Related

- #2017 (closed by #2038 on Direction 1)
- #2040 — the legacy path, which bypasses #2038's check entirely and is the more
  urgent of the two follow-ups

Rul1an:
PR #2044 attempted Direction 1 of this and was closed. Recording why, so the next attempt does not repeat it.

[...]

PR #2043, opened 2026-08-05
Failing validation was no worse than never being validated

### LINKED ISSUE #2040: A proof that fails validation falls through to the path that never validates
## Problem

`run_check` credits a PR when **either** path accepts:

```python
fallback_passed = valid_run is not None and sha_ok and gate_ok   # :1790
passed = attested.accepted or fallback_passed                     # :1791
```

The attested path downloads the proof pack and validates its manifest. The
legacy path does not. `find_valid_delegated_run` (`:1258`) reads only the
workflow run's `name`, `event`, `head_sha` and `conclusion`; `recorded_gate_ok`
(`:1594`) regexes the PR body or a comment for `gate: all`. No manifest is
fetched, so nothing the manifest says is checked.

This is not new and not accidental. `ci-lanes.md:75` calls it a compatibility
fallback for the transition, and it long predates any manifest validation.

What is new is that it now bypasses a check. #2017 was closed by requiring every
recorded gate to be `passed`, and that requirement lives entirely on the
attested path. A proof pack recording `gemini-google-genai-kernel-policy:
missing` is rejected there and credited here, because here the pack is never
read.

## Measured

Driving the real `run_check` with a stubbed API, a run whose pack records one
gate `missing`:

```
attested path accepted = False
commit status  = passed=True  "legacy delegated proof accepted: gates=all"
run_check exit = 0     (PR is credited)
```

Same outcome with no proof-pack artifact at all.

## Why this is a decision, not a patch

Closing the fallback means every runner-impacting PR must carry an attested
pack. That is almost certainly right now — the attested path has been the
default for a while, and every delegated run in recent memory produced a pack —
but it is a contract change for everyone, and it removes the escape hatch that
exists for when artifact attestation itself is unavailable.

Options, none chosen here:

1. **Remove the fallback.** Cleanest. Needs a check that no open PR relies on it,
   and a note in `ci-lanes.md`.
2. **Keep it, gate it.** Accept the fallback only when the attested path failed
   for an infrastructure reason (artifact missing, attestation API down) rather
   than for a validation reason. As written, a pack that fails validation falls
   through to the fallback and is credited — which is the worst ordering.
3. **Keep it, narrow it.** Require the recorded run to have produced a pack,
   even if the pack is not validated. Weak, but it removes the "no artifact at
   all" case.

Option 2 is the smallest change that removes the perverse ordering: today,
failing validation is not worse than never being validated.

## Provenance

Read at `origin/main` @ `9cd0b0e4` and reproduced against the PR branch for
#2038. Found by adversarial review of that PR.

Rul1an:
## Description

Closes #2040. **Stacked on #2038** — that PR added the per-gate execution check
this one stops being bypassable. Merge #2038 first.

`run_check` credits a PR on `attested.accepted or fallback_passed`, and the
fallback never downloads a manifest. `find_valid_delegated_run` reads the run's
name, event, head SHA and conclusion; `recorded_gate_ok` greps the PR body for a
gate name. Every check on the attested path — attestation statement, manifest
semantics, predicate, `gh attestation verify`, claims, and as of #2038 the
per-gate execution status — was bypassable by having a recorded run.

The perverse part is the ordering: a pack that **fails** validation falls
through to the branch that **never** validates. A rejected proof and an absent
proof reach the same verdict, and every strengthening of the attested path
widens the gap to the weaker one.

## Approach

The current consumer-side answer to this shape is a no-downgrade policy —
pnpm's `trustPolicy: no-downgrade` fails when a package's evidence level weakens
between versions, and exists because the 2025 s1ngularity attack shipped
versions lacking the provenance attestations their predecessors had. The
principle transfers: do not accept weaker evidence than was available.

Applied here, the distinction that matters is *why* the attested path failed.

`AttestedProofCheck` carries `evidence_rejected`, set at the five failure sites
that occur after a pack has been obtained and parsed — statement, semantics,
predicate, verify, claims — and not at the three that mean no evidence was
available: unreadable run, wrong workflow or conclusion, pack not downloadable.

`fallback_applies` returns `False` for the first class.

## The fallback keeps its purpose

`ci-lanes.md:75` describes it as a compatibility path for the transition.
Retention expiry, an unreachable attestation API, and runs predating the
attested lane all still reach it. What it loses is the case where evidence was
available and rejected.

| Situation | Fallback |
|---|---|
| no attested evidence, run recorded | applies |
| no attested evidence, nothing recorded | no |
| **pack obtained and rejected, run recorded** | **no** (was: yes) |
| pack obtained and rejected, nothing recorded | no |

## Verification

`fallback_applies` is a function rather than an expression inside `run_check`,
so `_test_rejected_evidence_does_not_fall_back` exercises what the consumer
actually calls.

**Verified to bite:** replacing the rule with `return eligible` fails the suite.
An earlier draft of this PR asserted on a copy of the expression and stayed
green after the rule was removed — the same shape as the missing coverage
#2038's review found, so it is worth naming rather than quietly fixing.

## Diagnostics

When the fallback is suppressed, the commit status says so. A maintainer reading
a red `lane-check/proof` is pointed at the attested diagnostics rather than left
wondering why a recorded run did not count.

## Scope

`Gate.NONE` — `scripts/ci/assay_runner_lane_check.py` is the proof consumer,
which runs on GitHub-hosted Ubuntu. No delegated dispatch required or claimed.

## Not addressed

#2041 — binding each gate key to the script path and digest as executed. Larger:
it changes the producer (`Gate.ALL`, so a delegated run) and the manifest shape.

## Checklist

### DIFF
--- a/docs/reference/runner/ci-lanes.md
+++ b/docs/reference/runner/ci-lanes.md
@@ -68,7 +68,9 @@ cleanup and checkout.
    `assay-runner-delegated-proof-pack-<run_id>` contains `manifest.json`,
    `subject-checksums.txt`, and `attestation-bundle.json`. Lane-check verifies
    the GitHub artifact attestation, the proof-pack subjects, the selected gate,
-   and the gated-path tree OIDs. During the transition, recording the workflow
+   the per-gate execution status recorded in the manifest, and the gated-path
+   tree OIDs. A pack in which any recorded gate is not `passed` is rejected,
+   so `gates=all` means five gates ran rather than that the label said so. During the transition, recording the workflow
    run URL, commit SHA, selected gate, and result in the PR body or a PR comment
    remains a compatibility fallback.
 4. Do not treat a delegated skip as success. In the delegated lane, exit `40`
--- a/docs/reference/runner/ci-lanes.md
+++ b/docs/reference/runner/ci-lanes.md
@@ -72,7 +72,10 @@ cleanup and checkout.
    tree OIDs. A pack in which any recorded gate is not `passed` is rejected,
    so `gates=all` means five gates ran rather than that the label said so. During the transition, recording the workflow
    run URL, commit SHA, selected gate, and result in the PR body or a PR comment
-   remains a compatibility fallback.
+   remains a compatibility fallback -- but only where no attested evidence was
+   available. A proof pack that was obtained and rejected on its contents is not
+   credited by the fallback, so failing verification is not equivalent to never
+   having been verified.
 4. Do not treat a delegated skip as success. In the delegated lane, exit `40`
    means the runner contract has drifted.
 5. Do not bypass required repository checks for runner-impacting changes.
--- a/scripts/ci/assay_runner_lane_check.py
+++ b/scripts/ci/assay_runner_lane_check.py
@@ -118,6 +118,12 @@ class AttestedProofCheck:
     accepted: bool
     run: dict[str, object] | None
     diagnostics: tuple[str, ...]
+    # True when a proof pack was obtained and its contents were rejected, as
+    # opposed to no attested evidence being available at all. The legacy
+    # fallback exists for the second case; letting it cover the first is a
+    # downgrade, and would make failing validation no worse than never being
+    # validated.
+    evidence_rejected: bool = False
 
 
 @dataclass(frozen=True)
@@ -1541,6 +1547,7 @@ def find_valid_attested_proof(
     expected: Gate,
 ) -> AttestedProofCheck:
     diagnostics: list[str] = []
+    evidence_rejected = False
     for run_id in run_ids:
         try:
             run = dict(api.request("GET", f"/actions/runs/{run_id}"))
@@ -1560,6 +1567,7 @@ def find_valid_attested_proof(
         statement, statement_diagnostics = validate_attestation_statement(pack)
         if statement is None:
             diagnostics.extend(f"run {run_id}: {line}" for line in statement_diagnostics)
+            evidence_rejected = True
             continue
 
         semantic_diagnostics = validate_proof_manifest_semantics(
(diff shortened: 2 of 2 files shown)


Rul1an:
Closing. This introduces a bypass worse than the one it closes, and the research says the design is the wrong shape.

**It let attacker-controlled bytes reach a security predicate.** The verify-stage classifier decides whether a failed `gh attestation verify` counts as a verdict by substring-matching gh's stderr. That stderr echoes the malformed value out of `attestation-bundle.json` — pack content, written by the PR author. Reproduced:

```
bundle mediaType contains 'not available'
  -> gh: 'unsupported media type: ...;v=not available'
  -> marker hit -> evidence_rejected=False -> fallback CREDITS the PR
benign rejection
  -> marker miss -> evidence_rejected=True  -> fallback denied
```

At the parent commit that site was unconditional, so this is a **regression on the axis the PR exists to protect**. And it is wording-matching, the technique the same commit disavows two paragraphs earlier while fixing the other defect structurally.

**Two further defects, both measured.** An oversized pack (>100 MiB) is bucketed as absence and credited — the cheapest bypass simply moved one branch up. And five of the six sites that set `evidence_rejected` are still individually and jointly deletable with the suite green; the commit names that gap in its own message and closes only the sixth site it added.

**The research says the shape is wrong, not the details.**

- This repository already answers "classify a verification failure": `ErrorClass` in `crates/assay-evidence/src/bundle/writer_next/errors.rs` is Integrity / Contract / Security / Limits. There is deliberately **no "could not check" class** — in the evidence path every failure is a verdict.
- The same bug class has a security advisory in the reference implementation: cosign GHSA-w6c6-c85g-mmv6, "reports false positive when payload parsing fails", root cause "a logic flaw in the error handling of the predicate type validation".
- Current guidance for an unreachable verifier is fail-closed, not fall back.

**The root cause is that the verifier is asked to infer whether evidence was *absent* or *bad*, from signals the interested party writes.** That is not a classification problem; it is an authorization one. Inference belongs nowhere near it.

The direction is recorded on #2040. Not built here: removing the inferred fallback needs an attributed override to replace it, and `assay-runner-lane-check.yml` has no override input — the privileged dispatch runs the same check, so today there would be no path to green for expired retention or a real outage.

Measured before deciding: **all four open PRs are `Gate.NONE`.** Nothing depends on this path right now, so closing costs nothing and shipping a bypass would.

Branch recovery SHA: `87a5b3d2`.

PR #2044, opened 2026-08-05
The proof named which gates ran, not what they ran

Rul1an:
## Description

Direction 2 of #2017. **Third in a stack — merge #2038, then #2043, then this.**

#2038 pinned each gate key to a script name in `GATE_SCRIPTS` and asserted it
against the workflow's `case` branch. That is parity between two files a PR can
edit, checked by a third that has to move in step. The proof itself — the thing
that is signed and attested — said nothing about which command ran.

A verifier reading the pack could see that a gate named `kernel-policy` passed.
To learn what `kernel-policy` *is*, it had to read a workflow file it did not
execute and was not attesting.

## What changed

- **`run_gate`** writes the repo-relative script path into the gate directory
  before executing it.
- **`collect_gate`** records that path with the digest of the file as it ran, so
  the pack states what each gate name executed.
- **`gate_command_diagnostics`** checks the pin against that record rather than
  against workflow text — the pin is now verified against attested evidence.

| Case | Result |
|---|---|
| gate records no script | **reject** |
| `script` is not an object | **reject** |
| gate ran a different script than its pin | **reject** |
| script recorded with no digest | **reject** |
| a gate the pin does not know | left to `gate_execution_diagnostics` |

Verified to bite: replacing `gate_command_diagnostics` with `return ()` fails
the suite.

## Path here, digest deliberately not

The recorded digest is checked for presence, not resolved against the blob at
the head. Resolving a blob is a content-provenance question and belongs with the
tree comparison, where the code that resolves refs already lives.

This is not a preference. An earlier draft put the git lookup inside
manifest-shape validation, and it forced three fixture edits chasing synthetic
SHAs through unrelated tests — the attested-accept fixture, the predicate
fixture, and the run fixture each needed a real ref for a reason that had
nothing to do with what they test. The third edit was the signal that the
placement was wrong rather than the fixtures. After the split, one fixture line
changed.

Binding the digest to the head tree is worth doing and is a separate change.

## Compatibility

A pack without the field is rejected rather than tolerated. Every pack predating
this change is already refused by the content-provenance tree comparison, since
this touches `scripts/ci` and the delegated workflow. There is no compatibility
case to keep open.

## Scope

**`Gate.ALL`** — touches `.github/workflows/runner-spike-delegated.yml` and
`scripts/ci/assay_runner_delegated_proof_pack.py`. A `Runner Spike Delegated`
dispatch with `gates=all` on this branch is required.

Not dispatched yet: this stacks on two open PRs, and a proof built on a base
that still moves would be invalidated by the merges below it.

## Related

- Direction 2 of #2017, tracked in #2041
- Stacked on #2043 (#2040), which is stacked on #2038

## Checklist

### DIFF
--- a/.github/workflows/runner-spike-delegated.yml
+++ b/.github/workflows/runner-spike-delegated.yml
@@ -214,6 +214,10 @@ jobs:
             echo "=== runner-spike delegated gate: ${label} ==="
             rm -rf "$gate_dir"
             mkdir -p "$gate_dir"
+            # Record which command this gate runs, as data the pack can attest.
+            # Without it the gate key binds to a script only through this file's
+            # case branch, which the proof cannot speak about (#2041).
+            printf '%s\n' "${script#"$PWD/"}" > "$gate_dir/script.txt"
             set +e
             sudo -E env \
               "PATH=$PATH" \
--- a/docs/reference/runner/ci-lanes.md
+++ b/docs/reference/runner/ci-lanes.md
@@ -68,7 +68,9 @@ cleanup and checkout.
    `assay-runner-delegated-proof-pack-<run_id>` contains `manifest.json`,
    `subject-checksums.txt`, and `attestation-bundle.json`. Lane-check verifies
    the GitHub artifact attestation, the proof-pack subjects, the selected gate,
-   and the gated-path tree OIDs. During the transition, recording the workflow
+   the per-gate execution status recorded in the manifest, and the gated-path
+   tree OIDs. A pack in which any recorded gate is not `passed` is rejected,
+   so `gates=all` means five gates ran rather than that the label said so. During the transition, recording the workflow
    run URL, commit SHA, selected gate, and result in the PR body or a PR comment
    remains a compatibility fallback.
 4. Do not treat a delegated skip as success. In the delegated lane, exit `40`
--- a/scripts/ci/assay_runner_delegated_proof_pack.py
+++ b/scripts/ci/assay_runner_delegated_proof_pack.py
@@ -107,12 +107,38 @@ def gate_status(gate_dir: Path) -> str:
     return "incomplete"
 
 
+def gate_script(gate_dir: Path, proof_root: Path) -> dict[str, Any] | None:
+    """Which command this gate ran, and the digest of that file as it ran.
+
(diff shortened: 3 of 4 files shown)


Rul1an:
Closing rather than patching. Adversarial review of the stack found that this reintroduces the exact deadlock #2038's own commit message removed, and #2043 closes the last way out.

**No ordering exists for renaming a gate script.** Measured:

| Attempt | Outcome |
|---|---|
| land the `GATE_SCRIPTS` pin alone (`Gate.NONE`) | base-side `--self-test` goes **red**, so `lane-check/proof` stops being posted for **every PR in the repo** |
| land the workflow rename first | `Gate.ALL`; its own pack is rejected by the base's pin |
| land both together | base-side check rejects |

`4d5ffaec` states the argument against shipping this: *"a PR adding or retiring a gate produced a pack its own verifier rejected in both directions, with no escape ... Measured."* That was removed for `GATE_SELECTIONS` and rebuilt here for `GATE_SCRIPTS`, with #2043 removing the fallback that would have been the escape.

**The design is also wrong a level down.** The record is written on the delegated host, inside the trust boundary of the gate it describes:

- `run_gate` passes `ASSAY_RUNNER_DELEGATED_PROOF_GATE_DIR` to the gate, which runs under `sudo`, so a gate can overwrite its own `script.txt` after it is written.
- `gate_script` hashes `script_repo_root() / recorded` — the file the record *names*, not the file that ran. A forged path therefore produces the correct digest of a script that never executed, and both checks return clean. Reproduced end to end.
- The pin compares basenames, so any directory satisfies it, including absolute paths that escape the repo root through `Path.__truediv__`.
- Neither self-test protects the producer: `gate_script -> return None` leaves both green, and the builder's `--self-test` is invoked nowhere in CI.

So the headline claim — that the proof now says what each gate executed — is false as built, and the honest version of it is documentation value rather than evidence value.

**What survives.** #2038's pin plus workflow parity is the right layer for drift: the workflow is content-addressed and the change shows in the diff. A host-side record adds nothing against a hostile gate and, as built, adds a deadlock.

#2041 is reopened with these findings so `commandSetHash` stays an open question rather than a half-answered one.

Branch recovery SHA: `d56c7463`.

Issue #2039, opened 2026-08-05
Close the unvalidated-string class by typing the boundary, not by fixing each site


Rul1an:
Four issues this week were the same defect: a CLI string with a fixed set of meanings, matched with a `_` arm that silently reinterpreted anything unrecognized. #2032 (`--fail-on none` was documented, shipped and never implemented), #2034 (seven `--format` arguments, three producing an empty artifact), #2025 and #2033 (severity, latent). Two were reachable from the command line and are fixed. This issue is about the ones nobody has found yet.

## The class is larger than the issues

Measured today:

| | count |
|---|---|
| bare `String` arguments under `crates/assay-cli/src/cli/` | 46 |
| of those, carrying a `value_parser` | 7 |
| `match … .as_str()` sites in `commands/` | 24 |
| files already using `clap::ValueEnum` | 14 |

Not all 46 are enumerations — paths, ids and free-form names are legitimately strings. But the seven that are validated are exactly the ones someone happened to look at, and the 14 `ValueEnum` files show the repo already knows the better shape and applies it inconsistently.

## What the fix so far does and does not do

#2032 and #2034 added `value_parser` lists. That rejects an unrecognized value at the boundary, which is the important half. It does **not** remove the `_` arm: the type stays `String`, so the match still needs a catch-all and still has to be kept in step with the parser by hand. `evidence/lint.rs` now carries a `debug_assert!` in that arm precisely because nothing structural prevents the two from drifting.

So the current state is a guard in front of the original defect, not the absence of it.

## The durable shape is not new

*Parse, Don't Validate* (Alexis King, 2019) states it: validation returns a boolean and leaves the data untyped, so every later use must re-check; parsing returns a **type**, and the proof travels with the value. Applied here: an argument with a fixed set of meanings should arrive as an enum, not as a string that is enumerated at each use.

What 2026 adds is urgency rather than technique. *Evaluation Blindness* (Bajaj, arXiv 2608.02786) classes broken measurement infrastructure as C4 and describes it as 100% silent by definition — the measurement returns a value consistent with a non-failing state and nothing flags the discrepancy. That is exactly what `--fail-on none` did for as long as it did. And because `assay_core::agentic::builder` matches on these codes as strings, a silent reinterpretation is not cosmetic: it is a branch a remediating agent cannot take.

## Proposal

[...]

PR #2050, opened 2026-08-05
Type coverage --format, the one argument two commands read differently

### LINKED ISSUE #2048: coverage --format means different things in its two modes and neither is declared
Split out of #2039, where it was the last of seven boundary arguments and the only one whose fix cannot be behaviour-preserving.

## Two commands behind one argument

`coverage/mod.rs:63` dispatches on `args.input.is_some()`: with `--input` the run goes to `generate`, otherwise to `legacy`. Both read the same `--format` string and give it different meanings.

**Input mode** — `coverage/generate.rs:162-169` validates and rejects with a message naming the accepted set:

```rust
"json" | "text"              => Ok(CoverageOutputFormat::Json),
"md" | "markdown" | "github" => Ok(CoverageOutputFormat::Markdown),
other => Err(format!("--format must be one of: json|md for --input mode (got '{other}')"))
```

**Legacy mode** — `coverage/legacy.rs:227` has arms for `json`, `markdown` and `github`, and a silent `_` that produces text.

So today:

| value | `--input` mode | legacy mode |
|---|---|---|
| `text` | JSON | text |
| `json` | JSON | JSON |
| `md` | Markdown | **text, silently** |
| `markdown` | Markdown | markdown |
| `github` | Markdown | github |

`text` and `md` mean different things depending on a flag that is not `--format`. Neither divergence is documented: `args/coverage.rs:45-46` describes the two sets in a doc comment, which is the closest thing to a declaration and is not one.

## Why an enum alone does not fix it

The other six arguments in #2039 became `ValueEnum`s with no behaviour change: same accepted values, same outputs, catch-all gone. Here a single enum over the union would carry the ambiguity into the type and make it look settled. `CoverageFormat::Md` would still mean Markdown on one path and text on the other.

## The precedent is in the same file

`coverage/mod.rs:57-58` already rejects a mode-inappropriate argument rather than reinterpreting it:

```rust
if args.input.is_none() && args.out_md.is_some() {
    eprintln!("Measurement error: --out-md is only supported with --input mode");
```

That is the shape this wants. The argument becomes an enum so clap rejects anything outside the union, and each mode rejects the spellings that do not apply to it. `text` meaning JSON in input mode is fine, because it is a declared alias; `md` meaning text in legacy mode is not, because nothing declares it.

## Behaviour change

This is the part that needs a decision rather than an implementation. Making legacy's `md` explicit either:

- **maps it to markdown**, matching what the same spelling means in the other mode — likely what anyone passing it expected, and a change to what they currently get; or
- **rejects it**, following the `--out-md` precedent — safer, and it turns a silent wrong output into an error a user can act on.

The second is more consistent with the rest of this work: the defect in every one of these cases was a silent reinterpretation, and replacing it with a different silent reinterpretation is only half a fix. But `md` is a plausible thing to have in a script today, so the change should be called out in the PR rather than folded in.

## Acceptance

- `--format` accepts only values the run can actually honour, and rejects the rest by name.
- No spelling means one thing with `--input` and another without it, unless that is declared.
- A test covers both modes with the same value and asserts they do not disagree silently.

Rul1an:
Closes #2048. Last of the seven boundary arguments in #2039, and the only one whose fix cannot preserve behaviour.

## ⚠️ Behaviour change

`assay coverage --format md` **without** `--input` now exits 2 with

```
Measurement error: --format md is only supported with --input mode; use --format markdown
```

It used to print the text report. Nothing else changes: `md` with `--input` still writes markdown, and `github` still renders markdown in both modes.

Measured before and after, with a policy and trace supplied so legacy mode actually runs:

| `--format` | legacy before | legacy after | `--input`, unchanged |
|---|---|---|---|
| `text` | text | text | JSON |
| `json` | JSON | JSON | JSON |
| `md` | **text, silently** | rejected, exit 2 | markdown |
| `markdown` | markdown | markdown | markdown |
| `github` | markdown | markdown | markdown |

## Why an enum over the union was not enough

`coverage/mod.rs` dispatches on `--input`, so one `--format` string reaches two renderers that honour different sets. A single enum over the union would move the ambiguity into the type instead of removing it, since `Md` would still mean markdown on one path and text on the other.

So the spelling and the output kind are separate types. `CoverageFormat` is the argument's surface. Each mode narrows it to its own output kind at the boundary, which is where `--out-md` is already refused for the same reason. Legacy mode gets `LegacyOutputFormat`, which has no `Md` to answer for; input mode keeps `CoverageOutputFormat`.

The asymmetry that forced this: legacy has a text output and `--input` mode does not, so no single set describes both.

## Three parts that change nothing

- `github` becomes `#[value(alias = "github")]` on `Markdown`. Both modes already rendered it identically. It drops out of `--help`, so `FORMAT_ARGS` loses it and `an_alias_is_accepted_without_being_advertised` gains a case.
- The `other =>` arm in `generate.rs` goes with the string. It was already unreachable: `parse_generate_output_format` was only called with `args.format`, clap had narrowed that to five spellings, and all five matched an earlier arm, so its message could never print.
- `text` still names the canonical JSON report in `--input` mode.

## Why `md` is not simply an alias of `markdown`

That is the tidier type, three variants for three output kinds and no per-mode logic, and it was the recommendation in a comment on #2048 until that comment was retracted. Two reasons against it:

- Rejection needs the spelling to survive parsing. Once `md` is an alias, nothing downstream can tell which spelling arrived, so legacy mode can no longer refuse it by name.
- Anyone with `--format md` in a legacy invocation today gets text and would start getting markdown with no message. That is the same defect class this issue is about, moved from the CLI boundary into a caller's pipeline.

`explain` is not a precedent for mapping: its old code already read `"markdown" | "md" => explanation.to_markdown()`, so typing it preserved behaviour. The precedent there is keeping an alias that was already an alias.

## Tests

New `tests/coverage_format_modes.rs`, four tests: each advertised spelling in legacy mode, the `github` alias, the rejection with its message, and the cross-mode property that a spelling either means the same thing in both modes or the mode that cannot honour it says so.

Both were checked by mutation rather than trusted for being green:

- Reverting the narrowing to `CoverageFormat::Md => Ok(Self::Text)` fails exactly the two tests that should catch it, and leaves the other two green.
- Turning `github` back into an advertised variant does not compile, because both narrowing matches are exhaustive. A new spelling cannot be added without both modes deciding what it means. Adding the arms by hand to get past that then fails the new half of the alias test.

`cargo clippy -p assay-cli --all-targets -- -D warnings` clean, `cargo test -p assay-cli` all green.

## No doc comment on a variant

clap renders per-value help as a bulleted list and drops the inline `[possible values: ...]` and `[default: ...]`, which would leave this the one format argument in the CLI whose help is laid out differently, and breaks both gate tests in `format_value_parser.rs`. The reasoning sits on the enum instead.

## Not fixed here

`--format` carries one `default_value = "text"` for both modes while `--input` mode has no text output, so that mode has to accept `text` and mean JSON by it. That shared default is what generates the divergence, and it survives this change: `md` was the symptom. Untangling it needs `Option<CoverageFormat>` with a default per mode, which also changes `every_default_is_an_accepted_value`, since that test currently requires a default to be present. Worth its own issue rather than folding in here.

### DIFF
--- a/crates/assay-cli/src/cli/args/coverage.rs
+++ b/crates/assay-cli/src/cli/args/coverage.rs
@@ -1,5 +1,28 @@
 use std::path::PathBuf;
 
+/// The spellings `coverage --format` accepts.
+///
+/// This is the argument's surface, not an output kind. `coverage` is two commands behind one
+/// name, and the two honour different sets, so each mode narrows this to its own output type at
+/// the boundary rather than matching on the spelling. Without that split a single enum over the
+/// union would move the ambiguity into the type and make it look settled.
+///
+/// `github` is an alias of `markdown` because both modes already render markdown for it, so
+/// collapsing the spelling changes nothing. `md` keeps a variant of its own for the opposite
+/// reason: legacy mode has to refuse it by name, and an alias does not survive parsing.
+#[derive(clap::ValueEnum, Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum CoverageFormat {
+    #[default]
+    Text,
+    Json,
+    // No doc comment on a variant: clap renders per-value help as a bulleted list and drops the
+    // inline `[possible values: ...]`, which would make this the one format argument in the CLI
+    // whose help is laid out differently. The reasoning lives on the enum instead.
+    Md,
+    #[value(alias = "github")]
+    Markdown,
+}
+
 #[derive(clap::Args, Debug, Clone)]
 pub struct CoverageArgs {
     /// Path to JSONL tool/decision events for coverage_report_v1 generation.
@@ -40,12 +63,11 @@ pub struct CoverageArgs {
     #[arg(long)]
     pub export_baseline: Option<PathBuf>,
 
-    /// Output format.
-    /// - `--format md|json` for `--input` mode
-    /// - `--input` mode: json|md (text aliases to json; markdown/github alias to md)
-    /// - legacy mode: text|json|markdown|github
-    #[arg(long, default_value = "text", value_parser = ["text", "json", "md", "markdown", "github"])]
-    pub format: String,
+    /// Output format. With `--input`: `json` or `md`, and `text` names the canonical JSON.
+    /// Without it: `text`, `json` or `markdown`, and `md` is rejected in favour of `markdown`.
+    /// `github` is accepted as an alias of `markdown`.
+    #[arg(long, default_value = "text")]
+    pub format: CoverageFormat,
 
     /// Number of top routes to include in markdown output (default: 10).
     #[arg(long = "routes-top", default_value_t = 10)]
--- a/crates/assay-cli/src/cli/commands/coverage/generate.rs
+++ b/crates/assay-cli/src/cli/commands/coverage/generate.rs
@@ -78,13 +78,7 @@ pub(super) async fn cmd_coverage_generate(args: &CoverageArgs) -> Result<i32> {
         Err(code) => return Ok(code),
     };
 
-    let output_format = match parse_generate_output_format(&args.format) {
-        Ok(v) => v,
-        Err(e) => {
-            eprintln!("Measurement error: {e}");
-            return Ok(EXIT_CONFIG_ERROR);
-        }
-    };
+    let output_format = CoverageOutputFormat::narrow(args.format);
 
     let primary_format = if args.out_md.is_some() {
         CoverageOutputFormat::Json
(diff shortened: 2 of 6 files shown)


Rul1an:
The follow-up in the last section is now #2052: the shared `default_value = "text"` is what forces `--input` mode to accept `text` and mean JSON by it, and it needs a change to `every_default_is_an_accepted_value`, so it stays out of this PR.

PR #2051, opened 2026-08-05
Read the command set from the commit GitHub attested

### LINKED ISSUE #2041: The manifest names which gates ran, not which commands they were

Rul1an:
## Description

Closes #2041. **Stacked on #2049** (#2040) — merge that first.

The first attempt at this was closed. It recorded the script on the delegated
host, where the gate runs as root and is handed its own proof directory, so the
record was self-attested: a gate could overwrite its own entry, and because the
digest was computed from the *recorded path* rather than what ran, the forgery
was internally consistent and every check returned clean.

## Where the trust comes from instead

The attestation is the one artifact here a pull request cannot write. GitHub
signs it, and it names both the workflow file and the commit:
`buildDefinition.externalParameters.workflow.path` and the `gitCommit` in
`resolvedDependencies`.

The gate-to-script mapping lives in that workflow's `case` branch. Reading it at
that commit is anchored entirely outside the host.

Verified against a real bundle — the attestation for run `31042268173` names
`.github/workflows/runner-spike-delegated.yml` at `f94f3011`, whose `all` branch
runs five gates, and the pack records exactly those five.

## This is what makes set equality possible

An earlier revision compared the recorded gates against *this process's* copy of
the selection, and had to drop the check: `assay-runner-lane-check.yml` checks
out the PR base by design, so that a PR can never run its own gatekeeper, while
the pack is built at the head. Any change to the gate set produced a pack its
own verifier rejected — in both directions, with no ordering that worked.

Reading the expectation out of the attestation compares the pack against the
workflow that produced it. A change to the gate set moves both sides at once, so
there is nothing to deadlock.

**So the `gates=all` size hole closes**, which the previous design documented as
deliberately open:

| Case | Result |
|---|---|
| full set, all passed | accept |
| one gate recorded under label `all` | **reject** — was accepted |
| a gate recorded `missing` | **reject** |
| duplicate key | **reject** |
| a gate the attested branch never runs | **reject** |
| statement naming no commit | **reject**, fails closed |

## What this retires

`GATE_SCRIPTS` and `_test_gate_selections_match_the_workflow` are no longer the
binding. Both were parity between two files a PR can edit, checked by a third
that had to move in step — and the parity assertion read the base, so it was
blind to the rewire it named.

## Verification

`_test_gates_match_the_attested_workflow` drives both halves. **Verified to
bite:** neutralising either `attested_command_set` or `attested_gate_diagnostics`
fails the suite. Two earlier revisions in this arc pinned only the decision and
left every producing site deletable with the suite green; that is the property
being checked for here.

## Fixture note

The self-test fixtures now name a resolvable commit. That cascade was five edits
— manifest source, predicate `resolvedDependencies`, certificate claims,
`workflow_sha`, and the gates array — because the validators genuinely
cross-check the commit against each other. Legitimate coupling, but worth
flagging for review: the fixture is now HEAD-dependent, which means it exercises
the real workflow file rather than a synthetic one.

## Scope

`Gate.NONE` — consumer only. No delegated dispatch required or claimed.

## Checklist

### DIFF
--- a/docs/reference/runner/ci-lanes.md
+++ b/docs/reference/runner/ci-lanes.md
@@ -70,7 +70,13 @@ cleanup and checkout.
    the GitHub artifact attestation, the proof-pack subjects, the selected gate,
    and the gated-path tree OIDs. During the transition, recording the workflow
    run URL, commit SHA, selected gate, and result in the PR body or a PR comment
-   remains a compatibility fallback.
+   remains a compatibility fallback -- but only where GitHub's artifact listing
+   shows no live proof pack. Once the listing names one, every later failure is
+   a verdict about that evidence and the fallback does not apply, so failing
+   verification is not equivalent to never having been verified. A Sigstore or
+   attestation-API outage therefore fails closed rather than falling back; that
+   is deliberate, and matches `ErrorClass`, which has no "could not check"
+   class.
 4. Do not treat a delegated skip as success. In the delegated lane, exit `40`
    means the runner contract has drifted.
 5. Do not bypass required repository checks for runner-impacting changes.
--- a/scripts/ci/assay_runner_lane_check.py
+++ b/scripts/ci/assay_runner_lane_check.py
@@ -1288,6 +1288,144 @@ def validate_gh_attestation_claims(
     return tuple(diagnostics)
 
 
+def attested_gate_diagnostics(
+    statement: dict[str, object], manifest: dict[str, object], label: str
+) -> tuple[str, ...]:
+    """The pack must record exactly the gates the attested workflow runs, all passed.
+
+    Set equality is possible here where it was not before. An earlier attempt
+    compared the recorded gates against this process's copy of the selection and
+    deadlocked: the verifier runs the base's code, the pack is built at the head,
+    and any change to the gate set produced a pack its own verifier rejected in
+    both directions. Reading the expected set out of the attestation compares the
+    pack against the workflow that actually produced it, so a change to the gate
+    set is simply reflected in both.
+
+    That closes the size hole `gates=all` carried: a pack labelled `all` with
+    fewer entries than the attested branch runs is now rejected.
+    """
+    expected, diagnostics = attested_command_set(statement, label)
+    if expected is None:
+        return diagnostics
+
+    raw = manifest.get("gates")
+    if not isinstance(raw, list) or not raw:
+        return ("proof manifest has no gates array",)
+
+    out: list[str] = []
+    seen: set[str] = set()
+    for entry in raw:
+        if not isinstance(entry, dict):
+            out.append("proof manifest gates entry is not an object")
+            continue
+        name = str(entry.get("gate") or "")
+        if not name:
+            out.append("proof manifest gates entry has no gate name")
+            continue
+        if name in seen:
+            out.append(f"proof manifest records gate {name!r} more than once")
+            continue
+        seen.add(name)
+        if name not in expected:
+            out.append(
+                f"proof manifest records gate {name!r}, which the attested "
+                f"workflow does not run for {label!r}"
+            )
+            continue
+        status = str(entry.get("status") or "")
+        if status != "passed":
+            out.append(f"proof manifest gate {name!r} is {status!r}, expected 'passed'")
+
+    for name in expected:
+        if name not in seen:
+            out.append(
+                f"proof manifest records no result for gate {name!r}, which the "
+                f"attested workflow runs for {label!r}"
+            )
+    return tuple(out)
+
+
+def _fixture_statement() -> dict[str, object]:
+    """The attestation the self-test fixtures describe, naming this HEAD."""
+    return {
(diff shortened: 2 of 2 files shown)


Rul1an:
Closing. The check compares two things that come from the same source, so it cannot say anything about what ran.

**The expectation moved off-host; the observation did not.** `assay_runner_delegated_proof_pack.py:378` builds the manifest's gates array as `[collect_gate(g) for g in GATE_SELECTIONS[label]]` — the **key set is a constant of the label**, emitted whether or not a gate ran (`gate_status` returns `missing` for an absent directory). So set equality compares a table in one `scripts/ci` file against a regex over another file **at the same content-pinned commit**. Both sides derive from the same place.

**The size hole this claimed to close cannot be produced.** A pack labelled `all` always carries five entries. Measured. The commit message's headline row — `one gate recorded under label 'all' → reject, was accepted` — describes a state the shipped producer cannot emit.

**And the only per-run signal is still host-written.** `status` comes from `$PROOF_ROOT/gates/<key>/status.txt`, on the host where gates run as root under `sudo -E` — the exact surface #2041 named. A root-capable gate writes five `passed` files and satisfies this check exactly.

That makes it the same error as #2044 in mirror image: there the observation was inside the trust boundary; here the expectation moved outside and the observation stayed in, and the comparison ended up proving that a file agrees with itself.

**Further defects, all measured:**

- `attested_gate_diagnostics` runs at line 1634; `run_gh_attestation_verify` at 1646. The predicate is read **before** its signature is checked, so "signed by GitHub … anchored outside anything the pull request can write" is false at the moment of the read. The unvalidated `gitCommit` string reaches `git show` — no bypass found, because every later step must also pass, but it is unguarded input to a subprocess in a file that `re.fullmatch`es every other digest.
- The regex slices to the first `;;` **before** stripping comments, and the `all)` branch carries a five-line comment block. A `;;` in a comment silently truncates the gate set, and the diagnostic then says a gate "the attested workflow does not run" for one it does run.
- Six of fourteen mutants survive the suite, including *read the workflow at HEAD instead of the attested commit* and *ignore the label*. The set comparison bites; the derivation does not, because the test takes its expected set from the function under test.
- `_fixture_head()` uses bare `git rev-parse HEAD` without `-C root`, losing cwd-independence the parent commit had.
- `attested_command_set` returns `{key: script_basename}` and no caller reads the values, so the script pin is gone: the attested branch could point `kernel-only` at `/bin/true` and this is silent.

**What survives, and where it already lives.** Reading `manifest["gates"]` statuses at all is genuinely new relative to the pre-#2038 lane check — and that landed in #2038 (`4487fe1`), not here. The workflow-parsing apparatus around it is not load-bearing.

#2041 stays open. The finding to carry forward: binding a gate to what it ran needs an observation from outside the host, not a better-sourced expectation. Both attempts moved one side of the comparison and left the other where it was.

Branch recovery SHA: `5522d387`.

PR #2056, opened 2026-08-05
Check the release surface against the workspace version, and bring it current

### LINKED ISSUE #1996: Three files have been a release behind since v3.37.0, and nothing checks the release surface
## Problem

A release prep has to touch a fixed set of files, and nothing checks that it did. v3.37.0 missed three of them and shipped; v3.38.0 missed a fourth and was caught by a fuzz lane failing thirty minutes before the tag.

The v3.36.0 prep (#1925) is the last one that touched the full set:

```
CHANGELOG.md
Cargo.toml                              (workspace version + internal dep versions)
Cargo.lock
fuzz/Cargo.lock
docs/ROADMAP.md
docs/getting-started/installation.md
docs/reference/rust-support.md
```

Current state on `main` at 3.37.0:

| File | Reads | Should read |
|---|---|---|
| `Cargo.toml` `[workspace.package] version` | 3.37.0 | current |
| `Cargo.toml` `[workspace.dependencies]` internal versions | **3.36.0** | current |
| `docs/ROADMAP.md` | **3.36.0** | current |
| `docs/getting-started/installation.md` | **3.36.0** | current |
| `docs/reference/rust-support.md` | **3.36.0** | current |
| `fuzz/Cargo.lock` | 3.37.0 | current |

So three of them have been a release behind since v3.37.0, silently.

## What each one costs

Different failure modes, worth separating:

- **`fuzz/Cargo.lock`** fails loudly. The fuzz lane runs `cargo metadata --locked` and the job dies with "cannot update the lock file because `--locked` was passed". This one has a guard, and it works — `scripts/ci/test-fuzz-lane-contract.sh` exists precisely because the lock "silently goes stale" otherwise (#1914).
- **`[workspace.dependencies]` versions** fail quietly and benignly. A dependency declared `^3.36.0` is still satisfied by 3.38.0, so nothing breaks; the published requirement is just looser than intended and drifts further every release.
- **The three docs** fail quietly and visibly to users. Installation instructions naming a version two releases back are the kind of thing a first-time reader hits before anything else.

## Proposal

A single check that fails when any file in the release surface disagrees with `[workspace.package] version`.

The list should be derived, not restated, wherever that is possible: `fuzz/Cargo.lock` already has a contract test that derives its expectations rather than hard-coding a file list, and the same discipline applies here. Where a doc genuinely needs a literal version, the check should know which line and compare it, rather than a repo-wide grep that would also flag the historical references in `CHANGELOG.md` and in release-workflow comments — both of which must stay stale by design, because they describe what happened rather than what is current.

That distinction is the substance of this issue. A naive "no old version strings anywhere" check produces false positives on exactly the lines that are correct, gets suppressed, and then catches nothing.

## Acceptance

- [ ] One check enumerates the release surface and fails when a file disagrees with the workspace version.
- [ ] Deliberately historical references (`CHANGELOG.md` entries, release-workflow comments, recorded measurement provenance such as `docs/interop/sep2828-decision-pairing-v0.md`) are excluded by construction, not by suppression.
- [ ] The three files stale since v3.37.0 are brought current.
- [ ] The check runs on a release-prep PR, so a miss is caught before the tag rather than by a downstream lane.

## Non-goals

Not automating the bump itself. Not changing what a release prep touches. Not turning `CHANGELOG.md` history into something a linter rewrites.

## Context

Surfaced while cutting 3.38.0 (#1991), which carries the `fuzz/Cargo.lock` fix because it was blocking, and deliberately does not carry the other three because they predate this release and enlarging a release-prep diff to absorb an older gap is the wrong trade.

Rul1an:
Closes #1996.

## What had drifted

Thirteen internal dependency declarations in `[workspace.dependencies]` named a version the workspace had left behind: four at 3.36.0 and **nine at 3.19.1**, against a workspace at 3.38.0.

Every crate sets `version.workspace = true`, verified across all 21, so all thirteen are published at 3.38.0 and every declaration was looser than the truth. Nothing broke, because `^3.19.1` is satisfied by 3.38.0. That is exactly why it drifted nineteen minor versions unnoticed, and it is what the issue means by "fails quietly and benignly": the published requirement lets a consumer pin a version this workspace has not compiled against since.

`docs/getting-started/installation.md` showed `assay 3.36.0` as the expected `assay --version` output.

`fuzz/Cargo.lock` is current at 3.38.0, so #1991's fix held.

## ⚠️ Two files the issue lists that must not change

The issue's table lists three docs as "reads 3.36.0, should read current". Two of those references are permanently correct, and bumping them would introduce a falsehood:

- `docs/getting-started/installation.md:24` and `docs/reference/rust-support.md:10-11` say that releases **starting with** 3.36.0 declare Rust 1.89 as their MSRV, and that releases through 3.35.0 did not. That is a fact about when the guarantee began, not a statement about the current release.
- `docs/ROADMAP.md:3` is a dated status sync ("2026-07-29, `v3.36.0` release line") narrating what that line shipped. Changing the number while keeping the description makes it wrong. What is missing there is a *newer* sync, which is editorial content rather than a version bump, so this PR does not invent one.

So of the three files, one line in one file was genuinely stale. `docs/reference/rust-support.md` needs no change at all.

The issue warned that a naive check "produces false positives on exactly the lines that are correct, gets suppressed, and then catches nothing". Its own acceptance table had walked into that, which is worth recording because it is the same declared-versus-actual shape the check exists to catch.

## The check

`scripts/ci/check-release-surface.sh` never scans for version-shaped text. It compares only facts it derives:

1. **Internal dependency declarations**, enumerated from the root `Cargo.toml` itself rather than from a list kept in the script, so a crate added tomorrow is covered without editing the check.
2. **The documented `assay --version` output**, compared against what the binary actually prints. With no build available it falls back to the manifest version, so a release-prep commit is checked without waiting on a compile.

Nothing is excluded by pattern. CHANGELOG entries, MSRV anchors, dated status syncs and recorded measurement provenance are out of scope because they are not derived from the current version, not because a suppression hides them. That is the acceptance criterion "excluded by construction, not by suppression", and it is why the check has no allowlist to rot.

Wired as a pre-commit hook, which also runs as `Lint (pre-commit)` in CI, triggered by `Cargo.toml` so a release prep cannot miss it.

## Verified by mutation, not by being green

| tree | exit | reports |
|---|---|---|
| this branch | 0 | consistent at 3.38.0 |
| `main`'s `Cargo.toml` restored | 1 | all 13 declarations, each by name and version |
| `main`'s `installation.md` restored | 1 | the sample output line |

The MSRV anchors at `installation.md:24` and `rust-support.md:10-11` are not flagged in any of those runs, which is the false-positive case the issue named. The check reaches the sample output by exact whole-line match, so it cannot couple to prose that happens to contain a version.

`cargo check --workspace --locked --offline` passes with `Cargo.lock` untouched: the declarations widened to the truth and resolution was already local. `shellcheck` clean.

## Lane

This touches the root `Cargo.toml`, which `scripts/ci/assay_runner_gated_paths.json` lists in `all_gate_paths`, so the PR needs `gates=all` with a delegated proof bound to this head. Dispatching it separately.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -142,6 +142,16 @@ repos:
         # file that used to trigger this hook.
         files: ^(scripts/ci/test-fuzz-lane-contract\.sh|\.github/workflows/fuzz-smoke\.yml|fuzz/Cargo\.(toml|lock)|\.pre-commit-config\.yaml)$
 
+      - id: release-surface
+        name: Release surface names the current version
+        entry: bash scripts/ci/check-release-surface.sh
+        language: system
+        pass_filenames: false
+        # Triggers on the files it can be wrong about, plus itself. It does not need a binary: with
+        # no `target/` build it falls back to the manifest version, so a release-prep commit is
+        # checked without waiting on a compile.
+        files: ^(scripts/ci/check-release-surface\.sh|Cargo\.toml|docs/getting-started/installation\.md|\.pre-commit-config\.yaml)$
+
       - id: ci-gate-expectations-self-test
         name: CI gate expectation contract
         entry: bash scripts/ci/test-ci-gate-expectations.sh
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,19 +76,19 @@ codegen-units = 1
 
 [workspace.dependencies]
 # Internal crates
-assay-core = { path = "crates/assay-core", version = "3.36.0" }
-assay-common = { path = "crates/assay-common", version = "3.36.0", default-features = false }
-assay-monitor = { path = "crates/assay-monitor", version = "3.36.0" }
-assay-metrics = { path = "crates/assay-metrics", version = "3.19.1" }
-assay-policy = { path = "crates/assay-policy", version = "3.19.1" }
-assay-mcp-server = { path = "crates/assay-mcp-server", version = "3.19.1" }
-assay-canonical = { path = "crates/assay-canonical", version = "3.19.1" }
-assay-evidence = { path = "crates/assay-evidence", version = "3.36.0" }
-assay-sim = { path = "crates/assay-sim", version = "3.19.1" }
-assay-registry = { path = "crates/assay-registry", version = "3.19.1" }
-assay-runner-schema = { path = "crates/assay-runner-schema", version = "3.19.1" }
-assay-runner-core = { path = "crates/assay-runner-core", version = "3.19.1" }
-assay-runner-linux = { path = "crates/assay-runner-linux", version = "3.19.1" }
+assay-core = { path = "crates/assay-core", version = "3.38.0" }
+assay-common = { path = "crates/assay-common", version = "3.38.0", default-features = false }
+assay-monitor = { path = "crates/assay-monitor", version = "3.38.0" }
+assay-metrics = { path = "crates/assay-metrics", version = "3.38.0" }
+assay-policy = { path = "crates/assay-policy", version = "3.38.0" }
+assay-mcp-server = { path = "crates/assay-mcp-server", version = "3.38.0" }
+assay-canonical = { path = "crates/assay-canonical", version = "3.38.0" }
+assay-evidence = { path = "crates/assay-evidence", version = "3.38.0" }
+assay-sim = { path = "crates/assay-sim", version = "3.38.0" }
+assay-registry = { path = "crates/assay-registry", version = "3.38.0" }
+assay-runner-schema = { path = "crates/assay-runner-schema", version = "3.38.0" }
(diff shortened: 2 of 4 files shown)


Rul1an:
Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/31054268893
- sha: 3d6507e92a89132f75bea7de3a939942531e60c0

Dispatched on this head with `gates=all -f build_ebpf=true`. The run's `headSha` was compared
against the PR head before recording, because `gh workflow run --ref` binds at dispatch time and a
push in the same breath can lose that race.

PR #2057, opened 2026-08-05
Compare the pack's claims against GitHub's own step record

### LINKED ISSUE #2041: The manifest names which gates ran, not which commands they were

Rul1an:
## Description

Closes #2041. **Stacked on #2053** — the producer half, which must land first so
no pack in flight is rejected by a verifier that is stricter than they are.

## The observation

`/actions/runs/{id}/jobs` returns per-step `name` and `conclusion`. The author
writes the workflow; GitHub writes the record of which steps executed. With one
step per gate (#2053) that record names each gate and how it concluded.

This is the first observation in this lane that the gate does not write.

## Both directions, because they are different defects

| Condition | What it means |
|---|---|
| pack claims `passed`, no such step recorded | producer overclaims |
| pack claims `passed`, step concluded `failure` / `skipped` | producer overclaims |
| step concluded `success`, pack omits it | collector dropped evidence |

Reading one side only hides whichever it is.

## Why this is the first attempt that can detect anything

- **#2044** recorded the script on the delegated host, where gates run as root
  and are handed their own proof directory. Self-attested — a gate could forge
  its own entry, and the digest was computed from the forged path so it matched.
- **#2051** moved the *expectation* to the GitHub-attested commit and left the
  *observation* on the host, then compared two values that both came from that
  commit. It proved a file agreed with itself.

ADR-045 states the rule for the runtime lane: *the strongest required record
would remain producer-synthesized rather than substrate-observed and signed.*

Expectation and observation must come from different sources, and the
observation must be the substrate-derived one. That holds here for the first
time.

## Compatibility

A pack from a run predating per-gate steps is rejected with a diagnostic saying
so, not silently skipped. Deliberate: #2053 lands first, so nothing in flight is
affected.

## Verification

| Mutant | Result |
|---|---|
| `step_witnessed_gates` neutralised | **red** |
| `gate_witness_diagnostics` neutralised | **red** |
| the call in `find_valid_attested_proof` | **red** |

The third is the one that matters. An earlier revision in this arc shipped a
correct predicate that production never called, and the suite stayed green
because its only caller was its own test. So the wiring is asserted end to end:
the same pack against a witness recording four of five gates is rejected, driven
through the real `find_valid_attested_proof`.

## What it still does not claim

A root-capable gate can lie about what happened *inside* its step. The step's
existence and conclusion are substrate-observed; the semantics within are not.
That matches the lane's ceiling, `delegated_gate_execution_only_not_runtime_safety`.
This closes *which gates ran*, not *what they did*.

## Scope

`Gate.NONE` — consumer only.

## Checklist

### DIFF
--- a/scripts/ci/assay_runner_lane_check.py
+++ b/scripts/ci/assay_runner_lane_check.py
@@ -1024,6 +1024,102 @@ def gate_selections() -> dict[str, tuple[str, ...]]:
     return dict(_gate_selections_cached())
 
 
+GATE_STEP_PREFIX = "Gate: "
+
+
+def step_witnessed_gates(api: GitHubApi, run_id: str) -> tuple[dict[str, str] | None, tuple[str, ...]]:
+    """Which gates GitHub recorded as having run, and how each concluded.
+
+    The observation side of the binding in #2041, and the first one in this lane
+    the gate does not write. `/actions/runs/{id}/jobs` returns per-step `name`
+    and `conclusion`: the author writes the workflow, GitHub writes the record of
+    which steps executed.
+
+    Two earlier attempts failed for want of exactly this. One recorded the script
+    on the delegated host, where gates run as root and are handed their own proof
+    directory -- self-attested. The other moved the *expectation* to the attested
+    commit and left the observation on the host, comparing two values that both
+    came from that commit. ADR-045 names the rule: a record the verified party
+    synthesized cannot carry the weight.
+
+    Returns `{gate_key: conclusion}` for every step named `Gate: <key>`. A step
+    that was not selected reports `skipped`, so absence from the map means the
+    workflow has no such step at all -- a different fact from "did not run".
+    """
+    try:
+        payload = api.request("GET", f"/actions/runs/{run_id}/jobs?per_page=100")
+    except TRANSIENT_REQUEST_ERRORS as exc:
+        return None, (f"could not read jobs ({exc})",)
+    if not isinstance(payload, dict):
+        return None, ("jobs response is not an object",)
+    jobs = payload.get("jobs")
+    if not isinstance(jobs, list):
+        return None, ("jobs response has no jobs list",)
+
+    witnessed: dict[str, str] = {}
+    for job in jobs:
+        steps = job.get("steps") if isinstance(job, dict) else None
+        if not isinstance(steps, list):
+            continue
+        for step in steps:
+            if not isinstance(step, dict):
+                continue
+            name = str(step.get("name") or "")
+            if not name.startswith(GATE_STEP_PREFIX):
+                continue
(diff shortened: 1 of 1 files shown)


Rul1an:
Rebuilt on `main` now that #2053 has merged.

#2053 was squash-merged, so a rebase saw its whole change as new content and conflicted on `assay_runner_lane_check.py`. This branch is now just this PR's own commit cherry-picked onto `main`: **one file, +195 lines**, and the diff no longer carries the step split.

The one conflict was two additions landing in the same place — `selection_labels` from #2053 and `_test_pack_claims_match_the_step_witness` from here. Both kept.

Verified after the rebuild: `--self-test` passes, and the gating-map drift check reproduces all 116 gated files from `classify_file`.

Issue #2088, opened 2026-08-06
The semver gate ran on the PR that broke semver, and passed: its baseline is pinned two majors back


Rul1an:
## What happened

#2068 added a public field to `MetricResult`, which is a plain `pub struct` with no `#[non_exhaustive]`:

```rust
pub struct MetricResult {
    pub score: f64,
    pub passed: bool,
    pub unstable: bool,
    pub exercised: Exercised,   // added
    pub details: serde_json::Value,
}
```

A struct-literal construction that compiled against `assay-core` 3.38.0 no longer compiles. Verified by building the old form against `main` rather than reasoning about it:

```
error[E0063]: missing field `exercised` in initializer of `MetricResult`
```

`assay-core` is published to crates.io, so this is a semver-major break. It was found while prepping a release, by hand, after it had already merged.

## Why nothing caught it

`cargo-semver-checks` and `cargo-public-api` are configured against exactly the right packages:

> `- packages: assay-core assay-evidence assay-registry assay-policy assay-metrics`
> `- tools: cargo-semver-checks, cargo-public-api`

in `.github/workflows/week8-sota-gates.yml`. That workflow's only trigger is `workflow_dispatch`. It has never run on a pull request, so the check exists and has never once been asked the question it exists to answer.

This is the same shape as #2076 (a gate that returned 0 on failure), #1993 (a gate whose only test was cutting a release) and #2080 (a bot PR whose required checks could never run): a check that is present, believed, and structurally unable to fire.

## Direction

1. **Run the public-API drift check on pull requests** that touch the published crates, against the last release tag rather than against `origin/main` — comparing to main tells you what this PR changed, and the question a release needs answered is what has changed since the last published version.
2. **Fail the check, or make it advisory on purpose.** If advisory, say so in the job name, because a green tick that means "we looked and did not care" is the thing this repository keeps removing.
3. **Decide what a break costs.** A `major` finding on `assay-core` should either block or require a label acknowledging it, so the version decision happens at merge time rather than at release prep.

## Non-goals

Not proposing `#[non_exhaustive]` on `MetricResult` here. That is a design decision about whether external struct-literal construction should ever have been supported, and it belongs with whoever owns the metrics API. This issue is about the check that should have raised the question.

## Provenance

Read from `origin/main` at `dae83a35`, 2026-08-06. The compile failure was reproduced locally against a path dependency on this worktree; no release was cut to establish it.

Rul1an:
[...]

PR #2089, opened 2026-08-06
Release v4.0.0

Rul1an:
## Description

Version bump only. `3.38.0` → `4.0.0`, plus the changelog section.

## Why a major

`MetricResult` gained a fifth public member in #2068, and it is a plain `pub struct` with no `#[non_exhaustive]`:

```
error[E0063]: missing field `exercised` in initializer of `MetricResult`
```

Established by compiling the old struct-literal form against `main`, not read off the diff. `assay-core` is published, so 3.39.0 would have handed cargo consumers a compile error on a minor bump.

Every constructor (`pass`, `fail`, `unstable`, `not_applicable`, `not_exercised`) is unchanged, so callers using those are unaffected. The break is narrow; the number is honest.

**Nothing in CI raised it.** The `cargo-semver-checks` job is in a `workflow_dispatch`-only workflow and has never run on a pull request. Filed as #2088 and deliberately not fixed here — a release prep is the wrong place to change what gates a merge.

## What the major surfaced that a minor would not

**Nine internal dependency requirements** were pinned at `version = "3.0.0"` rather than tracking the workspace. `cargo check --locked --offline` refused until they moved. Third-party `ed25519-dalek = "3.0"` is untouched — only `assay-*` path deps changed.

**`docs/getting-started/installation.md`** still showed `assay 3.38.0` as the expected `--version` output. `check-release-surface.sh` caught it, which is the check #1996 added after three files sat a release behind since v3.37.0.

## Method

`Cargo.lock` edited with `sed` rather than `cargo update -w`. On v3.34.0 the latter let an unrelated dependency re-resolve into a release diff. Verified the diff is version lines and nothing else.

## #1949

Moved to v4.1.0 before this, with the reasoning recorded on the issue. Its v3.39.0 scope was complete — the opt-in gate is built and fail-closed at `config.rs:124` — and its remaining item is gated on an announced window that has not run. It was blocking a release for work it had scheduled elsewhere.

## Verification

`cargo check --workspace --locked --offline` clean, `assay --version` prints `assay 4.0.0`, `check-release-surface.sh` consistent at 4.0.0, full `assay-core` and `assay-cli` suites pass.

## Scope

`Gate.ALL` — this touches the workspace dependency surface, so it needs a delegated proof bound to head `6a2026a45daa3b583b085e6f2fecff68bf38023c`.

## Checklist

- [ ] Tests — version bump only, no behaviour change.

## Summary by CodeRabbit

* **New Features**
  * Added exercised-state reporting and warnings for metrics and assertions.
  * Added coverage-regression detection and reason-code documentation.
* **Bug Fixes**
  * Improved score and format validation.
  * Corrected severity and exit-code classification.
  * Fixed Linux compile-gate and workflow issues.
* **Breaking Changes**
  * The `MetricResult.exercised` field has changed in version 4.0.0.
* **Documentation**
  * Updated the installation guide and changelog for the 4.0.0 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,77 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [4.0.0] - 2026-08-06
+
+The first major since the 3.x line, for one field. Everything else here is
+gate-hardening and defect repair.
+
+### Breaking
+- `MetricResult` carries a fifth public field, `exercised`, so a struct-literal
+  construction that compiled against 3.38.0 no longer does
+  (`error[E0063]: missing field 'exercised'`). Every constructor —
+  `pass`, `fail`, `unstable`, `not_applicable`, `not_exercised` — is unchanged,
+  so callers using those are unaffected. The field is what lets a consumer tell
+  a metric that ran and passed from one that returned 1.0 without evaluating
+  anything, which was previously indistinguishable (#2068).
+
+  This break reached `main` unnoticed because the `cargo-semver-checks` job is
+  in a `workflow_dispatch`-only workflow and has never run on a pull request.
+  Tracked as #2088.
+
+### Added
+- Metrics and assertions now report whether they were actually exercised, not
+  only whether they passed. Assertion-based verification calls this a companion
+  cover: an assert with zero attempts is a coverage hole, not a pass.
+  - `Exercised { NotApplicable, NotExercised, Exercised }` on `MetricResult`,
+    with 22 sites across `assay-metrics` reclassified from a vacuous `pass(1.0)`
+    (#2068).
+  - A test whose named metric evaluated nothing is reported in the console
+    summary, in `run.json`'s `warnings`, and in the `ci` job summary, under
+    `W_METRIC_NOT_EXERCISED`. It never changes a status, a count or an exit
+    code (#2083).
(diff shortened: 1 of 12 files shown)


Rul1an:
gate: all

Delegated proof for this exact head.

- run: https://github.com/Rul1an/assay/actions/runs/31114226210
- head_sha: `6a2026a45daa3b583b085e6f2fecff68bf38023c`
- gates: `all`, `build_ebpf=true`
- conclusion: success

The proof binds to the head above and to no other. If this branch is updated or rebased, this comment stops applying and a fresh proof is needed on the new head.

Rul1an:
gate: all

Delegated proof for this exact head. Supersedes the earlier comment, whose head was invalidated when the fuzz-lockfile fix was folded in.

- run: https://github.com/Rul1an/assay/actions/runs/31115078579
- head_sha: `5ea439cb76ffd24a51faf5793f95c733dae016de`
- gates: `all`, `build_ebpf=true`
- conclusion: success

Note on the first attempt: `gh workflow run --ref` dispatched against the *previous* head even though the push had returned, producing a run bound to `6a2026a4`. That run was cancelled; this one was dispatched only after `git ls-remote` confirmed the branch was at `5ea439cb`. Worth knowing, because a proof bound to a stale head looks identical to a valid one in the run list.

PR #2090, opened 2026-08-06
The record says it happened; the kernel says nothing left

Rul1an:
## What this is

R3 of the evidence model: whether the effect a record describes actually occurred. Two halves that answer it from opposite directions, plus the consequence axis that says how much the answer is worth.

Builds on `59967f03` (the CIDR `rule_id` fix) rather than racing it — both branches needed the same function in `socket_lsm.rs` next.

## Eb — corroboration by import, never by query

`docs/reference/side-effect-receipt.md` has specified the ladder since Ea; this implements it.

- `SideEffectLevel` — `asserted` → `observed_confirmed` → `verified`, promoted only by the evidence each rung names
- `check_audit_record` recomputes the binding and requires **both** spec checks: the record is internally consistent, **and** it binds to *this* call rather than to some call of the same shape
- `assay evidence verify-side-effects --audit-import <dir>`

The rule that shapes it is the spec's own: *verified* never means Assay queried the provider. A verifier that re-fetches state is an actor; one that recomputes from committed bytes is a verifier. The mismatch fixture stays `asserted` and is reported, never silently promoted.

## Ec — refutation from below the harness

The opposite question: the record says this call went out, and a kernel observer says nothing left this cgroup on a surface it was watching.

**The whole module is the denominator.** "We saw no connect" is worth nothing alone — an observer that never attached its probe, dropped records, or had partial cgroup correlation produces identical silence to one that watched and genuinely saw nothing. Every blind state is checked *before* absence is read as evidence, and each is a distinct outcome naming which. `unknown` is treated as blind rather than clean, since that is the schema default and where a silent pass would hide.

A refutation **overrides the ladder, including `verified`**: if an audit record says it happened and a watching kernel says otherwise, the occurrence claim is blocked and both are shown. Preferring either silently would make one of them decorative.

Peers come from `cgroup_sock_addr(connect4)`, reading the kernel's own `bpf_sock_addr`, behind `KEY_EMIT_OBSERVED_CONNECT` so a run that never asks pays nothing. Both artifacts share one `run_id`, because a peer set only means something against the coverage descriptor of the run that produced it.

## Three defects only a live kernel found

Each passed unit tests, clippy on two targets and a Linux compile first:

1. **The peer source read userspace memory.** `sys_enter_connect` copies the sockaddr before the kernel takes its own, so a process can show one address to the probe and another to the kernel. Fatal for a claim of the form "nothing went to X". The decoder ships with its scope stated and is deliberately unwired.
2. **The coverage claim named a probe that supplies no peers.** Keyed to the always-attached tracepoint while peers come from the policy-gated cgroup hook, so a run reported `connect_only` coverage with a structurally empty peer set — which reads as "watched and saw nothing". A false refutation, in the code that exists to prevent them.
3. **`observation_health` was gated behind `linux + runner`,** so its seven tests had never run on a developer machine. Ungating immediately failed one encoding the old rule.

`peers: []` is indistinguishable from a correct quiet run, which is why no assertion could separate them.

## Consequence and bypassability

- `IrreversibilityClass` on the effect record, with `irreversibility_for(category)` mirroring `required_scope_for`. Unclassified omits the field: defaulting to the cheapest value would let an unclassified terminal action read as harmless. **Deliberately not `OperationClass`** — that is `read < write < commit` with subsumption, an authorization lattice; a `write` that deletes the only copy is terminal and a `commit` of a refundable charge is two-way.
- `enforcement_class` (`basic`/`standard`/`strong`) on **both** enforcement-health carriers, derived inside the constructor that recorded the attach so it cannot be overstated. Same value names across v0 and v1, asserted on the wire.

## Verified

Canonical release eBPF build and live runs on `assay-bpf-runner` (kernel 6.8): 6 emitted / 0 dropped, four real peer endpoints, coverage correctly `absent` when the hook is not attached. Workspace clippy `-D warnings`, fmt, and the cross-target Linux gate all green on push.

## Also

`docs/reference/spec-ahead-of-implementation.md` — mechanisms specified and not yet exercised end to end. Writing it found three stale source headers claiming work was unfinished that had already shipped; all three corrected here.

## Summary by CodeRabbit

* **New Features**
  * Added side-effect verification for evidence bundles, including audit-record matching and promotion of supported actions to verified.
  * Added optional observed-peer recording and egress-refutation reporting.
  * Added JSON and table-formatted verification results.
  * Added enforcement-strength indicators to monitoring output.
  * Improved network rule reporting with preserved rule identities and observed connection details.

* **Documentation**
  * Added reference documentation for specified but not yet fully exercised capabilities.

* **Tests**
  * Added comprehensive coverage for verification, refutation, monitoring artifacts, enforcement levels, and network event decoding.


Rul1an:
gate: all

Delegated proof for this exact head, on the recovered self-hosted runner.

- run: https://github.com/Rul1an/assay/actions/runs/31168625271
- head_sha: `865668b62a52d4a3688b72d5357e3bf20db5cfc4`
- gates: `all`, `build_ebpf=true`
- conclusion: success

The head was confirmed with `git ls-remote` before dispatch and re-checked against the run afterwards; both agree. Earlier tonight a dispatch bound to a stale head produced a proof that looked identical in the run list and was not valid for the PR, so this is checked rather than assumed.

Rul1an:
gate: all

Delegated proof for this exact head. Supersedes the earlier comment, whose head was invalidated by the `CONNECT_PROBE` fix.

- run: https://github.com/Rul1an/assay/actions/runs/31169440611
- head_sha: `1eca3d651f982f15c111d927d90064c7634aede8`
- gates: `all`, `build_ebpf=true`
- conclusion: success

Head confirmed with `git ls-remote` before dispatch and re-checked against the run afterwards; both agree.

Worth recording why there are two proofs on this PR: the first was valid when taken, and then `-D warnings` on the merge with main surfaced `constant CONNECT_PROBE is never used`. Fixing that moved the head, so the first proof stopped describing this tree. That is the binding working — a proof that survived a source change would be the defect.

Rul1an:
gate: all

Delegated proof for this exact head. Supersedes the two earlier comments, whose heads were invalidated by fixes found after each dispatch.

- run: https://github.com/Rul1an/assay/actions/runs/31170887547
- head_sha: `8f667cdb6283ac41be203d7bc762dbc3e2111677`
- gates: `all`, `build_ebpf=true`
- conclusion: success

Head confirmed with `git ls-remote` before dispatch and re-checked against the run afterwards.

This is the third proof on this PR, and the reason is worth recording. Each of the two invalidations came from a defect that only exists after the merge with main and that a default `cargo build` cannot see: dead code needs the whole crate to compile, rustdoc needs `--deny warnings`, and the feature error needs a non-default feature set.

Before dispatching this one I ran the gates locally instead of one CI round-trip at a time, which found a fourth: a second private-item doc link in `assay-mcp-server`, hidden behind the first because rustdoc stops at the first crate it cannot document.

Verified on this head before dispatch: rustdoc `--workspace` under `-D warnings`, clippy `--workspace --all-targets -D warnings`, `cargo check` across seven feature combinations, `--workspace --all-targets`, and the cli/monitor/mcp-server test suites.

PR #2092, opened 2026-08-06
The wall a reproducer hits first was not in the pack

### LINKED ISSUE #1990: Ship the RFC 8785 vectors with the clean-room pack: canonicalization is the wall a reproducer hits first
## Problem

The clean-room reproduction invitation in #1840 has been open since 2026-07-24 with `candidate.3` published on 07-29. No implementation attempt has been declared.

We now know, empirically, what the hardest part of a from-spec reproduction of this profile actually is, and it is not the profile.

## What the evidence says

The one cross-language reproduction anyone in this ecosystem has completed — JM-Lab reproducing RGE-Bench in Java/Jackson — **failed on its first attempt because of canonicalization**: native insertion-order serialization produced a different digest. That is recorded in that project's own `REPRODUCTIONS.md`.

Investigating further turned up something sharper. That project's published `vectors_digest` is a Python `json.dumps(sort_keys=True, separators=(",",":"))` digest, not an RFC 8785 one: its corpus carries a float (`1.0`), which Python renders as `1.0` and RFC 8785 renders as `1`. The two digests differ completely, and nothing in that repository declares which scheme is in force (filed there as `rge-bench/rge-bench#14`).

So a reproducer coming to `privileged-mcp-action/v0` faces the same wall first: implement RFC 8785 correctly from prose, with no way to check the canonicalizer before discovering that fourteen bundles disagree.

## What changed

#1982 landed 31 RFC 8785 conformance vectors in `crates/assay-canonical/tests/vectors/rfc8785.json`, each with the exact expected bytes and the property it pins — number reformatting, both ES6 exponent boundaries, UTF-16 code-unit key ordering, and the absence of Unicode normalization. They were cross-validated against an independent implementation in another language.

That is precisely the tool a reproducer needs, and it currently sits where only this workspace can see it.

## Proposal

Cut `privileged-mcp-action-v0-candidate.4` carrying the RFC 8785 vectors alongside the existing `spec.md`, `descriptor.json` and fourteen opaque cases.

**This does not weaken the clean-room property.** The pack deliberately omits expected outcomes, semantic case names, the vector generator, and this workspace's implementation. Canonicalization vectors are none of those: they describe byte formation under a published RFC that the specification already requires (`canon: jcs-rfc8785`), and they say nothing about what any bundle should verify to. An implementer who checks their canonicalizer against them still has to derive every profile outcome themselves.

Concretely, the pack should gain:

- the vector file, unchanged;
- a short note stating that it is a prerequisite check, that passing it is not progress on the profile, and that the expectations were derived from the RFC rather than from this implementation;
- a pointer to run it before implementing the stages, because a canonicalizer that is wrong makes every later result uninterpretable.

## Corpus digest is unaffected

The corpus does not change: `sha256:cb58ce91…`, 14 cases, 5 accept / 9 reject. `candidate.4` re-releases the same corpus with better onboarding, so `candidate-release.json` carries the same `corpus_digest` under a new tag. A reproduction against `candidate.3` remains valid and in scope.

## Acceptance

- [ ] `candidate.4` published from `main` with the attested pack, containing the RFC 8785 vectors and the note above.
- [ ] `candidate-release.json` names the new tag with the corpus digest unchanged.
- [ ] #1840 repinned to `candidate.4`, stating that `candidate.3` attempts remain valid.
- [ ] The note makes explicit that the vectors are not answers and that agreement with them is not conformance.

## Non-claims

This does not assert that canonicalization is the only friction, only that it is the one we have observed break a real cross-language attempt. It does not change the profile, the corpus, or the independence bar.

Related: #1840, #1982, #1977.

Rul1an:
## Description

#1990. The clean-room pack shipped `spec.md`, `descriptor.json` and fourteen opaque cases, and left an implementer to derive RFC 8785 canonicalization from prose with nothing to check it against.

That is not a hypothetical friction. The one completed cross-language reproduction in this ecosystem **failed on its first attempt** because native insertion-order serialization produced a different digest. Content ids, mandate ids and the bundle run root are all sha256 over canonical bytes, so a wrong canonicalizer makes every later result uninterpretable — and you find out fourteen bundles later.

`candidate.4` carries the 31 vectors from #1982, **byte-identical** to the file this workspace tests its own canonicalizer against, plus a note.

## Same corpus

| | |
|---|---|
| `corpus_digest` | unchanged, `sha256:cb58ce91…` |
| cases | the same fourteen |
| a `candidate.3` reproduction | **remains valid, nothing to redo** |

`candidate.4` re-releases the same corpus with better onboarding. That is stated in the corpus README so nobody mid-attempt thinks they lost work.

## It does not weaken the clean-room property

The easy half of that claim is that the expectations come from a published RFC which `descriptor.json` already requires (`canon: jcs-rfc8785`), not from our implementation. The hard half is that the vectors stay free of profile answers — and that is the half that could quietly stop being true, so it is asserted rather than argued:

```
a profile case id      \bcase-\d
a corpus vector name   \b(ok|bad)-[a-z0-9-]{3,}
an outcome field       expected_outcome|\bverdict\b
```

The note says twice, in the pack itself, that passing them is not progress on the profile and that agreement with them is not conformance.

## Two things the work surfaced

**The scorer's pack contract is a fixed member count plus an explicit name set**, so adding two files failed it. Both moved deliberately rather than being derived from whatever the builder emits — a scorer taught to expect what it is handed cannot detect a pack that grew a member.

**My first leak check was a substring sweep and it was wrong.** It flagged `case-` inside *"case-insensitive"*, in a vector's own description of code-unit ordering. Anchored now.

Then the planted-leak mutation was caught by the **byte-identity** assertion rather than by the patterns, because the pack reads the file from the pinned commit and never saw my working-tree edit. So the patterns were still unproven — a check whose firing has never been demonstrated. They now self-test against a fixture carrying one instance of each, and the test also asserts they do *not* fire on the vectors' own prose.

## Verification

- 25 activation-kit tests pass
- pack built from a pinned commit, **byte-identical across two builds**
- packed vectors sha256 `64a71a5e…` == the file at `HEAD`
- 20 members: 14 cases + 4 original + 2 new

## Acceptance

- [x] pack contains the RFC 8785 vectors and the note
- [x] `candidate-release.json` names `candidate.4`, corpus digest unchanged
- [x] the note makes explicit that the vectors are not answers and agreement is not conformance
- [ ] `candidate.4` published from `main` — release controller, after merge
- [ ] #1840 repinned to `candidate.4` — after the release exists

Does not close #1990: the last two need the release to be cut.

## Scope

`Gate.NONE`.

## Checklist

## Summary by CodeRabbit

- **Documentation**
  - Updated clean-room reproduction instructions for candidate release `.4`.
  - Added guidance to run 31 RFC 8785 canonicalization vectors before profile work.
  - Clarified release contents, attested pack details, and disclosure timing.

- **Conformance**
  - Added canonicalization vectors and supporting documentation to clean-room packs.
  - Strengthened validation to confirm required materials are complete and answer-free.
  - Updated release references and download instructions to candidate `.4`.

### DIFF
--- a/conformance/privileged-mcp-action-v0/CONFORMANCE-PROTOCOL.md
+++ b/conformance/privileged-mcp-action-v0/CONFORMANCE-PROTOCOL.md
@@ -7,15 +7,19 @@ verifier logic, interpret the profile, or establish that an implementation is in
 ## Authorship order
 
 1. Obtain the clean-room pack.
-2. Read `spec.md` and `descriptor.json`.
-3. Implement a command that consumes one bundle path and emits the profile report.
-4. Freeze the implementation commit and record all materials consulted.
-5. Run the scorer. It snapshots the canonical oracle into scorer-private memory for tamper resistance,
+2. Run `canonicalization/rfc8785-vectors.json` against your canonicalizer before anything else. It
+   is a prerequisite check, not profile work: passing it is not progress and agreement with it is
+   not conformance. It is here because a wrong canonicalizer makes every later result
+   uninterpretable, and because that is what broke the one completed cross-language attempt.
+3. Read `spec.md` and `descriptor.json`.
+4. Implement a command that consumes one bundle path and emits the profile report.
+5. Freeze the implementation commit and record all materials consulted.
+6. Run the scorer. It snapshots the canonical oracle into scorer-private memory for tamper resistance,
    executes every opaque case without passing that oracle to the candidate, and only then compares.
-6. Publish the implementation source, run record, and completed implementation report.
+7. Publish the implementation source, run record, and completed implementation report.
 
 Reading Assay's verifier, `gen_vectors.py`, the canonical `MANIFEST.json`, or a prior scored report
-before step 4 changes the methodological classification. Disclose that access rather than describing
+before step 5 changes the methodological classification. Disclose that access rather than describing
 the run as blind.
 
 ## Candidate command
@@ -102,7 +106,7 @@ steps:
     env:
       GH_TOKEN: ${{ github.token }}
     run: |
-      tag=privileged-mcp-action-v0-candidate.3
+      tag=privileged-mcp-action-v0-candidate.4
       source_digest="$(gh api "repos/Rul1an/assay/commits/$tag" --jq .sha)"
       gh release download "$tag" \
         --repo Rul1an/assay \
@@ -133,7 +137,7 @@ The release controller runs only from `main`, validates `candidate-release.json`
 checked-out corpus, and creates the annotated tag after the pack, transformation check, and
 attestation succeed. Pack verification resolves that tag to the attested source commit. The
 composite action is pinned separately to the full commit carrying the action implementation; the
-`candidate.3` release-preparation change does not alter the invoked scoring path. Keep full-commit
+`candidate.4` release-preparation change does not alter the invoked scoring path. Keep full-commit
 pinning when updating the action used by a long-lived workflow.
 
 ## Claim ceiling
--- a/conformance/privileged-mcp-action-v0/README.md
+++ b/conformance/privileged-mcp-action-v0/README.md
@@ -27,15 +27,35 @@ repinned to verified release bytes.
 
 ### Clean-room path
 
-The active release target is `privileged-mcp-action-v0-candidate.3`. Once its GitHub
+The active release target is `privileged-mcp-action-v0-candidate.4`. Once its GitHub
 [Releases](https://github.com/Rul1an/assay/releases) entry is published, it contains a
-deterministic, attested clean-room pack with `spec.md`, `descriptor.json`, and fourteen opaque
-cases. It omits expected outcomes, semantic case names, the vector generator, and Assay's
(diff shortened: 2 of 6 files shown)


Rul1an:
Taken, and right on both points.

`_about` is metadata rather than a vector, and it is the key that made this count ambiguous when it was first measured — the file carries 32 keys and 31 vectors, which is exactly the sort of off-by-one a floor hides.

The floor is the weaker half of the problem though. The byte-identity assertion above already catches the packed file drifting from the repo's, so the case this assertion actually owns is the one that survives it: **both shrinking together**. Ten vectors deleted from the source would keep the pack byte-identical to the repo and pass every other check in this test.

Now `len(cases) == 31` with `_about` excluded.

One thing stated rather than implied: this is not verified by mutation. The pack builds from a git commit, so shrinking both sides needs a throwaway commit carrying the shrunk fixture, and the end-of-file hook rewrites that fixture before the commit can be made. I tried, backed it out, and left the tree clean. The reasoning above is why the assertion is there; the mutation that would prove it is not worth a hook bypass.

Issue #2076, opened 2026-08-06
The Linux compile gate returns 0 on failure, so it cannot catch the errors it exists for


Rul1an:
## Problem

`scripts/ci/check-linux.sh` exists to catch `cfg(target_os = "linux")` compile errors from macOS. Its header says so:

> Linux-only compile guard for macOS dev: catches `cfg(target_os="linux")` compile errors.

It cannot. Both branches of `run_target_check` end the same way:

```bash
cargo check --workspace --all-targets --target x86_64-unknown-linux-gnu || {
  echo "WARN: Linux cross-target check timed out or failed. Relying on CI."
  return 0
}
```

**`return 0` on failure.** A timeout, a missing cross-toolchain, a third-party build script, and *your code failing to compile on Linux* are all the same outcome: `WARN`, and the pre-push hook reports **Passed**.

## Observed, today

PR #2074 added four `MONITOR_STAT_SEND*` reads to `crates/assay-monitor/src/loader.rs` without importing the constants. The block is `cfg(target_os = "linux")`, so macOS never compiled it.

- `cargo build` on macOS: **clean**
- pre-push `linux compile gate (target/multipass)`: **Passed**
- delegated run [31090057720](https://github.com/Rul1an/assay/actions/runs/31090057720): **failed**, four `error[E0425]: cannot find value ... in this scope`

The gate ran, could not see the error, and said nothing was wrong. A delegated dispatch on the self-hosted host is the expensive way to find that out.

## Why it always fails on this machine

The workspace cross-check does not reach our crates at all:

```
error: failed to run custom build command for `ring v0.17.14`
error: failed to run custom build command for `aws-lc-sys v0.39.1`
```

Those build scripts need a C cross-compiler. So on a stock macOS dev box the check fails **every time**, takes the `return 0` path **every time**, and has never been able to report anything.

Narrowed to the crates that actually carry `cfg(linux)` code, it works and is fast:

```
cargo check -p assay-monitor -p assay-runner-core --target x86_64-unknown-linux-gnu
```

Without the fix: 4 `E0425`. With it: 0. That is the check reproducing the exact failure the delegated host found.

## Direction

Distinguish **"could not check"** from **"your code does not compile on Linux"**. The first may warn; the second must fail. That is the same distinction #1993 drew for the release gate, and the same one `ErrorClass` refuses to blur elsewhere in this repo.

Concretely, one of:

[...]

Issue #1993, opened 2026-08-04
The release gate cannot be tested without cutting a release, and it has failed open three times


Rul1an:
## Problem

The release-blocking gate added in #1985 is ~25 lines of inline bash inside `release.yml`. In a single day it acquired three fail-open defects, all of the same kind — the gate deciding it had nothing to check and exiting 0:

1. **Milestone lookup swallowed API failures.** `if ! gh api ... | grep -q` treats a rate limit as "no such milestone", because `set -e` is suspended inside an `if` condition and `!` inverts the pipeline's non-zero status. Caught in review.
2. **The lookup read only the first page.** `per_page=100` without `--paginate`, so a repository with more milestones than that disables its own gate. Caught in the same review.
3. **The version was prefixed twice.** `resolve-release-version.sh` returns the tag form with a leading `v`; the gate prefixed another one, looked for `vv3.38.0`, found no such milestone, and exited 0. Caught by running the resolver instead of assuming its output, hours before the first real release.

The third is the instructive one. The query logic had been verified against live repository state twice. What had never been checked was what the *variable* contained. Every one of these logs like success.

## Why inline bash is the wrong home for it

This step decides whether a release may proceed. It cannot be exercised without cutting a release, so its only test is production, and its failure mode is silence. That is the exact shape this repository refuses everywhere else: a check that cannot fail does not get to report a pass.

The repository already has the pattern — `scripts/ci/*.sh` with `test-*.sh` companions (`test-release-assets.sh`, `test-release-attestation-enforce.sh`, `test-release-channel-separation.sh`, and others).

## Proposal

Extract the gate to `scripts/ci/release_blocking_gate.sh`, taking the version and repository as inputs and shelling out for the two API calls, with `scripts/ci/test-release-blocking-gate.sh` beside it. `release.yml` calls the script.

Cases the test should pin, each of which is a defect that has already occurred or would pass silently today:

- `v3.38.0` and `3.38.0` resolve to the same milestone;
- an unrecognised version refuses rather than guessing;
- a milestone with an open `release-blocking` issue fails and names it;
- a milestone with none passes;
- **no milestone at all passes** — this is the one deliberate open path, and it should be pinned so it stays deliberate;
- a milestone listing that fails refuses rather than treating the failure as absence;
- pagination: a milestone beyond the first page is still found.

The last two need the API calls behind a seam the test can substitute. That seam is the point of the extraction, not a side effect of it.

## Acceptance

[...]

PR #2094, opened 2026-08-06
The semver gate skipped 254 checks and called it success

### LINKED ISSUE #2088: The semver gate ran on the PR that broke semver, and passed: its baseline is pinned two majors back

Rul1an:
## Description

Closes #2088.

`WAVE0_SEMVER_BASELINE_SHA` was pinned to `9cc23b4c` — `chore: release v2.18.0`, dated 2026-02-11.

`cargo semver-checks check-release` does not ask whether there is a breaking change. It asks whether the **declared version increment covers the changes it finds**. Against a 2.x baseline the manifest already declared 3.x, so:

```
Checking assay-core v2.18.0 -> v3.38.0 (major change)
Checked [0.000s] 0 checks: 0 pass, 254 skip
Summary no semver update required
```

Zero checks run, reported green. The job had been structurally unable to fail for the entire 3.x line — and a gate that skips every check looks exactly like a gate that finds nothing.

## Measured, not argued

Same tree (#2068's merge commit), baseline moved to the last release tag:

```
Checked [0.115s] 223 checks: 222 pass, 1 fail, 0 warn, 31 skip

--- failure constructible_struct_adds_field ---
  field MetricResult.exercised in crates/assay-core/src/metrics_api.rs:38

Summary semver requires new major version: 1 major and 0 minor checks failed
```

223 checks instead of 0, naming the exact field that reached `main` unnoticed and was found by hand during a release prep.

The other direction was run too: **4.0.0 against `v3.38.0` passes**, because the bump covers the break. A release PR stays green. Both cases were executed against the real tool before this was written.

## The fix

The baseline is `git tag --list 'v[0-9]*' --sort=-v:refname | head -n1`. **No tag is a hard failure**, not a skip — "could not check" and "found nothing" must not be spelled the same way, which is the rule #2076 and #1993 were both fixed to follow.

`scripts/ci/test-semver-gate.sh` holds it: cheap cases pre-push, planted-break case in the workflow's own job under `ASSAY_SEMVER_GATE_FULL=1`. It plants a public field on a public struct — the same shape as the break that got through — and requires a non-zero exit **and** the right lint name.

## Three things caught by the checks that seemed pedantic

**The self-test first expected a failure and got a pass.** `main` is at 4.0.0 with the newest tag at `v3.38.0`, so `check-release` sees a declared major and skips everything again. That is the gate being *right* — a break merged before 4.0.0 ships is licensed by the major that hasn't shipped — and my test being wrong. But a self-test whose verdict depends on whether a release happens to be pending tests nothing, so it equalises the versions first.

**Equalising the root version alone failed on `cargo metadata`**, because nine internal dependencies still declared 4.0.0. The gate failed on a resolver error rather than on the lint. That is exactly why the check asserts *why* it failed and not merely that it did.

**`BASELINE_TAG` went into the wrong job.** A first-match replace put it in `feature-matrix`, which has no `baseline` step. actionlint named it before it could ship.

## One property worth knowing

Between a release merge and its tag, the manifest declares a major the baseline doesn't have, so checks skip. That is semantically correct — those breaks are covered by the unshipped major — but it means the gate is quiet in that window. It closes when the tag lands, which is an argument for tagging promptly after a release merges.

## Scope

`Gate.NONE`.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release compatibility checks by automatically using the latest available release as the comparison baseline.
  * Added failure handling when no release is available, preventing misleading validation results.

* **Tests**
  * Added automated validation to confirm semver checks detect breaking public API changes.
  * Added lightweight pre-push validation and an optional comprehensive verification mode for CI.

### DIFF
--- a/.github/workflows/split-wave0-gates.yml
+++ b/.github/workflows/split-wave0-gates.yml
@@ -17,7 +17,6 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
-  WAVE0_SEMVER_BASELINE_SHA: "9cc23b4c684be7cfd81f170c4f66d59903dd76eb"
 
 jobs:
   detect-changes:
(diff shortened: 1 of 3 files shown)


Rul1an:
Thanks — three of the four are taken, and two of those were this gate committing the defect it was written to prevent.

**`git fetch --tags --quiet origin || true`** (line 416). Correct, and the sharpest of the four. A failed fetch fell through to whatever tags the checkout carried, so the job would resolve a stale baseline, compare against the wrong release, and report success. This whole PR exists because "could not check" and "found nothing" were spelled the same way; the suppression is gone.

**`semver gate: all cases pass` on the skipped path.** Also correct. Without `ASSAY_SEMVER_GATE_FULL` the only case that proves the gate reaches a verdict does not run, and printing "all cases pass" made a green line stand in for a check nobody performed. It now reads `semver gate: shape cases pass; the planted-break case was NOT run`.

**Isolated `CARGO_TARGET_DIR`.** Taken. The full self-test downgrades the workspace version and plants a breaking change, so its artifacts are keyed to a version and a source that do not exist and should not persist in the shared target directory.

**Pinning `cargo-semver-checks` — declined, with a reason.**

The concern is legitimate: an unpinned tool can change its verdict without a code change. But the planted-break case asserts both a non-zero exit *and* the lint name `constructible_struct_adds_field`, so if that lint is renamed or removed the self-test goes red on the next run. Tool drift surfaces as a failing check rather than as a quiet shift in what the gate accepts.

Pinning would suppress that signal and add a version constant somebody has to bump — the same shape as the hardcoded baseline SHA this PR removes, one level over. If the self-test is ever weakened so it no longer names the lint, pinning becomes the right call and this reasoning should be revisited.

Both modes re-verified: cheap mode now reports what it did not run; full mode still kills the planted break and names the lint that caught it.

Issue #2093, opened 2026-08-06
Build the run that emits and consumes a seal: the primitive exists and nothing produces one


Rul1an:
## What exists

ADR-045's seal primitive is built and checked:

| | |
|---|---|
| payload contract + fixture checker | `scripts/experiments/aee_landlock_seal_fixture.py` |
| production-shaped DSSE envelope | `crates/assay-cli/src/aee_seal_envelope.rs`, Ed25519 over the shared PAE |
| negative controls | 33, each required to fail for its own reason |
| rule/control coverage | bidirectional, rule set parsed from source |

## What does not

**Nothing consumes it.** `git grep 'aee_seal::'` across the workspace returns the two `pub mod` lines in `main.rs` and nothing else. No command constructs a seal, no command verifies one, no test outside the modules touches them.

So the honest description of today's state is that the primitive is *specified and tested*, not that a run produces one. The seal-eligibility refusal, the drop-proof models, the `checked`/`declared` basis — all of it is library code that no execution path reaches.

## Why this is filed now

The predicate author's instruction on [in-toto/attestation#570](https://github.com/in-toto/attestation/pull/570) was two clauses:

> "Build the seal first, **then a run that emits and consumes**, and I will treat anything it breaks as a defect in the specification until shown otherwise."

Clause one is done. Clause two has never been tracked anywhere — not in #1998 (which covered the design and is closed), not in the productization epics. It has lived in a PR body and a memory file, which is how work that someone is waiting on goes quiet.

He also made a standing offer that this work is the thing it depends on:

> "I will hold the shape of sealed and the run binding stable while you build, and if something has to move I will say so here first."

That offer is conditional on building, and it does not have an indefinite shelf life.

## Scope

**1. A command that emits a seal from a real run.** Landlock/TCP-connect first, since that is the collection path the production slice carries. It must go through `seal_eligibility` and refuse rather than emit when the drop accounting cannot be proved — that refusal is already code and has never executed.

**2. A consumer.** Verify the emitted seal, so the round trip is closed inside this repo before anything is claimed outside it.

**3. The second collection path.** The JSON-RPC proxy, as a second collection path into **one** substrate under **one** key. Two keys makes two substrates and the run binding stops resolving; that constraint came from the author directly.

This is the part worth doing for its own sake rather than for the integration: a two-vantage substrate under one operator is, in his words, "the case this text has the least evidence behind, and yours is the only one that exists." The spike carried two paths as synthetic fixtures. Production carries one.

[...]

PR #2099, opened 2026-08-06
A seal signed while the enforcement is still in force

Rul1an:
## Description

#2093, slice C. **Stacked on #2096 (A) and #2097 (B)** — it calls both, so it cannot build until they land, and its diff shrinks to slice C alone once they do. #2093 stays open after this: the second collection path and the `aeeVersion` 0.7/0.8 decision remain.

The seal is emitted from `sandbox/child.rs`, in the enforcing process, at the moment the health record is written. A later command signing a health artifact off disk would be claiming a property of a moment that has passed — the gap ADR-045's run-end requirement exists to close.

## The research changed what this slice had to be

ADR-045 permits `aeeStillArmed` to rest on Landlock's irrevocability. That invariant is real and documented — and it has an exception.

**[CVE-2024-42318](https://ubuntu.com/security/CVE-2024-42318)** ("Landlock Houdini", Jann Horn, 2024-07-24): Landlock implemented the `cred_prepare` LSM hook but not `cred_transfer`, so `keyctl(KEYCTL_SESSION_TO_PARENT)` dropped **every** Landlock restriction. Present from 5.13; fixed in v6.11-rc1, commit `39705a6c29f8`.

**`abi >= 4` does not exclude it.** ABI 4 arrives at kernel 6.7, four releases before the fix:

| | |
|---|---|
| `LANDLOCK_ACCESS_NET_CONNECT_TCP` (ABI 4) | 6.7 |
| CVE-2024-42318 fixed | 6.11 |
| what the record carried | `abi`, and no kernel version |

So every kernel in 6.7–6.11 passed every check `seal_eligibility` made, while the workload child — which by definition runs the code under test — could shed its sandbox with two syscalls.

## Why it is measured rather than gated on a version

Closing this by kernel version needs a per-distribution backport table, and a table of backport facts rots silently. Ubuntu Noble carries the fix at **`6.8.0-50.51`**, which no comparison against 6.11 would ever discover — our own runner would have been refused while patched.

`backend::landlock_shedding_probe` runs the CVE sequence against a throwaway process pair and reports **held / shed / inconclusive**. `seal_eligibility` requires `restrictions_held`; `inconclusive` is a probe that found nothing, which is not a finding that restrictions hold.

It needs **two** forks: `KEYCTL_SESSION_TO_PARENT` replaces the *parent's* credentials, so the process under test must be the caller's parent. The real process is never a target, which is why running the exploit deliberately does not weaken it.

Every enforcement-health record written before this field existed is now refused — by *fixture*, not a hand-built struct, because each one would otherwise have sealed with `aeeStillArmed: true` resting on an invariant nothing checked.

## Verified on the runner, not inferred

`6.8.0-136-generic`, patched at `50.51`. A real Landlock run, our own Ed25519 key:

```
restriction_shedding: restrictions_held
aeeStillArmed: true          assayDropProofBasis: declared
aeeObservedAttacks: []       assayCollectionPath: landlock-tcp-connect
aeeRunBinding: 50b143da...   945 payload bytes, 64 signature bytes
keyid sha256:af5dc9d8...
```

And without a run-end probe it refuses, loudly, writing nothing:

```
WARN: aee-seal: not seal-eligible: no run-end probe: a start-time restrict_self
confirmation proves the ruleset was applied, not that it was still applied at
run end; no seal was written
```

`assayDropProofBasis: declared` is deliberate and not dressed up: the synchronous-probe model's zero is this producer asserting its topology has no lossy channel, not a counter read.

## The posture-digest trap

Built declared-digest-first, the order ADR-045 names under *Field interpretation*: `aeePostureDigest` is the digest computed **before** the `digest` member is inserted, while `run_binding` hashes the object **including** it. `digest_json_public` shares the seal's own canonicalization rather than adding a second answer to "what is this object's digest".

## Two things this cost

**`SealSignError` had no `Display`** — invisible until the Linux path compiled for the first time on the runner.

**The three new args first carried long `///` comments**, which flip clap into multi-line help for the whole command and broke `sandbox --profile-format`'s `[possible values]` advertisement. Caught by #2086's own test, one day old. Same mistake as `doctor --format` that morning: why a flag exists belongs next to the code, not in `--help`.

## Boundary

The Linux **unit** tests did not run on the runner: `/opt/rust/registry` is root-owned there, so a test-only dev-dependency could not be fetched. The Linux path is verified by the end-to-end run above, which exercises more than a unit test would. The platform-independent suite (51 suites) passes on the dev box.

## Scope

`gates=all` — touches the sandbox enforcement path. Needs a delegated proof bound to its head.

## Checklist

### DIFF
--- a/crates/assay-cli/src/aee_seal.rs
+++ b/crates/assay-cli/src/aee_seal.rs
@@ -33,6 +33,12 @@ const BLOCKING_ERRNO: &str = "EACCES";
 /// that cannot express a TCP-connect restriction at all.
 const LANDLOCK_ABI_NET_CONNECT_TCP: u32 = 4;
 
+/// The one `restriction_shedding` value that lets a seal claim `aeeStillArmed`.
+///
+/// Same spelling as `backend::SheddingProbeOutcome::label`, and only that value: `inconclusive`
+/// means the probe found nothing, which is not the same as finding that restrictions hold.
+const RESTRICTIONS_HELD: &str = "restrictions_held";
+
 /// The operator-facing label for what this probe observed.
 fn observed_label(probe: &Probe) -> String {
     if probe.listener_reached {
@@ -74,6 +80,14 @@ pub enum NotSealEligible {
     UnprovenContextValue { field: &'static str },
     /// A digest the seal must derive could not be computed from the inputs given.
     DerivationFailed { what: &'static str },
+    /// The run did not establish that Landlock restrictions cannot be shed on this kernel, so
+    /// `aeeStillArmed` would rest on an invariant this run has no evidence for.
+    ///
+    /// ADR-045 lets still-armed rest on "a documented kernel-level invariant showing the applied
+    /// Landlock restrictions cannot be relaxed". CVE-2024-42318 is that invariant's exception --
+    /// `keyctl(KEYCTL_SESSION_TO_PARENT)` shed every restriction from 5.13 until v6.11-rc1 -- and
+    /// `abi >= 4` does not exclude it, because ABI 4 arrives four releases earlier at 6.7.
+    RestrictionSheddingNotEstablished { found: String },
     /// A counted-queue channel reported lost observations, so zero cannot be carried.
     ObservationsLost { channel: String, lost: u64 },
     /// A counted-queue model was declared with no channels, which proves nothing.
@@ -95,6 +109,9 @@ impl NotSealEligible {
             Self::RecordSelfContradictory { .. } => "record-self-contradictory",
             Self::UnprovenContextValue { .. } => "unproven-context-value",
             Self::DerivationFailed { .. } => "derivation-failed",
+            Self::RestrictionSheddingNotEstablished { .. } => {
+                "restriction-shedding-not-established"
+            }
             Self::ObservationsLost { .. } => "observations-lost",
             Self::DropAccountingUnnamed => "drop-accounting-unnamed",
         }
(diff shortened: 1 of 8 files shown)


Rul1an:
## The Linux probe has now been executed, not only compiled

`lane-check` passes this PR without a delegated proof, and that is correct — I ran `classify_file` over all nine files it touches and every one scores `Gate.NONE`.

But `Gate.NONE` for `backend.rs` means the new `landlock_shedding_probe()` would have merged having never *run*: this machine is macOS, `assay-cli` cannot be cross-checked locally (`ring`/`aws-lc-sys` need a C cross-compiler), and nothing in the crate's tests exercises it — the tests near it cover `classify_self_probe_exit`, which belongs to the older probe. CI's ubuntu build compiles it and stops there.

What I had proven earlier was a **C prototype** of the algorithm. That is not this code.

So the probe was extracted verbatim from this branch, compiled with real `rustc` on the `assay-bpf-runner` VM, and run against kernel `6.8.0-136-generic`:

```
outcome: RestrictionsHeld  label=restrictions_held
exit=0
```

Which matches the C prototype exactly, and matches Ubuntu's own record — Noble backported the CVE-2024-42318 fix at `6.8.0-50.51`, and this host is 86 ABI revisions past it.

**What this proves and what it does not.** Same source lines, different compilation unit: it establishes the probe's logic, its syscall usage, and its verdict on a real kernel. It does not establish integration with the rest of the crate, and it is not a delegated proof.

The three compile errors hit on the way were all artefacts of my extraction — a duplicated helper, a visibility, a missing `impl` — not defects in the probe.

Filed #2103 for the underlying gap: the gating map scores the file holding this repository's Landlock syscall wrappers as `Gate.NONE`, so kernel-behaviour changes reach `main` without a delegated proof. That is a question about the map, not about this PR.

Issue #2014, opened 2026-08-05
The seal names a scope and the checker has no rule on it


Rul1an:
Found by an adversarial review of #2013, in code that is already on `main` (#2008).

## The gap

`aee_landlock_seal_fixture.py` never validates `assaySealScope`. The member appears exactly once in the file — in `base_statement()`, where the fixture is *built* — and nowhere in `validate()`.

Verified by substituting the field in the positive fixture and re-running the checker:

```
assaySealScope = "filesystem_write_all"   -> credited   []
assaySealScope REMOVED entirely           -> credited   []
```

So a seal produced by a Landlock TCP-connect observation can name a filesystem enforcement scope, or name none at all, and the checker credits it as attested substrate evidence.

## Why it matters

`assaySealScope` is what tells a consumer *which bounded enforcement scope* the seal speaks for. ADR-045 lists it as required and the #2001 field contract says "First slice MUST equal `tcp_connect_landlock_port`". An unchecked scope member means the strongest record in the statement can claim a boundary it never observed — the checker's own `key-scope-collection-path-mismatch` rule exists to stop exactly this class of thing one field over.

## Scope of the fix

Add a payload-only rule to `validate()`: `assaySealScope` MUST be present, a non-empty string, and equal to the first-slice constant. Then add the negative control (`seal-scope-mismatch`) and its entry in `NEGATIVE_CONTROLS` so `--meta-test` proves it isolates its own reason.

Note the producer side is being fixed separately in #2013, which now compares the seal scope against the health record's scope. That fix does not close this: it constrains what *Assay's* producer emits, and the checker's job is to constrain what *anyone's* producer emits.

## Not in scope

Any change to `aeeVersion`, the fixture policy, or the trust-scope rules added in #2006.

Rul1an:
Widening this: `assaySealScope` is one of **eight** sketch rules the checker does not implement, and two of the others are worse. All verified against `validate()` by mutating the positive fixture and re-running.

## Missing rules

| Sketch / spec rule | In `validate()`? |
|---|---|
| Required field absent | no — no presence check on any member |
| Digest not lowercase 64-hex | **no** |
| `aeeVersion` != `0.7` | no |
| `aeeMethod` != `intercepted` | no |
| `assayCollectionPath` == literal | only against key scope, never the literal |
| `assayAttackRowAttributionSource` ∈ {2 values} | only the `substrate-runner` branch |
| `assayNonClaims` contains the required entries | no |
[...]

PR #2108, opened 2026-08-07
fix(adr-045): the six siblings the prefix rule already covered

Rul1an:
> **#2107 is merged**, so this now targets `main` directly. It introduced `PRODUCER_CREDIT_FIELDS` and `--producer-vocabulary-test`; this PR decides the six members it parked, and extends that test so it can see them.

## ADR Boundary Slice

- Canonical ledger: **none active.** AGENTS.md records that the previous programme ([#1866](https://github.com/Rul1an/assay/issues/1866)) closed on 2026-08-01 and no new one has opened. This is a standalone follow-up with no ledger to record to.
- Slice: ADR-045 producer-vocabulary prefix rule — the six sibling members #2107 parked
- Builder: Claude (Opus 5)
- Reviewers: 
- Final head SHA: `e78c4f502e7dadbf9cdd59f58ff2df95a7949407`
- ADR invariants affected: ADR-045 prefix rule (`assay`-prefixed members MUST NOT alter AEE structural validity); ADR-043's requirement that the three outcomes stay distinguishable

## Behavior

#2107 made `assayDropProofModel` inert to AEE structural validity and parked six siblings as "six contract questions, not one edit". They are decided here, and all six move to `PHASE_NOT_CREDITED`.

**The reason they were parked does not survive being written out.** It was that the normative paragraph at in-toto/attestation#570 governed members whose values a reader might rank, and these carry identities, instants and paths. That was true of the upstream paragraph **as it stood at `35ca6f899`**, and beside the point either way. ADR-045's own prefix rule is not scoped to rankable members:

> Fields beginning with `assay` in the sealed payload are Assay producer vocabulary. […] They MUST NOT alter AEE structural validity.

All of them, and it has said so since the ADR was written. Six local decisions were being held against an upstream sentence that could only ever be **narrower** than the rule we already had.

**Upstream then answered, and the earlier draft of this PR got that wrong.** When this was written, the question at [#570 (comment)](https://github.com/in-toto/attestation/pull/570#issuecomment-5216879286) — whether a producer member is inert whether or not its values rank — was open, and the body said so. It was answered at `237f83b9f`, **sixteen minutes** after we asked, in a commit titled *make producer territory inert, not merely unrankable*. The paragraph now reads:

> a producer-defined member MUST NOT affect structural validity, MUST NOT affect `result`, and MUST NOT affect the evidence tier, whether or not its values can be ordered

So the upstream rule and the local prefix rule now say the same thing, and this slice is what the two **jointly require** rather than what one of them permitted. Nothing about the six decisions changes — the reasoning never rested on the upstream paragraph, which is why it survived the paragraph moving — but a decision recorded as unconfirmed when it is confirmed reads as weaker evidence than it is. Corrected in `e78c4f5`, along with the tense of the scoping argument.

The upstream commit message names the gap as ours, unprompted: an implementer who "also emit[s] an unordered producer member, per-channel loss counters" (`assayDropChannels`), and a checker that "gated AEE validity on a producer member's value, which their own design document forbade in as many words". Both halves of #2107 and this PR.

**One boundary this does not settle**, named in the ADR rather than assumed. The new paragraph opens with a verifier reading *nothing* in producer territory, while the three obligations it states are structural validity, `result`, and the evidence tier. The not-credited verdict here is none of those three — it is a consumer trust decision of the same class as an untrusted signing key, which ADR-045's key-scope section already treats as the consumer's own, and AEE's `result` is a field this checker does not emit. That is a reading, not a quotation, and it is recorded as one so it gets re-checked if the opening sentence acquires normative weight.

### What was actually worth checking

Not whether the rule reaches each member — it always did. Whether moving it gives up what the structural phase was buying.

| Member | Rule(s) moved | What the check found |
|---|---|---|
| `assaySourceSchema` | `payload-source-schema-invalid` | Non-empty-string gate on a member the ADR table marks non-normative; no other rule reads the value. Nothing attached. |
| `assaySealScope` | `seal-scope-missing`, `seal-scope-mismatch` | Reading #2014 **settles** this rather than blocking it. The harm it names is that the checker "credits it as attested substrate evidence", and the rule it holds up as the model to copy is `key-scope-collection-path-mismatch` — already not-credited. Withholding credit is the whole of what #2014 asked for. |
| `assayCollectionPath` | `payload-collection-path-mismatch` | This member was being read at **both phases in one file**. The ADR's own key-scope section specifies the not-credited one: a key outside scope "is structurally valid but not credited". #2106 reached the same place from the producer side — a path types a vantage, and must not rank one above another. |
| `assayAttackRowAttributionSource` | `payload-attribution-source-unknown`, **`substrate-runner-observed-attacks-mismatch`** | See below. |
| `assayNonClaims` | `payload-non-claims-incomplete` | The prefix paragraph rules on this member by name: it "is only producer vocabulary and does not weaken required AEE checks". A rule that voids the record does the converse. A subset is still "a rejection rather than a warning" — not-credited is a rejection carrying a reason code. |
| `assaySealedAt` | `seal-instant-invalid` | The one with a real entanglement, so the one to check rather than assume. The parsed instant feeds `key-outside-validity-window`, and an unparsable value makes that check `continue` — so the structural phase was load-bearing for a bad instant not bypassing key expiry. It survives because *this* finding already withholds credit: no path reaches `credited`. The window is not weakened. |

### The finding the parked list did not contain

`assayAttackRowAttributionSource` had a **second** rule, and it is a different kind of violation. It is not only gated against a closed set — its value *selects how strictly an AEE member is checked*. Per ADR-045's attack-attribution rules 2–3, equality between `aeeObservedAttacks` and the caught row attack IDs is required under `substrate-runner` and not under `assembly-plane`.

Measured on the positive fixture at `ede1ee7c1`:

```
assayAttackRowAttributionSource = "assembly-plane"   (as fixtured) -> credited
assayAttackRowAttributionSource = "substrate-runner" (no other edit) -> malformed  ['substrate-runner-observed-attacks-mismatch']
```

One member, between its **two legal values**, moving a record across the structural boundary. That is the prefix rule's failure mode in its purest form: two consumers disagreeing about whether a record is well formed, because one reads Assay vocabulary and the other does not.

Moving only the closed-set rule would have left the property **false with `--producer-vocabulary-test` green**, because `substrate-runner` is neither absent, ineligible, nor ill-typed. Both rules move, and the test gained the ability to see it.

The baseline rule — every named attack must be supported by a caught row — reads only AEE members and stays structural. Only the tightening is Assay policy, so only the tightening withholds credit.

### Two checks added, because the existing ones could not see the move

1. **`PRODUCER_CREDIT_ALTERNATE_VALUES`** — mutate a member to one of its own *legal* values, asserting only that the record is not voided (a legal value may rightly leave it credited). This is the only mutation that reaches the selector case.
2. **`validate` read back** — every `assay`-prefixed member the function consults must appear in `PRODUCER_CREDIT_FIELDS`, so the next producer member to acquire a rule cannot acquire a phase by default. This is what makes "every producer member any rule reads is decided" a checked property rather than a comment.

Also: the generic ineligible value is now `""` rather than a borrowed one. `assaySourceSchema` accepts any non-empty string, so a plausible-looking wrong value passed straight through and the check read as "credit was not withheld". `""` is the one value none of the seven members' own rules accept.

### One deliberate divergence from #2107

The six **stay in `REQUIRED_SEAL_FIELDS`** (#2107 removed `assayDropProofModel` from it). That list states what the *Assay producer contract* obliges a seal to carry, and `--required-field-test` asserts absence is `!= CREDITED` — which not-credited satisfies. So "required" reads as required-for-credit for a producer member, which is the strongest thing the prefix rule permits it to mean. Keeping them preserves the per-member coverage proof, and keeps the ADR's existing "`REQUIRED_SEAL_FIELDS` names **six** of the ten producer members" sentence true without an edit.

### Where this leaves the rule

Seven of the ten producer members are read by some rule in the checker, all seven are enforced inert by test, and the remaining three (`assayObservedLabels`, `assayDropProofBasis`, `assayDropChannels`) are inert by having no rule — nothing can demote what nothing consults. **Every rule that can still return `malformed` reads an `aee*` member, the envelope, or the predicate structure.**

## Non-Claims

This changes only which of three existing outcomes a checker returns for a fixture. It is not a production exporter, and ADR-045 still blocks one.

## Test Evidence

All measurements on `e78c4f502e7dadbf9cdd59f58ff2df95a7949407` (rebased onto `main` after #2107 squash-merged; content identical to the reviewed tree plus the `e78c4f5` doc correction), worktree `friendly-matsumoto-e291ed`, macOS 25.6.0 (darwin), stable toolchain.

### RED

The property is a phase — one word passed to `add` — so the failing state is demonstrated by mutation rather than by a test written first. Each mutation applied in-memory against the committed source:

```
bite 1  assaySealScope removed from PRODUCER_CREDIT_FIELDS        -> CAUGHT
        FAIL assaySealScope: `validate` reads this producer member and it is not in PRODUCER_CREDIT_FIELDS

bite 2  payload-attribution-source-unknown back to PHASE_MALFORMED -> CAUGHT
        FAIL assayAttackRowAttributionSource absent: a producer member voided the record
        FAIL assayAttackRowAttributionSource ineligible: a producer member voided the record
        FAIL assayAttackRowAttributionSource wrong type: a producer member voided the record

bite 3  substrate-runner-observed-attacks-mismatch back to PHASE_MALFORMED -> CAUGHT
        FAIL assayAttackRowAttributionSource legal value 'substrate-runner': a producer member voided the record
```

Bite 3 is the one that matters: it is caught **only** by the new alternate-value mutation. The three generic mutations stay green through it, which is the gap this PR closes.

### GREEN

```
--meta-test                  PASS   33/33 controls isolate their reason
--required-field-test        PASS   16/16 required fields rejected when absent
--rule-coverage-test         PASS   33/33 reason codes have a control
--producer-vocabulary-test   PASS   7 credit fields x 22 mutations, all structurally-valid-not-credited;
                                    `validate` reads no producer member outside the tuple

cargo test -p assay-cli      53 test binaries ok, 0 failures
cargo fmt --all -- --check   clean
cargo clippy -p assay-cli --all-targets -- -D warnings   clean
git diff --check             clean
pre-commit (4 changed files) all hooks pass
scripts/ci/check-aee-seal-fixture-drift.sh  "fixtures reproduce from the emitter, with no extra files"
```

Documented smoke loop in `docs/experiments/aee-landlock-seal-fixtures-2026-08.md` re-run verbatim: positive credited, all 33 negative controls reject with their own reason, exit 0.

Outcome shift across the negative controls: exactly the 8 changed rule sites moved from `malformed` to `structurally-valid-not-credited` (`empty-source-schema`, `seal-scope-absent`, `seal-scope-other-boundary`, `seal-collection-path-other`, `unknown-attribution-source`, `substrate-runner-observed-attacks-mismatch`, `incomplete-non-claims`, `seal-instant-not-rfc3339`). No control changed reason code; the positive fixture is still `credited`.

`adr045_citations.rs` caught two paraphrases I had written inside quotation marks near an ADR-045 mention. Fixed, and every remaining quotation was re-grepped against the source it is attributed to (ADR-045, the checker's own comment, and issue #2014) before this was pushed.

### Not covered

No Rust production code changed, so the Rust producer's behaviour is unchanged and untested by this PR — `crates/assay-cli/src/aee_seal.rs` still emits the same twenty members. This moves only what the checker says about them. The Linux-only paths are untouched.

## Review Quorum

- [ ] One non-building agent review

The builder's self-review does not count, so the boxes above are for reviewers to fill. Auto-merge is deliberately **not** enabled: the quorum above is unmet.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -163,7 +163,8 @@ repos:
         # `--producer-vocabulary-test` says an Assay producer member withholds credit and never
         # voids the record. The other three are blind to it: a phase is one word passed to `add`,
         # so `assayDropProofModel` decided AEE structural validity for two slices while all three
-        # stayed green.
+        # stayed green, and six sibling members did for four. It also reads `validate` back, so a
+        # producer member that acquires a rule cannot acquire a phase by default.
         #
         # pre-commit, not pre-push: all four run in well under a second, and they are the tests
         # that tell you a rule you just wrote does not isolate.
--- a/docs/architecture/ADR-045-aee-substrate-signed-run-end-seal.md
+++ b/docs/architecture/ADR-045-aee-substrate-signed-run-end-seal.md
@@ -267,16 +267,83 @@ the checker's known-optional set names the other four, and the checker's positiv
 carries **eight** — the drop-proof basis and channels are permitted rather than fixtured, since the
 Rust producer's key-set test is what pins their emission.
 
-Bounded deliberately, and the bound is the honest part. Six sibling producer members —
-`assayCollectionPath`, `assaySealedAt`, `assaySourceSchema`, `assaySealScope`,
-`assayAttackRowAttributionSource`, `assayNonClaims` — still raise structural findings in this same
-checker, so the prefix rule below is enforced for one member and stated for eleven. They carry
-identities, instants, and paths rather than a rankable axis, which is why the normative paragraph
-in in-toto/attestation#570 reaches `assayDropProofModel` and not them, and each has a reason for
-being structural that predates this decision (`assaySealScope` carries #2014's rule that a seal
-must not claim a boundary nothing observed). Moving them is six contract questions, not one edit.
-They are named here so the decision is not read as wider than it is, and so the next pass has a
-list rather than a rediscovery.
+**The six siblings, decided 2026-08-07.** The passage above bounded itself to `assayDropProofModel`
+and parked six sibling producer members — `assaySourceSchema`, `assaySealScope`,
+`assayCollectionPath`, `assayAttackRowAttributionSource`, `assayNonClaims`, `assaySealedAt` — as six
+contract questions rather than one edit. They are decided here, and all six move.
+
+The reason they were parked does not survive being written out. It was that the normative paragraph
+in in-toto/attestation#570 governs members whose values a reader might rank, and these carry
+identities, instants and paths instead. That is true of the upstream paragraph and beside the point
+for us: the prefix rule below is not scoped to rankable members. It says *fields beginning with
+`assay`*, all of them, and has said so since this ADR was written. So six local decisions were being
+held against an upstream sentence that could only ever be narrower than the rule we already had. The
+question left open at in-toto/attestation#570 — whether a producer member is inert whether or not
+its values rank — is still unanswered upstream as of this decision, and is not load-bearing here. An
+answer would confirm the local rule, not extend it.
+
+What was worth checking per member was never whether the rule reaches them. It was whether moving
+each one gives up something the structural phase was buying.
+
+- **`assaySourceSchema`** (`payload-source-schema-invalid`). A non-empty-string gate on a member the
+  table above marks non-normative, and no other rule reads the value. Nothing attached, nothing lost.
+- **`assaySealScope`** (`seal-scope-missing`, `seal-scope-mismatch`). #2014 is why this rule exists,
+  and reading it settles the question rather than blocking it. The harm #2014 names is that the
+  checker "credits it as attested substrate evidence", and the rule it holds up as the model to copy
+  is `key-scope-collection-path-mismatch`, which is a not-credited rule. Withholding credit is the
+  whole of what was asked for. A seal naming `filesystem_write_all` is still refused; it is refused
+  as one this consumer will not credit rather than as a record no consumer can parse.
+- **`assayCollectionPath`** (`payload-collection-path-mismatch`). This member was already being read
+  at both phases in one file. `key-scope-collection-path-mismatch` withholds credit and
+  `payload-collection-path-mismatch` voided the record, and the key-scope side is the one this ADR
+  specifies itself: a credited signature requires that "the key's trusted scope includes the
+  payload's `assayCollectionPath`", and a key outside scope "is structurally valid but not credited".
+  #2106 reached the same place from the producer side — a path types a vantage, and must not rank one
+  above another.
+- **`assayAttackRowAttributionSource`** (`payload-attribution-source-unknown`, and
+  `substrate-runner-observed-attacks-mismatch`). Two rules, and the second is why this member was the
+  worst of the six rather than one more of them. It is not only gated against a closed set; its value
+  *selects how strictly an AEE member is checked*. Under the attack-attribution rules below, equality
+  between `aeeObservedAttacks` and the caught row attack IDs is required when the value is
+  `substrate-runner` and not when it is `assembly-plane`. Changing the positive fixture from one of
+  its two legal values to the other, with no other edit, turned a credited record into a malformed
+  one. Both rules move. The baseline — every named attack must be supported by a caught row — reads
+  only AEE members and stays structural; only the tightening is Assay policy, and only the tightening
+  withholds credit.
+- **`assayNonClaims`** (`payload-non-claims-incomplete`). The prefix paragraph below already rules on
+  this member by name: it "is only producer vocabulary and does not weaken required AEE checks". A
+  rule that lets an incomplete list void the record does the converse, letting producer vocabulary
+  strengthen a required check into a rejection every consumer must honour. The checker's own comment
+  asks that a subset be "a rejection rather than a warning", and it still is — not-credited is a
+  rejection carrying a reason code, not a warning.
+- **`assaySealedAt`** (`seal-instant-invalid`). The one with a real entanglement, and so the one to
+  check rather than assume. The parsed instant is the input to `key-outside-validity-window`, and an
+  unparsable value makes that check skip its record — so the structural finding was load-bearing for
+  a bad instant not bypassing key expiry. It survives the move because the finding it raises is
+  itself a not-credited finding: a seal whose instant will not parse cannot reach `credited` by any
+  path, whether or not the window check ran. The window is not weakened. The record merely stops
+  being called unreadable *by every consumer* on the strength of an Assay member.
+
+Two things deliberately unchanged. The six stay in `REQUIRED_SEAL_FIELDS`, because that list states
+what the Assay producer contract obliges a seal to carry, and the test over it asserts that absence
+is not credited rather than that absence is malformed — so "required" reads as required-for-credit
+for a producer member, which is the strongest thing the prefix rule permits it to mean. And
+`assayObservedLabels`, `assayDropProofBasis` and `assayDropChannels` needed no decision at all: no
+rule reads them, and nothing can demote what nothing consults.
+
+Two checks were added with the move, because the existing ones could not see it.
+`--producer-vocabulary-test` now also mutates a member to a *legal* alternate value, which is the
+only mutation that reaches the selector case above — absent, ineligible and ill-typed are all
+values `substrate-runner` is not. And it reads `validate` back, requiring every `assay`-prefixed member the
+function consults to be one this file has decided, so the next producer member to acquire a rule
+cannot acquire a phase by default. The generic ineligible value is now `""` rather than a borrowed
+one, since `assaySourceSchema` accepts any non-empty string and would have passed a plausible
+wrong value straight through.
+
+Where this leaves the prefix rule: **seven** of the ten producer members are read by some rule in the
+checker, all seven are enforced inert by test, and the remaining three are inert by having no rule.
+Every rule that can still return `malformed` reads an `aee*` member, the envelope, or the predicate
+structure.
 
 Fields beginning with `assay` in the sealed payload are Assay producer vocabulary. AEE consumers may ignore them unless their own policy understands them. They MUST NOT alter AEE structural validity. Any future AEE statement exporter MUST also carry predicate-level `doesNotAssert` for statement-level non-claims; `assayNonClaims` inside the sealed payload is only producer vocabulary and does not weaken required AEE checks.
 
--- a/docs/architecture/ADR-045-aee-substrate-signed-run-end-seal.md
+++ b/docs/architecture/ADR-045-aee-substrate-signed-run-end-seal.md
@@ -273,14 +273,32 @@ and parked six sibling producer members — `assaySourceSchema`, `assaySealScope
 contract questions rather than one edit. They are decided here, and all six move.
 
 The reason they were parked does not survive being written out. It was that the normative paragraph
-in in-toto/attestation#570 governs members whose values a reader might rank, and these carry
-identities, instants and paths instead. That is true of the upstream paragraph and beside the point
-for us: the prefix rule below is not scoped to rankable members. It says *fields beginning with
-`assay`*, all of them, and has said so since this ADR was written. So six local decisions were being
-held against an upstream sentence that could only ever be narrower than the rule we already had. The
-question left open at in-toto/attestation#570 — whether a producer member is inert whether or not
-its values rank — is still unanswered upstream as of this decision, and is not load-bearing here. An
-answer would confirm the local rule, not extend it.
+in in-toto/attestation#570 governed members whose values a reader might rank, and these carry
+identities, instants and paths instead. That was true of the upstream paragraph as it stood at
+`35ca6f899`, and beside the point either way: the prefix rule below is not scoped to rankable
+members. It says *fields beginning with `assay`*, all of them, and has said so since this ADR was
+written. So six local decisions were being held against an upstream sentence that could only ever be
+narrower than the rule we already had.
+
+That sentence is no longer narrower, and the tense above is load-bearing. The question we left open
+at in-toto/attestation#570 — whether a producer member is inert whether or not its values rank — was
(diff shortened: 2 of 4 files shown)


Rul1an:
#2107 is merged, so this should retarget to `main` on its own. One factual note on the replacement comment before it lands.

The new text says the scoping argument was that the upstream paragraph reaches members whose values rank, and calls that "true of the upstream paragraph". That stopped being true at 2026-08-07T12:25:38Z. Commit `237f83b9` on in-toto/attestation#570 generalised it, and the current spec text reads:

> a producer-defined member MUST NOT affect structural validity, MUST NOT affect `result`, and MUST NOT affect the evidence tier, whether or not its values can be ordered.

(spec at `237f83b9`, lines 1584-1586, restated in the changelog at 2176-2178.)

The conclusion is unaffected: all six belong in the tuple, and the argument here does not depend on the upstream scope since it rests on the local prefix rule. But the clause is now a false statement about upstream, and this comment is the place someone will read to find out why the change was made.

Worth knowing that the same staleness was in the text this replaces, for the same reason: both were written against the 09:02 version of the paragraph, and the 12:25 version generalises it specifically so that unordered members are covered. The per-channel loss counters we carry are exactly such a member, and his commit message names them as the gap that prompted the generalisation.

Suggested rewording of the clause: the upstream paragraph was once scoped to members whose values rank, and was generalised the same afternoon to cover unordered members too, so both it and the local prefix rule now reach all six.

Rul1an:
Confirmed at the primary source rather than taken on the report, and it holds. `237f83b9f` at 12:25:38Z, spec lines 1584-1586, changelog 2176-2178. Fixed in e78c4f5, in the ADR and in the checker comment, and the PR body above carried the same clause and now doesn't.

Two things beyond the reword, since checking it turned up more than a tense error.

The body also said the question was *still unanswered upstream*, which was true at 12:09Z and false sixteen minutes later. That is the worse of the two: a decision recorded as unconfirmed when it is confirmed reads as weaker evidence than it is. The ADR now records that both rules require this rather than that one of them permits it.

And the upstream commit message names the gap as ours without being asked to — an implementer who "also emit[s] an unordered producer member, per-channel loss counters", and separately "a checker that gated AEE validity on a producer member's value, which their own design document forbade in as many words". That is `assayDropChannels` and #2107's defect. I checked the direction rather than assuming it: our comment 12:09:24Z, the commit 12:25:38Z.

One thing I did not fold in. The generalised paragraph opens with a conforming verifier reading *nothing* in producer territory, while the three obligations it states are structural validity, `result`, and the evidence tier. This checker still reads producer members to withhold credit. I take that as outside the three — it is a consumer trust decision of the same class as an untrusted signing key, which ADR-045's key-scope section already treats as the consumer's own, and AEE's `result` is a field this checker does not emit. But that is a reading and not a quotation, so it is recorded in the ADR as one, to be re-checked if the opening sentence acquires normative weight of its own.

Rul1an:
## Review findings — one blocker, verified at `e78c4f502`

An independent non-building review ran against `b3d62a1bd` and returned APPROVE WITH FINDINGS. The head has since moved to `e78c4f502`, so that record does not carry under the quorum rule. I re-measured the two substantive findings myself at the current head rather than relaying them.

### Finding 1 — CONFIRMED at `e78c4f502`, and it blocks

`seal-scope-mismatch` is the one moved rule with no test coverage. Reverting line 793 from `PHASE_NOT_CREDITED` to `PHASE_MALFORMED` — precisely the regression this PR exists to prevent — leaves **all four modes green**:

```
--meta-test                 PASS   <-- regression undetected
--required-field-test       PASS   <-- regression undetected
--rule-coverage-test        PASS   <-- regression undetected
--producer-vocabulary-test  PASS   <-- regression undetected
```

Cause: all three generic mutations for `assaySealScope` (absent, `""`, `7`) land on the `if` at :789 and raise `seal-scope-missing`, so the `elif` at :790-793 is never reached. This is exactly the gap `PRODUCER_CREDIT_ALTERNATE_VALUES` was introduced to close, and `assaySealScope` is not in that table. It is also the member this PR argues hardest about, via #2014.

The one-line fix works, and I verified it in **both** directions rather than only that it passes:

```python
PRODUCER_CREDIT_ALTERNATE_VALUES: dict[str, tuple[str, ...]] = {
    "assayAttackRowAttributionSource": ("substrate-runner",),
    "assaySealScope": ("filesystem_write_all",),
}
```

- clean tree → `all 7 producer credit field(s) are inert to structural validity`
- regression reintroduced → `1 producer-vocabulary check(s) failed`

`SEAL_SCOPE` is `tcp_connect_landlock_port`, so any other non-empty string mismatches; the value above is arbitrary and legal.

### Finding 2 — CONFIRMED, but narrower than reported

The read-back at :1005 matches only double-quoted literals:

```python
consulted = {name for name in re.findall(r'"(assay[A-Z][A-Za-z]*)"', body)} - set(PRODUCER_CREDIT_FIELDS)
```

I removed `assaySourceSchema` from the tuple, re-malformed its rule, and switched its reference to single quotes. `--producer-vocabulary-test` **passed** — the read-back was defeated. The reviewer reported all four modes green in its construction; in mine `--required-field-test` still caught the regression, so the exposure is smaller than that report suggests. The check is still weaker than the claim: it enforces one spelling of one reference style inside one function, and `inspect.getsource(validate)` does not cover callees.

No formatter runs on this file (`.pre-commit-config.yaml`), so quote style is not normalised and this is a plausible accident rather than only an adversarial move.

Either harden it (AST walk over `validate` plus callees) or soften the wording in the ADR and PR body, which currently present this as making the property checked rather than commented.

### Finding 3 — relayed, not independently verified

Producer members reach `malformed` through interception records, which `--producer-vocabulary-test` never mutates (it only touches `observationRecords[2]`). The reviewer judged this not an ADR-045 breach, since `observed_set` hashes payload bytes opaquely and never interprets the member, so a consumer ignoring Assay vocabulary recomputes the identical digest. That reasoning looks right to me but I did not re-derive it. The point stands that the absolute phrasing — every rule that can return `malformed` reads an `aee*` member or the envelope — is true only under *reads* = *interprets*, and one clause saying so would save the next reader filing a bug.

### Process

The PR body records `Final head SHA: f908b3e0c…`, which is now two pushes stale. Every measurement in its Test Evidence section is attributed to a superseded head, and the quorum item about delegated proofs targeting the exact head cannot be ticked against it.

CI is green at `e78c4f502` (50 checks, 0 failing) and `mergeStateStatus` is `CLEAN`, so Finding 1 is the only thing I would hold merge on.

I am not counting this as a quorum review: I rebased and force-pushed this branch onto `main` after #2107 squash-merged, which is close enough to building that self-certifying would be wrong.

Rul1an:
## One push, then freeze

Both commissioned reviews are in. Rather than three separate reports, here is everything that still needs changing at `e78c4f502`, verified by me at that head, so it can go in a single commit. **After that push, please stop editing the branch** — every fix so far has voided the reviews measured on the previous head, and we have now done that three times.

### 1. `seal-scope-mismatch` has no test coverage — the only merge blocker

Reverting `scripts/experiments/aee_landlock_seal_fixture.py:793` from `PHASE_NOT_CREDITED` to `PHASE_MALFORMED`, which is exactly the regression this PR exists to prevent, leaves all four modes green. All three generic mutations for `assaySealScope` (absent, `""`, `7`) land on the `if` at :789 and raise `seal-scope-missing`, so the `elif` is never reached.

```python
PRODUCER_CREDIT_ALTERNATE_VALUES: dict[str, tuple[str, ...]] = {
    "assayAttackRowAttributionSource": ("substrate-runner",),
    "assaySealScope": ("filesystem_write_all",),
}
```

Verified in both directions at `e78c4f502`: clean tree gives `all 7 producer credit field(s) are inert`, and with the regression reintroduced it gives `1 producer-vocabulary check(s) failed`. `SEAL_SCOPE` is `tcp_connect_landlock_port`, so the value is arbitrary and legal.

### 2. "all of it used to be filed above" is false for four of the seven

`docs/experiments/aee-landlock-seal-fixtures-2026-08.md:177-178`. Measured against this document's history: `assaySealScope` and `assayNonClaims` have never appeared in it at all, and the `assayCollectionPath` and `assaySealedAt` bullets are new here too. They were structural *in the checker*; they were not filed above. The phrase carries a document-position meaning inherited from #2107, where it was literally true.

### 3. "four slices" is three under the qualifier the sentence gives

`.pre-commit-config.yaml:165-166` and the matching sentence in the fixture. The three modes arrived separately: `run_meta_test` in #2008 (`1619cad9d`), `run_required_field_test` in #2067 (`2639d12a3`), `run_rule_coverage_test` in #2072 (`53b41efe5`). Under the shared qualifier "while all three stayed green", the slices are #2072, #2095, #2107, so three. Four is reachable only on the different basis of all six rules existing from #2067, which is the basis the neighbouring "two" does not use. Pick one basis or name it.

### Already fixed, no action

The upstream-scoping clause and the "still unanswered upstream" sentence were both repaired in `e78c4f502` while the reviews were running, so those two blocking findings are stale. The PR body's stale `f908b3e0c` head SHA and its unmerged-stack language are also already gone.

### One review claim that does not hold

One review reported that `validate` already contains a single-quoted producer reference (`payloads[seals[0]].get('assaySealedAt')`), making the read-back gap live rather than hypothetical. I grepped the current head for `'assay[A-Z]` and there are zero matches. The gap in the guard is real — I defeated it myself by switching a reference to single quotes — but it is not already exercised in shipped code, so this is a choice rather than a repair: either widen the regex to `["']`, or narrow the prose in ADR-045, the experiments doc and `.pre-commit-config.yaml` to say it covers double-quoted literals inside `validate`. Both are acceptable; the current wording claims more than the code does.

### After the push

Freeze the branch. Both reviews will re-measure on the resulting SHA, and route 2 of the quorum closes on two non-building reviews at that head plus the bots being silent past thirty minutes. Copilot has posted nothing on `e78c4f502`; its earlier quota record sits on a superseded head and does not carry.

CI is green at `e78c4f502` (50 checks, 0 failing) and `mergeStateStatus` is `CLEAN`, so item 1 is the only thing holding merge.

PR #2119, opened 2026-08-07
The half we ship, written down so the half we do not can be pinned to it

Rul1an:
`RunnerClaimGate` answers *given how healthy this run's observation was, which claim kinds may I make*. A different rule has to answer *given this observer class and its declared probe set, which claim kinds can it support at all*. Two halves of one rule. Only the first is implemented anywhere, and it was not written down in a form anything outside this crate could check against — which is how you end up with an approximation instead of a call.

This adds `assay.runner.claim_support_parity.v0`: the health half as a table, verdict × claim kind, public and serializable. The class half will live in another language, so a parity test is the only fallback available across that boundary — the same reason `assay-evidence/tests/claim_gate_parity.rs` exists between two Rust vocabularies.

**The table derives every decision from `RunnerClaimGate`; it never restates one.** `table_derives_from_the_gate_rather_than_restating_it` fails if someone later inlines the fifteen cells as literals, which is the second implementation this module exists to prevent.

## Two things that fell out of writing it down

**`ExhaustiveSet` has no field on the fidelity gate.** `CoverageClaimKind` carries three kinds and `RunnerClaimGate` has fields for two: `measured_positive_claims` and `bounded_negative_claims`. The table reports the third as `NotModelled` rather than folding it onto a neighbour.

That is deliberate rather than lazy. Folding it onto `measured_positive_claims` would read an exhaustive-set claim as merely *degraded* under `Clipped`, when dropped ring-buffer records are precisely the thing that falsifies an exhaustive set. `CoverageDescriptor::claim_decision` does model the kind — `Allowed` on a full descriptor, `Degraded` on a partial one — so a caller asking about exhaustiveness has to ask that gate. The row names it, and a test asserts the row names it.

**Absence has never been more permissive than occurrence, and nothing checked it.** It holds by construction today at every verdict. A gate that let an absence claim through where it degraded an occurrence claim would have inverted the rule silently. There is now an assertion, plus a second test requiring the gap to be *strict* somewhere, so a gate that flattened the two kinds could not satisfy the first one vacuously.

## Scope

Additive. No existing type or decision changes; `RunnerClaimGate`, `CoverageDescriptor` and the fidelity classifier are untouched. The gating map is regenerated and the diff adds two lines and removes none, which is the direction that check cares about.

## Verification

- `cargo test -p assay-runner-schema` — 15 passed (9 new, 6 existing), on top of `e89bc1fc9`
- `cargo clippy -p assay-runner-schema --all-targets -- -D warnings` — clean
- Mutation-bitten both ways, and both bite:
  - `Clipped`'s `bounded_negative_claims` set to `Allowed` → 3 tests fail, including both asymmetry tests
  - `ExhaustiveSet` folded onto `measured_positive_claims` → `exhaustive_set_is_reported_unmodelled_for_every_verdict` fails

## Non-claims

The table declares four, and a test asserts they travel with it: it types no observer class, returns no blinding-cost verdict, computes no aggregate claim score, and does not replace the coverage descriptor gate. It is one half of a rule and says so.

---

Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/31215648848
- sha: ed5ea02e209a5601973ce1f7d065d0595842a750


Rul1an:
Recording what I checked on the two Bencher alerts rather than letting them pass unremarked.

**Not caused by this change, and the reason is structural.** Both alerting benchmarks live in `crates/assay-core/benches/store_write_heavy.rs`. `assay-core` does not depend on `assay-runner-schema` — the only crate this PR touches — so the changed code is not in that benchmark's compilation unit at all. There is no path by which an additive data module in a leaf crate reaches a SQLite write-heavy bench in a crate that cannot see it.

**These two alert chronically on this testbed.** #2114, an unrelated change, alerted on the same two benchmarks yesterday at +166% (`sw/12xlarge`) and +352% (`sw/50x400b`). This run reads +1,782% and +1,960% against baselines of 48 ms and 152 ms. Absolute numbers that small on a shared `ubuntu-latest` runner move by more than the threshold on scheduling noise alone, and a bigger breach here is consistent with a noisier instance rather than a slower binary.

`Criterion benches (store + suite)` passed, so the gate agrees.

**Not proposing a fix in this PR.** The signal worth acting on is that a threshold which fires on unrelated changes is not measuring what it is meant to measure — that belongs in its own change against the perf workflow, not smuggled into a PR about a claim-support table.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #2186, opened 2026-08-09
docs(agent): make Cargo.toml the version source

### LINKED ISSUE #2179: CLAUDE.md states the workspace version twice and the two disagree
Found while measuring the CLI surface at head `55e8103029583a16af96e1eafe73025192d69cd2` (`origin/main`).

`CLAUDE.md` states the workspace version twice and the two disagree:

```
Cargo.toml       version = "5.0.0"
CLAUDE.md    :5  Rust monorepo with workspace version `5.0.0` (21 crates).      correct
CLAUDE.md  :316  - Workspace version in root `Cargo.toml` (`version = "3.31.1"`)  stale
```

The header is right, so this reads as fixed at a glance. The stale copy sits under **Conventions**, which is the heading a reader consults precisely when they want the binding rule rather than the summary.

Worth a line rather than a shrug because of who reads this file. `CLAUDE.md` is loaded as authoritative context at the start of every agent session and is not cross-examined the way a stale README would be. A file that contradicts itself is a worse failure mode than one that is merely old: whichever line an agent reaches first wins, and nothing signals that a second answer exists.

## What would close this

Delete the version literal from the Conventions line rather than updating it. `Cargo.toml` is the source; restating the number in prose guarantees this recurs at the next bump. The line still carries its rule without it:

```
- Workspace version lives in root `Cargo.toml`
```

A grep gate for a hardcoded semver in `CLAUDE.md` would make the recurrence impossible, though that may be more machinery than one line deserves.

Rul1an:
## Summary

- remove the stale workspace version literal from `CLAUDE.md` conventions
- keep root `Cargo.toml` as the named version source
- avoid making the next workspace bump synchronize a second conventions literal

Closes #2179

## Scope

This changes one documentation line only. It does not bump the workspace version or alter runtime, schema, policy, evidence, MCP, or release behavior.

## Verification

Measured on exact head `8af9e1a646b2c735eefa3a15db7fec8c5218604c` in `/Users/roelschuurkes/.config/superpowers/worktrees/assay-2179-version-source`:

```text
rg check: no hardcoded Conventions semver remains
pre-commit run --files CLAUDE.md: passed
git diff --check: passed
```

## Risk

Low. The header still reports the current workspace version; the binding convention now points to the canonical `Cargo.toml` value instead of restating a stale literal.

## Summary by CodeRabbit

* **Documentation**
  * Updated the workspace version guidance to clarify that the version is defined in the root project configuration.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -313,7 +313,7 @@ See `docs/PERFORMANCE-ASSESSMENT.md` for full documentation, including measureme
 
 ## Conventions
 
-- Workspace version in root `Cargo.toml` (`version = "3.31.1"`)
+- Workspace version lives in root `Cargo.toml`
 - Internal crate deps use `workspace = true` with path + version
 - `#[deny(unsafe_code)]` on all crates except assay-ebpf
 - Error handling: `anyhow` for applications, `thiserror` for libraries


Rul1an (commented):
Read-only review bound to exact head `8af9e1a646b2c735eefa3a15db7fec8c5218604c`. One non-building agent review under route 2, with Copilot quota-limited on this head. No merge, no edits.

**No blockers.**

## What I verified

The diff is one line in `CLAUDE.md` and touches nothing else. It removes the stale literal rather than updating it, which is the correct shape: a number that is not restated cannot drift.

Measured at this head:

- The removed line carried `3.31.1`. The workspace is `5.0.0` (`Cargo.toml`), so the deleted copy was two majors behind.
- The only remaining version-shaped literals in `CLAUDE.md` are line 13, `workspace version 5.0.0`, which is **correct today**, and line 187, `SARIF 2.1.0`, which is an external format version and not ours.
- After this change the workspace version is stated once, which is exactly the defect #2179 opened on.

No behavioural surface is touched: no code, no generated artifact, no spec, no test. Public wording is unchanged apart from the deleted clause, and the replacement sentence makes no claim beyond naming where the value lives.

## One observation, non-blocking

The drift class is halved rather than closed. Line 13 still hard-codes `5.0.0` by hand and will go stale at the next release, in the same way the deleted line did. That is outside this PR's stated scope and #2179 is closed by what is here — worth recording only so nobody reads the title as meaning no hand-maintained copy remains.

If it is ever worth closing properly, this repository already has the mechanism: `docs/generated/` plus `scripts/ci/check-docs-generated-drift.sh` turn exactly this class of restatement into a gate.

Rul1an:
Independent non-building review record (route 2)

Reviewer: Franklin (Codex subagent; no write access to this branch)
Reviewed head: `8af9e1a646b2c735eefa3a15db7fec8c5218604c`
Base measured: `48beb09d1ba581f275da7775228262c778914a28`
Verdict: PASS, no findings.

Verified the one-line diff, root `[workspace.package].version` as the canonical source, release tooling reading that section, `git diff --check`, unchanged file mode, and no runtime/workflow/schema/policy/security/generated-artifact impact. Copilot declared a quota limit on this same head; Claude Code Desktop supplied the other exact-head non-building review.

Disposition of Claude’s nonblocking residual: the remaining `5.0.0` summary at `CLAUDE.md:13` is intentionally retained by #2179 and currently matches `Cargo.toml`; eliminating or generating that summary is a separate documentation-drift change and is not required for this one-line source-instruction correction.

Issue #2152, opened 2026-08-08
Productization: make the MCP tools installable where agents already run


Rul1an:
Parent: #1973 · Relates to #1975, #1979, #1840

## Problem

`assay-mcp-server` already exposes five production tools, and their descriptions are written for an agent rather than for a human reading docs. `assay_check_args` states when to reach for it and, unusually, what it does not establish: *"It never invokes the target tool and never proves the provider executed anything."*

**None of the three surfaces an agent runs in can install them**, because the repository ships no manifest. There is no `.mcp.json` at the root, no `.claude-plugin/`, and `.cursor/` carries only `BUGBOT.md` and settings.

So the work here is not building a capability. It is publishing the one that exists.

## What the surfaces need, measured

| Surface | Config | Distribution |
|---|---|---|
| Claude Code | `.mcp.json` at repo root, key `mcpServers` | `.claude-plugin/plugin.json` + a `marketplace.json`, installed with `/plugin marketplace add` then `/plugin install <name>@<marketplace>` |
| Cursor | `~/.cursor/mcp.json` global, `.cursor/mcp.json` per project | MCP directory; Team Marketplaces since 3.10 (June 2026) distribute across cloud agents, IDE and CLI |
| Codex | `~/.codex/config.toml` or project `.codex/config.toml`, `[mcp_servers.<name>]` with `command`/`args`/`env` | skills as `SKILL.md` |

Two facts make this cheap. `SKILL.md` is the same filename in Claude Code and Codex, and `AGENTS.md` is a cross-tool convention this repository already has.

## Steps

Each step is separately useful and separately revertible. Later steps assume the earlier ones only for content, not for mechanism.

**1. `.mcp.json` at the repo root.** One `mcpServers` entry invoking the separate `assay-mcp-server` binary over stdio with `--policy-root .`. Read by Claude Code and by Cursor project scope, so one file covers two surfaces. Smallest possible unit and the one that unblocks a reader who has already cloned the repo.

**2. A documented Codex entry.** The `[mcp_servers.assay]` block for `config.toml`, in the docs rather than as a file, because Codex reads user or project config rather than a repo manifest. Belongs beside step 1 so the three surfaces are described in one place.

[...]

PR #2187, opened 2026-08-09
test(agent): bound golden-path subprocesses

### LINKED ISSUE #2173: test: harden golden-path execution proof and process bounds
Origin: PR #2172, reviewed at `494a33ccad2b78382817daa5ba196bf020e32d66`.

## Problem

The runtime coverage gate is materially stronger than the original source-text scan, but four residuals remain:

1. `expected_outcome()` records a `(step, outcome)` lookup before the command is executed. The current scenarios still fail if their assertions fail, but the register proves "looked up by a running scenario" rather than the stronger claim "command output was observed".
2. Each orchestrator filters by its binary. A future contract row with a third binary is expected by neither orchestrator and can evade exact coverage.
3. CLI subprocesses and the Python availability probe have no local wall-clock/output ceilings. CI has a job timeout, but a stuck child gives slow and weak diagnostics.
4. The historical Task 2 RED checkpoint in the implementation plan says the whole machine contract is absent, although Task 1 had already created it. It should name missing MCP-owned outcomes instead.

## Acceptance

- Move outcome registration to a post-execution observation site. A mutation that performs only `expected_outcome()` without launching/observing the command must fail exact coverage.
- Assert the contract binary set is exactly `assay` and `assay-mcp-server`. A third-binary mutation must fail.
- Provide one shared bounded test-process runner for the CLI paths and Python preflight: deadline, bounded stdout/stderr, kill and reap on timeout/overflow, command context in diagnostics.
- Keep generated-contract reads classified correctly: they are tracked, generator-owned repository artifacts, not hostile external evidence. Do not add a ceiling merely by analogy unless a concrete failure mode justifies it.
- Correct the Task 2 RED text to assert missing/incomplete MCP rows, not a missing file.
- Preserve standalone scenario tests plus orchestrator execution; thread-local accounting makes this correct and the separate tests improve failure localization.

## Verification

- Focused CLI and MCP contract suites.
- Mutations for lookup-without-execution, unknown binary, hanging child, and output flood.
- Full affected crates, fmt, clippy, and generated-doc drift gate.

## Non-goals

- No production CLI/MCP behavior changes.
- No Python version pin in this slice; the fixture dependency policy belongs to #1975.

Rul1an:
## Summary

- add one shared bounded subprocess runner for golden-path CLI, MCP, and Python preflight tests
- register contract outcomes only after observed process output/status
- pin the production contract binary set to exactly `assay` and `assay-mcp-server`
- run the timeout/flood self-tests from shared support so the MCP crate executes them on Windows too
- correct the historical Task 2 RED description

Closes #2173.

## Scope and risk

- Test infrastructure and implementation-plan text only.
- No production CLI or MCP behavior changes.
- No public API, CLI flag, JSON schema, policy, or evidence-contract changes.
- The runner bounds direct child wall time plus captured stdout/stderr; it does not claim process-tree containment.

## Review packet

**Base:** `origin/main` at `f2ea96166ddb0ea760d6b31aac598286b53cc9c3`  
**Head:** `a45a51ed3ffe684937dcc4190c333e6cac5386c7`  
**Worktree:** `/Users/roelschuurkes/.config/superpowers/worktrees/assay-2173-process-bounds`  
**Toolchain:** `rustc 1.96.0`, `cargo 1.96.0`

### Contract changes

- `tests/support/agent_golden_path.rs` owns the single expected-outcome lookup and post-observation registration rule.
- `tests/support/bounded_process.rs` provides the shared deadline/output-bounded child runner and cross-platform self-tests.
- The real contract binary set is asserted as exactly `{assay, assay-mcp-server}`.
- Golden-path subprocess limits are 10 seconds and 1 MiB per output stream; diagnostic excerpts are bounded to 4096 bytes.

### TDD and mutation evidence

- RED: shared process API absent before implementation.
- Lookup-only mutation fails exact coverage with missing `("install-check", "success")`.
- Third-binary mutation fails with actual `{assay, assay-mcp-server, other}` versus the exact allowed set.
- Hanging-child test terminates and reaps the child with command context in the diagnostic.
- Stdout and asymmetric stderr flood tests terminate/reap the child and pin their independent ceilings.
- Mutating stderr to use the stdout ceiling fails the new stderr assertion.
- Raising the diagnostic excerpt ceiling fails the truncation assertion.
- Windows timeout uses the native indefinite `ping -t 127.0.0.1` form and is included through the non-excluded MCP integration crate.

### Verification

```text
cargo test -p assay-cli --test agent_golden_path_contract --locked
# 19 passed

cargo test -p assay-mcp-server --test agent_golden_path_contract --locked
# 13 passed

cargo test -p assay-cli --locked
cargo test -p assay-mcp-server --locked
cargo fmt --all -- --check
cargo clippy -p assay-cli -p assay-mcp-server --all-targets --all-features --locked -- -D warnings
python3 scripts/ci/test-agent-golden-path-skill.py
sh scripts/ci/check-docs-generated-drift.sh
cargo test -p assay-sim --locked
cargo test -p assay-core agentic::tests --locked
cargo test -p assay-core judge_internal::tests::contract --locked
git diff --check
pre-commit run --all-files
pre-commit run --hook-stage pre-push --all-files
# all passed for the changed tree; push hooks passed on final head
```

The simulation and AI-oriented checks include the full `assay-sim` suite, the core agentic tests, and judge contract tests.

### Failure policy

- Timeout or output overflow is fail-closed for the test: kill, reap, and return a bounded diagnostic.
- A scenario is not credited merely for looking up its expected outcome; it must record an observed result.
- Unknown binaries fail exact contract coverage instead of being ignored by per-binary filters.

### Review disposition

The first exact-head review found that the prior inline Windows `cmd` label was not a reliable hang and that stderr/truncation mutation coverage was incomplete. Both are fixed on this head. The broader process-tree/inherited-pipe and early-EPIPE semantics remain outside the safe current callsites and are tracked in #2189 before plugin commands may reuse this runner.

### Non-claims

- No Python version pin.
- No process-tree kill guarantee.
- No new ceiling on generated, tracked contract artifacts.

### DIFF
--- a/crates/assay-cli/tests/agent_golden_path_contract.rs
+++ b/crates/assay-cli/tests/agent_golden_path_contract.rs
@@ -1,5 +1,7 @@
 //! The #1975 journey as observed by a caller that reads only stdout and the exit status.
 
+#[path = "../../../tests/support/bounded_process.rs"]
+mod bounded_process;
 #[path = "../../../tests/support/agent_golden_path.rs"]
 mod runtime_coverage;
 
@@ -7,7 +9,9 @@ use serde_json::Value;
 use std::ffi::OsStr;
 use std::path::Path;
 use std::process::{Command, Output};
+use std::time::Duration;
 
+use bounded_process::{run_bounded, ProcessLimits, GOLDEN_PATH_LIMITS};
 use runtime_coverage::ExpectedOutcome;
 
 fn workspace_root() -> &'static Path {
@@ -136,12 +140,14 @@ fn assay<S: AsRef<OsStr>>(cwd: &Path, args: &[S]) -> Output {
             command.env_remove(name);
         }
     }
-    command
-        .current_dir(cwd)
-        .env("NO_COLOR", "1")
-        .args(args)
-        .output()
-        .expect("run assay binary")
+    command.current_dir(cwd).env("NO_COLOR", "1").args(args);
+    run_bounded(
+        &mut command,
+        b"",
+        GOLDEN_PATH_LIMITS,
+        "agent golden-path CLI command",
+    )
+    .unwrap_or_else(|error| panic!("{error}"))
 }
 
 fn assay_contract(cwd: &Path, expected: &Value, replacements: &[(&str, &str)]) -> Output {
@@ -407,4 +413,57 @@ fn every_cli_contract_outcome_is_executed_once() {
             offline_profile_verifier_keeps_both_outcomes_on_stdout,
         ],
     );
+}
+
+#[cfg(unix)]
+fn hanging_command() -> Command {
+    let mut command = Command::new("sh");
+    command.args(["-c", "while :; do :; done"]);
+    command
+}
+
+#[cfg(windows)]
+fn hanging_command() -> Command {
+    let mut command = Command::new("cmd");
+    command.args([
+        "/V:ON",
+        "/C",
+        "@echo off & :loop & set /a x+=1 >NUL & goto loop",
+    ]);
+    command
+}
+
+#[cfg(unix)]
+fn stdout_flood_command() -> Command {
+    let mut command = Command::new("sh");
+    command.args(["-c", "while :; do printf '0123456789abcdef'; done"]);
+    command
+}
+
+#[cfg(windows)]
+fn stdout_flood_command() -> Command {
+    let mut command = Command::new("cmd");
+    command.args(["/C", "for /L %i in (1,1,1000000) do @echo 0123456789abcdef"]);
+    command
+}
+
+#[test]
+fn bounded_runner_kills_timeout_and_reports_context() {
+    let mut command = hanging_command();
+    let limits = ProcessLimits::new(Duration::from_millis(100), 1024, 1024);
+    let error = run_bounded(&mut command, b"", limits, "hanging mutation")
+        .expect_err("hanging child must time out");
+    assert!(error.contains("hanging mutation"));
+    assert!(error.contains("deadline"));
+}
+
+#[test]
+fn bounded_runner_kills_output_flood() {
+    let mut command = stdout_flood_command();
+    let limits = ProcessLimits::new(Duration::from_secs(2), 1024, 1024);
+    let error = run_bounded(&mut command, b"", limits, "flood mutation")
+        .expect_err("output flood must exceed its ceiling");
+    assert!(error.contains("flood mutation"));
+    assert!(error.contains("stdout"));
+    assert!(error.contains("1024"));
 }
--- a/crates/assay-cli/tests/agent_golden_path_contract.rs
+++ b/crates/assay-cli/tests/agent_golden_path_contract.rs
@@ -9,9 +9,8 @@ use serde_json::Value;
 use std::ffi::OsStr;
 use std::path::Path;
 use std::process::{Command, Output};
-use std::time::Duration;
 
-use bounded_process::{run_bounded, ProcessLimits, GOLDEN_PATH_LIMITS};
+use bounded_process::{run_bounded, GOLDEN_PATH_LIMITS};
 use runtime_coverage::ExpectedOutcome;
 
 fn workspace_root() -> &'static Path {
@@ -413,57 +412,4 @@ fn every_cli_contract_outcome_is_executed_once() {
             offline_profile_verifier_keeps_both_outcomes_on_stdout,
         ],
     );
-}
-
-#[cfg(unix)]
-fn hanging_command() -> Command {
-    let mut command = Command::new("sh");
-    command.args(["-c", "while :; do :; done"]);
-    command
-}
-
-#[cfg(windows)]
-fn hanging_command() -> Command {
-    let mut command = Command::new("cmd");
-    command.args([
-        "/V:ON",
-        "/C",
-        "@echo off & :loop & set /a x+=1 >NUL & goto loop",
-    ]);
-    command
-}
-
-#[cfg(unix)]
-fn stdout_flood_command() -> Command {
-    let mut command = Command::new("sh");
-    command.args(["-c", "while :; do printf '0123456789abcdef'; done"]);
-    command
-}
-
-#[cfg(windows)]
-fn stdout_flood_command() -> Command {
-    let mut command = Command::new("cmd");
-    command.args(["/C", "for /L %i in (1,1,1000000) do @echo 0123456789abcdef"]);
-    command
-}
-
-#[test]
-fn bounded_runner_kills_timeout_and_reports_context() {
-    let mut command = hanging_command();
-    let limits = ProcessLimits::new(Duration::from_millis(100), 1024, 1024);
-    let error = run_bounded(&mut command, b"", limits, "hanging mutation")
-        .expect_err("hanging child must time out");
-    assert!(error.contains("hanging mutation"));
-    assert!(error.contains("deadline"));
-}
-
-#[test]
-fn bounded_runner_kills_output_flood() {
-    let mut command = stdout_flood_command();
-    let limits = ProcessLimits::new(Duration::from_secs(2), 1024, 1024);
-    let error = run_bounded(&mut command, b"", limits, "flood mutation")
-        .expect_err("output flood must exceed its ceiling");
-    assert!(error.contains("flood mutation"));
-    assert!(error.contains("stdout"));
-    assert!(error.contains("1024"));
 }
--- a/crates/assay-cli/tests/agent_golden_path_contract.rs
+++ b/crates/assay-cli/tests/agent_golden_path_contract.rs
@@ -8,6 +8,8 @@ use std::ffi::OsStr;
 use std::path::Path;
 use std::process::{Command, Output};
 
+use runtime_coverage::ExpectedOutcome;
+
 fn workspace_root() -> &'static Path {
     let root = Path::new(env!("CARGO_MANIFEST_DIR"))
         .parent()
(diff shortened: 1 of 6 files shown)


Rul1an (commented):
Read-only review bound to exact head `50a42cd87ab144010a5228cfaa8e2de781513b91`. Test infrastructure only; the six changed files do not include `CLAUDE.md`, so main's advance to `f2ea9616` via #2186 does not interact. No edits, no merge.

**No blockers.**

## Correctness and deadlocks

The three-thread structure in `tests/support/bounded_process.rs` is the correct one and each leg is load-bearing:

- stdin is written on its own thread and the handle is dropped there, so the child gets EOF and the main thread never blocks on a full stdin pipe;
- stdout and stderr each get a draining reader, so a child cannot wedge itself on a full output pipe while the main thread polls;
- the main thread polls `try_wait` on a 10 ms interval against a deadline.

The overflow path is the interesting one. `spawn_reader` takes `limit + 1` and then stops, which on its own would let the child block on a full pipe forever. It is rescued because the reader signals overflow *before* returning and the poll loop checks that channel at the top of each iteration, so the kill happens within one poll interval. Boundary handling is right: exactly-at-limit passes, `limit + 1` fails.

Kill and reap is correct on both exit paths, and `kill_and_reap` preserves the kill error into the wait diagnostic rather than discarding it.

## The grandchild question, and why it does not bite here

`run_bounded` forces `Stdio::piped()` on stderr. A command that spawns a child with inherited stderr would hand that pipe to a grandchild, and a surviving grandchild would leave `join_reader` blocked after the child is already reaped — defeating the timeout it just enforced.

That does not happen at this head, and not by luck: `proxy-enforce`, the one path whose upstream is spawned with `.stderr(Stdio::inherit())` (`crates/assay-mcp-server/src/proxy/mod.rs:88`), deliberately bypasses this runner and keeps `Conn::attach`, which bounds every read with `recv_timeout` and carries its own reap timeout. Two mechanisms for two interaction shapes — one-shot write-and-drain versus interactive JSON-RPC where you cannot read to EOF — rather than one runner stretched over both. That is the right call and the module docs on `jsonrpc_conn` already explain the reasoning.

## Accounting

Both #2173 residuals are delivered with self-tests rather than assertions about themselves:

- `record_observation(&ExpectedOutcome)` moves registration to the observation site, and `lookup_only_scenario` proves a scenario that only looks up fails `assert_exact`.
- `assert_contract_binaries` pins the binary set, with a mutation test proving a changed set fails.

## Diagnostics

Bounded at 4096 bytes with the true total reported, so a flood cannot fill CI logs. Worth naming a property that may be incidental: `excerpt` renders through `{:?}` on the lossy string, which escapes control characters, so a child cannot inject terminal sequences into CI output. That matches the sanitisation posture the repository applies to rendered evidence.

## Non-blocking advice

**1. Write down the constraint that currently holds by call-site choice.** `run_bounded` must not be used for a command that spawns children with inherited stdout/stderr, for the reason above. Today that is satisfied because `proxy-enforce` uses `Conn`. A future caller routing it through the shared runner would silently lose the bound, and nothing in the file says so. One doc comment.

**2. EPIPE on an early-exiting child is reported as a runner failure.** `stdin_result` is checked after status and both streams, so a child that exits before reading stdin surfaces as `write stdin: ...` rather than as the child's own outcome. Unreachable at this head — the only `run_bounded` caller passing non-empty stdin drives a command that does read it — but the golden path explicitly contains early-exit rows, so it is a plausible future trigger.

**3. Process-group kill.** `child.kill()` signals the direct child only, so a timeout leaves any grandchild orphaned. A leak on an already-failing path, not a hang, and consistent with the separation above.

## Non-claims, checked

Nothing here changes production CLI or MCP behaviour, no output schema, no exit code, and no evidence or policy surface. The 10 s ceiling sits well inside nextest's `slow-timeout` of 60 s with `terminate-after = 2`, so a stuck child now yields a named failure with command and cwd instead of a harness SIGTERM at 120 s — which is the stated point of the change.

Rul1an:
Review disposition for Claude Code Desktop’s three nonblocking observations on exact head `50a42cd87ab144010a5228cfaa8e2de781513b91`:

- Verified as non-triggering for the current one-shot CLI/MCP/Python callsites; `proxy-enforce` remains on the interactive bounded `Conn` path.
- No process-tree cleanup claim is made in this PR or its review packet.
- The inherited-pipe/grandchild, direct-child kill, API documentation, and early-EPIPE semantics are tracked together in #2189.
- #2152 plugin implementation will not route Claude/plugin install commands through this helper until that process shape is explicitly proven safe or a separate bounded strategy is used.

No current-head change is required; pushing a doc-only amendment here would invalidate exact-head reviews without changing the delivered #2173 behavior.

Rul1an:
@coderabbitai review

Rul1an:
@coderabbitai review final head `a45a51ed3ffe684937dcc4190c333e6cac5386c7`

Rul1an (commented):
Final read-only re-review bound to exact head `a45a51ed3ffe684937dcc4190c333e6cac5386c7`, against base `f2ea96166ddb0ea760d6b31aac598286b53cc9c3`. The prior head `50a42cd87` is superseded and does not count toward this verdict. No edits, no merge.

**No blockers.**

## Delta since the superseded head

Three files: `CLAUDE.md` (+1/−1, the main merge, no overlap with this PR's work), `crates/assay-cli/tests/agent_golden_path_contract.rs` (+1/−55), `tests/support/bounded_process.rs` (+89). The self-tests moved out of one test binary and into the shared module.

## The relocation is not tidying — it is the only way those tests reach Windows

I checked whether the stated rationale actually holds, because "moved so Windows executes them" is a CI-reachability claim rather than a code one.

`.github/workflows/ci.yml:665`, in the job whose matrix is `[ubuntu-latest, macos-latest, windows-latest]`, runs:

```
cargo test --locked --workspace --exclude assay-ebpf --exclude assay-it --exclude assay-monitor --exclude assay-cli
```

**`assay-cli` is excluded on non-Linux.** So while the self-tests lived in that crate's test file they could never run on Windows. `assay-mcp-server` is not excluded, and its test binary includes the module via `#[path = "../../../tests/support/bounded_process.rs"]`. Both test binaries include it, so the self-tests now run on every platform the cross-OS job covers.

The claim holds, and the mechanism is the exclusion list rather than anything about the tests themselves.

## Windows command validity

`ping -t 127.0.0.1` is a real system binary that pings indefinitely, so it exercises the deadline rather than an argument-parse failure — which is what the previous inline label did. With a 100 ms limit it times out long before its roughly one-line-per-second output approaches any ceiling.

The flood commands use `cmd /C for /L %i in (1,1,1000000) do @echo ...`. Single `%i` is correct for command-line context; the doubled `%%i` form would be wrong here and is the usual mistake. Both variants are valid.

## Stream-specific assertions

`stderr_flood_uses_its_own_ceiling_and_bounded_diagnostic` is the strongest test in the file, and the asymmetry is why. Ceilings of stdout 1024 / stderr 8192 mean a plumbing bug that applied `max_stdout_bytes` to both streams would report a `1024-byte ceiling` and fail this assertion. Symmetric limits would have let that bug pass.

`... (8193 bytes total)` pins the reader's `take(limit + 1)` boundary exactly, and `error.len() < 5000` catches a removed truncation. Together they cover both the detection and the bounding, which the earlier head asserted only for stdout.

## Process and deadlock behaviour

Unchanged from the superseded head and re-confirmed: three-thread structure, stdin dropped in its own thread, both output streams drained continuously, overflow signalled before the reader returns so the poll loop kills within one interval, and `kill_and_reap` preserving the kill error into the wait diagnostic. `proxy-enforce` still bypasses this runner and keeps `Conn`, so the grandchild-holds-the-pipe case does not arise.

## #2189 disposition

Correct, and better scoped than my advice was. It captures all three residuals — descendant survival, a descendant holding the pipe past kill, and EPIPE ordering hiding the child's own outcome — binds them to the originating SHA, and states the forward reason: the boundary starts to matter when #2152 step 4 drives commands that may spawn `git` or other descendants. Leaving them out of this PR is the right call.

## Non-blocking

**Platform-dependent headroom in the truncation assertion.** `error.len() < 5000` is a magic threshold whose margin is not the same everywhere. On Unix the flood is printable ASCII, so 4096 kept bytes render as ~4096 characters through `{:?}`. On Windows `echo` appends CRLF and `{:?}` renders each of those two bytes as two characters, giving roughly 11% expansion — about 4550 characters for the same 4096 bytes, plus context, command and suffix. Still under the bound, but with roughly half the Unix margin. Deriving the ceiling from `MAX_DIAGNOSTIC_BYTES` rather than a literal would make the assertion platform-neutral and remove a future Windows-only flake.

**Housekeeping:** #2189 currently carries no labels, while the rest of the open set now has size and cluster labels.

Rul1an:
Independent non-building final-head review record

Reviewer: Franklin (Codex subagent; read-only, no branch write)
Head: `a45a51ed3ffe684937dcc4190c333e6cac5386c7`
Base: `f2ea96166ddb0ea760d6b31aac598286b53cc9c3`
Verdict: PASS, no findings.

Verified that the shared self-tests are included by the Windows-tested MCP integration crate, Windows uses indefinite `ping -t 127.0.0.1`, the asymmetric stderr mutation pins the independent 8192-byte ceiling/8193-byte overflow capture/bounded diagnostic, and the current-main merge adds only `CLAUDE.md` outside the reviewed PR files. Focused final-head suites: CLI 19/19, MCP 13/13; `git diff --check` passed and the worktree remained clean. Copilot’s quota-limit record is on the superseded head and is not counted for this head.

PR #2190, opened 2026-08-09
test(process): bound subprocess trees and preserve early exits

### LINKED ISSUE #2189: test: define process-tree and early-EPIPE semantics for bounded runner
## Origin

Nonblocking exact-head review advice on PR #2187 (`50a42cd87ab144010a5228cfaa8e2de781513b91`). The current golden-path callsites are safe and #2173 remains mergeable, but the helper's boundary matters before step 4 of #2152 drives Claude/plugin commands that may spawn git or other descendants.

## Problem

`tests/support/bounded_process.rs` is a one-shot direct-child runner. Three limits are currently implicit:

1. `child.kill()` terminates only the direct child; descendants may survive.
2. A descendant inheriting stdout/stderr can keep the pipe open after the direct child is killed/reaped, leaving `join_reader()` blocked and defeating the wall-clock bound.
3. If a child exits before consuming non-empty stdin, the writer's `BrokenPipe`/EPIPE is currently returned as a runner failure after the child outcome was already observed, potentially hiding the more useful child result.

Current #2173 callsites do not trigger these cases. `proxy-enforce` deliberately keeps the interactive `Conn` path instead of using the one-shot runner.

## Acceptance

- Document the one-shot runner's supported process shape at its API boundary.
- Add a negative contract test where a descendant retains an inherited output handle; the test must itself remain bounded.
- Decide and implement a cross-platform process-tree policy for test subprocesses before using the helper for Claude/plugin install/update commands:
  - Unix process group/session termination;
  - Windows Job Object or an explicitly equivalent bounded strategy.
- Add a test for early child exit with non-empty stdin and define whether EPIPE is ignored when a child status/output is available or returned as a secondary diagnostic.
- Preserve bounded diagnostic excerpts and avoid environment dumps or secret-bearing argument expansion.
- Keep interactive JSON-RPC on its existing bounded `Conn` mechanism unless a shared abstraction can preserve incremental request/response semantics.

## Non-goals

- No production process supervisor.
- No claim that a direct-child timeout currently proves process-tree cleanup.
- Do not stretch the one-shot runner over interactive MCP traffic solely to reduce helper count.

Rul1an:
## Summary

- own each bounded one-shot test command through a Unix process group or Windows Job Object
- terminate descendants on timeout, output overflow, and normal direct-child completion
- preserve an observed child status when the stdin writer receives `BrokenPipe`
- add bounded cross-platform regressions for inherited-output descendants, quiet descendants, early stdin close, and platform-specific `EPERM`

Closes #2189.

## Scope / non-scope

Changed behavior is confined to the shared integration-test helper used by the CLI and MCP golden-path contracts. No production process supervisor, CLI schema, MCP protocol, or interactive JSON-RPC path changes.

Processes that deliberately escape the owned process group or Job Object remain outside the supported shape.

## Design

`process-wrap 9.1` is pinned once in workspace dependencies and enabled per target:

- Unix: `ProcessGroup::leader()`
- Windows: `JobObject`, which starts suspended and assigns the process before resume
- both dependencies are dev-only

The helper waits for child status, stdin completion, and bounded stdout/stderr readers under one deadline. Output overflow closes the reader pipe at `limit + 1` and remains under that deadline. Tree termination is followed by a separate one-second polling grace; no `Child::wait()` remains. `BrokenPipe` is ignored only at the final ordering point, after status and both output streams have been collected, so it cannot replace the child outcome.

## Review packet

**Base:** `main`

**Head measured:** `207da8ce958621c2f10d81c281387d51054f6772`

**Worktree:** `/Users/roelschuurkes/.config/superpowers/worktrees/assay-2189-process-tree`

**Hot file:** `tests/support/bounded_process.rs`

**Risk flags:**

- production behavior: no
- public output/schema: no
- test process isolation: yes
- cross-platform behavior: yes
- new dependency: dev-only, `cargo deny` accepted

## TDD and mutation proof

RED before implementation:

- inherited-output descendant: runner escaped its wall-clock bound after 2.01 seconds
- 8 MiB stdin with child exit 23: runner returned `Broken pipe (os error 32)` instead of status 23

Mutations killed by the final tests:

- removing `ProcessGroup::leader()` reproduces the bounded outer timeout
- rejecting `BrokenPipe` reproduces the status-23 masking failure
- replacing normal-completion tree cleanup with a no-op leaves the quiet descendant alive and fails its PID assertion
- accepting `EPERM` in the strict tree-absence classifier fails the platform boundary assertion
- the inherited-output test pins the spawned PID and fails if it survives tree termination

## Review dispositions

Review on the prior head found four blockers; all are addressed on the measured head:

1. Windows fixture values use environment variables rather than trailing PowerShell `$args`; the liveness probe also avoids the read-only, case-insensitive `$PID` automatic variable by using `$targetPid`. The Windows inner deadline is 5 seconds with a separate 10-second test guard.
2. The guard failure path performs no unbounded join or cleanup command. It fails immediately; the normal result path joins only after the worker has sent its terminal result.
3. A quiet descendant closes inherited stdio while its parent exits normally; the test preserves the parent outcome and proves the descendant is killed. The no-op mutation fails.
4. `EPERM` is never process-tree absence. It is accepted only on macOS, only in the post-status/post-I/O normal-completion path; Linux and forced termination accept only `ESRCH`.

CodeRabbit's poll-error finding is also fixed: post-kill reaping is a bounded `try_wait()` loop, not `wait()`. Its PID-reuse nit is dispositioned by removing cleanup signals entirely; the short liveness poll can now only produce a conservative test failure, never signal a reused unrelated PID. A positive calibration test pins that the platform liveness probe recognizes the current test process, so an always-dead probe cannot make descendant assertions pass.

## Verification

On the measured tree committed as the head above:

- `cargo test -p assay-cli` - pass
- `cargo test -p assay-mcp-server` - pass
- CLI bounded runner module - 8/8 pass
- MCP bounded runner module - 8/8 pass
- `cargo clippy -p assay-cli -p assay-mcp-server --all-targets --all-features -- -D warnings` - pass
- `cargo fmt --all -- --check` - pass
- `cargo deny check` - pass; existing duplicate/yanked/unmatched-license warnings only
- exact-file pre-commit - pass
- pre-push workspace clippy, RustSec audit, and Linux cross-target gate - pass
- `git diff --check` - pass

Dedicated feature proof in an isolated target from the first implementation commit remains applicable because later commits touch only the shared helper:

- `cargo test -p assay-mcp-server --features test-outbound --locked no_passthrough` - 1/1 pass

Delegated exact-head proof in progress:

- gate: `all`
- run: https://github.com/Rul1an/assay/actions/runs/31334297252
- sha: `207da8ce958621c2f10d81c281387d51054f6772`

## Known test-harness interaction

A synthetic combined command over `assay-cli` and `assay-mcp-server` with `--all-features` can fail `no_passthrough_e2e`: the existing CLI `e2e_mcp_wrap_assert_cmd` performs a nested featureless server build into the same Cargo uplift path. The test file already documents this Cargo behavior. The canonical package suites and isolated feature gate pass; this PR does not weaken or alter that gate.

## Platform non-claim

The local host proved macOS behavior and the pre-push gate proved Linux compilation. No Windows Rust target is installed locally, so Job Object runtime behavior is delegated to the final-head Windows GitHub Actions job and is not claimed before that job passes.

## Fail policy

- timeout/output overflow: fail closed after process-tree termination and bounded diagnostics
- process-tree termination/reap error: fail closed within a bounded grace; direct-child kill is cleanup and the tree error remains primary
- observed child status plus stdin `BrokenPipe`: preserve the child outcome after all observable output is collected
- other stdin failures: fail with bounded diagnostic context

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -272,6 +272,7 @@ dependencies = [
  "once_cell",
  "pkcs8 0.11.0",
  "predicates",
+ "process-wrap",
  "proptest",
  "rand 0.8.7",
  "ratatui",
@@ -421,7 +422,9 @@ dependencies = [
  "clap",
  "hex",
  "jsonwebtoken",
+ "libc",
  "moka",
+ "process-wrap",
  "rand 0.8.7",
  "reqwest 0.12.28",
  "rsa",
@@ -3833,6 +3836,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "process-wrap"
+version = "9.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2e842efad9119158434d193c6682e2ebee4b44d6ad801d7b349623b3f57cdf55"
+dependencies = [
+ "indexmap",
+ "nix 0.31.3",
+ "windows 0.62.2",
+]
+
+[[package]]
 name = "procfs"
 version = "0.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5136,7 +5150,7 @@ dependencies = [
  "ntapi",
  "once_cell",
  "rayon",
- "windows",
+ "windows 0.52.0",
 ]
 
 [[package]]
@@ -5970,6 +5984,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "windows"
+version = "0.62.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "527fadee13e0c05939a6a05d5bd6eec6cd2e3dbd648b9f8e447c6518133d8580"
+dependencies = [
+ "windows-collections",
+ "windows-core 0.62.2",
+ "windows-future",
+ "windows-numerics",
+]
+
+[[package]]
+name = "windows-collections"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "23b2d95af1a8a14a3c7367e1ed4fc9c20e0a26e79551b1454d72583c97cc6610"
+dependencies = [
+ "windows-core 0.62.2",
+]
+
+[[package]]
 name = "windows-core"
 version = "0.52.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5989,6 +6024,17 @@ dependencies = [
  "windows-link",
  "windows-result",
  "windows-strings",
+]
+
+[[package]]
+name = "windows-future"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e1d6f90251fe18a279739e78025bd6ddc52a7e22f921070ccdc67dde84c605cb"
+dependencies = [
+ "windows-core 0.62.2",
+ "windows-link",
+ "windows-threading",
 ]
 
 [[package]]
@@ -6018,6 +6064,16 @@ name = "windows-link"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f0805222e57f7521d6a62e36fa9163bc891acd422f971defe97d64e70d0a4fe5"
+
+[[package]]
+name = "windows-numerics"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6e2e40844ac143cdb44aead537bbf727de9b044e107a0f1220392177d15b0f26"
+dependencies = [
+ "windows-core 0.62.2",
+ "windows-link",
+]
 
 [[package]]
 name = "windows-result"
@@ -6119,6 +6175,15 @@ dependencies = [
  "windows_x86_64_gnu 0.53.1",
  "windows_x86_64_gnullvm 0.53.1",
  "windows_x86_64_msvc 0.53.1",
+]
+
+[[package]]
+name = "windows-threading"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3949bd5b99cafdf1c7ca86b43ca564028dfe27d66958f2470940f73d86d75b37"
+dependencies = [
+ "windows-link",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,6 +102,7 @@ clap = { version = "4", features = ["derive", "env"] }
 uuid = { version = "1.23", features = ["v4"] }
 chrono = { version = "0.4", features = ["serde"], default-features = false }
 tempfile = "3"
+process-wrap = { version = "9.1", default-features = false }
 regex = "1"
 async-trait = "0.1"
 sha2 = "0.11"
--- a/crates/assay-cli/Cargo.toml
+++ b/crates/assay-cli/Cargo.toml
@@ -123,3 +123,9 @@ predicates = "3.1.4"
 proptest = "1.10"
 serial_test = "3.5.0"
 criterion = "0.8"
+
+[target.'cfg(unix)'.dev-dependencies]
+process-wrap = { workspace = true, features = ["process-group", "std"] }
+
+[target.'cfg(windows)'.dev-dependencies]
+process-wrap = { workspace = true, features = ["job-object", "std"] }
--- a/crates/assay-cli/tests/agent_golden_path_contract.rs
+++ b/crates/assay-cli/tests/agent_golden_path_contract.rs
@@ -141,7 +141,7 @@ fn assay<S: AsRef<OsStr>>(cwd: &Path, args: &[S]) -> Output {
     }
     command.current_dir(cwd).env("NO_COLOR", "1").args(args);
     run_bounded(
-        &mut command,
+        command,
         b"",
         GOLDEN_PATH_LIMITS,
         "agent golden-path CLI command",
--- a/crates/assay-mcp-server/Cargo.toml
+++ b/crates/assay-mcp-server/Cargo.toml
@@ -51,3 +51,10 @@ base64 = "0.22"
 # uuid = { version = "1.0", features = ["v4"] } # Removed in favor of SystemTime
 rsa = "0.9"
 rand = "0.8"
+
+[target.'cfg(unix)'.dev-dependencies]
+libc = "0.2"
+process-wrap = { workspace = true, features = ["process-group", "std"] }
+
+[target.'cfg(windows)'.dev-dependencies]
+process-wrap = { workspace = true, features = ["job-object", "std"] }
--- a/crates/assay-mcp-server/tests/agent_golden_path_contract.rs
+++ b/crates/assay-mcp-server/tests/agent_golden_path_contract.rs
@@ -145,7 +145,7 @@ fn run_server<S: AsRef<OsStr>>(cwd: &Path, args: &[S], stdin: &[u8]) -> Output {
     let mut command = clean_server_command();
     command.current_dir(cwd).args(args);
     run_bounded(
-        &mut command,
+        command,
         stdin,
         GOLDEN_PATH_LIMITS,
         "agent golden-path MCP command",
@@ -166,7 +166,7 @@ fn required_python() -> &'static str {
     let mut command = Command::new(interpreter);
     command.arg("--version");
     let output = run_bounded(
-        &mut command,
+        command,
         b"",
         GOLDEN_PATH_LIMITS,
         "protected-action Python preflight",
--- a/tests/support/bounded_process.rs
+++ b/tests/support/bounded_process.rs
@@ -82,7 +82,6 @@ pub fn run_bounded(
     let mut child = command
         .spawn()
         .map_err(|error| format!("{context}: spawn {command_display}: {error}"))?;
-
     let mut child_stdin = child
         .stdin()
         .take()
@@ -121,19 +120,18 @@ pub fn run_bounded(
     let mut early_failure = None;
     let mut observed_status = None;
     let status = loop {
-        if let Ok(stream) = overflow_rx.try_recv() {
-            early_failure = Some(format!(
-                "{} exceeded its {}-byte ceiling",
-                stream.name(),
-                stream_limit(limits, stream)
-            ));
-            break terminate_and_reap(
-                child.as_mut(),
-                observed_status.take(),
-                context,
-                &command_display,
-            )?;
+        if early_failure.is_none() {
+            if let Ok(stream) = overflow_rx.try_recv() {
+                // The reader closes its pipe after limit + 1 bytes. Keep the
+                // child under this same deadline while it observes the close.
+                early_failure = Some(format!(
+                    "{} exceeded its {}-byte ceiling",
+                    stream.name(),
+                    stream_limit(limits, stream)
+                ));
+            }
         }
+
         if observed_status.is_none() {
             match child.try_wait() {
                 Ok(Some(status)) => observed_status = Some(status),
@@ -159,7 +157,9 @@ pub fn run_bounded(
         }
 
         if Instant::now() >= deadline {
-            early_failure = Some(format!("deadline of {:?} expired", limits.timeout));
+            if early_failure.is_none() {
+                early_failure = Some(format!("deadline of {:?} expired", limits.timeout));
+            }
             break terminate_and_reap(
                 child.as_mut(),
                 observed_status.take(),
@@ -212,6 +212,8 @@ pub fn run_bounded(
             &stderr,
         ));
     }
+    // Status and both output streams are already collected, so BrokenPipe here
+    // records an early stdin close without replacing the child's outcome.
     stdin_result.or_else(|error| {
         if error.kind() == std::io::ErrorKind::BrokenPipe {
             return Ok(());
@@ -300,7 +302,7 @@ fn finish_process_tree(
 ) -> Result<(), String> {
     match child.start_kill() {
         Ok(()) => Ok(()),
-        Err(error) if process_tree_absent(&error) => Ok(()),
+        Err(error) if process_tree_quiescent_after_io(&error) => Ok(()),
         Err(error) => Err(format!(
             "{context}: terminate remaining process tree for {command}: {error}"
         )),
@@ -309,13 +311,25 @@ fn finish_process_tree(
 
 #[cfg(unix)]
 fn process_tree_absent(error: &std::io::Error) -> bool {
-    // macOS may report EPERM when the owned group has only an already-reaped leader.
-    matches!(error.raw_os_error(), Some(libc::ESRCH) | Some(libc::EPERM))
+    error.raw_os_error() == Some(libc::ESRCH)
 }
 
 #[cfg(windows)]
 fn process_tree_absent(_error: &std::io::Error) -> bool {
     false
+}
+
+#[cfg(unix)]
+fn process_tree_quiescent_after_io(error: &std::io::Error) -> bool {
+    // macOS may report EPERM for an already-empty group. This is accepted only
+    // after child status, stdin, stdout, and stderr are all complete, so it can
+    // never lead into a blocking join on a live inherited pipe.
+    matches!(error.raw_os_error(), Some(libc::ESRCH) | Some(libc::EPERM))
+}
+
+#[cfg(windows)]
+fn process_tree_quiescent_after_io(error: &std::io::Error) -> bool {
+    process_tree_absent(error)
 }
 
 fn stream_limit(limits: ProcessLimits, stream: CapturedStream) -> usize {
@@ -366,6 +380,8 @@ fn excerpt(bytes: &[u8]) -> String {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(unix)]
+    use super::{process_tree_absent, process_tree_quiescent_after_io};
     use super::{run_bounded, ProcessLimits};
     use std::path::Path;
     use std::process::{Command, Stdio};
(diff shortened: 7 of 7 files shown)


Rul1an:
@coderabbitai review

Rul1an:
Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/31332887387
- sha: c6aa9226103573e843c08d977e90d37a4ae843c2

This record is informational; merge credit depends on the attested proof pack and lane-check/proof status.

Rul1an:
Read-only implementation review bound to exact head `c6aa9226103573e843c08d977e90d37a4ae843c2`. No edits, no commits, no merge.

**No blockers. Two MEDIUM findings, one of which the Windows job will settle better than I can.**

## Findings

### 1. MEDIUM — `$args` is not populated by `powershell.exe -Command` (inferred, not executed)

Both Windows helpers pass a trailing argument and read `$args[0]`:

- `inherited_output_descendant_command` — `-Command "... [IO.File]::WriteAllText($args[0], [string]$p.Id) ..."` then `.arg(pid_file)`
- `descendant_is_alive` — `-Command "if (Get-Process -Id $args[0] ...)"` then `&pid.to_string()`

PowerShell populates `$args` for `-File script.ps1 arg`, not for `-Command`. With `-Command`, trailing tokens are appended to the command text rather than bound, so `$args[0]` is `$null`. If that holds, `WriteAllText($null, ...)` throws, the pid file is never written, and `terminates_inherited_output_descendant` fails at `.expect("descendant must publish its pid before inheriting output")`.

**State of the evidence.** I reviewed from macOS and did not run PowerShell, so this is a reading of documented behaviour, not a measurement. `Build + Test (windows-latest)` is IN_PROGRESS on this head and is the authority. The good news either way: the test **panics** on a missing pid rather than skipping, so on Windows this fails loudly instead of passing decoratively — which is exactly what question 4 asks and the answer is that it is not decorative.

If that job goes green, this finding is void and I would rather be shown wrong by execution than argue it. If it goes red, the fix is `-File` with a temporary script, or interpolating the values into the command string rather than binding them.

### 2. MEDIUM — treating EPERM as "tree absent" can leave an unbounded join

`process_tree_absent` returns true for both `ESRCH` and `EPERM`. `ESRCH` is unambiguous. `EPERM` is not: the comment names the benign macOS case — an owned group whose only member is an already-reaped leader — but `killpg` also returns `EPERM` when a member cannot be signalled, and in that case a descendant is alive.

The consequence is specific. On the deadline path: `terminate_and_reap` → `start_kill()` returns EPERM → filtered as absent → `tree_error` is `None` → the loop breaks with a status → then `join_reader` on a pipe a live descendant still holds **blocks indefinitely**. The deadline has already fired and nothing bounds what follows. That is the deadlock #2189 exists to prevent, reachable through the one branch that decides a kill did not need to happen.

Not reachable by the current corpus — nothing in these tests changes uid — so this is a latent path rather than a live defect.

The cleanest fix is not a better EPERM probe. **The three joins after the loop are the only unbounded operations left in the function**; bounding them makes `run_bounded` unconditionally bounded regardless of what `start_kill` reports, and retires this class rather than this instance.

### 3. LOW — the PR body claims a guard the code does not have

The body says BrokenPipe is ignored "only after child status and output readers are available". The code ignores any `BrokenPipe` on `stdin_result` unconditionally; status and readers merely happen to be bound by that point. That is ordering, not a condition. The behaviour is defensible — a BrokenPipe on stdin means the child closed its read end, which is the child's prerogative — but the sentence describes a check that is not there.

## Verified, and the central one is right

**The four-way completion condition is the actual fix.** The loop now breaks normally only when the status is observed **and** the stdin writer **and** both readers have finished. Previously it broke on child exit and then blocked in `join_reader`; now a descendant holding an inherited pipe keeps a reader unfinished, so the deadline fires instead. That is #2189's second constraint closed at the right place, and it is what makes the descendant test meaningful rather than the process-group wrapper alone.

Also confirmed by inspection: `observed_status` is latched once so `try_wait` stops after the first `Some`; `finish_process_tree` terminates lingering descendants on the **success** path too, which is #2189's first constraint; every non-BrokenPipe stdin error still produces a full diagnostic with status and both excerpts.

**Dependency hygiene is exactly as claimed** — measured, not taken on trust: one workspace pin, `process-wrap = { version = "9.1", default-features = false }`, under `[target.'cfg(unix)'.dev-dependencies]` and `[target.'cfg(windows)'.dev-dependencies]` in both crates, with `process-group` / `job-object` respectively, and `libc` unix-dev-only, matching the `#[cfg(unix)]` use in `process_tree_absent`.

**Non-claims hold.** The doc comment on `run_bounded` states the escape limitation plainly rather than implying total containment, and the diff touches test support and manifests only — no production supervisor, CLI schema, MCP protocol, or JSON-RPC path.

## CI note

Two checks are red on this head: `lane-check` and `lane-check/proof`. Those bind to a SHA and normally need re-running against the head that merges; worth confirming that is all it is. `Build + Test (windows-latest)` is still running and is the one that decides finding 1.

Rul1an:
Reviewed exact head `c6aa9226103573e843c08d977e90d37a4ae843c2` against `origin/main` (`7afe2bef2b00db97b971ed6afd07710cd356a60a`).

**Verdict: FINDINGS (BLOCK)**

1. **[P1] The Windows process-tree fixture is killed before it can establish the descendant it claims to test.** `tests/support/bounded_process.rs:425`, `tests/support/bounded_process.rs:585`, `tests/support/bounded_process.rs:608`
   The fixture cold-starts PowerShell, calls `Start-Process ping.exe`, and writes the PID, but `run_bounded` starts a 250 ms deadline immediately. On this exact head, [Build + Test (windows-latest)](https://github.com/Rul1an/assay/actions/runs/31332759602/job/93293535685) fails at line 609 with "descendant must publish its pid before inheriting output." Use a deterministic/ready-signalled fixture or a realistic Windows startup deadline while retaining a separate wall-clock guard.

2. **[P1] The regression test's outer timeout can itself hang indefinitely.** `tests/support/bounded_process.rs:590`, `tests/support/bounded_process.rs:595`, `tests/support/bounded_process.rs:515`
   After the two-second receive timeout, cleanup and a second bounded receive are followed by an unbounded `worker.join()`; Windows cleanup also launches unbounded `taskkill`. If the mutation under test deadlocks `run_bounded`, this failure path deadlocks the test/CI instead of enforcing the claimed wall-clock bound. Run the probe in a killable subprocess or otherwise ensure no unbounded join/cleanup remains after the guard expires.

3. **[P2] Normal direct-child completion cleanup is unpinned and can regress decoratively.** `tests/support/bounded_process.rs:150`, `tests/support/bounded_process.rs:155`, `tests/support/bounded_process.rs:575`
   The only descendant test intentionally keeps inherited output open, so it always reaches the deadline/`terminate_and_reap` path. Replacing `finish_process_tree` with a no-op still leaves all added tests passing. Add a silent descendant that closes inherited stdio while the direct parent exits normally, then assert the direct outcome is preserved and the descendant is gone.

4. **[P2] A real Unix `EPERM` tree-kill failure is classified as absence on every Unix target.** `tests/support/bounded_process.rs:310`, `tests/support/bounded_process.rs:313`
   `EPERM` means the process group can exist while the caller lacks permission to signal it; treating it like `ESRCH` can suppress a failed tree termination and permit a wait/join hang. The comment documents a macOS-specific observation, but the predicate applies to Linux as well. Restrict any proven exception to the affected platform/state and fail closed elsewhere.

The EPIPE path itself is ordered correctly: non-BrokenPipe write errors remain failures, BrokenPipe is considered only after a child status has been observed, and timeout/overflow failures win. Target-specific `process-wrap` features are development-only and configured coherently.

Rul1an:
Final-head delegated proof dispatched.

Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/31333986304
- sha: f32c466e877b1f55be7b1152ae42527326493659

Rul1an:
@coderabbitai review

Rul1an:
CI triage on final head `f32c466e877b1f55be7b1152ae42527326493659`:

- `Lint (pre-commit)` failed before repository hooks ran because the hosted runner's Rust 1.96 toolchain had no `cargo` component.
- `Kernel Matrix Summary` is only the downstream aggregation of that failure.
- Classification: hosted-runner/toolchain infrastructure, not a code/test regression.
- Action: reran only the failed jobs; no workflow or product code was changed.

Rul1an:
Read-only re-review bound to exact head `f32c466e877b1f55be7b1152ae42527326493659`. Reviews on `c6aa9226` and `0d76704c` are void. Delta since `c6aa9226` is one file, `tests/support/bounded_process.rs` (+149/−59). No edits, no commits.

**No blocker remains in the runner.** One MEDIUM on Windows test validity, which the still-running job will settle better than I can.

## Findings

### 1. MEDIUM — `$pid` collides with PowerShell's read-only automatic variable (inferred, not executed)

`bounded_process.rs:522`:

```
"$pid = [Environment]::GetEnvironmentVariable('ASSAY_DESCENDANT_PID'); if (Get-Process -Id $pid -ErrorAction SilentlyContinue) { exit 0 } else { exit 1 }"
```

`$PID` is a PowerShell automatic variable holding the current process id, and it is read-only; PowerShell variable names are case-insensitive, so `$pid` is `$PID`. The assignment raises *Cannot overwrite variable PID because it is read-only or constant*.

Both outcomes are bad, and the quieter one is worse:

- if the error is non-terminating, `$pid` keeps PowerShell's **own** id, `Get-Process` finds it, the helper returns "alive" unconditionally, `wait_for_descendant_exit` never succeeds, and both descendant tests fail loudly;
- if it is terminating, the script exits non-zero, `descendant_is_alive` returns **false unconditionally**, `wait_for_descendant_exit` returns `true` on its first poll, and `kills_descendant_that_holds_inherited_output_open` and `kills_quiet_descendant_after_normal_parent_exit` **pass without verifying anything**.

The second branch is the decorative case the rest of this PR is careful to avoid.

Two fixes, and I would take both. Rename to a non-automatic identifier (`$descendantId`), and add a **positive control**: assert the descendant is alive *before* the bounded run completes. Without that, a helper that always reports "dead" is indistinguishable from a working one — the same reasoning that put an anchor guard on the docs-drift self-test.

I could not execute PowerShell from this host and `Build + Test (windows-latest)` is IN_PROGRESS at this head, so this is a reading of documented behaviour, not a measurement. If that job goes green, discard this finding.

### 2. LOW — PID reuse is not dispositioned

Nothing in the file addresses it. The pid is read from a file and probed later, so in principle it can be recycled between publication and probe. Bounded and unlikely in a test with second-scale windows, but the disposition is currently absent rather than recorded.

### 3. Note — `lane-check` and `lane-check/proof` are FAILURE at this head

Those bind to a SHA and normally need re-running against the head that merges. Worth confirming that is all it is before merge; everything else is 37 SUCCESS with 10 in progress.

## Prior findings: closed, and one of them better than I proposed

**EPERM.** This is the fix I would highlight. Rather than bounding the joins as I suggested, the predicate was **split**:

- `process_tree_absent` (`:329`) is now **ESRCH only** on Unix and is what the timeout and overflow paths use, so `terminate_and_reap` no longer accepts EPERM on the one path where a live descendant could still hold a pipe;
- `process_tree_quiescent_after_io` (`:338`) accepts EPERM on macOS only, and is called **exclusively** from `finish_process_tree`.

I checked the precondition rather than the comment: that call site is reached only when `observed_status.is_some()` and all three of `stdin_writer`, `stdout_reader` and `stderr_reader` report `is_finished()` (`:148-153`). The subsequent joins therefore return immediately, so the accepted EPERM cannot lead into a blocking join. That removes the cause instead of adding a second bound, which is the stronger of the two fixes.

`eperm_is_quiescent_only_on_macos_after_all_io_is_complete` (`:709`) pins both predicates, including that EPERM is *not* absent — the right granularity, since a real EPERM is impractical to provoke.

**Bounded post-kill reap.** `reap_after_tree_kill` (`:295`) polls `try_wait` against `REAP_GRACE` of 1 s and errors on expiry. Grepped for `.wait()` across the file: **none remain**.

**Unbounded joins.** Still plain `.join()` at `:174-177`, and now unreachable while blocking: the normal path requires all three threads finished, and the timeout and overflow paths terminate the tree with ESRCH-only tolerance first.

**Windows value passing.** Both scripts now read `ASSAY_DESCENDANT_PID_FILE` / `ASSAY_DESCENDANT_PID` through `[Environment]::GetEnvironmentVariable`, which sidesteps the `-Command` binding problem at the cause rather than working around it. `$path` and `$p` are safe names; only `$pid` collides, per finding 1.

**BrokenPipe.** Unchanged in behaviour, and the comment at `:216-217` now describes what the code does — status and both streams are collected before the check — instead of claiming a guard that was only ordering. Non-BrokenPipe stdin errors still produce a full diagnostic.

**Non-decorative coverage.** `kills_quiet_descendant_after_normal_parent_exit` (`:675`) exercises `finish_process_tree` with a descendant that does *not* hold the pipe, which is the normal-completion cleanup that the timeout tests cannot reach. `preserves_child_outcome_when_stdin_closes_early` (`:695`) writes 8 MiB into a child that exits 23 immediately, so EPIPE is guaranteed by construction rather than hoped for, and asserts exit code 23 survives.

**Realistic deadlines.** Per-platform startup and outer guards at `:554-569` replace the single literal.

## #2189 acceptance

All three constraints are met: descendants are terminated on timeout, overflow **and** normal completion; the inherited-pipe case can no longer defeat the wall-clock bound; and an early exit preserves the child's outcome instead of surfacing the writer's BrokenPipe.

Rul1an:
The final Windows liveness probe now uses `$targetPid`; PowerShell treats `$PID` case-insensitively as a read-only automatic variable.

New exact head: `668bcb425a3f1d8c904f3adf39bfbe0eadfb2b10`

Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/31334219928
- sha: 668bcb425a3f1d8c904f3adf39bfbe0eadfb2b10

Rul1an:
Final test-oracle hardening: the liveness probe must positively recognize the current test process before it can be trusted to report a descendant as gone. This prevents an always-dead Windows probe from passing cleanup assertions.

New exact head: `207da8ce958621c2f10d81c281387d51054f6772`

Assay-Runner delegated proof:
- gate: all
- run: https://github.com/Rul1an/assay/actions/runs/31334297252
- sha: 207da8ce958621c2f10d81c281387d51054f6772

Rul1an:
@coderabbitai review

Rul1an:
Read-only re-review bound to exact head `207da8ce958621c2f10d81c281387d51054f6772`. Delta since `f32c466e` is two commits touching one file, `tests/support/bounded_process.rs` (+9/−1).

**No blocker. The prior MEDIUM is closed, and closed in a way that no longer depends on my diagnosis being right.**

## The two changes

`:522` — `$pid` becomes `$targetPid`. Not an automatic variable, so the read-only collision cannot arise.

`:542-549` — a new `liveness_probe_recognizes_the_current_test_process` asserting `descendant_is_alive(std::process::id())` is true.

## Why the calibration test is the stronger half

My finding named two branches, and the quiet one was the problem: if the probe script failed for any reason, `descendant_is_alive` would return false unconditionally, `wait_for_descendant_exit` would succeed on its first poll, and both cleanup assertions would pass without verifying anything.

Both branches are now caught, and by different mechanisms:

- an always-**dead** oracle fails `liveness_probe_recognizes_the_current_test_process`, because the current process is demonstrably alive;
- an always-**alive** oracle leaves `wait_for_descendant_exit` never returning true, so the two cleanup tests fail loudly, as they already did.

That is worth stating plainly: **the calibration test makes my PowerShell inference irrelevant.** I read `$PID` as a read-only automatic variable and could not execute PowerShell to confirm it. Whether or not that reading was correct, an oracle that cannot recognise a live process now fails on its own terms. Defending the property beats diagnosing the cause, and this is the second time in this PR that the fix landed one level above the finding.

The test carries no `#[cfg]`, so it compiles and runs on both platforms, and it exercises the real oracle on each: `kill -0` against the test binary's own pid on Unix, `Get-Process -Id` on Windows. Both succeed for a live self, so it is a genuine control rather than a tautology.

## Prior runner findings, re-checked at this head

Unchanged by this delta and still correct: the ESRCH-only `process_tree_absent` on the timeout and overflow paths with `process_tree_quiescent_after_io` confined to `finish_process_tree` after all I/O is complete; `reap_after_tree_kill` bounded by `REAP_GRACE` with no `Child::wait` anywhere in the file; the joins unreachable while blocking; BrokenPipe accepted only with status and both streams already collected, and its comment describing what the code does; the quiet-descendant and 8 MiB early-stdin regressions non-decorative by construction.

The one remaining item from my last review is unchanged and unaddressed by design: **PID reuse is still not dispositioned** anywhere in the file. Low severity for second-scale windows, and a note rather than a request.

## CI at this head

`Build + Test` for all three platforms is QUEUED, so Windows has still not reported — but as above, the merge decision no longer hangs on it for this finding. `lane-check/proof` is FAILURE with `lane-check` in progress; those bind to a SHA and two pushes just moved the target, so re-running them against the head that merges is the expected step rather than a signal.

Rul1an:
Independent non-building agent review (Franklin), read-only and bound to exact head `207da8ce958621c2f10d81c281387d51054f6772`.

**Findings:** no blockers.

Verified:
- Windows uses `$targetPid`, environment-based argument passing, realistic deadlines, and positive liveness calibration.
- timeout guard failure performs no stuck-worker join; normal-completion cleanup has discriminatory coverage.
- forced termination accepts only `ESRCH`; macOS `EPERM` is confined to completed status/I/O.
- post-kill reaping is bounded by one-second polling grace.
- `BrokenPipe` is considered only after status/output collection; timeout and overflow retain precedence.
- clean exact-head worktree, complete diff against merge base `7afe2bef2b00db97b971ed6afd07710cd356a60a`, and `git diff --check`.

**Verdict: READY.** Exact-head platform CI was still running and is not claimed by this review.

Rul1an:
@copilot Please perform a read-only code review of exact head `207da8ce958621c2f10d81c281387d51054f6772`; do not make changes. Report findings or an explicit no-blocker verdict bound to that SHA.

Rul1an:
## Final-head merge quorum

Recorded at `2026-08-09T21:02:00Z` for exact head `207da8ce958621c2f10d81c281387d51054f6772`.

### Checks and delegated proof

- Current PR checks: zero pending and zero failing; PR is `CLEAN` and mergeable.
- Windows, macOS, and Ubuntu build/test jobs passed.
- Assay-Runner delegated `all` proof passed on this exact SHA: https://github.com/Rul1an/assay/actions/runs/31334297252
- The earlier red lane-check was rerun after proof publication and passed: https://github.com/Rul1an/assay/actions/runs/31334313064

### Review route

AGENTS.md route 2 is satisfied by two independent non-building reviews on the exact final head:

1. Claude Code Desktop: no blocker, bound to `207da8ce958621c2f10d81c281387d51054f6772`: https://github.com/Rul1an/assay/pull/2190#issuecomment-5233693175
2. Franklin: READY/no blockers, bound to `207da8ce958621c2f10d81c281387d51054f6772`: https://github.com/Rul1an/assay/pull/2190#issuecomment-5233695550

Bot availability on the exact final head:

- CodeRabbit explicitly reported `Review rate limited`: https://github.com/Rul1an/assay/pull/2190#issuecomment-5233689737
- Copilot was explicitly requested at `2026-08-09T20:31:04Z` and produced no current-head artifact during more than 30 minutes of unchanged-head silence: https://github.com/Rul1an/assay/pull/2190#issuecomment-5233699802
- Neither bot is counted as a pass.

All actionable CodeRabbit threads are resolved. The PID-reuse nit is technically dispositioned in the PR body: cleanup signals were removed; remaining liveness polling can only fail conservatively and cannot signal a reused unrelated PID.

PR #2203, opened 2026-08-10
docs(build): name the worktree target as the review-build default

### LINKED ISSUE #2202: Review rounds allocate one cargo target dir per head; document one per branch
## Measured

On 2026-08-10 the disk filled to 99% (139 MiB free of 460 GB) twice in one day. The mass was not worktree targets or the pre-push cache — both already covered by existing hygiene — but **per-head `CARGO_TARGET_DIR` scratch under `/tmp/assay-*`**, left by agent review rounds:

```
18G   /tmp/assay-target-2189
7.3G  /tmp/assay-review-2187-target
3.9G  /tmp/assay-target-2189-feature-proof-2
3.3G  /tmp/assay-review-f1303706-target
3.3G  /tmp/assay-review-ea41604-target
2.6G  /tmp/assay-target-2176-b-compact
1.7G  /tmp/assay-2152-plugin-packaging-target
```

Seven directories, all belonging to merged pull requests, roughly 40 GB. Note rows four and five: **two heads of the same pull request, two separate 3.3 GB targets.**

Two facts make this unbounded rather than self-limiting:

- macOS does not reap `/tmp` on the development machine. `/etc/periodic/daily` is empty and files from days earlier survive, so nothing expires on its own.
- Isolation per review round is deliberate and correct — a review build should not disturb the working tree. It is the *per-head naming* that duplicates, not the isolation.

## Why this is a documentation change rather than tooling

A daily reaper now removes `/tmp/assay-*` directories that carry `CACHEDIR.TAG` and are older than three days, so the symptom is handled and nobody has to remember it. This issue is about the cause.

A shared target directory per branch is the established mitigation for exactly the multi-worktree case, and it buys more than disk: the second review round of a pull request builds warm instead of cold. Cargo serialises concurrent builds against one target directory, which is the documented trade-off, and it costs nothing here because review rounds are sequential.

## Acceptance

- `CLAUDE.md`'s build section documents the canonical scratch target for review rounds: one per branch, not one per head.
- The wording is a build fact beside the existing build commands, not a housekeeping reminder. It belongs where an agent already reads how to build this repository.
- No enforcement check. A gate that inspects how someone sets an environment variable would cost more than it saves, and the reaper already bounds the consequence.

## Non-goals

- No change to the pre-push hook, the worktree layout, or `.cargo/config.toml`. A repository-level `target-dir` would apply to CI and to every contributor, which is a different decision with a different blast radius.
- No global `~/.cargo/config.toml` change; that would silently affect unrelated Rust work on the same machine.
- Not a replacement for existing worktree hygiene. Worktree `target/` directories belong to a branch rather than to an age, so they stay a deliberate cleanup.

## Provenance

Measured 2026-08-10 on the development machine. Directory sizes from `du -sh /tmp/assay-* | sort -rh`; the absence of OS-level reaping from an empty `/etc/periodic/daily` and surviving files dated three days earlier.

Rul1an:
Closes #2202.

## Summary

One paragraph in `CLAUDE.md`'s Key Commands section: review builds use the worktree's own `target/`, and `CARGO_TARGET_DIR` is set only when a build must not warm the branch's cache, to one stable path rather than one per reviewed head.

## Why

Review rounds in this programme improvised a scratch target per reviewed head — `/tmp/assay-review-<sha>-target`, `/tmp/assay-target-<issue>`. Two measurements, both on `4125d81b`:

- **The per-head form is inconsistent, not unsourced.** My original claim that it had no source was wrong: I grepped the wrong ref. Against `4125d81b`, `git grep -hoI "CARGO_TARGET_DIR=[^ "]*"` returns 13 distinct values across tracked files, a majority of them per-issue `/tmp` paths in plan documents. What the repository lacks is not a source but a stated default.
- **The scratch was not needed at all.** `.gitignore:29` is `/target`, so an in-tree build leaves `git status --porcelain` empty. Every read-only review in this programme asserted exactly that after building, so the property a scratch directory was protecting was never at risk.

The cost is real but the mechanism is narrower than I first wrote. macOS **does** reap `/tmp`, daily, through `/usr/libexec/tmp_cleaner` — files untouched for three days, then directories that have become empty. My original evidence was void: `/etc/periodic` does not exist on this machine, so an empty listing was an absent directory rather than an absence of jobs. A target still being built against keeps a fresh atime and never ages out while its pull request is open, which is why per-head paths accumulate across concurrent reviews.

This also folds an unmanaged class into a managed one. Worktree targets already have hygiene: clean them when the pull request merges. The `/tmp` scratch had none.

## Verification

Every factual claim in the added paragraph was checked against this tree rather than recalled:

- `grep -c "^/target$" .gitignore` → 1
- `grep -c 'CARGO_TARGET_DIR:-/tmp/assay-target' scripts/ci/phase5-check.sh` → 1
- `scripts/precommit/cargo-clippy.sh:57` sets `${common_dir}/prepush-target`, and `.pre-commit-config.yaml:317` declares `stages: [pre-push]`, so the hook attribution is accurate
- `pre-commit run --files CLAUDE.md` → all applicable hooks pass

## Non-goals

- No `.cargo/config.toml` change. A repository-level `target-dir` would apply to CI and to every contributor, which is a different decision with a different blast radius.
- No global `~/.cargo/config.toml` change; that would silently affect unrelated Rust work on the same machine.
- No enforcement check. A gate inspecting how someone sets an environment variable would cost more than it saves.
- No change to the pre-push hook, the worktree layout, or existing worktree hygiene.

## Non-claims

This does not reclaim space by itself and does not replace worktree cleanup at merge time. It states a default so the unmanaged class stops being created.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for configuring build output directories.
  * Clarified when to use an external build cache and recommended a stable repository path.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -61,11 +61,13 @@ cargo xtask build-ebpf                      # Build eBPF (Linux)
 ```
 
 Build in the worktree's own `target/`, including for read-only reviews: `/target` is git-ignored,
-so an in-tree build leaves `git status --porcelain` empty and needs no scratch directory. Set
-`CARGO_TARGET_DIR` only when a build genuinely must not warm the branch's cache, and then to one
-stable path per repository — `scripts/ci/phase5-check.sh` and the pre-push hook are the idiom. Not
-one path per reviewed head: those are never reaped (`/tmp` has no expiry on macOS) and reached
-~40 GB across seven merged pull requests on 2026-08-10.
+so an in-tree build leaves `git status --porcelain` empty and needs no scratch directory. That
+already satisfies `AGENTS.md`'s rule against sharing a target between active worktrees, so a
+scratch path is not required to honour it. If you do set `CARGO_TARGET_DIR`, keep one path per
+worktree — never one per reviewed head. macOS does reap `/tmp` daily through
+`/usr/libexec/tmp_cleaner`, but only files untouched for three days and then only directories that
+have become empty, so a target still being built against never ages out while its pull request is
+open; per-head paths therefore accumulate one full target per review round.
 
 ## CLI Entry Points
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,6 +60,13 @@ cargo clippy --workspace --all-targets -- -D warnings  # Lint
 cargo xtask build-ebpf                      # Build eBPF (Linux)
 ```
 
+Build in the worktree's own `target/`, including for read-only reviews: `/target` is git-ignored,
+so an in-tree build leaves `git status --porcelain` empty and needs no scratch directory. Set
+`CARGO_TARGET_DIR` only when a build genuinely must not warm the branch's cache, and then to one
+stable path per repository — `scripts/ci/phase5-check.sh` and the pre-push hook are the idiom. Not
+one path per reviewed head: those are never reaped (`/tmp` has no expiry on macOS) and reached
+~40 GB across seven merged pull requests on 2026-08-10.
+
 ## CLI Entry Points
 
 All commands defined in `crates/assay-cli/src/cli/args/mod.rs`, dispatched in `crates/assay-cli/src/cli/commands/mod.rs`. The table below is a representative subset; the CLI has ~40 subcommands (see `commands/` for the full set, including `import`, `project-otel`, `inventory`, `discover`, and the `verify-*` evidence family).


Rul1an (commented):
## Independent adversarial review

Posted by a Claude Code subagent acting as an independent reviewer. The author of this pull request could not self-review it, so this stands in as the review. My brief was to **refute** the change, not to confirm it.

- **Reviewed SHA:** `a8b59dc200576c66e47182bd1a6eb7bff04c8a86` against base `4125d81b00847a7dc9a2276c9a7d0c44a247e569`.
- **Worktree:** `/Users/roelschuurkes/.config/superpowers/worktrees/assay-2202-target/worktree`, HEAD confirmed at `a8b59dc2`. `git status --porcelain` empty before and after all work, including after two cargo builds.
- Everything below marked **[ran]** I executed; **[read]** I only read.

---

# Findings

## BLOCKING

### F1 — The paragraph states a false fact about macOS. `/tmp` does have an expiry. `CLAUDE.md:65-66`

The merged text says per-head targets `are never reaped (`/tmp` has no expiry on macOS)`. That parenthetical is the only justification the paragraph offers for its prohibition, and it is wrong.

**[ran]**

```
$ plutil -p /System/Library/LaunchDaemons/com.apple.tmp_cleaner.plist
  "Label" => "com.apple.tmp_cleaner"
  "ProgramArguments" => [ 0 => "/usr/libexec/tmp_cleaner" ]
  "StartCalendarInterval" => { "Hour" => 0 }

$ launchctl print system/com.apple.tmp_cleaner
	path = /System/Library/LaunchDaemons/com.apple.tmp_cleaner.plist
	type = LaunchDaemon

$ strings /usr/libexec/tmp_cleaner
daily_clean_tmps_dirs="/tmp"
daily_clean_tmps_days="3"
args="-atime +$daily_clean_tmps_days -mtime +$daily_clean_tmps_days"

$ man 8 tmp_cleaner
NAME  tmp_cleaner - Remove old content from /tmp
HISTORY  tmp_cleaner first appeared in macOS 14.0
```

macOS ships a daily `/tmp` reaper with a three-day threshold, loaded on this machine. The accurate statement is narrower and more interesting than the one in the paragraph: `tmp_cleaner` deletes `-type f` older than three days but removes directories only when `-empty`, so a cargo target's *contents* do drain while the directory skeleton lingers across several runs. "No expiry" is not a rounding error on that; it is the opposite claim.

The origin of the belief is checkable and it is also void. **[ran]** `ls /etc/periodic/` → `No such file or directory`. Issue #2202 and `~/.local/bin/assay-scratch-reaper.sh` both cite "`/etc/periodic/daily` is empty" as the evidence. That directory does not exist on this machine at all, so the cited evidence never supported the conclusion; macOS moved this to `com.apple.tmp_cleaner` in 14.0. The wrong belief already has two carriers. Merging this makes `CLAUDE.md` the third, and `CLAUDE.md` is the file agents read as ground truth.

Fix: drop the parenthetical, or replace it with what is actually true. The recommendation does not need this claim to stand.

### F2 — The escape-hatch clause instructs what `AGENTS.md` forbids, in a file that declares it cannot override `AGENTS.md`

`CLAUDE.md` preamble (unchanged, lines 3-5):

> Cross-agent work must first read and follow [AGENTS.md](AGENTS.md). ... This file adds Assay-specific technical context and **does not override that contract**.

`AGENTS.md:49`:

> Do not share `target/` directories between active worktrees. Use `CARGO_TARGET_DIR`.

New `CLAUDE.md:65-66`:

> Set `CARGO_TARGET_DIR` only when a build genuinely must not warm the branch's cache, and then to one stable path **per repository**

One stable path per repository is, by construction, shared between active worktrees. This is not hypothetical about the idiom being cited: **[ran]** from inside this worktree, `git rev-parse --path-format=absolute --git-common-dir` → `/Users/roelschuurkes/assay/.git`, and `du -sh /Users/roelschuurkes/assay/.git/prepush-target` → `2.7G` — one directory shared by all 85 worktrees `git worktree list` reports.

Note also that this drifted from the issue it closes. #2202's acceptance says "one per branch, not one per head". Per-repository is a third option, and it is precisely the one `AGENTS.md` rules out. Either narrow the wording to per-branch/per-worktree, or say openly that the two documents disagree and amend `AGENTS.md` in the same change.

## IMPORTANT — not in the merged text, but the stated rationale does not survive

### F3 — "The per-head form had no source" is refuted by re-running the author's own command

The body claims the grep "returns three values and all three use one stable path". **[ran]** verbatim at the base SHA:

```
$ git grep -hoI "CARGO_TARGET_DIR=[^ ]*" 4125d81b -- docs scripts | sort | uniq -c | sort -rn
   9 CARGO_TARGET_DIR=/tmp/assay-target-2159
   4 CARGO_TARGET_DIR=/tmp/assay-2173-process-bounds-target
   4 CARGO_TARGET_DIR=/tmp/assay-2152-target
   3 CARGO_TARGET_DIR=/tmp/assay-target-2152
   2 CARGO_TARGET_DIR=/tmp/assay-2176-durability-target
   2 CARGO_TARGET_DIR="$(mktemp
   1 CARGO_TARGET_DIR=/tmp/assay-target-2159`
   1 CARGO_TARGET_DIR=/tmp/assay-target-2152`.
   1 CARGO_TARGET_DIR=/tmp/assay-2176-durability-target`;
   1 CARGO_TARGET_DIR=/tmp/assay-2173-process-bounds-target`;
   1 CARGO_TARGET_DIR="${common_dir}/prepush-target"
   1 CARGO_TARGET_DIR="${CARGO_TARGET_DIR:-/tmp/assay-target}"
   1 CARGO_TARGET_DIR="$scratch_target"
```

Thirteen distinct values, not three, and the majority are exactly the per-issue `/tmp` scratch the paragraph forbids, carried by six tracked files under `docs/`. So "Nothing in the repository asked for that" and "The per-head form had no source" are both false against the tree they were measured on.

Worse for the change as merged, `docs/guides/agent-golden-path.md:64` is a current *guide* rather than a dated plan, and it prescribes a fresh directory **per invocation**, which is stricter than per-head:

```bash
export CARGO_TARGET_DIR="$(mktemp -d "${TMPDIR:-/tmp}/assay-target-2154.XXXXXX")"
```

After this merges, `CLAUDE.md` contradicts a tracked guide that this PR does not touch. An agent following the golden path will do the thing `CLAUDE.md` now names as the failure. Update `docs/guides/agent-golden-path.md` here, or name it as a deliberate exception.

### F4 — "Folds an unmanaged class into a managed one" is not established; the receiving class has already overflowed

**[read]** `~/.local/bin/cargo-target-sweep`, header:

> a repo with dozens of worktrees accumulates one multi-GB target/ per worktree. Five idle ones held 41 GB and filled a 460 GB disk to 100%, which broke builds with "ld: write() failed, errno=28".

**[ran]** `~/Library/Logs/cargo-target-sweep.log`:

```
2026-08-09T07:51:07 free 0G is below 20G — using emergency threshold 1d instead of 7d
2026-08-09T07:51:54 sweep done: swept=4 kept=2 reclaimed=20.5G free=0G->29G
2026-08-10T07:29:45 free 0G is below 20G — using emergency threshold 1d instead of 7d
2026-08-10T07:30:23 sweep done: swept=0 kept=5 reclaimed=0G free=0G->5G
```

Worktree `target/` directories filled the disk to 0G free on 2026-08-09, the day *before* the event this PR cites, and 20.5 GB had to be reclaimed from them. Current state **[ran]**: 85 worktrees, `assay-2200-cfg-argv/worktree/target` at 9.4G, `assay-2173-process-bounds/target` at 3.1G.

Two consequences for the argument. First, the body's "Worktree targets already have hygiene: clean them when the pull request merges" is not what actually holds the line — an unattended 7-day sweeper does, with a 1-day emergency threshold, and even that found nothing to sweep on 2026-08-10. Second, the direction of the advice is against the disk argument, not with it: `/tmp` is reaped at three days by `tmp_cleaner` (F1), worktree `target/` at seven, or one under pressure. This moves builds from the more-reaped location to the less-reaped one.

### F5 — The concurrency objection is already answered, against this PR, by the author's own tooling

**[read]** `~/.local/bin/cargo-target-sweep`, same header:

> Why not a shared CARGO_TARGET_DIR: cargo locks the target directory for the duration of a build, so worktrees sharing one would serialize. With several agent sessions building at once that is a worse trade than the disk.

That is a direct rebuttal of F2's "one stable path per repository", written previously by the same author, and the PR does not engage with it. #2202 waves it off with "it costs nothing here because review rounds are sequential" — with 85 worktrees and concurrent agent sessions that premise is asserted, not shown.

### F6 — The historical numbers: what is checkable now and what is not

I am splitting these rather than accepting them.

- **Verified [ran].** The disk was full on 2026-08-10: `2026-08-10T07:29:45 free 0G is below 20G` in the sweep log.
- **Not verifiable, and it should not be treated as established.** The ~40 GB across seven `/tmp` directories. I checked all seven by name; every one is ABSENT. `du -sh /tmp` now reports `0B`; free space is 42Gi. The figure rests entirely on the author's contemporaneous `du` output, which is reasonable to trust but is gone. `CLAUDE.md` would carry "~40 GB ... on 2026-08-10" permanently as an assertion no future reader can check. Consider dropping the number and keeping the instruction.
- **Refuted [ran].** #2202 says "A daily reaper now removes `/tmp/assay-*` directories". `~/.local/bin/assay-scratch-reaper.sh` exists but is scheduled nowhere: absent from `~/Library/LaunchAgents`, `/Library/LaunchAgents`, `/Library/LaunchDaemons`; `crontab -l` contains only the bpf-runner health check; `launchctl list` has no matching entry. It is a manual script. Not in the merged text, but it is load-bearing for the "the symptom is handled" disposition.

## MINOR

- Body says "`.gitignore:29` is `/target`". **[ran]** `git grep -n "^/target$" 4125d81b -- .gitignore` → `.gitignore:37:/target`. Not in the merged text.

---

# Refutations I attempted and could not make stick

Stating these deliberately: a failed attack is worth more than agreement.

1. **The core claim — an in-tree build leaves the tree clean.** I attacked this by execution rather than by reading `.gitignore`, which is what the instruction asks readers to rely on. From a clean worktree at `a8b59dc2`: **[ran]** `cargo build -p assay-cli` (Finished in 1m 13s, seven crates compiled) → `git status --porcelain` empty. Then **[ran]** `cargo test -p assay-evidence -p assay-canonical` → passing, `git status --porcelain` empty. I could not dirty the tree. The load-bearing claim holds.
2. **A nested-crate gitignore gap.** `/target` is anchored, so a nested crate's `target/` would escape it. **[ran]** `assay-python-sdk` is a workspace member (root `Cargo.toml` `members`), so it shares the root `target/`; `find . -maxdepth 3 -name target -type d` outside root returns nothing. Python artifacts are covered: `.gitignore` carries `__pycache__/`, `.pytest_cache/`, `.venv/`. No gap.
3. **Alternate spellings that would contradict the "stable path" characterisation.** **[ran]** Checked `--target-dir`, `[build] target-dir`, `CARGO_BUILD_TARGET_DIR`, and `.cargo/config.toml`. The only hits are prose in a doc comment. `.cargo/config.toml` sets an alias and darwin rustflags, no `target-dir`. Nothing contradicts the characterisation of the two cited scripts.
4. **The hook attribution.** **[ran]** `scripts/precommit/cargo-clippy.sh:55-57` sets `CARGO_TARGET_DIR="${common_dir}/prepush-target"` only when unset; `.pre-commit-config.yaml` declares `cargo-clippy` with `stages: [pre-push]`, as does `cargo audit`. `scripts/ci/phase5-check.sh:9` is `export CARGO_TARGET_DIR="${CARGO_TARGET_DIR:-/tmp/assay-target}"`. Both are stable paths, both run where the paragraph says. The attribution is accurate.
5. **Hook cleanliness.** **[ran]** `pre-commit run --files CLAUDE.md` → every applicable hook Passed, exit 0, tree still clean.

**Is it correct but useless?** No. It is specific, actionable, and sits where an agent already reads how to build this repository, so it would change behaviour. My objection is the inverse: it would change behaviour on the strength of a rationale that does not survive checking.

---

# Verdict

**Request changes.** Not on the recommendation — the central instruction survived every refutation I could construct, and I verified by execution that an in-tree build and an in-tree test run both leave `git status --porcelain` empty. That part is right and worth landing.

Two defects are in the merged text itself and should block:

- **F1**, a false factual claim about macOS `/tmp` that would enter `CLAUDE.md` as ground truth, when the truth is available from `man 8 tmp_cleaner`.
- **F2**, an escape-hatch that instructs sharing a target directory between active worktrees, which `AGENTS.md:49` forbids, inside a file that explicitly disclaims overriding `AGENTS.md`.

**F3** should also be resolved before merge: `docs/guides/agent-golden-path.md:64` prescribes exactly the pattern this paragraph forbids, and leaving both in the tree replaces one ambiguity with a contradiction.

**F4**–**F6** do not change the recommendation but do undercut the reasoning offered for it. I would drop the ~40 GB figure rather than commit an unverifiable number to a durable document, and either narrow the "one stable path per repository" clause or reconcile it with the serialisation argument the author already wrote down in `cargo-target-sweep`.

Rul1an:
Revised at `b1dc8f198`, addressing all three blocking findings from the adversarial review on `a8b59dc20057`. I verified each one against this machine rather than accepting it, and each held.

## 1. The macOS claim was false, and so was its evidence

`/usr/libexec/tmp_cleaner` exists and is scheduled daily at hour 0. Its own text sets `daily_clean_tmps_dirs="/tmp"`, `daily_clean_tmps_days="3"`, and reaps with `-atime`/`-mtime`/`-ctime` for files and `-empty -mtime` for directories.

The evidence I gave was worse than the claim. `/etc/periodic` **does not exist on this machine** — my `ls` returned nothing and I read an absent directory as an absence of jobs. That is the same error this repository catches everywhere else: an empty result treated as a measurement rather than as a question that was never asked.

The paragraph now states the narrower mechanism that is actually true: reaping happens, but a target still being built against keeps a fresh atime and never ages out while its pull request is open.

I also dropped a first-draft attempt to quote the flags as `-atime +3 -mtime +3`. There is a third predicate, `-ctime +3`. Quoting two of three flags in a durable document is worse than quoting none, so the paragraph now describes the behaviour and names the script for anyone who wants to check.

## 2. The escape hatch contradicted AGENTS.md

`AGENTS.md:49` — "Do not share `target/` directories between active worktrees." My line recommended one stable path **per repository**, which is shared by construction, inside a file whose own header defers to `AGENTS.md`. The cited `${common_dir}/prepush-target` is indeed one shared directory across every worktree.

Replaced with one path per worktree, and the paragraph now says explicitly that an in-tree target already satisfies that rule — turning a contradiction into a stated consistency.

## 3. "The per-head form had no source" was false

I grepped the wrong ref. Against `4125d81b` the search returns 13 distinct values across tracked files, a majority of them per-issue `/tmp` paths in plan documents. The body now says what is actually true: the form is inconsistent rather than unsourced, and what the repository lacks is a stated default.

## 4. The unverifiable number is out

The ~40 GB figure is gone from `CLAUDE.md`. The directories no longer exist, so committing it to a durable document would have made it permanently uncheckable. The disk-full event stays in the pull request body, where it is provenance rather than doctrine.

## What survived the attack, and is worth recording as such

The reviewer tried to dirty a worktree with both `cargo build` and `cargo test` and could not: `git status --porcelain` stayed empty. No nested-crate gap, no alternative spelling of the variable, and the hook attribution checked out. The core claim of this change is the one thing that was attacked directly and held.

## Separately: an overstatement in issue #2202

The reviewer also refuted a claim in the issue, not this pull request: it says a daily reaper "now removes `/tmp/assay-*`". The script exists, but it is scheduled in an application-level scheduler, not in `launchd` or `crontab`. Nothing runs it if that application is closed. I will correct the issue text; it does not affect this change.

Rul1an (commented):
## Independent adversarial review (second pass)

Posted by a Claude Code subagent acting as an independent reviewer. The author of this pull request could not self-review it, so this stands in as the review. My brief was to **refute** the revision, and specifically to assume that fixing the three earlier defects introduced new ones.

- **Reviewed SHA:** `b1dc8f19893d676fef333e1b8b5612cc2de9f98f` against base `4125d81b00847a7dc9a2276c9a7d0c44a247e569`.
- **Worktree:** `/Users/roelschuurkes/.config/superpowers/worktrees/assay-2202-target/worktree`, HEAD confirmed at `b1dc8f19`. `git status --porcelain` empty before and after all work, including after a build, a test run and a `pre-commit` run.
- **[ran]** = I executed it. **[read]** = I only read it.

The three earlier defects are genuinely closed. Two new ones took their place in the same two sentences, and one earlier finding was dropped without being mentioned.

---

# Findings

## BLOCKING

### F1 — The `therefore` at `CLAUDE.md:69-70` does not follow from the premise the same sentence establishes

```
macOS does reap /tmp daily through /usr/libexec/tmp_cleaner, but only files untouched for
three days and then only directories that have become empty, so a target still being built
against never ages out while its pull request is open; per-head paths therefore accumulate
one full target per review round.
```

The premise is correct and I verified it. **[ran]**

```
$ cat /usr/libexec/tmp_cleaner
daily_clean_tmps_dirs="/tmp"
daily_clean_tmps_days="3"
args="-atime +$daily_clean_tmps_days -mtime +$daily_clean_tmps_days"
args="${args} -ctime +$daily_clean_tmps_days"
dargs="-empty -mtime +$daily_clean_tmps_days"

$ launchctl print system/com.apple.tmp_cleaner
	type = LaunchDaemon
	runs = 7
	last exit code = 0
	descriptor = { "Hour" => 0 }
```

A directory that is still being built against is non-empty, so it is never removed. That much holds.

The conclusion does not attach to it. Under per-head naming, round *N+1* creates a **new** path, which makes round *N*'s target one that is **no longer being built against**. The premise covers only targets still under active build — at most one per open pull request. Every earlier round's target falls into exactly the case the reaper *does* handle: its files match `-atime +3 -mtime +3 -ctime +3` and are deleted, and the emptied directory goes on a later run. The mechanism the paragraph names explains the opposite of what it concludes.

Worth naming the shape, because it is why this recurred. Revision 1 said "`/tmp` has no expiry, therefore per-head paths accumulate". That premise was refuted. Revision 2 says "`/tmp` is reaped but not in this case, therefore per-head paths accumulate". The conclusion was held fixed while a second premise was fitted underneath it. The 40 GB observation is real; what has never been shown is that it survives a reaping window, and since the seven directories are gone it can no longer be shown at all.

Fix: drop the second clause, or state the reason that is actually load-bearing and does not depend on the reaper — per-head naming makes each round's target unreusable by name, so it forfeits the warm cache and creates a directory nothing will ever look for again.

### F2 — `CLAUDE.md:65-66` asserts consistency with `AGENTS.md` while quoting half its rule and contradicting the other half

**[read]** `AGENTS.md:49`, in full:

> - Do not share `target/` directories between active worktrees. Use `CARGO_TARGET_DIR`.

**[read]** `CLAUDE.md:65-66`:

> That already satisfies `AGENTS.md`'s rule against sharing a target between active worktrees, so a scratch path is not required to honour it.

The rule is two sentences. The paragraph paraphrases it as if only the first exists. The second is an imperative — *use* `CARGO_TARGET_DIR` — and the new default is to set no `CARGO_TARGET_DIR` at all. So the advice satisfies the rule's intent while contradicting its letter, inside a file whose own header (`CLAUDE.md:3-5`) says it "does not override that contract".

This is a harder version of the previous F2, not a softer one. Revision 1 was silently inconsistent with `AGENTS.md`. Revision 2 makes an explicit claim of consistency, so a reader has no reason to go check. Two agents reading the two documents literally will do different things and each can cite a governing file.

`AGENTS.md:49` is one line. Amend it here, or say plainly that the in-tree default is an agreed exception to its second sentence.

## IMPORTANT

### F3 — The previous review's golden-path finding is unresolved, and the disposition comment does not mention it

**[read]** `docs/guides/agent-golden-path.md:64`:

```bash
export CARGO_TARGET_DIR="$(mktemp -d "${TMPDIR:-/tmp}/assay-target-2154.XXXXXX")"
```

One path per *invocation* — strictly more extreme than per-head, and squarely against `CLAUDE.md:66-67` "keep one path per worktree — never one per reviewed head". It sits in the hand-written `## Verification` section, below the generated-table marker at line 45, so it is a live instruction rather than a dated plan. The disposition comment lists four addressed items and this is not among them; the diff touches `CLAUDE.md` only.

I attacked the guide's premise directly, by running its own command with no `CARGO_TARGET_DIR` at all, in-tree, at `b1dc8f19`. **[ran]**

```
$ git status --porcelain | wc -l
0
$ cargo test -p assay-cli --test agent_golden_path_contract
test result: ok. 24 passed; 0 failed; 0 ignored; finished in 0.38s
$ git status --porcelain | wc -l
0
```

The scratch the guide mandates is unnecessary for the exact command it guards, by this pull request's own argument.

Two further wrinkles. **[ran]** `echo $TMPDIR` → `/var/folders/8_/…/T/`, and `tmp_cleaner`'s own config is `daily_clean_tmps_dirs="/tmp"` — so on macOS that scratch lands outside the mechanism `CLAUDE.md:68` cites entirely. (Whether `/var/folders` is reaped by `com.apple.bsd.dirhelper`, which is loaded here, I did not verify.)

And this bears on whether the change does anything at all. The paragraph's first sentence prescribes cargo's default, so the only readers whose behaviour it changes are ones currently setting `CARGO_TARGET_DIR` — and the one tracked *guide* telling them to is the file this pull request leaves alone.

### F4 — "keep one path per worktree" is stated without exception, and the repository's own tooling breaks it

**[ran]**

```
$ sed -n '55,57p' scripts/precommit/cargo-clippy.sh
if [[ -z "${CARGO_TARGET_DIR:-}" ]]; then
    common_dir="$(git rev-parse --path-format=absolute --git-common-dir …)"
    export CARGO_TARGET_DIR="${common_dir}/prepush-target"

$ git rev-parse --path-format=absolute --git-common-dir   → /Users/roelschuurkes/assay/.git
$ git worktree list | wc -l                               → 85
$ du -sh /Users/roelschuurkes/assay/.git/prepush-target   → 2.7G

$ grep -n CARGO_TARGET_DIR scripts/ci/phase5-check.sh
9:export CARGO_TARGET_DIR="${CARGO_TARGET_DIR:-/tmp/assay-target}"
```

One directory shared by 85 worktrees, plus one fixed `/tmp` path. The disposition comment identifies the first correctly ("indeed one shared directory across every worktree") and then removes the *citation* rather than the *conflict*. `CLAUDE.md` now carries a per-worktree rule that two tracked scripts break, with the non-goals explicitly excluding any change to the pre-push hook. An agent reads `CLAUDE.md:66-67`, runs the pre-push hook, and watches the doc be contradicted by the tooling. Either scope the sentence to hand-run review builds, or name the hooks as the known exception.

Separately, the escape hatch lost its referent in this revision. "Keep one path per worktree" does not say *where*, and the obvious reading — the worktree's own `target/` — is the default the preceding sentence already prescribes. The only non-trivial reading is a path outside the worktree, and the `phase5-check.sh` idiom the first draft pointed at was dropped. A reader who wants the hatch now has nothing concrete to copy.

## MINOR

### F5 — The pull request body still says `.gitignore:29`

**[ran]** `git grep -n "^/target$" b1dc8f19 -- .gitignore` → `.gitignore:37:/target`. Flagged as MINOR against the previous head and still present in the **Why** section. The body's own **Verification** section is correct (`grep -c "^/target$" .gitignore` → 1, which I reproduced), so the body contradicts itself two sections apart.

### F6 — Timestamp wording is loose in the direction of overstating the reaper

`find`'s `-atime +3` matches `n > 3` where `n = floor(age / 86400)`, so the effective threshold is four days, not three; and the predicate is a **conjunction** over atime, mtime *and* ctime, so a file survives if any one clock is fresh. `dargs` also carries `-mtime +3` on directories, which "only directories that have become empty" omits. Every one of these makes the reaper *less* aggressive than the sentence says, so none changes the conclusion — but `three days` is the script's variable name, not its behaviour. Describing rather than quoting flags was the right call after the first draft; this is the residue of it.

---

# Refutations I attempted and could not land

Stated deliberately. A failed attack is worth more than agreement.

1. **The core claim — an in-tree build leaves the tree clean.** Attacked by execution, not by reading `.gitignore`. **[ran]** at `b1dc8f19`: `git status --porcelain` empty → `cargo build -p assay-cli` (Finished, 1.13s warm) → empty → `cargo test -p assay-cli --test agent_golden_path_contract`, 24 passed → empty. I also hashed `Cargo.lock` before and after (`c60113399a35488a20273f5e0230df89` both times) in case a build resolved a dependency into a tracked file. It did not. I could not dirty the tree.
2. **`/target` anchoring inside a linked worktree.** `/target` is anchored to the repo root, so a linked worktree's target could plausibly escape it. It does not: the pattern anchors to the directory holding the `.gitignore`, which in a worktree is the worktree root. Demonstrated rather than argued — this worktree carries a 4.6 GB `target/` and still reports clean.
3. **Every factual component of the `tmp_cleaner` sentence.** Existence, loading, scope, threshold, and configurability all held. **[ran]** the script exists as `-rwxr-xr-x root:wheel` POSIX sh on the sealed read-only system volume, so it cannot be edited locally; the daemon is loaded with `runs = 7` and `last exit code = 0`; `StartCalendarInterval → Hour 0` is daily; scope is `/tmp` only; and the three-day value is hardcoded with no config file sourced, so there is nothing to override it. Naming the script instead of quoting its flags was the correct call.
4. **The atime mechanism the body relies on.** If the data volume were mounted `noatime`, "keeps a fresh atime" would be false. **[ran]** `mount` → `/System/Volumes/Data (apfs, local, journaled, nobrowse, protect, root data)`, no `noatime`; and a direct test showed atime advancing on a read (`11:20:53` → `11:20:55`) while mtime stayed put. The mechanism is real.
5. **"13 distinct values, a majority per-issue `/tmp` paths".** **[ran]** verbatim at `4125d81b`: exactly 13 distinct values, of which 9 are per-issue `/tmp` paths. Accurate, and the correction of the earlier "no source" claim is itself correct.
6. **Hook cleanliness.** **[ran]** `pre-commit run --files CLAUDE.md` at `b1dc8f19` → every applicable hook Passed, tree clean afterwards.

# Not verifiable, and it should not be treated as settled

- Whether the seven `/tmp/assay-*` directories would have survived a reaping window. They are gone; `/tmp` no longer holds them. That is precisely the gap F1 turns on, and it cannot be closed after the fact. On this machine no target directory is currently old enough to test the predicate against: **[ran]** across five worktree targets (10 GB, 4.6 GB, 3.1 GB, 636 MB, 532 MB), `find -type f -atime +3 -mtime +3 -ctime +3` returns **0** in every one, because all were touched within the last day.
- Whether `/var/folders` TMPDIR scratch is reaped, and on what schedule.

---

# Verdict

**Request changes.**

Again, not on the recommendation. The central instruction survived every refutation I could construct, and I verified by execution — build, test, and `Cargo.lock` hash — that an in-tree run leaves `git status --porcelain` empty. That part is right and should land.

What blocks is that both sentences carrying the *justification* still do not hold:

- **F1**, a `therefore` that inverts the scope of its own premise, in the sentence that states why the rule exists. The pattern across two revisions — conclusion fixed, premise refitted — is the thing to fix, not just the wording.
- **F2**, an explicit claim of consistency with `AGENTS.md:49` that is reached by quoting one of its two sentences and contradicting the other, in a file that disclaims overriding it.

**F3** should be resolved in this change: `docs/guides/agent-golden-path.md:64` prescribes a fresh target per invocation, which is the pattern `CLAUDE.md` would now name as the failure, and it is the file that actually causes the behaviour this pull request is trying to change. It was raised on the previous head and the disposition comment does not mention it.

**F4** does not block but leaves the new rule visibly unenforced by the repository's own hooks. **F5** and **F6** are corrections, not objections.

Rul1an:
Converted to draft at `b1dc8f19893d676fef333e1b8b5612cc2de9f98f`. Not a third revision of the same paragraph: two of the findings say the change is aimed one level too low, and I verified both myself rather than accepting them.

## Why draft rather than another push

**F2 is the reason.** `AGENTS.md:49` is two sentences — "Do not share `target/` directories between active worktrees. **Use `CARGO_TARGET_DIR`.**" My paragraph paraphrased the first and asserted consistency, while its default — do not set the variable — contradicts the second, inside a file whose own header defers to `AGENTS.md`.

That is not a wording defect. Either the recommendation is wrong, or `AGENTS.md`'s second sentence should change. The second is a governance decision with its own pull request and its own reviewer, and it cannot be smuggled in through a `CLAUDE.md` paragraph. The previous revision was silently inconsistent; this one *claims* consistency, which is worse, and the reviewer was right to say so.

**F3 makes the change underpowered even if F2 were resolved.** `docs/guides/agent-golden-path.md:64` exports `CARGO_TARGET_DIR="$(mktemp -d …/assay-target-2154.XXXXXX)"` — a target per *invocation*, which is the behaviour this change exists to stop, prescribed in a live guide. Documenting a default in `CLAUDE.md` while another document instructs the opposite changes nothing. The earlier pass flagged this and I let it stand without comment; that was my omission, not an oversight in the finding.

## The other findings, checked

**F1 is a real thinking error and worth naming precisely.** My premise covered targets still being built against; my conclusion was about abandoned targets from merged pull requests — which the reaper does handle, after three days. The honest mechanism is that the reaper's window is longer than the rate at which review rounds create full targets, not that they never age out. Across two revisions I held the conclusion fixed and fitted a second premise underneath it.

**F4** — `scripts/precommit/cargo-clippy.sh:57` and `scripts/ci/phase5-check.sh:9` both set one path shared across worktrees, which "one path per worktree" forbids. I removed the citation instead of the conflict.

**F5** — `.gitignore:37`, not 29. My line number came from the main checkout on a different branch; the same wrong-ref mistake I made with the earlier `git grep`.

## What survived

The core claim, attacked directly and twice: an in-tree build leaves `git status --porcelain` empty, verified by execution with both `cargo build` and `cargo test`, an unchanged `Cargo.lock`, and 24 passing contract tests. Every factual component of the `tmp_cleaner` sentence held, including that the three-day window is not configurable and that atime does advance. The recommendation is sound; its justification and its reach are not.

## What has to happen first

1. Decide, in `AGENTS.md`, whether per-worktree isolation is satisfied by the worktree's own `target/` or whether `CARGO_TARGET_DIR` remains mandatory. That decision governs everything else here.
2. Fix `docs/guides/agent-golden-path.md:64` and reconcile the two shared-path scripts, or state why each is an exception.
3. Only then is a `CLAUDE.md` line worth writing, and it will be shorter because it will have something to point at.

I will record this on #2202 so the issue carries the corrected scope rather than the version I first wrote.

Rul1an:
## Review note: #2205 merged first, and it rewrote the rule this PR cites

Read-only review from another writer's branch; I have not touched this one.

#2205 merged as `360c4963a` and is what `main` currently points at. The two PRs do **not** conflict — #2205 changed `AGENTS.md` and `docs/guides/agent-golden-path.md`, this one changes `CLAUDE.md` — and they agree substantively: both say to build in the worktree's own `target/`. So there is nothing to reconcile in the guidance itself.

One line here has gone stale, though, and it is the load-bearing one:

> That already satisfies `AGENTS.md`'s rule against sharing a target between active worktrees, so a scratch path is not required to honour it.

`AGENTS.md` no longer has a rule "against sharing". #2205 replaced it:

```diff
-- Do not share `target/` directories between active worktrees. Use `CARGO_TARGET_DIR`.
+- Build in the worktree's own `target/`. Cargo creates one per worktree on first build, so
+  worktrees do not share one unless told to, and `/target` is git-ignored so an in-tree build
+  leaves a read-only review's tree clean. A `CARGO_TARGET_DIR` you export by hand, you remove when
+  the work ends.
```

Two consequences for this PR, which is `BEHIND` and needs an update anyway:

1. The citation now describes wording that is gone. More than cosmetic: the sentence's rhetorical work is to argue that an in-tree build *also* satisfies a rule that seemed to require `CARGO_TARGET_DIR`. That tension no longer exists, because the merged rule now says to build in the worktree's own target directly. The argument is answering a rule nobody holds.
2. The first sentence's rationale — `/target` is git-ignored, so an in-tree build leaves a read-only review's tree clean — is now stated in `AGENTS.md` itself. Restating it in `CLAUDE.md` puts one rule in two documents, which is the drift shape `AGENTS.md` warns about. `CLAUDE.md`'s own preamble says it "adds Assay-specific technical context and does not override that contract", so the version that survives an update is probably a pointer to the merged rule plus only what `AGENTS.md` does not say.

What `AGENTS.md` does **not** say, and what looks like this PR's remaining unique content, is the operational part: keep one `CARGO_TARGET_DIR` per worktree and never one per reviewed head, with the macOS `/usr/libexec/tmp_cleaner` detail explaining why per-head paths accumulate — three-day untouched threshold, empty directories only, so a target being built against never ages out while its PR is open. That is concrete, checkable, and stated nowhere else. It is worth keeping on its own rather than as support for a citation that no longer resolves.

Rul1an:
Closing without merging. The decision this pull request argued for is now in `AGENTS.md` via #2205, and repeating it here would be the defect this repository names "one rule, one function".

## Why not revive it

`CLAUDE.md`'s own header says it "adds Assay-specific technical context and does not override" `AGENTS.md`. The rule that landed — build in the worktree's own `target/`; a `CARGO_TARGET_DIR` you export by hand, you remove when the work ends — is a cross-agent working rule, not Assay-specific technical context. A second statement of it in `CLAUDE.md` would create a place for it to drift, in a file that disclaims authority over it.

Measured on `origin/main` at `360c4963`: `CLAUDE.md` says nothing about target directories at all, so there is nothing there that contradicts the new rule and nothing to reconcile.

## What this pull request was actually worth

Not the paragraph. Two adversarial passes here established, by execution rather than by reading `.gitignore`, that an in-tree build leaves `git status --porcelain` empty — with `cargo build` and `cargo test`, an unchanged `Cargo.lock`, and 24 passing contract tests. That is the load-bearing fact under the merged rule, and it was attacked directly twice and held both times.

The passes also caught three false claims of mine before they could ship: that macOS does not reap `/tmp`, on evidence that was void because `/etc/periodic` does not exist here; that the per-head form had no source, from a grep run against the wrong ref; and a `.gitignore` line number taken from a different branch.

## Disposition of what stays open

Step 2 of #2202 closed with #2205: the golden-path guide's per-invocation `mktemp` export is gone from `main`, and the final rule is silent about script-set paths rather than permissive, so the two shared-path scripts need no exception. Step 3 is declined for the reason above, and recorded as declined rather than pending.

The branch is left in place; nothing on it is worth carrying forward.

Issue #2177, opened 2026-08-09
ax(cli): migrate remaining command failures to structured agent envelopes


Rul1an:
Parent: #1975 · Closed original instances: #2160, #2161, #2162, #2163, #2164, #2165, #2166 · Adjacent: #2167, #2168, #2201, #2439

## Current problem

The original 2026-08-09 measurement on Assay 5.0.0 is historical. The seven filed command contracts have landed, the shared `Summary` renderer and `CliFailure` funnel exist, PR #2437 added `assay ci --format json`, and PR #2438 classified two driven legacy coverage JSON failures.

The remaining P1 is a regression-test gap on exact `origin/main` `f2a558d1ce16cc31d565ef6ed6a7773e81002339`: `json_format_reports_failures_on_stdout.rs` drives representative `run`, `ci`, and coverage failures, but no explicit expected-set contract states which gate commands must remain covered. A row can disappear without making the test fail.

`assay validate` already emits `assay.validate_report.v1` with diagnostic codes and suggested actions. Whether every machine report must also carry top-level `reason_code` and `next_step` is a schema-shape decision, not a zero-byte-stdout defect, and stays outside this bounded residual.

## Completed slices

- [x] PR #2437: `assay ci --format json` publishes the existing `assay.run_summary.v1` document for early and completed paths, preserves `summary.json` and exit semantics, and leaves default/text stdout empty.
- [x] PR #2438: two legacy `assay coverage --format json` argument failures publish exit 2 / `E_INVALID_ARGS` Summary documents on stdout; text mode remains stdout-clean and `--input` mode remains on its file-sink contract. This is explicitly 2 of 11 known legacy early/error paths, not completeness.

## Remaining acceptance criteria

- [ ] One table-driven process contract names the intended gate-command set explicitly; deleting a command row fails an expected-set assertion.
- [ ] Every row drives a deterministic non-zero command path under its advertised JSON format and asserts non-empty, parseable stdout.
- [ ] Summary-envelope rows pin `schema`, `reason_code`, `next_step`, and exit identity; command-specific reports pin their documented schema and actionable diagnostic fields.
- [ ] Text/default controls keep stdout free of machine JSON unless the command already documents otherwise.
- [ ] Deliberate stdout sink failures are excluded from this producer test. #2439 owns the correct exit-3/zero-delivered-byte behavior when a machine document cannot be written.
[...]

Issue #2160, opened 2026-08-08
Doctor JSON failure lacks the golden-path reason and next step


Rul1an:
Parent: #2177 · Origin: #2154

## Driven observation

Measured on `18abbebf025e4a9243b1ec80f798db5ccad2afed` with the built `assay` binary:

```text
assay doctor --format json --config <missing.yaml>
exit: 1
stdout: valid assay.doctor_report.v0 JSON
config_error.code: E_CFG_PARSE
reason_code: absent
next_step: absent
```

A caller restricted to stdout and the exit code can identify a parse failure, but cannot use the frozen top-level `reason_code` / `next_step` contract, and the exit is `1` rather than the frozen configuration/usage class `2`.

## Acceptance

- The JSON failure report exposes a registered top-level `reason_code` and non-empty `next_step`.
- The exit mapping is reconciled with the frozen external exit-code contract, or the exception is made explicit and tested.
- A binary-level test drives the failure and reads stdout plus the exit code only.

## Non-goal

No changes to the human text renderer in this issue.

Rul1an:
Reopened after exact-head review of PR #2175 found the tracker state contradicted current repository and binary evidence.

Measured on `8dc465185b6d314373830a1f153332c4dcb4c1f3`:

- `crates/assay-cli/src/diagnostics/report.rs` still exposes no top-level `reason_code` or `next_step`.
- `crates/assay-cli/tests/agent_golden_path_contract.rs` still pins this gap to #2160 and passes.
- No commit on `origin/main` references #2160.
- The generated machine contract and project skill therefore still describe the behavior as a gap.

The prior `CLOSED / COMPLETED` state had no implementation or disposition attached. This issue remains open until the acceptance criteria are implemented or an explicit contract decision supersedes them.

Rul1an:
Measured confirmation of the second half of this issue, the part the title does not name.

Head `55e8103029583a16af96e1eafe73025192d69cd2` (`origin/main`), own debug build, `rustc 1.96.0`, `NO_COLOR=1`, no TTY.

```
$ printf 'not: [broken\n' > bad.yaml
$ assay doctor --format json --config bad.yaml
exit=1   stdout=1480B (valid JSON)   stderr=0B

config_error = {"code": "E_CFG_PARSE",
                "message": "ConfigError: failed to parse YAML: unknown field `not` ..."}
```

Two findings, and they point in opposite directions.

[...]

PR #2204, opened 2026-08-10
fix(cli): encode config recovery as JSON argv

### LINKED ISSUE #2200: Encode E_CFG_PARSE recovery as argument-safe JSON argv
Parent: #2177 · Found while resolving PR #2198 review

## Driven defect

`ReasonCode::ECfgParse.next_step(context)` interpolates a caller-controlled config path into shell prose:

```text
Run: assay doctor --config {path}
```

Whitespace breaks the argument boundary. Shell metacharacters become executable if a consumer treats `next_step` as a command. PR #2198 moves `E_POLICY_PARSE` to platform-neutral JSON argv; leaving this arm unchanged makes the same public field polymorphic and preserves an already-shipped injection hazard.

## Acceptance

- Publish `E_CFG_PARSE` remediation as JSON argv using the same representation as `E_POLICY_PARSE`.
- A test parses the argv JSON and passes it directly to `Command::args`.
- Paths containing whitespace and shell metacharacters remain exactly one argv element.
- Audit all remaining `ReasonCode::next_step` arms and record why non-command prose does not need argv conversion.
- Update generated/public references through their owner, not by hand.

## Sequencing

Complete before the next #2177 child adds another dynamic recovery command. No new envelope field or schema identity.

Rul1an:
## Summary

- encode `E_CFG_PARSE` recovery as compact JSON argv instead of shell-interpolated prose
- use one private formatter for the dynamic `E_CFG_PARSE` and `E_POLICY_PARSE` command recoveries
- prove a path containing whitespace and shell metacharacters remains one argument and executes through `Command::args` without a shell
- align stderr/run-output parity and the hand-written AI/public contract references

Closes #2200
Parent: #2177

## Scope / DoD

- [x] `E_CFG_PARSE` publishes `Run argv: ["assay","doctor","--config",path]`
- [x] binary contract parses the JSON and passes `argv[1..]` directly to `Command::args`
- [x] hostile path remains exactly one argument and creates no shell sentinel
- [x] all `ReasonCode::next_step` arms audited: dynamic executable caller data uses JSON argv; dynamic trace guidance is prose; remaining command-like strings contain no caller data
- [x] generated golden-path output remains generator-owned and unchanged because #2160 still owns doctor typed output
- [x] no envelope field or schema identity changes

## Compatibility

This intentionally changes the `next_step` bytes for `E_CFG_PARSE` from `Run: ...` to `Run argv: <JSON array>`. A consumer matching the old shell string will stop rather than execute an injection-prone command; the schema and field type remain unchanged. Parse the JSON array and invoke it directly without a shell.

The existing `unwrap_or(E_CFG_PARSE)` fallback can still overstate config remediation for an unclassified error. That pre-existing classification issue is not hidden in this slice and is recorded on #2177: https://github.com/Rul1an/assay/issues/2177#issuecomment-5237917917

## Smoking Gun

```text
Run argv: ["assay","doctor","--config","cfg file;$(touch should-not-exist).yaml"]
```

The bounded test executes those parsed elements directly, reaches `doctor` (`Config Status: FAILED`), emits no Clap `Usage:`, and verifies `should-not-exist` was not created.

## Threat Model Delta

- blocks shell interpretation of caller-controlled config paths in published recovery guidance
- preserves argument boundaries across whitespace and shell metacharacters
- does not claim the config is trustworthy or that `doctor` repairs it
- does not change classification, exit code, summary schema, or generated golden-path gaps
- malformed config remains fail-closed with exit 2; unavailable/unclassified behavior is not converted to success

## TDD / Mutation Proof

RED on the pre-change tree:

- unit test failed at `config recovery must publish JSON argv`
- bounded binary contract failed at the same missing prefix

Mutation after implementation: temporarily restoring the old `Run:` arm made both tests fail independently (`exit 101`); restoring the JSON argv arm made both green.

## Verification

Exact committed head: `11f9ad035d526b143032cd89bc35235e8d1d7fc7`
Toolchain: `rustc/cargo 1.96.0`

The verification below was first recorded on `4b993474e8a0`. Three commits followed; the re-run on the final head is in the [final-head verification comment](https://github.com/Rul1an/assay/pull/2204#issuecomment-5238794463).
Worktree: `/Users/roelschuurkes/.config/superpowers/worktrees/assay-2200-cfg-argv/worktree`
Toolchain: workspace Rust toolchain, `assay-cli 5.0.0`

```bash
CARGO_TARGET_DIR="$PWD/target" cargo test -p assay-cli
CARGO_TARGET_DIR="$PWD/target" cargo clippy -p assay-cli --all-targets -- -D warnings
CARGO_TARGET_DIR="$PWD/target" cargo package -p assay-cli --allow-dirty --no-verify
cargo fmt --all -- --check
bash scripts/ci/check-docs-generated-drift.sh
git diff --check
```

All pass on the exact head. Pre-commit and pre-push hooks also pass, including workspace clippy and the Linux cross-target gate.

## Workflow / AI Review

- real-binary user workflow: malformed config -> JSON failure -> parsed argv -> direct `doctor` invocation
- adversarial input: whitespace, semicolon, `$()`, and sentinel side-effect check
- agent workflow: JSON argv is deterministic and process-API ready; no shell reconstruction
- Claude Code Desktop, existing chat `Issue prioritering en overzicht`: advisory design review found no blockers, confirmed the generated non-change, requested the compact wire-format pin, and explicitly does not count toward formal quorum

## Review Focus

1. Does the shared formatter preserve the previously shipped compact `E_POLICY_PARSE` bytes?
2. Is the binary test genuinely shell-free and mutation-sensitive?
3. Are any remaining public E_CFG_PARSE shell examples still reachable?
4. Is the safe compatibility break described precisely enough for downstream agents?

## Summary by CodeRabbit

* **Bug Fixes**
  * Recovery commands now use safely encoded JSON argument arrays, preserving paths—including those with special characters—as single arguments.
  * Configuration and policy recovery actions no longer require shell interpretation, preventing unintended execution of path content.

* **Documentation**
  * Updated recovery guidance and output contracts to describe the new `Run argv` JSON format and direct process execution.

* **Tests**
  * Added coverage for malformed configurations, hostile paths, argument preservation, and end-to-end recovery execution.

### DIFF
--- a/crates/assay-cli/Cargo.toml
+++ b/crates/assay-cli/Cargo.toml
@@ -21,8 +21,9 @@ exclude = [
     "tests/e2e_mcp_wrap_assert_cmd.rs",
     # Workspace-only contract: reads generated docs and shared test support outside this package.
     "tests/agent_golden_path_contract.rs",
-    # Workspace-only contract: uses the shared bounded-process helper outside this package.
+    # Workspace-only contracts: use the shared bounded-process helper outside this package.
     "tests/policy_validate_machine_contract.rs",
+    "tests/config_recovery_argv_contract.rs",
 ]
 
 [lints]
--- a/crates/assay-cli/src/cli/commands/pipeline_error.rs
+++ b/crates/assay-cli/src/cli/commands/pipeline_error.rs
@@ -235,6 +235,14 @@ mod tests {
         )
     }
 
+    fn recovery_argv(step: &str) -> Vec<String> {
+        serde_json::from_str(
+            step.strip_prefix("Run argv: ")
+                .expect("command recovery must publish JSON argv"),
+        )
+        .expect("command recovery argv must parse")
+    }
+
     /// The property this issue exists to establish: an operator grepping stderr and
     /// a CI job parsing run.json see the same code for the same failure.
     #[test]
@@ -275,8 +283,8 @@ mod tests {
         let config = RunError::config_parse(Some("suites/prod.yaml".to_string()), "bad indent");
         let diagnostic = diagnostic_for(&config, ReasonCode::ECfgParse);
         assert_eq!(
-            diagnostic.fix_steps[0],
-            "Run: assay doctor --config suites/prod.yaml"
+            recovery_argv(&diagnostic.fix_steps[0]),
+            ["assay", "doctor", "--config", "suites/prod.yaml"]
         );
         assert_eq!(
             outcome_as_written(&config, ReasonCode::ECfgParse).next_step,
@@ -302,8 +310,8 @@ mod tests {
         let run_error = RunError::config_parse(None, "could not locate a config");
         let diagnostic = diagnostic_for(&run_error, ReasonCode::ECfgParse);
         assert_eq!(
-            diagnostic.fix_steps[0],
-            "Run: assay doctor --config <config.yaml>"
+            recovery_argv(&diagnostic.fix_steps[0]),
+            ["assay", "doctor", "--config", "<config.yaml>"]
         );
         assert_eq!(
             outcome_as_written(&run_error, ReasonCode::ECfgParse).next_step,
@@ -440,10 +448,19 @@ mod tests {
             .fix_steps
             .iter()
             .any(|s| s.contains("Regenerate baseline")));
-        assert!(
-            folded.fix_steps.iter().any(|s| s.contains("assay doctor")),
-            "the reason's own next step survives too: {:?}",
-            folded.fix_steps
+        let own_recovery = folded
+            .fix_steps
+            .iter()
+            .find(|step| step.starts_with("Run argv: "))
+            .unwrap_or_else(|| {
+                panic!(
+                    "the reason's own next step survives too: {:?}",
+                    folded.fix_steps
+                )
+            });
+        assert_eq!(
+            recovery_argv(own_recovery),
+            ["assay", "doctor", "--config", "assay.yaml"]
         );
 
         let rendered = folded.format_plain();
--- a/crates/assay-cli/src/exit_codes.rs
+++ b/crates/assay-cli/src/exit_codes.rs
@@ -521,6 +521,31 @@ mod tests {
     }
 
     #[test]
+    fn dynamic_recovery_argv_round_trips_json_significant_paths() {
+        let path = "cfg \"quoted\"\\nested\nline\ttab\u{0007}.yaml";
+        let cases = [
+            (
+                ReasonCode::ECfgParse,
+                vec!["assay", "doctor", "--config", path],
+            ),
+            (
+                ReasonCode::EPolicyParse,
+                vec!["assay", "policy", "validate", "--input", path],
+            ),
+        ];
+
+        for (reason, expected) in cases {
+            let next_step = reason.next_step(Some(path));
+            let encoded = next_step
+                .strip_prefix("Run argv: ")
+                .expect("dynamic recovery must publish JSON argv");
+            let argv: Vec<String> =
+                serde_json::from_str(encoded).expect("recovery argv must remain valid JSON");
+            assert_eq!(argv, expected, "{reason:?} must preserve one path argument");
+        }
+    }
+
+    #[test]
     fn test_reason_code_display() {
         assert_eq!(
             format!("{}", ReasonCode::ETraceNotFound),
--- a/crates/assay-cli/src/exit_codes.rs
+++ b/crates/assay-cli/src/exit_codes.rs
@@ -106,6 +106,10 @@ pub enum ReasonCode {
     EArgSchema,
 }
 
+fn format_recovery_argv(args: &[&str]) -> String {
+    format!("Run argv: {}", serde_json::json!(args))
+}
+
 impl ReasonCode {
     /// Get the corresponding exit code for this reason, respecting version
     pub fn exit_code_for(&self, version: ExitCodeVersion) -> i32 {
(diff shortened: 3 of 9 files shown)


Rul1an:
## Final-head verification update

Exact head: `11f9ad035d526b143032cd89bc35235e8d1d7fc7`
Worktree: `/Users/roelschuurkes/.config/superpowers/worktrees/assay-2200-cfg-argv/worktree`
Toolchain: `rustc/cargo 1.96.0`

The final review finding was valid: the recovery subprocess bypassed the contract test's sanitized command builder. The fix makes both the initial failure and published recovery invocation use one bounded, env-scrubbed `assay(...)` path. A focused test pins case-insensitive `ASSAY_*` removal.

TDD / verification on this exact committed tree:

- RED: `assay_command_scrubs_assay_environment_case_insensitively` failed to compile because the shared builder did not exist.
- GREEN: `cargo test -p assay-cli --test config_recovery_argv_contract -- --nocapture` — 10 passed.
- `cargo test -p assay-cli` — pass.
- `cargo clippy -p assay-cli --all-targets -- -D warnings` — pass.
- `cargo fmt --all -- --check` — pass.
- `git diff --check HEAD^` — pass.
- Worktree clean.

Prior review/verification records are not carried across this push. Fresh independent review remains required on this SHA.

Rul1an:
## Non-building agent review — exact head `11f9ad035d526b143032cd89bc35235e8d1d7fc7`

Reviewer: Cursor (Opus 5). Non-building basis: this change and its plan
(`docs/superpowers/plans/2026-08-10-cfg-parse-json-argv.md`) were authored by Codex; I authored
neither the change nor any normative specification governing it. The GitHub account is shared
across agents, so the independence claimed here is at the agent level, not the account level.
That is why this is a SHA-bound comment rather than an `APPROVED` review record: GitHub refuses
self-approval on the authoring account.

**Verdict: approve for merge.** No blocking findings.

### What I verified independently, rather than reading the checklist

- **The audit claim is true.** The PR asserts every `ReasonCode::next_step` arm was audited. I
  re-derived it: exactly two arms place caller-controlled data in an executable command position,
  `ECfgParse` and `EPolicyParse`, and both now route through `format_recovery_argv`. `ETraceNotFound`
  interpolates a caller path but into prose (`Check trace file exists: {}`), not a command. Every
  other `Run: ` string is static with a `<test-id>` placeholder, so none carries an argument boundary.
  No `Run: ...{...}` interpolation survives anywhere in the crate.
- **One rule, one function is satisfied, and this PR is what created it.** Before this change
  `EPolicyParse` built its JSON argv inline while `ECfgParse` emitted shell prose — two answers to
  one question. `format_recovery_argv` is now the single site, which is the shape `AGENTS.md`
  requires rather than a parity test standing in for it.
- **Checks are bound to this head.** 53 success, 9 skipped, nothing pending or failed on
  `11f9ad03`, including `Build + Test (windows-latest)`, which matters here because the hostile
  fixture filename `cfg file;$(touch should-not-exist).yaml` uses only characters Windows permits.
  The 9 skips are not a blind spot: required check `CI` is the gate that maps skips against the
  scope outputs, and it passed.
- **Process bounds match the sibling contract.** `ProcessLimits::new(5s, 64 KiB, 64 KiB)` is
  identical to `policy_validate_machine_contract.rs`, so this adds a case to an existing convention
  instead of a second one.

### Findings

**1. Stale measurement provenance in the PR body (LOW, fixed).** The `Verification` block named
`4b993474e8a0` as the exact committed head while the head had advanced three commits. The
final-head comment was correct, but a reader of the body alone would attribute the verification to
the wrong tree — the attribution failure `AGENTS.md` describes, where the number is right and the
tree it names is not. Disposition: fixed by the reviewer in the PR body; the body now names
`11f9ad03` and links the final-head comment. No code change.

**2. The contract test hard-pins `doctor`'s exit code, and #2160 is about changing it (INFO,
recorded not fixed).** `config_parse_recovery_argv_contract` asserts
`recovered.status.code() == Some(1)`. Issue #2160 is scoped to `doctor` using the wrong exit class,
so that slice will have to update this assertion. That is correct behaviour for today and not a
defect here; it is worth carrying to #2160 so it arrives as an expected edit rather than a
surprise red test. I will note it on #2160.

### One observation, no action requested

The `should-not-exist` assertion is described in the body as the smoking gun for blocking shell
interpretation. It cannot fail on the caller side, because `Command::args` never invokes a shell —
so it does not prove the published guidance is safe against a consumer that uses one, and no test
here could. What it does cover is the callee: it would fail if `assay doctor` itself shelled out
with the config path, which is a real property worth pinning. The load-bearing evidence for the
issue is the `assert_eq!` that the hostile path survives as exactly one argv element. The test is
right; only the framing claims more than it can falsify.

Rul1an:
## Final-head review quorum

Exact head: `11f9ad035d526b143032cd89bc35235e8d1d7fc7`
Base reviewed: `4125d81b00847a7dc9a2276c9a7d0c44a247e569`

Bot disposition on this head:

- CodeRabbit declared its review limit while explicitly selecting the delta through this exact head. This is unavailable infrastructure, not a passing review.
- Copilot has no usable final-head review; its quota record is bound to an older commit and does not carry.

AGENTS.md fallback route therefore uses two independent non-building reviews:

- **Popper: READY**, no findings. Reviewed all nine files, issue acceptance, argv/escaping, direct process execution, env isolation, docs, package boundary and generated-doc drift. Focused tests and package-shape checks passed. Non-claims recorded.
- **Gauss: READY**, no findings. Independently reviewed all nine files, cross-platform source/fixture behavior, hostile JSON values, mutation sensitivity, command isolation and package exclusion. Pipeline parity (13) and binary/bounded-process contract (10) passed. Non-claims recorded.

Both reviews are read-only and bound only to `11f9ad035d526b143032cd89bc35235e8d1d7fc7`; neither reviewer implemented or designed the slice. No actionable findings remain.

Builder exact-head verification is recorded in the prior comment. Merge remains conditional on zero pending/failed required CI checks; a newly rerunning Criterion comparison is not treated as green until it completes.

Issue #2211, opened 2026-08-10
golden-path hardening harness re-implements the gate's issue-reference rule


Rul1an:
Origin: #2160 · Related: #2176, #2207

## Observed

Verified statically, by reading both implementations at
`e931cb285493d8070120e3080c7eefcb992e70d1` (branch `cursor/2160-doctor-json-contract`, PR #2207).
The harness was not executed, so the claim below is that the two rules differ in shape and in the
inputs they accept, not that a run disagreed.

The new allow case `append_existing_contract_evidence_issue`
(`scripts/ci/test-agent-golden-path-skill-hardening.sh:154-199`) picks an issue the shipped contract
already vouches for, rather than naming one — a deliberate improvement, and its own comment at
lines 162-163 says why: naming one "would state the allowed-issue set a second time, and it went
stale the moment a gap closed."

It then derives that set with its own regex, at line 172:

```python
re.findall(r"(?:#|/issues/)([0-9]+)", claim)
```

The gate it exercises computes the same set in `contract_issue_numbers`
(`scripts/ci/test-agent-golden-path-skill.py:249-284`), which routes through `issue_references`
(lines 230-234) and the stricter `ISSUE_REFERENCE_PATTERN` (lines 120-123):

```python
r"(?<!\]\()#(?P<short>[0-9]+)(?![A-Za-z0-9_-])"
r"|/issues/(?P<url>[0-9]+)(?![A-Za-z0-9_-])"
```

The gate's pattern refuses a Markdown link target via the negative lookbehind `(?<!\]\()` and
refuses a numeric prefix of a longer token via the trailing `(?![A-Za-z0-9_-])`. The harness copy
does neither.

They agree on today's contract. The defect is shape rather than outcome: commit `e931cb28` exists
partly to stop restating the allowed-issue set, and it replaced one restatement with a second
implementation of the rule in another language. That is the drift the `AGENTS.md`
one-rule-one-function section describes, where "the second implementation starts as a faithful
approximation and diverges one edit at a time".

## Why this is LOW and fails closed

The divergence direction is safe. The harness pattern is the looser of the two, so it can only
select an issue number the gate would reject — never admit something the gate would accept. A
`non_claims` entry containing a Markdown anchor like `](#2160)` would be extracted by the harness,
written into the skill, and then rejected by the gate, so the allow case fails and CI goes red on a
contract that is actually fine. A false CI failure, not a missed violation. The gate's own behaviour
[...]

PR #2218, opened 2026-08-10
ci: state the toolchain and the cargo-deny version instead of inferring them

### LINKED ISSUE #2214: Lint job installs stable but cargo install selects the pinned 1.96.0, reddening PRs at random
## Driven observation

The `Lint (pre-commit)` job in `.github/workflows/kernel-matrix.yml` fails non-deterministically. It reddened PR #2213 at head `200378965af82567c8ab3006d6950d2b14316b2f` while the *same branch* had been green four minutes earlier:

```text
2026-08-10T11:48  failure  pull_request  cursor/2199-evidence-integrity-reason
2026-08-10T11:43  success  pull_request  cursor/2199-evidence-integrity-reason
2026-08-10T11:47  success  pull_request  cursor/2160-doctor-json-contract
```

Failing step: `Install cargo-deny (for pre-commit hook)`. Job [93443840963](https://github.com/Rul1an/assay/actions/runs/31385161830/job/93443840963):

```text
error: 'cargo' is not installed for the toolchain '1.96.0-x86_64-unknown-linux-gnu'.
##[error]Process completed with exit code 1.
```

## Root cause

Two facts about the same job disagree:

- `.github/workflows/kernel-matrix.yml:66-69` installs **stable** — the log confirms `rustc 1.97.1 (8bab26f4f 2026-07-14)`, with `rustfmt, clippy`.
- `rust-toolchain.toml` pins **1.96.0**.

`cargo install --locked cargo-deny` (line 80) runs inside the repository, so rustup honours `rust-toolchain.toml` and selects 1.96.0 — a toolchain the job never installed. Whether the step succeeds then depends on whether the runner image happens to ship 1.96.0 *with cargo*, which is why the same branch flips between green and red with no code change.

This is the exact failure mode `AGENTS.md` names under version pinning:

> an invocation can select a different toolchain through an alias while the log still names the pinned one

Here the log names 1.97.1 and the invocation took 1.96.0. Reading the log gives you the wrong answer about what ran, which is why this went unnoticed until the runner image stopped covering for it.

## Why this is MVP-blocking

It is not a slow check or a flaky test; it is a **required check that can redden any PR at random**. Every MVP PR in the queue inherits a coin flip, and the failure text points at a dependency install rather than at the toolchain mismatch, so each occurrence costs a fresh investigation. It also trains readers to re-run red checks, which is how a real failure gets waved through.

## Second defect in the same area

`cargo-deny`'s version is not pinned at all, and it is installed independently in two places:

- `.github/workflows/ci.yml:160` — `cargo install --locked cargo-deny`
- `.github/workflows/kernel-matrix.yml:80` — `cargo install --locked cargo-deny`

Both resolve to whatever is latest at run time, so two jobs in the same PR can enforce two different rule sets, and a new `cargo-deny` release can fail CI on a PR that changed nothing related. `--locked` pins cargo-deny's *own* dependencies, not the version of cargo-deny being installed — a distinction that reads as a pin without being one. `AGENTS.md` asks for the version to live in one place that both the install and the invocation read, and for the run to echo it so a version claim is checkable against the log.

## Acceptance

- The `cargo install` in the lint job selects a toolchain that the job actually installed, and does so **explicitly** rather than relying on what the runner image happens to carry. The fix must remove the dependence on runner-image contents, not merely make the current image work.
- `cargo-deny`'s version is stated once and read by every install site, with the resolved version echoed in the run so the log can be checked against the intended pin.
- A negative control: the change is verified against the condition that produced the failure — a runner where the pinned toolchain has no `cargo` — rather than only against a run that happens to pass.

## Non-goals

- No change to the pinned toolchain in `rust-toolchain.toml`.
- No change to what `cargo deny` checks or to any hook's rule set.
- No relabelling, weakening, or removal of the required check to get a branch mergeable.

Rul1an:
Closes #2214.

## Provenance

Branch `cursor/2214-pin-cargo-deny`, head `2be4026ee` on top of `aaf57f94c`, worktree `/Users/roelschuurkes/wt-2214`, based on `main` at `360c4963a8f978513948349204e12fc735595730`. Every observation below was made in that worktree.

## The defect

The `Lint (pre-commit)` job reddened pull requests at random. Same branch, no code change:

```text
2026-08-10T11:48  failure  pull_request  cursor/2199-evidence-integrity-reason
2026-08-10T11:43  success  pull_request  cursor/2199-evidence-integrity-reason
```

Failing step `Install cargo-deny`, job [93443840963](https://github.com/Rul1an/assay/actions/runs/31385161830/job/93443840963):

```text
error: 'cargo' is not installed for the toolchain '1.96.0-x86_64-unknown-linux-gnu'
```

The job installs **stable** — its own log says `rustc 1.97.1` — but `cargo install` runs inside the repository, where rustup reads `rust-toolchain.toml` and selects **1.96.0**, which the job never installed. Whether the step worked therefore depended on what the runner image happened to carry.

Measured here rather than reasoned about:

```text
$ rustup show active-toolchain
1.96.0-aarch64-apple-darwin (overridden by '/Users/roelschuurkes/wt-2214/rust-toolchain.toml')

$ RUSTUP_TOOLCHAIN=stable rustup show active-toolchain
stable-aarch64-apple-darwin (overridden by environment variable RUSTUP_TOOLCHAIN)
```

This is the failure mode `AGENTS.md` names under version pinning — "an invocation can select a different toolchain through an alias while the log still names the pinned one." The log named 1.97.1 while 1.96.0 ran, so reading the log gave the wrong answer about what was selected. That is why it survived unnoticed until the image stopped covering for it.

## What changed

**The toolchain selection is now stated, not inferred.** `scripts/ci/install-cargo-deny.sh` exports `RUSTUP_TOOLCHAIN` (default `stable`, overridable) before installing, so the selection no longer depends on runner-image contents. The `cargo-audit` install in the same job gets the same treatment for the same reason — it had the identical latent failure, and fixing one would have left the other to fire later.

**The invocation had the same hole, and that is easy to miss.** `cargo deny check` re-enters the same toolchain resolution, so the check would have inherited exactly the failure the install step was fixed for. CI and the pre-commit hook now invoke the `cargo-deny` binary directly, which removes `cargo` from the path entirely.

**The version is pinned in one place.** Two workflows installed cargo-deny independently and unpinned, so two jobs in one pull request could enforce two different rule sets, and a new cargo-deny release could redden a PR that changed nothing related. `0.20.2` is what an unpinned install resolves today, so this freezes current behaviour rather than changing it.

**`--locked` is why the absence of a pin was invisible**, and the script says so in a comment, because this is the part a future reader will undo: `--locked` pins cargo-deny's *own dependency versions* from its published lockfile, not the version of cargo-deny being installed. It reads as a pin without being one. Anyone deleting `--version` because `--locked` looks sufficient reintroduces the drift.

**The run echoes the resolved version and fails if it differs from the pin**, so the number in the log is one that was checked rather than one that was printed.

## Verification

`RUSTUP_TOOLCHAIN` override: quoted above, both directions.

Version guard, both directions, using a stub `cargo` and `cargo-deny` first on `PATH` so no minutes-long install was needed:

```text
mismatch:  cargo-deny resolved: cargo-deny 0.19.0
           cargo-deny reports 'cargo-deny 0.19.0', which is not the pinned 0.20.2
           the pin lives in scripts/ci/install-cargo-deny.sh and is the only place to change it
           exit=1

match:     cargo-deny resolved: cargo-deny 0.20.2
           exit=0
```

The stub also confirms the argv is what it should be: `install --locked --version 0.20.2 cargo-deny`.

Other gates: `shellcheck` clean on the new script; `cargo-deny check --help` accepted, confirming the direct-invocation form is valid; the full pre-push hook set passed, including `shellcheck`, `cargo fmt --check`, workspace clippy with `-D warnings`, and the cross-target Linux compile gate.

## The second commit, and why it is separate

The first push was **refused by a gate**, which is worth showing rather than folding away:

```text
inline run anchor is not unique: 'run: cargo install --locked cargo-deny'
```

The hardening harness's inline-run parser probe anchors on that exact line in `kernel-matrix.yml` and requires exactly one occurrence. Replacing the install with a script made the count zero, so the probe's guard fired. The guard did its job — it refused to run a probe whose subject had disappeared, instead of passing vacuously — and only the anchor needed to move.

Recorded because the shape recurs: the probe needs *an* inline run step in the lint job, not this specific command, so pinning the literal text lets an unrelated workflow edit block a push. This is the second anchor in this harness to break on a legitimate change today. #2211 already covers the theme of the harness restating what it could derive; generalising this probe belongs there rather than here.

## What this does not prove

The negative control I could not run is the real one: a runner where the pinned toolchain has no `cargo`. The stub proves the version guard and `rustup show` proves the override, but the original failure was environmental, and the only complete proof is CI staying green across runs that would previously have flipped. Judge this change on the mechanism, and treat one green run as consistent with the fix rather than as evidence the flake is gone.

No change to `rust-toolchain.toml`, to what `cargo deny` checks, or to any hook's rule set. No required check weakened, relabelled, or removed.

## Summary by CodeRabbit

* **Chores**
  * Standardized dependency checks on a pinned, verified cargo-deny version.
  * Improved CI consistency by using the stable Rust toolchain by default.
  * Updated pre-commit and workflow checks to use the standalone cargo-deny command.
  * Added validation to detect unexpected cargo-deny versions during installation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,6 +152,23 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    env:
+      # Stated once for the whole job, because every cargo command below runs inside the repository,
+      # where rustup reads `rust-toolchain.toml` and selects 1.96.0 -- a toolchain this job never
+      # installs, so whether a step works depends on what the runner image happens to carry:
+      #
+      #   error: 'cargo' is not installed for the toolchain '1.96.0-x86_64-unknown-linux-gnu'
+      #
+      # Both installs and both `cargo audit` invocations took that resolution, so stating it per
+      # step would write `stable` once per cargo command; the job is the smallest scope that covers
+      # them all with one statement. `install-cargo-deny.sh` honours an inherited value, so this is
+      # the selection there too rather than being shadowed by the script's own default.
+      #
+      # Scope, deliberately: no step in this job builds or tests the workspace -- checkout, two tool
+      # installs, a `sed`-only alignment check, `cargo-deny check` and two `cargo audit` runs. This
+      # selects the toolchain that runs the dependency tooling, never one that compiles a crate, and
+      # `rust-toolchain.toml` stays the source for every job that does compile.
+      RUSTUP_TOOLCHAIN: stable
     steps:
       - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
         with:
@@ -159,16 +176,15 @@ jobs:
       - name: Install cargo-deny
         run: ./scripts/ci/install-cargo-deny.sh
       - name: Install cargo-audit
-        # Same toolchain selection as the cargo-deny install, and for the same reason: this job
-        # installs no toolchain, so without RUSTUP_TOOLCHAIN rustup reads rust-toolchain.toml and
-        # picks 1.96.0, which the runner image is not required to carry a cargo for.
-        run: RUSTUP_TOOLCHAIN=stable cargo install --locked cargo-audit
+        run: cargo install --locked cargo-audit
       - name: Check Aya/eBPF dependency alignment
         run: ./scripts/ci/check-aya-ebpf-alignment.sh
       - name: cargo deny (advisories, licenses, bans, sources)
-        # `cargo-deny` rather than `cargo deny`: invoking it through cargo re-enters the same
-        # toolchain resolution that broke the install, so the check would inherit the failure the
-        # install step was just fixed for. The binary is the one the pinned install asserted.
+        # `cargo-deny` rather than `cargo deny`: invoking it through cargo re-enters the toolchain
+        # resolution that broke the install. The job-level RUSTUP_TOOLCHAIN now states that
+        # resolution, so this form no longer carries the job on its own; it is kept because it does
+        # not depend on that variable surviving a later edit. The binary is the one the pinned
+        # install asserted.
         run: cargo-deny check advisories bans licenses sources
       - name: cargo audit (RustSec)
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,13 +157,19 @@ jobs:
         with:
           persist-credentials: false
       - name: Install cargo-deny
-        run: cargo install --locked cargo-deny
+        run: ./scripts/ci/install-cargo-deny.sh
       - name: Install cargo-audit
-        run: cargo install --locked cargo-audit
+        # Same toolchain selection as the cargo-deny install, and for the same reason: this job
+        # installs no toolchain, so without RUSTUP_TOOLCHAIN rustup reads rust-toolchain.toml and
+        # picks 1.96.0, which the runner image is not required to carry a cargo for.
+        run: RUSTUP_TOOLCHAIN=stable cargo install --locked cargo-audit
       - name: Check Aya/eBPF dependency alignment
         run: ./scripts/ci/check-aya-ebpf-alignment.sh
       - name: cargo deny (advisories, licenses, bans, sources)
-        run: cargo deny check advisories bans licenses sources
+        # `cargo-deny` rather than `cargo deny`: invoking it through cargo re-enters the same
+        # toolchain resolution that broke the install, so the check would inherit the failure the
+        # install step was just fixed for. The binary is the one the pinned install asserted.
+        run: cargo-deny check advisories bans licenses sources
       - name: cargo audit (RustSec)
         run: |
           rm -rf ~/.cargo/advisory-db
--- a/.github/workflows/kernel-matrix.yml
+++ b/.github/workflows/kernel-matrix.yml
@@ -77,7 +77,7 @@ jobs:
           pip install pre-commit==4.4.0
 
       - name: Install cargo-deny (for pre-commit hook)
-        run: cargo install --locked cargo-deny
+        run: ./scripts/ci/install-cargo-deny.sh
 
       - name: Run pre-commit (all files)
         shell: bash
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -293,7 +293,12 @@ repos:
       # --- Dependency security (shift-left) ---
       - id: cargo-deny
         name: cargo deny (advisories, licenses, bans, sources)
-        entry: bash -lc 'cargo deny check advisories bans licenses sources'
+        # `cargo-deny` rather than `cargo deny`: going through cargo resolves the toolchain from
+        # rust-toolchain.toml, which in CI selects a toolchain the lint job never installed. The
+        # binary runs the same checks without that resolution. No version assertion here -- the pin
+        # is enforced where CI installs it, and asserting it in a local hook would fail every
+        # developer whose cargo-deny predates the pin.
+        entry: bash -lc 'cargo-deny check advisories bans licenses sources'
         language: system
         pass_filenames: false
         # Only run when deps/policy change (avoids noise on unrelated commits)
--- a/scripts/ci/install-cargo-deny.sh
+++ b/scripts/ci/install-cargo-deny.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+# Install cargo-deny at the pinned version, from a toolchain this job actually has.
+#
+# The version lives here and nowhere else. Two workflows install cargo-deny -- the dependency
+# security job in ci.yml and the pre-commit lint job in kernel-matrix.yml -- and before this script
+# each ran its own `cargo install --locked cargo-deny`, resolving whatever was latest at run time.
+# Two jobs in one pull request could therefore enforce two different rule sets, and a new cargo-deny
+# release could redden a pull request that changed nothing related.
+#
+# `--locked` did not prevent that, and it is worth being explicit about why, because it reads like a
+# pin: it pins cargo-deny's *own dependency versions* from its published lockfile, not the version
+# of cargo-deny being installed. Anyone removing `--version` below because `--locked` looks
+# sufficient will reintroduce the drift.
+set -euo pipefail
+
+CARGO_DENY_VERSION="0.20.2"
+
+# Select the toolchain explicitly rather than letting rustup choose.
+#
+# Both jobs run inside the repository, where `rust-toolchain.toml` pins 1.96.0. rustup honours that
+# file, so `cargo install` selected 1.96.0 -- a toolchain neither job installs. The lint job installs
+# `stable` (observed as 1.97.1 in its own log) and the dependency job installs nothing at all, so
+# whether the step worked depended on what the runner image happened to carry:
+#
+#   error: 'cargo' is not installed for the toolchain '1.96.0-x86_64-unknown-linux-gnu'
+#
+# The same branch passed at 11:43 and failed at 11:48 on 2026-08-10 with no code change. The log
+# named 1.97.1 while 1.96.0 ran, which is why reading the log gave the wrong answer about what was
+# selected. RUSTUP_TOOLCHAIN overrides the file, so the selection is now stated rather than inferred.
+export RUSTUP_TOOLCHAIN="${CARGO_DENY_TOOLCHAIN:-stable}"
+
+echo "installing cargo-deny ${CARGO_DENY_VERSION} using toolchain ${RUSTUP_TOOLCHAIN}"
+cargo install --locked --version "${CARGO_DENY_VERSION}" cargo-deny
+
+# Echo the resolved version and hold it against the pin, so the number in the log is one that was
+# checked rather than one that was printed. An install that silently produced a different version
+# would otherwise leave a log that looks like evidence of the pin.
+resolved="$(cargo-deny --version)"
+echo "cargo-deny resolved: ${resolved}"
+if [[ "${resolved}" != "cargo-deny ${CARGO_DENY_VERSION}" ]]; then
+  echo "cargo-deny reports '${resolved}', which is not the pinned ${CARGO_DENY_VERSION}" >&2
+  echo "the pin lives in scripts/ci/install-cargo-deny.sh and is the only place to change it" >&2
+  exit 1
+fi
--- a/scripts/ci/install-cargo-deny.sh
+++ b/scripts/ci/install-cargo-deny.sh
@@ -27,7 +27,13 @@ CARGO_DENY_VERSION="0.20.2"
 # The same branch passed at 11:43 and failed at 11:48 on 2026-08-10 with no code change. The log
 # named 1.97.1 while 1.96.0 ran, which is why reading the log gave the wrong answer about what was
 # selected. RUSTUP_TOOLCHAIN overrides the file, so the selection is now stated rather than inferred.
-export RUSTUP_TOOLCHAIN="${CARGO_DENY_TOOLCHAIN:-stable}"
+#
+# An inherited value wins. ci.yml's dependency job states the selection once for every cargo command
+# it runs; overriding that here would make the job's statement decorative for this step, and would
+# make `stable` a value written in two places that have to agree -- the defect this script exists to
+# remove. `stable` below is the fallback for a caller that states nothing, which is the lint job in
+# kernel-matrix.yml and anyone running the script by hand.
+export RUSTUP_TOOLCHAIN="${RUSTUP_TOOLCHAIN:-stable}"
 
 echo "installing cargo-deny ${CARGO_DENY_VERSION} using toolchain ${RUSTUP_TOOLCHAIN}"
 cargo install --locked --version "${CARGO_DENY_VERSION}" cargo-deny
--- a/scripts/ci/test-agent-golden-path-skill-hardening.sh
+++ b/scripts/ci/test-agent-golden-path-skill-hardening.sh
@@ -1256,7 +1256,7 @@ import os
 from pathlib import Path
 
 path = Path(os.environ["CASE_ROOT"]) / ".github/workflows/kernel-matrix.yml"
-source = "run: cargo install --locked cargo-deny"
+source = "run: ./scripts/ci/install-cargo-deny.sh"
 replacement = "run: echo inline-parser-sentinel"
 text = path.read_text(encoding="utf-8")
 if text.count(source) != 1:


Rul1an:
# Independent adversarial review — head `2be4026ee2bda84fc11b1e5e17c8fc4ffefa57d9`

I did not author this change and did not author issue #2214. Review performed in
`/Users/roelschuurkes/wt-2214` with `CARGO_TARGET_DIR=/tmp/assay-review-2218-target`.
Head verified at start and at end:

```text
$ git -C /Users/roelschuurkes/wt-2214 rev-parse HEAD
2be4026ee2bda84fc11b1e5e17c8fc4ffefa57d9
$ git status --porcelain     # after the whole mutation battery
(empty)
$ git diff --exit-code >/dev/null && echo "diff clean"
diff clean
```

All findings below apply to `2be4026ee2bda84fc11b1e5e17c8fc4ffefa57d9`.

## Verdict: REQUEST CHANGES

Two blocking findings, both about completeness rather than about the mechanism. The mechanism is
right, the diagnosis is right, and I confirmed both from primary evidence. What is wrong is that
the same PR that establishes "invoking a cargo plugin through `cargo` re-enters the toolchain
resolution that broke the install" leaves one such invocation in the *same job it edited*, and
leaves five more install sites elsewhere in `.github/workflows/` without saying so.

---

## BLOCKING 1 — `.github/workflows/ci.yml:177`: `cargo audit` is still on the path this PR calls broken

The PR removes `cargo` from the deny path in the `deps-security` job:

```yaml
# .github/workflows/ci.yml:168-172
      - name: cargo deny (advisories, licenses, bans, sources)
        # `cargo-deny` rather than `cargo deny`: invoking it through cargo re-enters the same
        # toolchain resolution that broke the install, so the check would inherit the failure the
        # install step was just fixed for. The binary is the one the pinned install asserted.
        run: cargo-deny check advisories bans licenses sources
```

Five lines later, in the same job, unchanged:

```yaml
# .github/workflows/ci.yml:173-177
      - name: cargo audit (RustSec)
        run: |
          rm -rf ~/.cargo/advisory-db
          # RUSTSEC-2023-0071 (rsa Marvin Attack): no fixed upgrade; assay-mcp-server JWT (see deny.toml)
          cargo audit --ignore RUSTSEC-2023-0071
```

This runs from the repository root, so it takes exactly the resolution the quoted comment
describes. I checked this rather than reasoning about it. Negative control: a directory whose
`rust-toolchain.toml` names a toolchain that cannot supply `cargo`, which is the condition the
issue says was never reproduced:

```text
$ cat /tmp/negctl/rust-toolchain.toml
[toolchain]
channel = "no-cargo-probe"

$ cd /tmp/negctl && cargo install --locked cargo-deny        # the pre-PR install form
error: custom toolchain 'no-cargo-probe' specified in override file '/private/tmp/negctl/rust-toolchain.toml' is not installed
exit=1

$ cargo deny check --help >/dev/null                          # the pre-PR invocation form
error: custom toolchain 'no-cargo-probe' ... is not installed
exit=1

$ cargo audit --version                                       # STILL IN ci.yml:177
error: custom toolchain 'no-cargo-probe' ... is not installed
exit=1

$ RUSTUP_TOOLCHAIN=stable cargo --version                     # what the new script exports
cargo 1.92.0 (344c4567c 2025-10-21)
exit=0

$ cargo-deny --version                                        # the new invocation form
cargo-deny 0.19.0
exit=0

$ cargo-audit audit --version                                 # the direct form for audit
cargo-audit-audit 0.22.0
exit=0
```

`cargo audit` fails under precisely the condition `cargo deny` was rescued from, and the direct
binary survives it. Note also that the *fuzz* audit at `ci.yml:197` is already safe, because it
runs `( cd "${RUNNER_TEMP}" && cargo audit ... )` outside the repository — so the change already
reasons about working-directory sensitivity two steps below the line it left alone.

Why it matters: the `deps-security` job is a required check. `1.96.0` is a real distributable
toolchain that rustup will usually auto-install, so this does not fail every time — it fails when
the runner image leaves 1.96.0 partial, which is exactly the coin-flip profile #2214 was opened
about. The PR moves the flake from the deny path to the audit path in the same job.

What I would accept: `cargo-audit audit --ignore RUSTSEC-2023-0071` (verified above that the
`cargo-audit audit` form parses), or `RUSTUP_TOOLCHAIN=stable` on that step. Either is one line.

## BLOCKING 2 — the remaining exposure is real and undisclosed

The PR body scopes the problem to "two workflows install cargo-deny". That is true of cargo-deny
and false of the failure mode. Every one of these is `cargo install` executed from the repository
root after `dtolnay/rust-toolchain@stable`, with no `RUSTUP_TOOLCHAIN`, i.e. byte-for-byte the
latent failure this PR fixed twice:

| File:line | Command | Fixed here? |
|---|---|---|
| `.github/workflows/ci.yml:160` | `./scripts/ci/install-cargo-deny.sh` | yes |
| `.github/workflows/ci.yml:165` | `RUSTUP_TOOLCHAIN=stable cargo install --locked cargo-audit` | toolchain yes, version no |
| `.github/workflows/kernel-matrix.yml:80` | `./scripts/ci/install-cargo-deny.sh` | yes |
| `.github/workflows/split-wave0-gates.yml:202` | `cargo install --locked cargo-nextest` | no |
| `.github/workflows/split-wave0-gates.yml:203` | `cargo install --locked cargo-hack` | no |
| `.github/workflows/split-wave0-gates.yml:396` | `cargo install --locked cargo-semver-checks` | no |
| `.github/workflows/release.yml:368` | `cargo install cargo-cyclonedx --version 0.5.9 --locked` | no |
| `.github/workflows/fuzz-smoke.yml:87` | `cargo install cargo-fuzz --locked --version 0.13.2` | no |

And the matching invocations, all cargo-proxied from the repo root:
`split-wave0-gates.yml:238,271,289,307` (`cargo nextest`), `:240,272,291` (`cargo hack`),
`:461` (`cargo semver-checks`), `scripts/ci/optional-public-api-drift.sh:64` (`cargo semver-checks`),
`scripts/ci/mutation-smoke-pure-modules.sh:12,25,28,31` (`cargo mutants`),
`scripts/ci/review-dependency-perf-hygiene-pr4.sh:36` (`cargo deny check advisories bans sources`),
`.pre-commit-config.yaml:318` (`cargo audit`).

I am not asking for all of them in this PR — the `split-wave0-gates` and `release` jobs are outside
#2214 and a bigger change deserves its own review. I am asking that the PR stop reading as if the
class is closed. `AGENTS.md` says never turn absence of evidence into a clean result, and a PR
titled "state the toolchain … instead of inferring them" that leaves six sites inferring it needs
to say so. This is blocking but cheap to clear: a paragraph in the PR body naming the residual
sites plus a follow-up issue would satisfy me. No code change required.

Related, same paragraph: the body says "The `cargo-audit` install in the same job gets the same
treatment for the same reason." It gets the *toolchain* treatment. It does not get the version pin,
the echo, or the assertion. A reader finishes that sentence believing cargo-audit is pinned; it
resolves to whatever is latest, which is the second half of the rationale the PR itself gives for
pinning cargo-deny ("a new release could redden a pull request that changed nothing related").
Please either pin it or narrow the sentence.

---

## Non-blocking findings

### N1 — `deny.toml:4` now states the invocation the PR just replaced

```text
$ sed -n '1,4p' deny.toml
# deny.toml - cargo-deny configuration
# Docs: https://embarkstudios.github.io/cargo-deny/
#
# Run: cargo deny check advisories bans licenses sources
```

This is the file a reader opens to learn how the check is run, and it now disagrees with both CI
and the hook. Under one-rule-one-function this is the cheap kind of second statement: it costs one
character to fix (`cargo-deny`). `scripts/ci/review-dependency-perf-hygiene-pr4.sh:36` has the same
issue in executable form; I found no CI or hook reference to that script, so its blast radius is a
human running it by hand.

### N2 — the script's own guarantees are enforced by nothing but a comment

`scripts/ci/install-cargo-deny.sh:10-13` warns that a future reader will delete `--version` because
`--locked` looks sufficient. I removed `--version` and every gate stayed green (mutation M2 below).
It stays green until upstream publishes 0.20.3, at which point the assertion starts firing and the
log will point at the install rather than at the deletion.

The verification for this already exists — the PR body's stub run — but it was not committed, so it
runs once, in a PR body, and never again. This repo already has the pattern for making that durable
(`scripts/ci/test-semver-gate.sh` is a self-test of a gate, wired in at `.pre-commit-config.yaml:330`).
A ~20-line self-test that puts a stub `cargo` and `cargo-deny` on `PATH` and asserts (a) argv
contains `install --locked --version 0.20.2 cargo-deny` and (b) a mismatching stub exits 1 would
convert three of my four silent mutations into caught ones. I reproduced the PR body's stub exactly,
so the material is already written:

```text
===== MATCH case (stub reports 0.20.2) =====
installing cargo-deny 0.20.2 using toolchain stable
STUB cargo argv: install --locked --version 0.20.2 cargo-deny
STUB RUSTUP_TOOLCHAIN=stable
cargo-deny resolved: cargo-deny 0.20.2
exit=0
===== MISMATCH case (stub reports 0.19.0) =====
cargo-deny resolved: cargo-deny 0.19.0
cargo-deny reports 'cargo-deny 0.19.0', which is not the pinned 0.20.2
the pin lives in scripts/ci/install-cargo-deny.sh and is the only place to change it
exit=1
```

### N3 — the assertion checks "a cargo-deny on PATH", not "the one just installed"

`install-cargo-deny.sh:38` resolves `cargo-deny` through `PATH`, with no relationship to where
`cargo install` wrote. In my MATCH run above, the stub `cargo` installed nothing at all and the
script still exited 0, because an unrelated `cargo-deny` reporting `0.20.2` was on `PATH`. On a
GitHub runner this needs `CARGO_INSTALL_ROOT` set, or `~/.cargo/bin` shadowed, so I rate it low.
`command -v cargo-deny` echoed alongside the version, or asserting against
`"${CARGO_HOME:-$HOME/.cargo}/bin/cargo-deny"`, closes it.

### N4 — the pre-commit asymmetry is live today, not hypothetical

The stated reason for omitting the assertion from the hook is that it "would fail every developer
whose cargo-deny predates the pin". I was one of them:

```text
$ cargo-deny --version
cargo-deny 0.19.0
```

So at the moment this PR was written, the hook and CI did enforce two different cargo-deny
versions — the condition the PR body names as the defect it is fixing, relocated from
CI-vs-CI to local-vs-CI. I accept the reasoning: a hard failure in a local hook is hostile, and CI
is the authority. But "not fatal" and "not mentioned" are different choices. A non-fatal
`warning: local cargo-deny is X, CI pins Y` would keep the hook usable and stop the drift being
invisible; it also gives the pin a second reader, which is the direction `AGENTS.md` asks for.
I checked that 0.20.2 is green on this tree, so the divergence is currently benign:

```text
$ cargo-deny --version && cargo-deny check advisories bans licenses sources
cargo-deny 0.20.2
advisories ok, bans ok, licenses ok, sources ok
```

### N5 — nit

`docs/superpowers/plans/2026-08-09-golden-path-durability.md:217` still instructs a reader to anchor
the probe on `run: cargo install --locked cargo-deny`. It is a historical plan, so this may be
deliberate; flagging only because it is written imperatively.

---

## Mutations

Gates run after each mutation: `shellcheck scripts/ci/install-cargo-deny.sh`,
`pre-commit run actionlint --all-files`, `pre-commit run agent-golden-path-skill-contract --all-files`
(a `pre-commit`-stage hook, so it is what CI's lint job runs), and
`bash scripts/ci/test-agent-golden-path-skill-hardening.sh` (reached only via the `pre-push`-stage
hook at `.pre-commit-config.yaml:136-143`). Every mutation was reverted with `git checkout -- .`
and the tree re-verified clean.

| # | Mutation | Caught by | Verdict |
|---|---|---|---|
| M1 | Revert the harness anchor at `test-agent-golden-path-skill-hardening.sh:1259` to `run: cargo install --locked cargo-deny` | hardening harness (exit 1) | caught — **local pre-push only** |
| M2 | Drop `--version "${CARGO_DENY_VERSION}"` from the `cargo install` line | nothing | **SILENT** |
| M3 | Replace the version comparison with `if false; then` | nothing | **SILENT** |
| M4 | Delete `export RUSTUP_TOOLCHAIN="${CARGO_DENY_TOOLCHAIN:-stable}"` | nothing | **SILENT** (reinstates the original defect) |
| M5 | `ci.yml:172` back to `cargo deny check advisories bans licenses sources` | nothing | **SILENT** (reinstates the original defect) |
| M6 | `kernel-matrix.yml:80` back to inline `cargo install --locked cargo-deny` | hardening harness (exit 1) | caught — **local pre-push only** |
| M7 | Delete `scripts/ci/install-cargo-deny.sh` while both workflows still call it | no gate; repo `shellcheck` hook still reports `Passed` | **SILENT locally**, deterministic CI failure at runtime |
| M8 | Change the pin to `0.19.0` | nothing | **SILENT** (expected — single source of truth, nothing asserts *which* version) |
| M9 | `.pre-commit-config.yaml:301` back to `cargo deny check …` | nothing | **SILENT** (reinstates the original defect) |

Raw results:

```text
RESULT|M1 revert harness anchor to old literal|shellcheck=0|actionlint=0|golden-path-contract(CI stage)=0|hardening-harness(pre-push)=1
RESULT|M2 drop --version from cargo install|shellcheck=0|actionlint=0|golden-path-contract(CI stage)=0|hardening-harness(pre-push)=0
RESULT|M3 make version comparison always pass|shellcheck=0|actionlint=0|golden-path-contract(CI stage)=0|hardening-harness(pre-push)=0
RESULT|M4 remove RUSTUP_TOOLCHAIN export|shellcheck=0|actionlint=0|golden-path-contract(CI stage)=0|hardening-harness(pre-push)=0
RESULT|M5 ci.yml back to 'cargo deny check'|shellcheck=0|actionlint=0|golden-path-contract(CI stage)=0|hardening-harness(pre-push)=0
RESULT|M6 kernel-matrix back to inline cargo install|shellcheck=0|actionlint=0|golden-path-contract(CI stage)=0|hardening-harness(pre-push)=1
RESULT|M7 delete install-cargo-deny.sh (refs remain)|shellcheck=2|actionlint=0|golden-path-contract(CI stage)=0|hardening-harness(pre-push)=0
RESULT|M8 change the pin to 0.19.0|shellcheck=0|actionlint=0|golden-path-contract(CI stage)=0|hardening-harness(pre-push)=0
RESULT|M9 pre-commit hook back to 'cargo deny'|shellcheck=0|actionlint=0|golden-path-contract(CI stage)=0|hardening-harness(pre-push)=0
=== FINAL TREE STATE ===
diff clean
2be4026ee2bda84fc11b1e5e17c8fc4ffefa57d9
```

The `shellcheck=2` on M7 is my direct invocation failing to open a deleted file, not a gate; the
repo's own `shellcheck` hook reported `Passed` with the script gone.

Two structural conclusions from the table. First, M4, M5 and M9 each restore the exact defect this
PR exists to remove, and no gate anywhere notices — the fix is correct and entirely unprotected
against being undone. Second, the only two mutations that were caught were caught by a hook with
`stages: [pre-push]`, and no workflow in `.github/workflows/` runs `pre-commit --hook-stage pre-push`
(`rg -n "pre-push" .github/workflows/` returns nothing). The anchor guard the PR credits with
refusing the first push protects a local `git push`, not the repository.

## What held up under attack

- **The diagnosis is right, and I confirmed it from the primary log rather than from the PR body.**
  Job 93443840963 is a first attempt inside run 31385161830 (the run now reads green because it was
  re-run; `gh api .../runs/<id>/jobs` returns only the latest attempt, which is why the run looks
  clean). `gh api repos/Rul1an/assay/actions/jobs/93443840963/logs` gives, in order:

  ```text
  11:51:48.5108260Z   stable-x86_64-unknown-linux-gnu unchanged - rustc 1.97.1 (8bab26f4f 2026-07-14)
  11:51:48.5109266Z info: note that the toolchain '1.96.0-x86_64-unknown-linux-gnu' is currently in use (overridden by '/home/runner/work/assay/assay/rust-toolchain.toml')
  ...
  11:51:54.7044896Z ##[group]Run cargo install --locked cargo-deny
  11:51:54.7197542Z info: syncing channel updates for 1.96.0-x86_64-unknown-linux-gnu
  11:51:54.7899341Z info: downloading component clippy
  11:51:55.4689075Z error: 'cargo' is not installed for the toolchain '1.96.0-x86_64-unknown-linux-gnu'.
  ```

  rustup itself names the override file. This is not merely plausible, it is stated by the tool.
  I also confirmed no alternative explanation in the workflow: `dtolnay/rust-toolchain` sets
  `CARGO_HOME`, `CARGO_INCREMENTAL` and `CARGO_TERM_COLOR` into the environment and does **not** set
  `RUSTUP_TOOLCHAIN` (visible in every subsequent step's `env:` block), so nothing else was
  competing for the selection.
- **`RUSTUP_TOOLCHAIN` does override `rust-toolchain.toml`**, reproduced independently:
  ```text
  $ rustup show active-toolchain
  1.96.0-aarch64-apple-darwin (overridden by '/Users/roelschuurkes/wt-2214/rust-toolchain.toml')
  $ RUSTUP_TOOLCHAIN=stable rustup show active-toolchain
  stable-aarch64-apple-darwin (overridden by environment variable RUSTUP_TOOLCHAIN)
  ```
  I also checked the failure mode the rule warns about — a toolchain named but absent. rustup 1.29.0
  auto-installs it (`RUSTUP_TOOLCHAIN=1.70.0 cargo --version` → `info: syncing channel updates` →
  `cargo 1.70.0`), so `stable` being unavailable on a runner is not a new way for this to break.
- **The pin genuinely lives in one place.** I searched the whole tree for a second literal and there
  is none: `CARGO_DENY_VERSION="0.20.2"` at `install-cargo-deny.sh:16` is the only occurrence of the
  version, and both the install argument and the assertion string interpolate that variable rather
  than restating it. `deny.toml` carries no version. This is the part of `AGENTS.md` the change
  satisfies cleanly.
- **The version guard fails closed in every direction I attacked it in.** Binary absent → `line 38:
  cargo-deny: command not found`, exit 127. Version on stderr only → `resolved` empty → exit 1.
  Extra line before the version → exit 1. `cargo install` itself failing → exit 101 before the
  assertion is reached. `set -euo pipefail` behaves as intended on the command substitution at
  line 38. `shellcheck 0.11.0` is clean (exit 0).
- **The direct-binary form is valid and equivalent, including at the pinned version.** I installed
  0.20.2 and ran it against this tree: `advisories ok, bans ok, licenses ok, sources ok`.
  `cargo-deny check --help` and `cargo-deny deny check --help` print the same usage, so the binary
  tolerates both the bare and the cargo-passed argv. I found no `cargo deny` left in CI or in the
  hook.
- **`0.20.2` really is what an unpinned install resolves to today**, observed twice rather than
  asserted: the successful re-run compiled `cargo-deny v0.20.2` from the unpinned command, and my
  own `cargo install --locked cargo-deny` downloaded `cargo-deny v0.20.2`.
- **The anchor probe is not vacuous after the move.** `run: ./scripts/ci/install-cargo-deny.sh`
  occurs exactly once in `kernel-matrix.yml` and is inside the lint job; the probe's
  `if text.count(source) != 1` guard fires on both M1 and M6, and the assertion at line 1330 still
  requires exactly one retained `echo inline-parser-sentinel` step, so a stale anchor cannot pass
  silently. The probe needs *an* inline `run:` step in the lint job and it still has one. Baseline:
  `agent golden-path hardening: 86 case(s) observed; 22 structural probe(s) observed`, exit 0.
- **Gates I ran, all green at this head:** `cargo fmt --all -- --check` (exit 0),
  `shellcheck scripts/ci/install-cargo-deny.sh` (exit 0), `pre-commit run actionlint --all-files`
  (Passed), `pre-commit run shellcheck --all-files` (Passed),
  `pre-commit run agent-golden-path-skill-contract --all-files` (Passed),
  `bash scripts/ci/test-agent-golden-path-skill-hardening.sh` (exit 0).
- **On "what this does not prove":** the admission is honest, but it is more conservative than it
  needed to be. A *sufficient* negative control is runnable locally in about two minutes — point a
  `rust-toolchain.toml` in a scratch directory at a toolchain that cannot supply `cargo`, then show
  that every cargo-proxied command fails there while `RUSTUP_TOOLCHAIN=stable` and the direct
  binaries do not. That is the control quoted under BLOCKING 1, and it is what found BLOCKING 1.
  It is not the *exact* CI failure: I could not reproduce `'cargo' is not installed for the
  toolchain` locally, because that message requires a partially-installed distributable toolchain,
  whereas a linked custom toolchain makes rustup fall back to another cargo instead
  (`info: 'cargo' is unavailable for the active toolchain / info: falling back to …`). So the
  irreducible gap the PR names is real; the gap it treated as irreducible was smaller than stated.

## What I did NOT check

- I did not run this branch in GitHub Actions. Every CI claim here is read from the logs of runs
  that already happened on other branches, or reproduced locally on aarch64-apple-darwin with
  rustup 1.29.0. The Linux runner behaviour of `RUSTUP_TOOLCHAIN=stable` in the `deps-security` job,
  which installs no toolchain of its own, is inferred, not observed.
- I did not verify that cargo-deny 0.20.2 and 0.19.0 apply *identical rules* — I verified only that
  0.20.2 passes on this tree. The claim that pinning "freezes current behaviour rather than
  changing it" is consistent with what I saw but I did not diff the two versions' semantics.
- I did not run the full `pre-commit run --all-files`, only the four hooks named above plus
  `shellcheck`. Drift checks, generated-docs checks, typos, and the Rust hooks were not exercised.
- I did not run clippy, the workspace test suite, or the cross-target Linux compile gate. The change
  touches no Rust, so I judged the risk near zero; that is a judgement, not a measurement.
- I did not review `.github/workflows/split-wave0-gates.yml`, `release.yml` or `fuzz-smoke.yml`
  beyond locating their `cargo install` and plugin-invocation lines. I have not checked whether
  those jobs have some other protection I did not see.
- I did not test `cargo-audit audit --ignore RUSTSEC-2023-0071` end-to-end — I verified only that
  the `cargo-audit audit` argv form is accepted (`cargo-audit-audit 0.22.0`), not that the full
  command produces the same report as `cargo audit --ignore …`.
- I did not evaluate whether this PR should also fix #2211's theme (the harness restating what it
  could derive). I accepted the PR's deferral at face value.
- Environment note, for completeness rather than as a finding: reproducing the unpinned resolution
  replaced my local `cargo-deny 0.19.0` with `0.20.2`. Nothing in the repository was modified; the
  worktree is byte-identical to `2be4026ee2bda84fc11b1e5e17c8fc4ffefa57d9`.

Rul1an:
# Writer record — new head `edb0afbd250782ac2d3a821be8ce8ef373751746`

Relayed on behalf of the implementing agent, whose token could not post to this PR
(`Resource not accessible by integration`). Everything below was measured by that agent on
`edb0afbd250782ac2d3a821be8ce8ef373751746`; I have not re-measured it and say so rather than
presenting it as my own verification.

## Both blocking findings are fixed, not deferred

Two reviews landed on `2be4026e`, both naming the same two blockers: `cargo audit` still routing
through the unqualified `cargo` proxy, and `stable` written twice (`ci.yml:165` and
`install-cargo-deny.sh:30`). A job-level `env: RUSTUP_TOOLCHAIN: stable` closes both at once.

The obvious risk with a job-level toolchain variable is that it silently changes which toolchain
compiles or tests something. That was checked rather than assumed: no step in `deps-security`
builds or tests anything. The job is checkout, two tool installs, a `sed`-only alignment check,
`cargo-deny check`, and two `cargo audit` runs. `install-cargo-deny.sh` now defers to the inherited
value, so the job's `env:` is the single statement of the selection.

Structural diff against `main`: every job's `name`, `if`, `needs`, `continue-on-error` and every
step `if:` are unchanged. The added `env` block is the only structural difference, so no required
context was renamed, weakened, or made conditional.

## Verification on `edb0afbd`

- 50 checks pass, 13 skipping, all measured on this SHA.
- The `deps-security` log carries `RUSTUP_TOOLCHAIN: stable` in every step's environment.
- `1.96.0` appears nowhere in that job's log, which is the observable that distinguishes "the pinned
  toolchain was selected" from "the repo default was inherited".

## Review status

**Both review records on `2be4026e` are invalidated by this push.** They do not carry: the advance
is not an upstream merge, so the equivalence check in `AGENTS.md` does not apply.

This PR needs one non-building agent review on `edb0afbd`. CodeRabbit is a bot and does not count
toward that quorum. The delta since the reviewed head is narrow — one `env:` block and the install
script's deferral — so a review scoped to the five claims above is sufficient; the rest of the
change was already reviewed at `2be4026e` and is unchanged.

## Follow-up issues filed

- #2223 — `docs/PINNED-ACTIONS.md`. Re-measured: **15 of 17** SHAs match no call site. The earlier
  figure understated this.
- #2224 — six residual tool-install sites (the LOW findings from the second review).
- #2225 — bpf-linker `0.10.3` and `nightly-2026-01-01` restated across five files.
- #2226 — the pin is unprotected against its own removal; reproduced rather than argued.

Two nits were dropped as documented decisions: the dated-plan reference and the local-hook version
asymmetry.

## Non-claims

- I did not re-run the checks; the 50/13 figure is the implementing agent's measurement on this SHA.
- Nothing here establishes runner behavior beyond what the job log shows.
- The four follow-up issues are filed, not fixed.

Rul1an:
# Independent review — head `edb0afbd250782ac2d3a821be8ce8ef373751746`

No review record existed on this head when I started: the two `REQUEST CHANGES` reviews are bound to
`2be4026ee2bda84fc11b1e5e17c8fc4ffefa57d9`, the only comment on `edb0afbd` is the relayed **writer**
record (builder-side, not quorum), the Copilot entry says it could not review because of a quota
limit — which per `AGENTS.md` is unavailable infrastructure, not a verdict — and CodeRabbit is not
part of the quorum. So this comment is the quorum review, bound to the SHA above.

I am a non-building reviewer: I authored and edited neither this change, nor issue #2214, nor any
specification governing it, and I have pushed nothing to this branch. Measured in
`/Users/roelschuurkes/assay` at `edb0afbd250782ac2d3a821be8ce8ef373751746`
(`git status --porcelain` empty; the toolchain probe below ran in `/tmp`, not in the worktree).
Scope: the delta `2be4026e..edb0afbd`, plus the whole diff against `main` for the structural claim.

## Verdict: **APPROVE** — both prior blockers fixed; one non-blocking comment-accuracy finding

### 1. Nothing in `deps-security` compiles a **workspace** crate — but two steps do compile crates

The job is: checkout, `install-cargo-deny.sh`, `cargo install --locked cargo-audit`,
`check-aya-ebpf-alignment.sh`, `cargo-deny check`, two `cargo audit` runs.
`check-aya-ebpf-alignment.sh` contains no `cargo`/`rustc`/`rustup` reference at all, so the
"`sed`-only" description holds. But both installs build from source under the job-level toolchain,
and the log for this head proves it:

```text
job 93478415223 (run 31395807322), head edb0afbd
434:  Compiling cargo-deny v0.20.2
977:  Compiling cargo-audit v0.22.2
```

The safety case survives, and for a stronger reason than the comment gives: what
`rust-toolchain.toml` governs is how *this repository's* crates are built, and no step here builds
one. Compiling two third-party tools under `stable` cannot affect a workspace artifact or the MSRV
check (which lives in its own job, unchanged). It is also not optional — `1.96.0` has no `cargo` on
this image, which is the failure being fixed.

**Non-blocking finding (comment accuracy, not behavior):** the inline comment's clause
"never one that compiles a crate" is false as written; `cargo install` compiles crates. The adjacent
clause "no step in this job builds or tests the workspace" is exact and is the one the argument
actually needs. I am not blocking on a word in a comment whose surrounding claim verifies, and I
record it here rather than costing another round: **disposition — accept as-is, correct opportunistically.**

### 2. Unset `RUSTUP_TOOLCHAIN` still selects `stable`; the literal was removed, not hidden

Ran the script at this head with `cargo`/`cargo-deny` stubbed on `PATH` (in `/tmp`, no repo writes):

| inherited env | selected |
|---|---|
| both unset | `installing cargo-deny 0.20.2 using toolchain stable` |
| `RUSTUP_TOOLCHAIN=1.97.1` | `... using toolchain 1.97.1` — inherited value wins |
| `CARGO_DENY_TOOLCHAIN=beta`, `RUSTUP_TOOLCHAIN` unset | `... using toolchain stable` |

So every caller that states nothing (kernel-matrix.yml's lint job, pre-commit, a developer by hand)
gets exactly what `${CARGO_DENY_TOOLCHAIN:-stable}` gave it before — no silent substitution, and the
remaining `stable` is a default for callers that state nothing rather than a second statement that
must agree with `ci.yml`. Change `ci.yml` and the script follows; that is one rule, one place.
The dropped `CARGO_DENY_TOOLCHAIN` knob (row 3, now ignored) breaks no caller:
`git grep CARGO_DENY_TOOLCHAIN` returns nothing at either `edb0afbd` or `origin/main`.

### 3. No required context was renamed or made conditional

Diffing every job key, `name:`, `if:`, `needs:`, and `continue-on-error:` between `origin/main` and
`edb0afbd` in `ci.yml` returns empty. `kernel-matrix.yml` changes one `run:` line, name untouched.
`deps-security` keeps `name: Dependency Security (cargo-deny + audit)`, `needs: [scope]`, and its
existing `if:`. The `env:` block adds a key; it renames nothing.

### 4. `1.96.0` appears nowhere in the job log for this head

`grep -c '1\.96\.0'` over the full 5401-line log of job 93478415223 (head `edb0afbd`) returns `0`,
and line 140 reads `installing cargo-deny 0.20.2 using toolchain stable`. That is the observable
separating a stated selection from an inherited `rust-toolchain.toml` default, and it is clean.

### 5. Check counts are bound to this SHA

From `repos/Rul1an/assay/commits/edb0afbd.../check-runs`: **50 success, 15 skipped, 0 failing,
0 pending**; the PR-level rollup reads 52 success / 15 skipped (two extra status contexts) with
nothing outside success/skipped. Note the PR body's "13 skipping" understates the skip count by two;
the API figures above are the measured ones. No check is failing or pending on this head either way.

## Auto-merge preconditions

- required checks green — yes, on `edb0afbd` (above), `mergeStateStatus: CLEAN`;
- delegated runner proof on the final head — the relayed writer record targets `edb0afbd`, and every
  load-bearing claim in it is independently re-derived above rather than taken on trust;
- quorum — satisfied by this review, on this head;
- findings — both `2be4026e` blockers fixed and verified; the one new finding is non-blocking with the
  disposition recorded above.

Merging by squash.

PR #1335, opened 2026-05-23
chore(release): public-crate-policy update for runner substrate (v3.11.3)

Rul1an:
## Summary

This is a **policy PR**, not a manifest hot-fix. `v3.11.0`, `v3.11.1`, and `v3.11.2` are all partial-publish lines on crates.io: `assay-cli` did not publish on any of them. `v3.11.2` specifically blocked at the release-workflow policy gate (`scripts/ci/check-public-crate-policy.sh`) because the policy allow-list still listed only the original 10 public crates, while the manifest flip in `v3.11.2` had made the four runner crates `publish = true`.

The gate worked as intended. Making the runner crates publicly available on crates.io is a real public-surface decision, not a one-line manifest fix. This PR lands that decision deliberately, with all the matching policy, doc, and PR-CI guardrail wiring in one place.

## What changes

- **`scripts/ci/check-public-crate-policy.sh`**: extend `public_crates` to include `assay-runner-schema`, `assay-runner-core`, `assay-runner-linux`, `assay-runner-spike`. Comment block documents the framing — published because `assay-cli` depends on them; package descriptions explicitly say *"Internal/experimental substrate for Assay measured-run workflows … No standalone product guarantee; API surface remains narrow and intentionally undocumented for third-party use; semver tracks the Assay workspace."*; intentionally **not** in the Wave 0 library semver allowlist; adding any new public crate here is a deliberate public-surface decision.
- **`.github/workflows/ci.yml`**: new `Public crate policy` PR-CI job that runs `check-public-crate-policy.sh` on every PR. Until now the script only ran inside the release workflow on tag push, so a divergence between manifest and allow-list only surfaced after merge. This is the same defense-in-depth pattern as the `Publish-shape guardrail (assay-cli)` job added in `v3.11.1`.
- **`docs/contributing/WAVE0-GATES.md`**: document the runner crates as published-but-not-Wave-0-semver-checked; note the new PR-CI guardrail.
- **`CHANGELOG [3.11.3]`**: full reasoning, known-issue note covering the `v3.11.0/1/2` partial-publish lines, and "what changes for consumers" section.
- Workspace pins 3.11.2 → 3.11.3.

## Test plan

- [x] Local `bash scripts/ci/check-public-crate-policy.sh` — passes
- [x] Local `bash scripts/ci/check_assay_cli_publish_shape.sh` — passes (defense-in-depth)
- [ ] PR CI green incl. new `Public crate policy` job
[...]

PR #1333, opened 2026-05-23
chore(release): prepare v3.11.1 (publish-path hot-fix for assay-cli)

Rul1an:
## Summary

Hot-fix release that makes `assay-cli` publishable to crates.io again. The `v3.11.0` release published 8 of the 9 workspace crates but failed on `assay-cli` because Slice 6B (PR #1325) had introduced direct, non-optional dependencies on the internal `assay-runner-{schema,core,linux}` crates, all of which are `publish = false`. Cargo refused the publish with `no matching package named "assay-runner-core" found`. The release workflow only exercises `cargo publish` on tag push, so PR CI never saw the failure mode at v3.11.0 time.

This PR restores the `cargo install assay-cli` install path that was incomplete in the `v3.11.0` line. **No behavioural change** for repo / workspace consumers or for the GitHub Release binary tarballs.

## What changes

- `crates/assay-cli/Cargo.toml`: `assay-runner-{schema,core,linux}` deps become `optional = true`, behind a new `runner` feature in the default feature set. Workspace builds + binary releases unchanged.
- `crates/assay-cli/src/cli/commands/{mod,dispatch}.rs` and `crates/assay-cli/src/cli/args/mod.rs`: the internal `runner-spike` command is gated with `#[cfg(feature = "runner")]`. `cargo install assay-cli` from crates.io (which deactivates `runner`) ships a CLI without this internal-only command. This matches the existing CHANGELOG framing that `assay runner-spike` is outside the public CLI contract.
- `scripts/ci/publish_idempotent.sh`: `assay-cli` is published with `--no-default-features --features tui,sim`. All other crates publish unchanged.
- `.github/workflows/runner-spike-{delegated,sdk}.yml`: the runner gate builds add `--features runner` so `assay runner-spike` is still compiled when building with `--no-default-features`.
- `.github/workflows/ci.yml` + `scripts/ci/check_assay_cli_publish_shape.sh`: new `Publish-shape guardrail (assay-cli)` PR-CI job that statically verifies no non-optional dep in `crates/assay-cli/Cargo.toml` targets a workspace crate marked `publish = false`. Negative-tested locally; catches the exact regression shape that caused this hot-fix.
- Workspace pins bump 3.11.0 → 3.11.1.
- CHANGELOG entry for `[3.11.1]` documents the known issue with the `v3.11.0` crates.io line and the fix.

## Test plan

- [x] `cargo check -p assay-cli --no-default-features --features tui,sim` — builds clean without runner feature
[...]

PR #2213, opened 2026-08-10
feat(cli): register E_EVIDENCE_INTEGRITY as a content-only integrity reason

Rul1an:
Closes #2199. Unblocks #2164 and #2165.

## Provenance

- Branch `cursor/2199-evidence-integrity-reason`. This body names the SHA each measurement was taken on rather than "the current head", because the head moves and a description that tracks it goes stale silently — which is the failure this body is being corrected for. Read `gh pr view 2213 --json headRefOid` for the head under review.
- Earlier in this branch's history, `200378965af82567c8ab3006d6950d2b14316b2f` merged `origin/main` at `360c4963a8f978513948349204e12fc735595730` into the change commit `ec7e0be42aef904c0271679665056367d4260551`. Main's advance touched no file this PR changes — verified with `comm -12` over the two name lists, empty.
- **Earlier revisions of this body reported every number on `200378965`, which is not the head being merged.** They are re-stated below against the head each was actually measured on, per the measurement-provenance rule in `AGENTS.md`. Nothing in this body is now attributed to a tree it was not measured on.
- `--bins` is 544 where earlier revisions said 542. That is the platform, not the tree: those ran on macOS and the current figures are Linux, so `cfg(target_os = "linux")` tests are compiled in.

## What this is

Registry-only. It adds one reason code, `E_EVIDENCE_INTEGRITY`, and registers it in the spec. No command emits it yet; #2164 and #2165 are the consumers, and their opt-in is where the boundary below becomes enforceable at runtime.

## The decision the issue asked for

**One identity, and a boundary drawn where the measurement stops.**

The code says: content that was read does not verify — a member hash, an event's `content_hash`, a manifest entry, or the run integrity chain (`run_root`) disagrees with what the bundle records. It says nothing about how the content came to differ, so it is not a tampering or intent claim.

`run_root` is a hash chain, not a tree. `compute_run_root` (`crates/assay-evidence/src/crypto/id.rs:152-160`) is a flat `sha256` over each content hash followed by a newline, so there are no inclusion proofs and no sub-range verification. An earlier revision of this body called it "the Merkle root"; that named a structure the implementation does not build and is withdrawn.

[...]

PR #2207, opened 2026-08-10
fix(cli): publish the doctor config failure as a registered diagnosis

Rul1an:
## Summary

`assay doctor` reported an unloadable explicit config with a literal `E_CFG_PARSE` string, no
top-level `reason_code` or `next_step`, and the test-failure exit class. #2160 called the exit code
"the sharper defect", and driving the binary showed why: a caller that branched on the published
exit table read a malformed config as a failing test, and the recovery `assay run` itself publishes
— `assay doctor --config <path>` — downgraded the class from `2` to `1`.

The fix routes the whole decision through one helper that asks the reason-code registry, so the
reason identity, the recovery step and the exit class are decided once for the JSON report, the
text path, and `assay run` alike. The literal `"E_CFG_PARSE"` disappears from `doctor`.

Closes #2160.

## Scope / DoD

| #2160 acceptance | Where |
|---|---|
| Failure report exposes a registered top-level `reason_code` and non-empty `next_step` | `implementation.rs` reads both from `RunOutcome::from_reason` |
| Exit mapping reconciled with the frozen contract | `ECfgParse` → `EXIT_CONFIG_ERROR`, identical under the V1 and V2 profiles |
| Binary-level test reads stdout plus the exit code only | `agent_golden_path_contract.rs`, data-driven from the regenerated contract |
| Non-goal: no change to the human text renderer | the text `println!` lines are byte-identical |

## Compatibility

- `assay doctor` with an unloadable explicit config changes exit `1` → `2` on both output formats
  **when `--fix` is absent**. Only the exit class moves in text mode; the rendered text is
  byte-identical. An earlier revision of this line said "both channels" without qualification,
  which was untrue — see the `--fix` disclosure in Review focus 4.
- The JSON failure report gains `reason_code` and `next_step`. The success report is untouched:
  neither key appears when there is no config failure.
- `config_error.code` keeps the value `E_CFG_PARSE`; it is now read from the registry instead of a
  literal, so the two cannot drift.
- `assay.doctor_report.v0` keeps its schema id. Document identity stays #2167's subject.
- `docs/generated/agent-golden-path.json` and its five derived copies were regenerated by
  `scripts/docs/generate-agent-golden-path.py`; none was hand-edited.

## Smoking Gun

Two observations that a caller reading the contract could not have defended itself against.

**The recovery reclassified the failure it recovers from.** On one malformed file, one build:

```text
assay run    --config bad.yaml --format json   -> exit 2, reason_code E_CFG_PARSE
[...]

PR #2236, opened 2026-08-10
fix(ci): wire the two unenforced release guardrails into the required CI gate

### LINKED ISSUE #2230: Two ci.yml release guardrails sit outside the required CI gate, so their failure blocks nothing
`publish-shape-cli` and `public-crate-policy` run on every pull request and are absent from the `needs:` list of the `CI` job that aggregates results into the only required context, so either can fail while `CI` reports success and the pull request stays mergeable.

Measured on `360c4963a8f978513948349204e12fc735595730`.

## Evidence

The required set, from the artifact the guard treats as source of truth:

```
.github/rulesets/main-required-ci-contexts.json
"required_status_checks": [
  { "context": "CI", ... },
  { "context": "lane-check/proof", ... },
  { "context": "host-capability-check", ... }
]
```

`CI` is a single aggregation job whose comment states its purpose:

```
.github/workflows/ci.yml:900-905
  # Single job named "CI" so branch protection (required status check "CI") is satisfied.
  # Branch protection matches check run name; other jobs use descriptive names, this one reports "CI".
  ci:
    name: CI
    ...
    needs: [scope, deps-security, clippy, rustdoc, public-msrv, distribution-boundary, vendored-packs, release-asset-contract, mcp-registry-foundation, perf, test, ebpf-smoke-ubuntu]
```

Comparing the jobs `ci.yml` defines against that list:

```
jobs defined in ci.yml (name @ line):
   scope (24), deps-security (148), clippy (193), rustdoc (234), public-msrv (283),
   distribution-boundary (320), publish-shape-cli (332), public-crate-policy (354),
   vendored-packs (378), release-asset-contract (390), mcp-registry-foundation (412),
   perf (430), test (495), ebpf-smoke-ubuntu (667), ebpf-smoke-self-hosted (736), ci (901)

JOBS NOT AGGREGATED BY THE REQUIRED `CI` CONTEXT:
   publish-shape-cli        (.github/workflows/ci.yml:332)
   public-crate-policy      (.github/workflows/ci.yml:354)
   ebpf-smoke-self-hosted   (.github/workflows/ci.yml:736)
```

Neither missing guardrail is reachable transitively: both are leaf jobs with no `needs:` and no dependents, verified by reading their definitions at `ci.yml:332-352` and `ci.yml:354-377`.

`ebpf-smoke-self-hosted` is listed for completeness and is **not** part of this report. ADR-043's consequences record that kernel validation running post-merge rather than on pull requests is consistent with its supporting status, so its absence looks deliberate.

## Why these two specifically

Both jobs were added to block merges, and say so:

```
ci.yml:333-336 (publish-shape-cli)
    # Guardrail: catch v3.11.0-style regressions where `assay-cli` accumulates
    # a direct dependency on a `publish = false` workspace crate, which would
    # only surface as a release-time `cargo publish` failure on tag push.
```

```
ci.yml:361-364 (public-crate-policy)
    # `check-public-crate-policy.sh` previously only ran inside the release
    # workflow, so a divergence ... only surfaced at tag time after merge. This
    # PR-CI mirror runs the same script on every PR so the gate fires
    # before tag, not after.
```

The stated purpose of `public-crate-policy` is to move a failure from tag time to pull-request time. Because it is outside the aggregator, it moves the *notification* but not the *block*: the merge still proceeds and the release still breaks at tag.

This is drift rather than a decision. The `needs:` list dates to #455; `publish-shape-cli` arrived in #1333 and `public-crate-policy` in #1335, and neither added itself to the gate:

```
$ git log --oneline -S "publish-shape-cli:" -- .github/workflows/ci.yml | tail -1
bdfa6841 chore(release): prepare v3.11.1 (publish-path hot-fix for assay-cli) (#1333)
$ git log --oneline -S "public-crate-policy:" -- .github/workflows/ci.yml | tail -1
5d0005f0 chore(release): public-crate-policy update for runner substrate (v3.11.3) (#1335)
$ git log --oneline -S "needs: [scope, deps-security" -- .github/workflows/ci.yml | tail -1
f65eac5f ci: harden docs-only CI publication and perf triggering (#455)
```

## Where the existing compensating control stops

The `CI` job is otherwise unusually careful about this class of failure. Its body pairs every job it lists with the condition under which that job was expected to run, so a `skipped` result is only accepted when the scope classifier said it was not needed, and its comment explains that a gate which "never asks WHY a job was skipped cannot tell 'the classifier said this PR does not need it' from 'the job's own `if:` broke'". `scripts/ci/check-required-contexts.py` separately pins the required-context *set* across the ruleset, `CI-CONTRACT.md` and the runbook, and self-tests its own parsers.

Both controls answer questions one level away from this one. The first asks whether the jobs in the list produced a real result; the second asks whether the three named contexts agree across three files. Neither asks whether the list of jobs the aggregator waits on covers the jobs the workflow defines, which is where these two fell out.

## Smallest correct fix

Add `publish-shape-cli` and `public-crate-policy` to the `ci` job's `needs:` list and to the result table below it, with expectation `required` — they have no `if:`, so `success` is the only acceptable result and no scope output is involved.

Because the same omission can recur, the durable version is a check that reads `ci.yml`, subtracts the aggregator's `needs:` from the workflow's job list, and fails on any residue not carrying an explicit opt-out marker. That belongs next to `check-required-contexts.py`, which already owns "the places that name the required set must agree".

## What I did not establish

- **I could not read live branch protection.** `gh api repos/Rul1an/assay/branches/main/protection` returns HTTP 403 for this token. The required set above is taken from the checked-in ruleset, which is the artifact `check-required-contexts.py` calls the source of truth because it is the one that can be imported; I did not confirm it is what is currently applied.
- I did not observe a merge that actually happened with one of these jobs red. The claim is that the configuration permits it, read off the workflow.
- I make no claim about the other 46 workflow files, or about workflow count as such. Sprawl is not a defect and this issue does not propose reducing it.
- I did not check whether `lane-check/proof` or `host-capability-check` have an analogous gap. `host-capability-check`'s always-run-and-self-skip design is documented in its own header as deliberate, and I found nothing to contradict that.

Rul1an:
## Description

`publish-shape-cli` and `public-crate-policy` ran on every pull request and were absent from the `needs:` of the `ci` job, which reports the only required context `CI`. Either could fail while `CI` reported success and the pull request stayed mergeable, so both release guardrails have been reporting without blocking. Reported as [#2230](https://github.com/Rul1an/assay/issues/2230).

**Provenance and duration.** The rollup's `needs:` list arrived in [#455](https://github.com/Rul1an/assay/pull/455) (`f65eac5f`, merged 2026-02-27). `publish-shape-cli` arrived in [#1333](https://github.com/Rul1an/assay/pull/1333) (`bdfa6841`, 2026-05-23) and `public-crate-policy` in [#1335](https://github.com/Rul1an/assay/pull/1335) (`5d0005f0`, 2026-05-23). Neither added itself to the earlier gate. Both were therefore unenforced from **2026-05-23 to 2026-08-10 — 79 days**. This is drift, not a decision: `public-crate-policy`'s own comment says it exists so the gate "fires before tag, not after", and outside the rollup it moved the notification while the merge still proceeded.

**A third job was outside the rollup.** `ebpf-smoke-self-hosted` is also absent from `needs:`, and the issue set it aside as probably deliberate. It is not wired in: its `if:` excludes `pull_request`, so on the event branch protection judges it has no result to contribute, and ADR-043's consequences record kernel validation running post-merge as consistent with its supporting status. It now carries an explicit opt-out marker, so that absence is a decision written on the page rather than a gap that looks identical to the other two. Those three are the complete set — the workflow defines 16 jobs, the rollup is one, and the other 12 were already wired.

## Blast radius

**Nothing is newly blocked.** Both jobs are green on the GitHub runner at this PR's head, `3467d9af470c6ce5273ec1137464502cd15b9682` — `CI / Publish-shape guardrail (assay-cli)` and `CI / Public crate policy` both `success`, and the `CI` rollup itself is `success` with the two new dependencies wired in ([run 31399999206](https://github.com/Rul1an/assay/actions/runs/31399999206)). The same scripts also pass locally on `main` at `360c4963a8f978513948349204e12fc735595730`:

```
$ ./scripts/ci/check_assay_cli_publish_shape.sh          # exit 0
Publish-shape guardrail OK: assay-cli has no non-optional deps on
publish = false crates (7 workspace crates checked).
Publish-shape guardrail OK: no packaged file #[path]-includes outside the package
(4 excluded path(s) honoured).

$ bash scripts/ci/check-public-crate-policy.sh           # exit 0
Public crate policy is consistent.
```

So wiring the gate does not block any open PR today. It means the next regression in either guardrail stops a merge instead of being noticed after one.

**One thing this did surface**, and it is worth stating because it is what a working gate looks like. On the pre-merge head `ab59758a`, `CI / Build + Test (windows-latest)` failed and correctly took the `CI` rollup red with it — annotation `Required CI dependency test ended with failure`. That job is untouched by this PR and is an existing intermittent failure on `main`: `Build + Test (windows-latest)` is `failure` on `4125d81b` (#2198) and `5ddc9010` (#2204), and `success` on `cd8dc5d6` (#2218), read from `/commits/{sha}/check-runs`. It passed on re-run at `3467d9af`. Not fixed here and not investigated further; if it recurs it deserves its own issue.

## The fix

Both jobs join `needs:` and the rollup's `name|result|expectation` decision table with expectation `required`.

**Is a skip a new fail-open?** No. Read against the rollup's logic: the `skipped` branch increments `failures` whenever the expectation is `required`, and only an expectation computed from a scope output can relax it. Neither job has an `if:` and neither has `needs:`, so no scope output can skip them and `required` is the only correct expectation — a skip means the job's own definition broke, and the gate now fails on it. Executed as well as read: `test-ci-gate-expectations.sh` runs both jobs through its unconditional-skip cases and requires the "was skipped, but this run required it" message, plus a new case pinning the #2230 outcome directly, that a red guardrail turns the required context red. The coverage guard below independently rejects a computed expectation on a job with no `if:`, so the fail-open cannot be reintroduced by relaxing the table.

## The new guard, and why it cannot itself drift

`scripts/ci/check-ci-gate-coverage.py`, following the `semver-gate` and `linux-gate` self-test precedent in `.pre-commit-config.yaml`. It runs as a **step of the `CI` job itself**, so the required context refuses a workflow whose rollup does not cover it, not only a local hook someone can skip; and as a pre-commit hook for fast feedback. It executed inside the required job on the runner at `3467d9af` and passed.

The rule is derived from the workflow, never restated beside it — a hardcoded list of job names would be a second statement of the same rule, free to drift from the first, which is this defect reintroduced one level up:

- Every job the workflow defines either appears in the rollup's `needs:` **and** its decision table, or carries `# ci-gate: not-required -- <reason>` at job indentation with a reason of at least 30 characters. A job added tomorrow starts with neither, so it is covered by construction.
- Both halves are required: `needs:` membership alone makes the rollup wait, and the table is what makes it judge. Membership without a table entry is waited on and ignored; an entry naming another job's result variable judges the wrong job.
- Expectations are checked: a job with no `if:` can only legitimately end in `success`, so its expectation must be the literal `required`.
- The rollup itself is not named. It is whichever job reports a context the checked-in ruleset requires, so renaming that job without amending the ruleset fails this check instead of silently detaching branch protection.
- Anything unparseable raises rather than passes. A restructured workflow surfaces as a failure to read it, not as a guard that quietly stopped looking.

`test-ci-gate-expectations.sh` loses its three-name `REQUIRED` list and three `*_RESULT` greps: those were the hand-maintained form of the rule the guard now derives, and keeping both would violate one-rule-one-function.

## Verification

Local measurements are from this VM, `/workspace`, python 3.12, actionlint 1.7.10, shellcheck 0.10.0.

**Before**, at `5987d862b4b1c4b990cd8e4b707010e06a870b31` — the guard-only commit, whose `ci.yml` is byte-identical to `main` at `360c4963`:

```
$ python3 scripts/ci/check-ci-gate-coverage.py          # exit 1
ci-gate-coverage=failed
  `publish-shape-cli` is absent from `ci`'s `needs:`, so its failure cannot make the required context red. ...
  `public-crate-policy` is absent from `ci`'s `needs:`, so its failure cannot make the required context red. ...
  `ebpf-smoke-self-hosted` is absent from `ci`'s `needs:`, so its failure cannot make the required context red. ...

A job outside the rollup's `needs:` can fail while the required context reports success, so the merge it was written to block goes through (#2230).
```

**After**, at `3467d9af470c6ce5273ec1137464502cd15b9682`:

```
$ python3 scripts/ci/check-ci-gate-coverage.py               # exit 0
ci-gate-coverage=passed
$ python3 scripts/ci/check-ci-gate-coverage.py --self-test   # exit 0
ci-gate-coverage self-test=passed
$ bash scripts/ci/test-ci-gate-expectations.sh               # exit 0
... 12 ok lines ...
PASS: CI gate expectation contract
```

**A future unwired gating job still fails it.** A dummy `supply-chain-attestation` job was added to `ci.yml` in the workspace, the guard run, and the job reverted:

```
ci-gate-coverage=failed
  `supply-chain-attestation` is absent from `ci`'s `needs:`, ...
```

`actionlint` was clean on that same dummy tree, so the finding is about the wiring and not about a malformed job. The `--self-test` covers the same property mechanically across 13 cases, each pinning the message it must produce rather than merely that something went red, including that the opt-out marker works — without which the unwired case would be a finding about newness rather than about wiring.

**Other checks**, at `3467d9af`:

- GitHub Actions: 62 check runs, 0 failed, `CI` (required), `lane-check/proof` (required) and `host-capability-check` (required) all `success`.
- `actionlint -config-file .github/actionlint.yaml` on `.github/workflows/ci.yml`: clean.
- `pre-commit run --files <the five touched files>`: all hooks pass. (`docs-auto-update-reconcile-self-test` needed `ruby` installed in this VM; it exited 127 `command not found` before and passes after, so its green is a result and not a missing interpreter.)
- `pre-commit run --hook-stage pre-push --files <same>` at `ab59758a`, unchanged by the merge: all pass, including `cargo clippy (workspace, deny warnings)`, `cargo fmt --check`, `shellcheck`, and the linux cross-target compile gate.
- `git diff --check`: clean.
- **The `CI` context name is unchanged.** The diff touches no `name: CI` line, `.github/workflows/ci.yml` still declares `    name: CI`, and the ruleset is unmodified. The guard now cross-checks the two, so a future rename fails rather than silently detaching protection.

`3467d9af` merges `origin/main` (`cd8dc5d6`, [#2218](https://github.com/Rul1an/assay/pull/2218)) into `ab59758a`. That commit edits `deps-security` and the `cargo-deny` pre-commit hook; it does not touch the rollup, the two guardrail jobs, or anything this PR changes, and the merge was textually clean. It was taken before any review existed, so no review record is affected.

## Type of Change
- [x] Bug fix

## Non-claims

- **I did not read live branch protection.** `gh` has no working credentials in this environment and the authenticated API returned 401, so the required set is taken from the checked-in ruleset, as `check-required-contexts.py` does. I did not confirm the applied protection matches it. If live protection requires a context the ruleset does not name, the guard's identification of the rollup is correct only by coincidence.
- **I did not observe a merge that actually happened with one of these jobs red.** The claim is that the configuration permitted it, read off the workflow — the same non-claim #2230 makes.
- **The self-test cannot run on a failing tree.** It refuses when the baseline already fails, following `check-required-contexts.py`, so it is measured only at the post-fix SHAs. The plain check is measured at both.
- **The windows flake is characterised, not diagnosed.** I read its conclusion across four commits from the check-runs API; I could not read its logs (403 without repository admin rights) and did not determine the cause.
- I make no claim about the other 46 workflow files, or about jobs in any workflow other than `ci.yml`. `lane-check/proof` and `host-capability-check` are separately required contexts and this guard says nothing about whether they have an analogous gap.
- I make no claim that the two guardrails are correct, only that their result now reaches the gate. Whether `check_assay_cli_publish_shape.sh` and `check-public-crate-policy.sh` catch what they intend to is out of scope.
- No evidence-ingestion, evidence-verification, or MCP-authorization code is touched, so the ADR-042/043 boundary template does not apply.

## Review notes

The builder's self-review does not count; this needs one non-building agent review on head `3467d9af470c6ce5273ec1137464502cd15b9682`, and bots do not count toward that quorum. Do not merge and do not enable auto-merge.

Start with the skip semantics: read the `skipped` branch of the gate loop in `ci.yml` against the two new `required` triples and confirm that a skipped `publish-shape-cli` fails rather than passes. Then check that the coverage guard's rule really is derivable — that no job name is hardcoded anywhere in `check-ci-gate-coverage.py` outside the self-test's mutations. Third, the `ebpf-smoke-self-hosted` opt-out is the one judgement call here: it is the only job the guard is being told not to enforce, and if that reasoning is wrong the right fix is to wire it in, not to widen the marker.

Closes #2230

<div><a href="https://cursor.com/agents/bc-aa0367af-0bce-4e05-b61a-901b448fd139?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-aa0367af-0bce-4e05-b61a-901b448fd139&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **CI Improvements**
  * Strengthened CI validation so required checks are consistently included in the overall status.
  * Added detection for missing, renamed, or silently skipped required checks.
  * Added coverage validation for publish-shape and public crate policy checks.
  * Improved handling of conditionally applicable lightweight, eBPF, and MCP registry checks.

* **Developer Experience**
  * Added automated pre-commit validation for CI gate coverage.
  * Added comprehensive self-tests and clearer diagnostics for CI configuration issues.


Rul1an:
## Merge of `main` into this branch — head is now `fd68ed3a333c36077aaba181cf70bb1781b1912a`

I am this push's **builder, not a reviewer**. **No review exists on `fd68ed3a3`.** The only
review record on the PR is `copilot-pull-request-reviewer` `COMMENTED` on `ab59758af`, which is
neither this head nor part of the quorum. This head needs exactly one non-building agent review.

Measurement provenance: every result below was taken in worktree `/Users/roelschuurkes/wt-2236`
on the tree committed as `fd68ed3a333c36077aaba181cf70bb1781b1912a`, with `actionlint` 1.7.10 and
`pre-commit` running the repo's pinned hook revisions.

### What conflicted, and how it was resolved

**Nothing conflicted.** The premise that `ci.yml` would conflict was already spent: the branch's
merge base with `main` was `cd8dc5d6f` — that *is* #2218 — so #2218 was already an ancestor of
`3467d9af4`, the prior head's own merge commit, and its `ci.yml` resolution had already happened
there. The only upstream advance left was `1a58c933e` (#2207, doctor config diagnosis), which
touches `crates/assay-cli/**` and docs and no workflow file. `git merge --no-ff origin/main`
therefore reported "Merge made by the 'ort' strategy" over 15 files with zero conflicts, and
`git diff --stat 3467d9af4 HEAD -- .github/workflows/ci.yml` is empty: this merge did not touch
`ci.yml` at all.

Both intents are nevertheless confirmed present on the merged tree, since "no conflict" is a
claim about the merge and not about the file:

- **`main`'s intent (#2218)** — `deps-security` carries the job-level
  `env: RUSTUP_TOOLCHAIN: stable` at `ci.yml:171`, so that job's `cargo` invocations state their
  toolchain instead of inheriting 1.96.0 from `rust-toolchain.toml`.
- **this branch's intent** — `publish-shape-cli` and `public-crate-policy` are both in the `CI`
  rollup's `needs:` (`ci.yml:930`) *and* in its `name|result|expectation` decision table;
  `ebpf-smoke-self-hosted` carries `# ci-gate: not-required --` with its reason (`ci.yml:759`);
  the rollup's first step runs `python3 scripts/ci/check-ci-gate-coverage.py`.

### The guard on the merged tree

```
$ python3 scripts/ci/check-ci-gate-coverage.py
ci-gate-coverage=passed
```

Exit 0. The merge added no job to `ci.yml`, so nothing new needed wiring or a marker — but that
is the guard's answer, not an inference from the diff.

### Two deliberate breakages, on the merged tree

**1. A job that neither reaches the rollup nor carries the marker.** Inserted
`breakage-probe-unwired` ahead of the rollup:

```
ci-gate-coverage=failed
  `breakage-probe-unwired` is absent from `ci`'s `needs:`, so its failure cannot make the required
  context red. Wire it in, or mark it `# ci-gate: not-required -- <reason>` at the job's own
  indentation if it is genuinely not a merge gate

A job outside the rollup's `needs:` can fail while the required context reports success, so the
merge it was written to block goes through (#2230).
```

Exit 1. Reverted; guard back to `ci-gate-coverage=passed`.

**2. The rollup job renamed.** `ci:`/`name: CI` → `ci-rollup:`/`name: CI rollup`:

```
ci-gate-coverage=failed
  .github/workflows/ci.yml: no job reports a required context ['CI', 'host-capability-check',
  'lane-check/proof']. Either the rollup job was renamed, which detaches branch protection, or the
  ruleset no longer names the context it reports.
```

Exit 1 — it failed rather than silently passing, because the guard locates the aggregator by
matching a job's reported name against the checked-in ruleset instead of hardcoding `ci`. Reverted.

Tree clean afterwards: `git status --porcelain` empty, `git diff --exit-code` exit 0.

### Verification

| Check | Result |
|---|---|
| `actionlint` on merged `ci.yml`, repo's own invocation (`-config-file .github/actionlint.yaml -ignore 'unexpected key "queue" ...'`) | exit 0, no findings |
| `actionlint` bare, no repo args | one finding, `queue: max` at `ci.yml:770` — **identical on `origin/main`** (`main-ci.yml:767`), i.e. pre-existing and not introduced here; it is the finding `.pre-commit-config.yaml` ignores for rhysd/actionlint#657 |
| `pre-commit run --files` over the five touched files | 22 hooks Passed, 2 Skipped (no files), 0 Failed — including the two new ones, `CI gate expectation contract` and `Required CI gate waits on every gating job` |
| `scripts/ci/test-ci-gate-expectations.sh` | 12 `ok:` cases, `PASS: CI gate expectation contract` |
| `check-ci-gate-coverage.py --self-test` | `ci-gate-coverage self-test=passed` (9 defeat mutations detected, 4 unreadable-structure cases raised) |
| `git diff --check` and `git diff --check origin/main...HEAD` | exit 0 |
| pre-push hooks (all, on the pushed commit) | Passed, including `cargo clippy (workspace, deny warnings)`, `cargo fmt --check`, `shellcheck`, `linux compile gate (cross-target)` |

No check was weakened, bypassed, or relabelled. The push was a normal push; no force-push.

### One note on `main` moving again

`main` advanced to `c2c41b038` (#2213) during the ~6 minutes the pre-push hooks took, so GitHub
reports this head `BEHIND` again. I deliberately did not chase it, because chasing is unbounded and
unnecessary here: `c2c41b038` touches `crates/assay-cli/src/exit_codes*`, one spec doc and one test,
and **none of this PR's five files**. So a review taken on `fd68ed3a3` *can* carry across a later
routine merge of `c2c41b038` under the revalidation rule — the merge introduces nothing not already
on `main`, and the advance touched no reviewed file — provided both checks are recorded in the review
record at that time. That is the reason to review this head now rather than wait for another merge.

Rul1an:
## Independent non-building review — head `fd68ed3a333c36077aaba181cf70bb1781b1912a`

I am a **non-building reviewer**. I authored and edited neither this change nor any specification
or plan governing it, and I have not pushed to this branch. My reproduction ran in a worktree I
created for the purpose, `/Users/roelschuurkes/rev-2236`, detached at
`fd68ed3a333c36077aaba181cf70bb1781b1912a`, which is the head `gh pr view 2236 --json headRefOid`
reports and the head every number below is bound to. The builder's evidence was not accepted; each
claim was re-derived.

For the quorum record, the two bot records on this PR are worth what `AGENTS.md` says they are
worth. `copilot-pull-request-reviewer` `COMMENTED` on `ab59758af` with "unable to review this pull
request because the user who requested the review has reached their quota limit" — a record that
says it did not review, on a head that is not this one. `coderabbitai` `COMMENTED` on `fd68ed3a3`
with one actionable finding, adjudicated below. Neither is part of the quorum; this review is.

### Verdict: **APPROVE**. No blocking findings. Merging.

The defect #2230 describes is closed, not relocated. Three follow-up findings, none of which
justifies invalidating a green head under the one-review-per-head cap.

---

## Findings

### Blocking

**None.**

### Follow-up

**F-1 — MEDIUM — `continue-on-error` is invisible to the new guard.**
`scripts/ci/check-ci-gate-coverage.py:188-241`. Filed as #2242.

The guard establishes that every job reaches `needs:` and the decision table with a non-permissive
expectation. It never asks whether the job can *report* failure. A `continue-on-error: true` on any
step of `publish-shape-cli` makes that step's conclusion `success` when it fails — documented, not
inferred: "When a `continue-on-error` step fails, the `outcome` is `failure`, but the final
`conclusion` is `success`." The job then concludes `success`, the rollup reads `success`, and the
required context is green while the guardrail is red. That is #2230's outcome reached by a different
road, and the guard this PR adds is the natural place to close it. I reproduced the blind spot at
job level (case F in the table below): the guard reports nothing.

Smallest fix: in `check()`, add a problem for any job in `needs:` whose body matches
`continue-on-error:\s*true` at either job or step indentation, plus one `detected` self-test case.
Roughly four lines in the function that already walks each job body.

I did not empirically confirm the *job*-level `needs.<job>.result` value under
`continue-on-error: true`; GitHub documents the conclusion-substitution for steps and is silent for
jobs. The step-level vector is documented and sufficient on its own.

**F-2 — MEDIUM — the checked-in ruleset is a transcription, and nothing reconciles it with what is
actually enforced.** `.github/rulesets/main-required-ci-contexts.json`, newly load-bearing via
`scripts/ci/check-ci-gate-coverage.py:47,119-132`. Pre-existing; not introduced here. Filed as #2243.

`gh api "repos/Rul1an/assay/rulesets?includes_parents=true"` returns `[]`. **No ruleset is applied
to this repository at all.** Live enforcement is classic branch protection, and the file is an
importable artifact that has not been imported. `scripts/ci/check-required-contexts.py:52` calls it
"the source of truth: it is the one that can be applied" — which is true of its *shape* and false of
its *status*.

The sets agree today, so this is not the HIGH finding it would otherwise be. Live protection requires
`CI`, `host-capability-check`, `lane-check/proof`; the file names those three. One field already
disagrees: the file carries `"integration_id": 15368` for `CI` where live reports `"app_id": null`.

What holds the agreement is also worth stating precisely, because the failure directions are not
symmetric. Rename the rollup job without amending the file, or amend the file without renaming the
job, and the guard raises `CoverageError` and the required context goes red — both fail closed, and
I verified both (cases B and H). A rename propagated to the file but not to live protection leaves
live protection requiring a context nothing reports, which blocks every merge. The residual is
one-directional: if live protection is *loosened* — `CI` dropped from the required set — no artifact
in this repository notices, and every guard in this PR keeps passing while gating nothing. That is
#2230's shape at the outermost layer.

Smallest fix: a scheduled job that reads `repos/Rul1an/assay/branches/main/protection` and fails when
the live required-context set differs from the checked-in artifact — one call and a set comparison,
and it makes the file authoritative in the only way a checked-in file can be.

**F-3 — LOW — no `timeout-minutes` on the rollup, or on 14 other jobs.** `.github/workflows/ci.yml:926`.
This is CodeRabbit's finding, adjudicated in full below. Filed as #2244.

### Informational, no action

An opt-out marker placed on the rollup job itself is silently inert: `check()` skips
`name == aggregator` before reading markers, so the marker neither takes effect nor is rejected
(case K). Harmless — the rollup's coverage is not a question the guard asks — but a future reader
may expect it to be refused.

Provenance nit: the merge record states "10 skipped". At `fd68ed3a3` the count is **12**. The 54
successes and the zero failures are exact. Most likely measured before the two eBPF check runs
registered at 15:35:08Z.

---

## The six angles

### 1. Did the fail-open move rather than close? — **CLEAN. It closed.**

The rollup's evaluation is a three-way `case` on each dependency's result
(`.github/workflows/ci.yml:1025-1043`). For a dependency carrying expectation `required`:

| result | rollup concludes |
|---|---|
| `success` | pass |
| `skipped` | **fail** — `::error::<name> was skipped, but this run required it to execute`, `failures+1` |
| `failure`, `cancelled`, anything else | **fail** via the catch-all `*)` arm, `failures+1` |

So a skipped required dependency turns the required context red. `skipped` relaxes the gate only
where the expectation is a *computed* one and the scope output that computes it is the literal
`false`.

Both newly wired jobs carry the **literal** `required`, not a computed expectation
(`ci.yml:1011-1012`), so no scope rule can reach them. That is belt and braces over a second fact:
`publish-shape-cli` (`ci.yml:354`) and `public-crate-policy` (`ci.yml:376`) declare neither `needs:`
nor `if:`, so they always start and `skipped` is unreachable for them in the first place. And the
guard enforces the pairing rather than trusting it — `check()` at line 236 rejects a computed
expectation on any job with no `if:`, which I reproduced.

Proved by construction, not by reading intent:

```
$ bash scripts/ci/test-ci-gate-expectations.sh
ok: a job with no condition may not be skipped even on a docs-only run
ok: a failed release guardrail fails the required gate
PASS: CI gate expectation contract
```

The first line runs `PUBLISH_SHAPE_CLI_RESULT=skipped` and `PUBLIC_CRATE_POLICY_RESULT=skipped`
under `LIGHTWEIGHT_ONLY=true` — the most permissive scope the workflow can produce — and the gate
still fails. The second runs each at `failure` and requires the gate to name the job.

### 2. Is the `ebpf-smoke-self-hosted` opt-out honest? — **CLEAN, and nothing relies on it.**

The stated reason checks out three ways. The `if:` at `ci.yml:764` opens with
`github.event_name != 'pull_request'`, so on the event branch protection judges the job cannot
reach a conclusion other than skipped. Empirically, at `fd68ed3a3` the check run
`eBPF monitor smoke (Linux - Self-Hosted)` has `"conclusion": "skipped"`. And the ADR the marker
cites says what it says — `ADR-043:133`: "The kernel-observation posture is unchanged: eBPF
validation running post-merge rather than on pull requests remains consistent with its supporting
status."

The harder question does not turn up a gap wearing a marker. On a pull request the eBPF smoke is
gated by `ebpf-smoke-ubuntu` (`ci.yml:689`), which *is* in the rollup's `needs:` and carries
`${ebpf_expectation}`, derived from `ebpf_smoke_required` — and the harness proves
skipped-while-required fails (`ok: ebpf-smoke-ubuntu skipped while required fails the gate`). So
the marked job is a post-merge *addition* to a PR-path gate that exists, not a substitute for one.
Nothing claims otherwise: `rg 'ebpf-smoke-self-hosted|Self-Hosted' CI-CONTRACT.md
docs/BRANCH-PROTECTION-SETUP.md` returns zero hits, so no document asserts it is required and no
required context depends on it.

### 3. Is the checked-in ruleset authoritative? — **No. It is a transcription.** See F-2.

It agrees with live protection today on the set of context names, so no HIGH finding; nothing
enforces that it keeps agreeing, which is F-2 at MEDIUM.

### 4. Can the guard pass vacuously? — Both builder breakages reproduced independently; five
further cases attempted, one real gap (F-1).

I mutated the workflow text in memory and called `check()` directly, so the reviewed tree was never
written to. Harness: `/tmp/rev2236/attack.py`.

| # | case | guard | note |
|---|---|---|---|
| A | unwired dummy job *(builder claim 1)* | **DETECTED** | ``​`dummy-guardrail` is absent from `ci`'s `needs:` `` — reproduced independently, fails for the stated reason |
| B | rollup renamed, ruleset untouched *(builder claim 2)* | **fails closed** | `CoverageError: no job reports a required context [...]` — reproduced |
| C | job in `needs:`, absent from the decision table | **DETECTED** | ``waits on `dummy-guardrail` and then never judges it`` |
| D | table entry reads another job's result variable | **DETECTED** | ``judged on `${VENDORED_PACKS_RESULT}`, which is not the variable bound`` |
| E | conditional job, computed expectation from the *wrong* scope output | passes | **covered by the sibling harness** — see below |
| F | `continue-on-error: true` on a wired required job | **passes** | **F-1** |
| G | real gate dropped from `needs:` + table, wearing a plausible 30+ char reason | passes | by design; see below |
| H | ruleset names a context no job reports | **fails closed** | `CoverageError` |
| I | guardrail added to a *different* PR-triggered workflow | passes | stated scope; see below |
| J | rollup `needs:` with a trailing comment | **fails closed** | `keep it a single-line flow list` |
| K | opt-out marker on the rollup job itself | passes, inert | informational above |

On **E** — the guard validates the *shape* of a computed expectation, not its *correspondence* to
the job's `if:`. I swapped `mcp-registry-foundation`'s expectation from `${registry_expectation}` to
`${code_gated_expectation}`; the guard is silent. But the sibling harness is not, and I confirmed it
empirically rather than assuming it:

```
$ bash scripts/ci/test-ci-gate-expectations.sh   # under mutation E
FAIL: explicit false relaxes both: expected the gate to pass, it exited 1
```

The coverage guard asks whether the wiring exists; the expectations harness asks whether the wiring
decides correctly. Two questions, two mechanisms — not two statements of one rule, so
"one rule, one function" is satisfied and E is **invalid** as a finding.

On **G** — any marker scheme reduces to "a human typed a sentence a reviewer must read". The
`MIN_REASON_CHARS = 30` floor buys visibility, not truth. I record this as the design's honest floor
rather than a defect: the marker's job is to convert a silent omission into a reviewable claim, and
it does that.

On **I** — the guard reads `ci.yml` only, which the docstring and `CI-CONTRACT.md:500` both state
("every job in `ci.yml`"). Worth knowing how wide the uncovered area is: 23 workflows trigger on
`pull_request` and 3 contexts are required, so roughly twenty PR checks can fail without touching a
required context. `CI-CONTRACT.md:508-516` documents that as deliberate — advisory lanes, matrix
leaves, host-capability jobs. The two required contexts produced by *other* workflows,
`lane-check/proof` and `host-capability-check`, need no coverage guard because they gate directly
rather than through a rollup. So **not a finding against this PR**, and its scope claim is accurate.

### 5. CodeRabbit's `timeout-minutes` finding — **VALID, and I endorse the builder's deferral.**
Recorded as F-3, follow-up, not blocking.

The hazard is real in kind. Without `timeout-minutes` a job inherits GitHub's 360-minute default, so
a hung rollup holds the required context pending for six hours, and this PR newly gives the rollup
something that can hang — an `actions/checkout` over the network. It is not real in degree here: the
rollup completed in 8 seconds at this head (`started_at 15:35:00Z`, `completed_at 15:35:08Z`), and it
is one of **15 of 17** jobs in `ci.yml` carrying no timeout — only the two eBPF jobs have one, at 15
and 60 minutes. Timing the rollup alone would leave the fourteen that share the exposure and would
make the file less consistent, not more. The coherent fix is a workflow-wide policy, which is a
different change than this PR.

So I **overrule the framing and endorse the disposition**: it is not a "quick win" scoped to one
job, and re-verifying a green head for a hygiene nit is exactly the churn the review cap exists to
prevent. Filed as follow-up.

### 6. Was anything weakened? — **CLEAN.**

Diffing every job key, `name:`, `if:`, `needs:` and `continue-on-error:` between `main`
(`c2c41b038`) and this head:

```
$ diff <(git show c2c41b038:.github/workflows/ci.yml | rg '^  [A-Za-z0-9_-]+:$|^    (name|if|needs|continue-on-error):') \
       <(git show fd68ed3a3:.github/workflows/ci.yml | rg '^  [A-Za-z0-9_-]+:$|^    (name|if|needs|continue-on-error):')
```

The only substantive difference is the rollup's `needs:` gaining `publish-shape-cli` and
`public-crate-policy`. Everything else is line-number displacement from the three added comment
lines. **No job renamed. No `if:` added, removed or altered. No `continue-on-error:` introduced
anywhere — `ci.yml` still has none at any level.** The change to the gate is strictly additive: it
can now refuse in two cases where it previously could not, and in no case can it now accept where it
previously refused. The added `permissions: contents: read` on the rollup narrows rather than widens,
against a workflow-level `permissions: {}`.

CI counts, bound to this exact SHA rather than to a branch name:

```
$ gh api "repos/Rul1an/assay/commits/fd68ed3a333c36077aaba181cf70bb1781b1912a/check-runs?per_page=100"
{"total": 66, "by_conclusion": {"skipped": 12, "success": 54}, "by_status": {"completed": 66}}
```

Zero failures, zero cancelled, zero neutral, nothing still running. `CI` itself:
`{"name": "CI", "conclusion": "success", "head_sha": "fd68ed3a333c36077aaba181cf70bb1781b1912a"}`.

---

## Verification

Every command below ran in `/Users/roelschuurkes/rev-2236`, a worktree detached at
`fd68ed3a333c36077aaba181cf70bb1781b1912a`, with `python3` 3.x and `bash` from macOS 25.6.0.

```
$ git rev-parse HEAD
fd68ed3a333c36077aaba181cf70bb1781b1912a

$ python3 scripts/ci/check-ci-gate-coverage.py
ci-gate-coverage=passed                                    # exit 0

$ python3 scripts/ci/check-ci-gate-coverage.py --self-test
ci-gate-coverage self-test=passed                          # exit 0

$ bash scripts/ci/test-ci-gate-expectations.sh
ok: a complete green run passes
ok: a documentation-only run passes with its jobs scoped out
ok: a code-gated job that silently did not run fails the gate, and is named
ok: ebpf-smoke-ubuntu skipped while required fails the gate
ok: mcp-registry-foundation skipped while touched fails the gate
ok: a job with no condition may not be skipped even on a docs-only run
ok: a failed release guardrail fails the required gate
ok: an outright failure fails, and a skipped scope fails
ok: a failed or missing public-msrv result fails closed and names the job
ok: an empty lightweight_only is treated as the strict case, not the permissive one
ok: empty or wrong-case == 'true' outputs are strict, so a disarmed job cannot read green
ok: an explicit false still scopes its job out
PASS: CI gate expectation contract                          # exit 0
```

Live enforcement, for angle 3:

```
$ gh api "repos/Rul1an/assay/rulesets?includes_parents=true"
[]

$ gh api repos/Rul1an/assay/branches/main/protection --jq '.required_status_checks.contexts'
["CI", "host-capability-check", "lane-check/proof"]
```

The breakage table in angle 4 is the output of `/tmp/rev2236/attack.py`, which mutates the workflow
text in memory and calls `guard.check()` — nothing was written to the reviewed tree. The single
on-disk mutation, case E's re-run of the expectations harness, was made in a throwaway copy under
`/tmp`, not in the worktree.

Tree clean afterwards:

```
$ git status --porcelain && echo CLEAN
CLEAN
$ git diff --exit-code   # exit 0
```

### Revalidation, recorded before the merge

`main` has advanced to `c2c41b038` (#2213) and this branch shows `BEHIND` under
`strict_required_status_checks_policy`. `AGENTS.md` carries a review across an upstream advance only
on a **recorded equivalence check with two conditions**. I verified both against a local probe merge
rather than taking the argument on trust; the builder's reasoning happens to be right, but "touches
none of the five files" is only the second condition and the rule requires both.

One commit is missing: `c2c41b038`, touching `crates/assay-cli/src/exit_codes.rs`,
`crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md`,
`crates/assay-cli/tests/spec_reason_code_registry.rs`, `docs/architecture/SPEC-PR-Gate-Outputs-v1.md`.

Probe: `git merge --no-ff --no-commit c2c41b038` in the review worktree → "Automatic merge went well",
`git diff --name-only --diff-filter=U` empty, so **no conflicts**. Tree written as probe commit
`9cc3baebf0a6199eca00294af3e9296fe0be9c44` (local only, never pushed).

**Condition 1 — the new head introduces no change, to any file, that is not already on `main`.**
`git diff --name-only c2c41b038 9cc3baeb` returns exactly the five files this PR changes, and the
diff is byte-identical to the reviewed one:

```
$ diff <(git diff c2c41b038...fd68ed3a3) <(git diff c2c41b038 9cc3baeb) && echo IDENTICAL
IDENTICAL
```

This is a whole-tree comparison, not a file list, so nothing outside the reviewed files can ride the
carry.

**Condition 2 — the advance touched no file the review covered.**
`git diff --name-only fd68ed3a3 9cc3baeb` returns exactly the four `c2c41b038` files above. The
intersection with this PR's five reviewed files is empty.

Both conditions hold, so this review carries to a head that is `fd68ed3a3` plus `c2c41b038` and
nothing else. The worktree was reset to `fd68ed3a3` and is clean.

---

## Non-claims

- **Static analysis only, not executed under GitHub Actions.** Every statement about what the rollup
  concludes comes from reading `ci.yml` and from running the extracted gate body under injected
  environments via `test-ci-gate-expectations.sh` locally. I did not observe a real `publish-shape-cli`
  failure turn the `CI` context red on GitHub's runners. The harness executes the same shell the
  workflow embeds, which is the strongest local evidence available, and it is not the same thing as
  the live behaviour.
- **F-1's job-level variant is documentation-derived, not measured.** GitHub documents
  conclusion-substitution for `continue-on-error` *steps*; it does not document
  `needs.<job>.result` under a job-level `continue-on-error`. My case F establishes only that the
  guard is silent about the construct. The step-level fail-open stands on the documented sentence.
- **I did not verify that live branch protection will still require `CI` tomorrow.** F-2 is a
  statement about what no artifact checks, read off one API response taken today.
- **`lane-check/proof` and `host-capability-check` were confirmed present and green at this head, and
  not audited further.** Whether either gate is internally sound is outside this review.
- **No claim about the ~20 non-required PR checks.** I counted PR-triggered workflows and read the
  contract section that designates advisory lanes; I did not evaluate whether each is correctly
  classified.
- **I did not run the Rust test suite or clippy.** The change touches no Rust; the five changed files
  are a workflow, a hook config, a contract document, and two scripts. `cargo` verification would
  have measured nothing about this diff.

## Merge

The three auto-merge preconditions in `AGENTS.md` are met on `fd68ed3a3`: required checks green
(`CI`, `lane-check/proof`, `host-capability-check` all `success`, zero failures across 66 check runs),
this review posted on that head as the one non-building agent review, and every actionable finding
either fixed, dispositioned, or filed. No required check was weakened, bypassed, or relabelled to get
here — the `BEHIND` state is being resolved by satisfying the strict-update requirement, not by
routing around it, and the review carries under the recorded equivalence check above. Squash-merging.

F-1 is #2242, F-2 is #2243, F-3 is #2244. None blocks this merge.

Rul1an:
## Revalidation record — review carried from `fd68ed3a3` to `e95c692fac3f9bf46aefda1b0f9a623c73b965d5`

Same non-building reviewer. `gh pr merge --squash` refused on `fd68ed3a3` with "the head branch is
not up to date with the base branch". `--admin` would have bypassed a required check and is therefore
not available under `AGENTS.md`, so the strict-update requirement was **satisfied rather than routed
around**: I updated the branch from `main` with `expected_head_sha=fd68ed3a3` and am carrying the
review under the recorded equivalence check. This push introduces no change of mine — it is the merge
of `c2c41b038` (#2213) that the strict policy demanded.

The new head is exactly the probe merge I verified before posting the review, which is the strongest
form the equivalence check can take:

```
$ git rev-list --parents -n 1 e95c692fa
e95c692fac3f9bf46aefda1b0f9a623c73b965d5 fd68ed3a333c36077aaba181cf70bb1781b1912a c2c41b038e89ff2a8029c84334261126fdade987

$ git rev-parse e95c692fa^{tree}   # actual new head
237955d29971e9b9332a6c5a01df0b0e55042e8a
$ git rev-parse 9cc3baebf^{tree}   # probe merge from the review above
237955d29971e9b9332a6c5a01df0b0e55042e8a
```

Two parents, no third; the tree is byte-identical to the probe. Both conditions re-verified against
the real head rather than inherited from the probe:

**Condition 1 — introduces no change, to any file, that is not already on `main`.**

```
$ git diff --name-only c2c41b038 e95c692fa
.github/workflows/ci.yml
.pre-commit-config.yaml
CI-CONTRACT.md
scripts/ci/check-ci-gate-coverage.py
scripts/ci/test-ci-gate-expectations.sh

$ diff <(git diff c2c41b038...fd68ed3a3) <(git diff c2c41b038 e95c692fa) && echo DIFF IDENTICAL
DIFF IDENTICAL
```

A whole-tree comparison, not a file list: exactly the five reviewed files differ from `main`, and the
diff is the one I reviewed.

**Condition 2 — the advance touched no file the review covered.**

```
$ git diff --name-only fd68ed3a3 e95c692fa
crates/assay-cli/src/exit_codes.rs
crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md
crates/assay-cli/tests/spec_reason_code_registry.rs
docs/architecture/SPEC-PR-Gate-Outputs-v1.md
```

Empty intersection with the five reviewed files.

Both hold, so the review above counts for `e95c692fa`. This is an upstream-advance carry, not a
rewritten history: `fd68ed3a3` is a parent of the new head, so no rebase or squash is involved.

Guards re-run on the new head in `/Users/roelschuurkes/rev-2236` at `e95c692fa` — not inferred from
the unchanged diff:

```
$ python3 scripts/ci/check-ci-gate-coverage.py
ci-gate-coverage=passed
$ python3 scripts/ci/check-ci-gate-coverage.py --self-test
ci-gate-coverage self-test=passed
$ bash scripts/ci/test-ci-gate-expectations.sh
PASS: CI gate expectation contract
$ git status --porcelain   # empty
```

Merging by squash once the required contexts report green on `e95c692fa`. Findings unchanged: none
blocking, follow-ups #2242, #2243, #2244.

---

### Merged

Squash-merged with `--match-head-commit e95c692fac3f9bf46aefda1b0f9a623c73b965d5`, so the merge is
pinned to the head this review carries to and could not have landed a head that moved underneath it.

- merge commit on `main`: **`802a033005b308ee0d9437ae4d6953f02361147a`** — "fix(ci): wire the two
  unenforced release guardrails into the required CI gate (#2236)"
- #2230 closed as `COMPLETED` at 16:03:58Z
- required contexts on `e95c692fa`: `CI=success`, `host-capability-check=success`,
  `lane-check/proof=success`; 62 check runs, 49 success, 13 skipped, **zero failures**

Worth recording that the `CI` rollup ran and passed on `e95c692fa` *with the new guard as its first
step*, which is the one thing local reproduction could not establish: the guard now executes inside
the required context, not only in a pre-commit hook.

The branch `cursor/wire-unenforced-ci-guardrails-d139` and the builder's worktree at
`/Users/roelschuurkes/wt-2236` were left in place. `AGENTS.md` gates their removal on recording the
merge in the programme ledger, and no programme is active, so there is no ledger to record it to —
and the worktree is the builder's, not mine. My own review worktree has been removed.

PR #2238, opened 2026-08-10
ci: invoke cargo-audit as the binary, and gate the cargo-deny install pin

### LINKED ISSUE #2237: cargo-audit still goes through cargo, and pre-commit has no env-var cover for it (#2214 residual)
Follow-up to #2214, which is closed by #2218 (merged as `cd8dc5d6f`). The defect #2214 named is
fixed for `cargo-deny`. The same defect class survives for `cargo-audit`, and the two halves are now
defended by mechanisms of different strength. All observations below are read at `cd8dc5d6f` on
`origin/main`, the tip at the time of filing.

## What #2218 fixed, and how

Two independent remedies landed:

1. `scripts/ci/install-cargo-deny.sh` pins the version in one place, exports `RUSTUP_TOOLCHAIN`, and
   asserts the resolved version against the pin. The invocation is `cargo-deny check ...` — the
   binary, not `cargo deny` — so it never re-enters `rust-toolchain.toml` resolution.
2. `.github/workflows/ci.yml` sets `RUSTUP_TOOLCHAIN: stable` for the whole `deps` job (line 171),
   which states the resolution for every cargo command in that job.

Remedy 1 holds regardless of the environment. Remedy 2 holds only while that `env:` block survives.
`ci.yml:184-187` says so itself: the direct-binary form "is kept because it does not depend on that
variable surviving a later edit."

## Residual 1 — `cargo audit` is still invoked through `cargo`, and pre-commit has no env-var cover

Three sites still go through `cargo`:

- `.github/workflows/ci.yml:193` — `cargo audit --ignore RUSTSEC-2023-0071`
- `.github/workflows/ci.yml:213` — `cargo audit --file .../fuzz/Cargo.lock`
- `.pre-commit-config.yaml:318` — `entry: bash -lc 'cargo audit --ignore RUSTSEC-2023-0071'`

The two CI sites are covered by the job-level `RUSTUP_TOOLCHAIN`, so they are not failing today; they
are one edit away from failing again, and the edit that breaks them is a narrowing of an `env:` block
that mentions neither of them by name.

**The pre-commit site has no such cover.** It runs on a developer machine inside the repository, where
`rust-toolchain.toml` selects 1.96.0 and nothing sets `RUSTUP_TOOLCHAIN`. Its immediate neighbour is
the `cargo-deny` hook at `:294-301`, which carries a comment explaining precisely why going through
`cargo` is wrong:

> `cargo-deny` rather than `cargo deny`: going through cargo resolves the toolchain from
> rust-toolchain.toml, which in CI selects a toolchain the lint job never installed.

Same file, adjacent hooks, same defect class, one defended and one not. That is the "one rule, one
function" failure mode from `AGENTS.md`: the rule is answered in two places and the second answer
drifted.

**The direct form is verified, not assumed.** Measured locally with `cargo-audit 0.22.0` at
`/Users/roelschuurkes/.cargo/bin/cargo-audit`, from `/Users/roelschuurkes/assay` at `cd8dc5d6f`:

```
$ cargo-audit audit --version
cargo-audit-audit 0.22.0
$ cargo-audit audit --ignore RUSTSEC-2023-0071
... warning: 5 allowed warnings found
```

So `cargo-audit audit ...` accepts the same flags and produces the same audit as `cargo audit ...`.

## Residual 2 — `cargo-audit` is installed unpinned while its sibling is pinned

`.github/workflows/ci.yml:179` is `cargo install --locked cargo-audit`, two lines below a pinned
`cargo-deny` install. `--locked` pins the tool's *dependencies*, not the tool version, so the version
installed is whatever crates.io publishes on the day the job runs. `AGENTS.md` states the rule this
crosses:

> Pin a tool version in one place, and have both the install and the invocation read that one value.

There may be a deliberate reason to float this one — an advisory scanner arguably wants the newest
release, since a pinned scanner silently stops learning about new advisories. That trade-off is
defensible; it is currently undocumented, which is the actual defect. An unexplained asymmetry beside
a pinned sibling reads as an oversight.

## Residual 3 — `install-cargo-deny.sh` has no self-test

`scripts/ci/install-cargo-deny.sh` carries the pin, the toolchain export, and the version assertion,
and nothing exercises any of them. Mutation controls run against a local copy show what that costs:
deleting `--version`, deleting the `export RUSTUP_TOOLCHAIN`, and rewriting the version assertion to
`if false; then` are each silent — the script still exits 0 and still prints a version line that
reads as evidence of a pin.

A self-test with stub `cargo` and `cargo-deny` binaries on `PATH` catches all three. One caveat worth
recording for whoever writes it: the obvious shape of such a script is fail-open. A first draft here
printed `FAIL`, then aborted on an unset variable under `set -u`, and **exited 0** — a gate that
reports a failure and returns success. The abort path needs an exit code of its own.

## Why this is `horizon:post-mvp`

#2214 was MVP-blocking because it reddened PRs at random and blocked merges. This is not that: CI is
green today, and the pre-commit exposure surfaces as a clear toolchain error on a `Cargo.lock` change
rather than as a silent pass. It is a robustness and consistency defect that makes the #2214 class
able to return, so it wants fixing before someone edits that `env:` block — not before MVP.

## Acceptance

- [ ] All three `cargo audit` sites invoke the binary directly, so the fix does not depend on an
      `env:` block that names neither of them.
- [ ] The `cargo-audit` install either reads a pin from one place, or states in the workflow why its
      version deliberately floats.
- [ ] `install-cargo-deny.sh` has a self-test that fails when the pin, the toolchain export, or the
      version assertion is removed, and whose own abort path exits non-zero.
- [ ] The self-test runs in CI or as a pre-commit hook, so it is a gate and not a script someone can
      forget to run.

Rul1an:
Closes #2237. Follow-up to #2214, which #2218 closed as `cd8dc5d6f`.

All measurements below were taken at commit `2dd803defb9463a8811b3e0f9c19af655d753e70`, in worktree
`/Users/roelschuurkes/wt-2214`, macOS aarch64, `rustc`/`cargo 1.96.0` from `rust-toolchain.toml`,
`cargo-deny 0.20.2` and `cargo-audit 0.22.0` from `~/.cargo/bin`.

## Why this exists

#2218 fixed the `cargo-deny` half of #2214 two independent ways, and the two are not equally strong:

1. `scripts/ci/install-cargo-deny.sh` pins the version in one place and invokes `cargo-deny` directly,
   so it never re-enters `rust-toolchain.toml` resolution. This holds regardless of environment.
2. `.github/workflows/ci.yml:171` sets `RUSTUP_TOOLCHAIN: stable` for the whole dependency job. This
   holds only while that `env:` block survives.

`cargo-audit` was left resting entirely on the second, and `ci.yml` says as much about its neighbour:
the direct-binary form "is kept because it does not depend on that variable surviving a later edit."

**The pre-commit hook has no `env:` block at all.** It runs on a developer machine inside the
repository, where `rust-toolchain.toml` selects 1.96.0 and nothing overrides it. Its immediate
neighbour, the `cargo-deny` hook, carries a comment explaining exactly why going through `cargo` is
wrong. Same file, adjacent hooks, same defect class, one defended and one not — the
one-rule-one-function failure mode from `AGENTS.md`, with the second answer already drifted.

## What changed

**Three `cargo audit` sites now invoke the binary**: `.github/workflows/ci.yml:193` and `:213`, and
`.pre-commit-config.yaml:318`. The `cargo-audit` hook gains the reasoning its neighbour already had, so
the next reader does not have to rediscover why the form matters.

**The invocation form is measured, not assumed.** This is a security gate, so the form was checked
against the real lockfile before the workflow was edited:

```
$ cargo-audit audit --version
cargo-audit-audit 0.22.0
$ cargo-audit audit --help | head -3
Audit Cargo.lock files for vulnerable crates
Usage: cargo audit [OPTIONS] [COMMAND]
$ cargo-audit audit --ignore RUSTSEC-2023-0071      # from the repo root
... warning: 5 allowed warnings found
```

**`install-cargo-deny.sh` gains a self-test, wired as a gate.** The script carried the pin, the
toolchain export and the version assertion with nothing exercising any of them. Three mutations
applied to a local copy at this commit, each reverted after measuring:

| mutation | before | after |
|---|---|---|
| delete `--version "${CARGO_DENY_VERSION}"` from the install | silent, exit 0 | caught, exit 1 |
| turn `export RUSTUP_TOOLCHAIN=…` into a plain assignment | silent, exit 0 | caught, exit 1 |
| rewrite the version assertion to `if false; then` | silent, exit 0 | caught, exit 1 |

"Silent" means the script still exited 0 *and still printed a version line that reads as evidence of a
pin* — the worst shape for this defect, because the log looks like proof.

The self-test drives stub `cargo` and `cargo-deny` binaries on `PATH`, so it needs no network and no
install; it runs in about 0.4s. It is registered as the `install-cargo-deny-self-test` pre-commit hook,
triggered by changes to either script or to `.pre-commit-config.yaml`, so it is a gate rather than a
script someone has to remember.

**The install script also reports which binary answered**, not only which version. `cargo-deny
--version` resolves through `PATH` with no relationship to where `cargo install` wrote, so an unrelated
`cargo-deny` reporting the pinned version would have satisfied the assertion while the install did
nothing. Asserting the path would be stricter and was deliberately not done — `CARGO_INSTALL_ROOT` is
unset in both jobs, and shadowing `~/.cargo/bin` on a runner is not a condition this repository has
met. The comment records that reach so the next reader knows what the check does not cover.

## One defect found in this PR's own gate, and fixed

The first draft of the self-test was fail-open. Under the toolchain-export mutation the install script
died on an unbound variable; the self-test printed `FAIL pinned: exit 1, expected 0`, then died itself
on an unset case variable under `set -u`, **and exited 0**:

```
== the install names the pinned version explicitly ==
FAIL pinned: exit 1, expected 0
      …/install-cargo-deny.sh: line 38: RUSTUP_TOOLCHAIN: unbound variable
…/test-install-cargo-deny.sh: line 83: CASE_LOG: unbound variable
SELFTEST_EXIT=0
```

A gate that reports a failure and returns success is worse than no gate. Two causes, both fixed: the
case log path is now published before the exit-code comparison rather than after it, so a failing case
cannot leave the next assertion reading an unset variable; and an `ERR` trap names the abort so it
cannot be read as a pass. The table above was re-measured after both fixes.

## Verification at `2dd803def`

| command | outcome |
|---|---|
| `bash scripts/ci/test-install-cargo-deny.sh` | 10 cases pass, exit 0 |
| the three mutations above, each applied and reverted | exit 1 each, table above |
| `pre-commit run install-cargo-deny-self-test --all-files` | `Passed` |
| `pre-commit validate-config .pre-commit-config.yaml` | exit 0 |
| `shellcheck scripts/ci/test-install-cargo-deny.sh` | exit 0 |
| `cargo-audit audit --ignore RUSTSEC-2023-0071` | runs, `5 allowed warnings found` |
| full pre-commit suite on commit | all hooks pass, including the new gate |
| pre-push: `cargo clippy` workspace `-D warnings`, linux compile gate | both pass |

## Not claimed

- **The unpinned `cargo-audit` install is left unpinned.** It now states why: an advisory scanner is the
  one tool where a pin ages into a silent gap rather than a loud failure. That reasoning is stated, not
  settled — #2237 carries the question of whether it survives scrutiny.
- I did not verify the workflow change on a GitHub runner. The invocation form was measured locally;
  whether the `deps` job stays green is CI's answer, not mine.
- The self-test asserts the install script's *argv and its version gate*. It does not verify that
  `cargo install` actually placed a binary, because it stubs `cargo` — by design, so the test needs no
  network. That boundary is stated in the script's own header.
- Six other `cargo install` sites in CI workflows are untouched and unpinned. They are outside this
  change and are not claimed to be fixed; #2237's acceptance list scopes only the `cargo-audit` one.
- Nothing here touches evidence ingestion, evidence verification, or MCP authorization, so this is not
  an ADR-042/043 boundary change. No public-facing string makes a compliance, certification or
  whole-action claim.


Rul1an:
## Interim review record — reproduction published before analysis completes

**Reviewer status:** non-building. I authored no commit on `cursor/2237-cargo-audit-direct-binary`,
and no line of #2237, #2238, #2218 or `AGENTS.md`. I will not edit this branch; if I conclude a code
change is required I will say so and stop.

**Head measured:** `2dd803defb9463a8811b3e0f9c19af655d753e70` (confirmed via
`gh pr view 2238 --json headRefOid`). Base `main` at `802a033005b308ee0d9437ae4d6953f02361147a`
(the #2236 merge). Merge base `cd8dc5d6f43ac32d1dc7a987f193c0204ccdf073`.
Review worktree `/Users/roelschuurkes/rev-2238` (detached at the head above), macOS aarch64,
`cargo-audit 0.22.0` and `cargo-deny 0.20.2` from `~/.cargo/bin`.

This is published now rather than at the end because a review is an artifact bound to a SHA and
unpublished analysis is not a review. A final record follows on whatever head this PR merges from.

### Angle 1 — the `main` overlap on `ci.yml`: established, and the carry is refused

#2236 (merged as `802a033`) changed **`.github/workflows/ci.yml` and `.pre-commit-config.yaml`** —
**both** files this PR changes. So the second `AGENTS.md` equivalence condition ("the advance touched
no file the review covered") is **not satisfied**, and nothing carries. I am reviewing the merged
result directly, as instructed.

Local merge of `802a033` into `2dd803def`: **clean, no conflicts**, `Auto-merging` both overlapping
files. The two changes occupy disjoint regions of `ci.yml` — #2238 edits steps *inside* the existing
`deps-security` job (lines ~178–218); #2236 edits the `ci` rollup's `needs:`/decision table
(lines ~756, ~927–1010) and adds a marker comment on `ebpf-smoke-self-hosted`. In
`.pre-commit-config.yaml`, #2236 inserts at ~262 and #2238 at ~308–339. Neither undoes any part of
the other.

**#2236's guard run on the combination** (not on either change alone):

```
$ python3 scripts/ci/check-ci-gate-coverage.py --self-test
ci-gate-coverage self-test=passed      # exit 0
$ python3 scripts/ci/check-ci-gate-coverage.py
ci-gate-coverage=passed                # exit 0
```

The guard's rule is that every job in `ci.yml` reaches the rollup's `needs:` and decision table or
carries a `# ci-gate: not-required` marker. #2238 **adds no job** — it changes steps within a job
already in the rollup — so the combination is coherent by construction, and the guard confirms it
rather than being taken on faith.

### Angle 3 — the toolchain-bypass claim: reproduced, with a discriminating control

The PR claims invoking the binary directly never re-enters `rust-toolchain.toml` resolution. I did not
accept it. `rustup` does not own these binaries:

```
$ rustup which cargo-fmt
/Users/…/.rustup/toolchains/1.96.0-aarch64-apple-darwin/bin/cargo-fmt
$ rustup which cargo-audit
error: unknown binary 'cargo-audit' in toolchain '1.96.0-aarch64-apple-darwin'
$ rustup which cargo-deny
error: unknown binary 'cargo-deny' in toolchain '1.96.0-aarch64-apple-darwin'
```

`~/.cargo/bin/cargo-audit` is the real binary; `~/.cargo/bin/cargo-fmt` is a rustup shim. Both are
`Mach-O` executables in the same directory, so file type does not distinguish them — which is why the
**control** below matters: it shows the test can fail.

```
$ RUSTUP_TOOLCHAIN=9.9.9-bogus cargo-fmt --version        # rustup-shimmed control
error: override toolchain '9.9.9-bogus' is not installed…   → exit 1
$ RUSTUP_TOOLCHAIN=9.9.9-bogus cargo audit --version      # through cargo
error: override toolchain '9.9.9-bogus' is not installed…   → exit 1
$ RUSTUP_TOOLCHAIN=9.9.9-bogus cargo-audit audit --version # direct binary
cargo-audit-audit 0.22.0                                    → exit 0
```

**The pre-commit case, where there is no `env:` cover at all**, is the one that matters most, so I
reproduced it against the hook's exact `entry:` string with `RUSTUP_TOOLCHAIN` unset, by mutating
`rust-toolchain.toml` to a channel that is genuinely not installed:

| step | command | result |
|---|---|---|
| mutate | `channel = "1.96.0"` → `"1.94.0-not-installed"` in `/Users/roelschuurkes/rev-2238/rust-toolchain.toml` | — |
| old form (`main`) | `env -u RUSTUP_TOOLCHAIN bash -lc 'cargo audit --ignore RUSTSEC-2023-0071'` | **exit 1** — `error: custom toolchain '1.94.0-not-installed' specified in override file '…/rust-toolchain.toml' is not installed` |
| new form (this PR, `:323`) | `env -u RUSTUP_TOOLCHAIN bash -lc 'cargo-audit audit --ignore RUSTSEC-2023-0071'` | **exit 0** — `warning: 3 allowed warnings found` |
| neighbour (already fixed by #2218) | `env -u RUSTUP_TOOLCHAIN bash -lc 'cargo-deny --version'` | **exit 0** — `cargo-deny 0.20.2` |
| restore | `cp /tmp/rev2238-rust-toolchain.toml.bak rust-toolchain.toml` — **exact path, no bare-tree command** | sha256 `fa0addad…4d30c0`, matches pre-mutation; `git diff -- rust-toolchain.toml` empty |

So the defect is real, the fix removes it, and the fix is not resting on an `env:` block. The guard
bites: the old form fails where the new form passes, in exactly the environment the pre-commit hook
runs in.

*(Note on numbers: I measure `3 allowed warnings`, the PR body measured `5`. The advisory database is
live and mine is a different snapshot — not a discrepancy in the change.)*

### Angle 2 — literal count for the pin: **one**

```
$ git grep -n -- '0\.20\.2' 2dd803def -- . ':!*.lock' ':!Cargo.toml'
scripts/ci/install-cargo-deny.sh:16:CARGO_DENY_VERSION="0.20.2"
```

Exactly one literal repo-wide. The new self-test does **not** restate it — it reads it back out of
the one place with
`PIN="$(sed -n 's/^CARGO_DENY_VERSION="\(.*\)"$/\1/p' "${UNDER_TEST}")"` and fails loudly if the
assignment's shape moves. That is one-rule-one-function done as the rule asks, not a parity test as
the fallback.

Angles 4, 5 and 6, and the remaining verification, follow in the final record.

Rul1an:
## Final review record — non-building review, head `d95ae972df1c77b4565e1c1dc4eb6583b25435ee`

**Reviewer status: non-building.** I authored no commit on `cursor/2237-cargo-audit-direct-binary`
and no line of #2237, #2238, #2218, `AGENTS.md` or `CI-CONTRACT.md`. I made no edit to this branch.
This record supersedes my [interim comment](https://github.com/Rul1an/assay/pull/2238#issuecomment-5242839459),
which was published at first reproduction on the pre-update head `2dd803def` so the analysis could not
be lost; both stand, and this one is the record the merge rests on.

**Verdict: APPROVE — merge.** One Medium finding in the PR's own new code, dispositioned as
follow-up and filed as #2250 with its measurement; nothing blocking.

**Provenance.** Head `d95ae972df1c77b4565e1c1dc4eb6583b25435ee`, confirmed with
`gh pr view 2238 --json headRefOid`. Reviewed head tree `a90fa3a2997442784e3b15debaeb9056540a2aad`.
Parents `2dd803def` (the reviewed PR head) and `802a033` (`main`, the #2236 merge). Merge base
`cd8dc5d6f`. Review worktree `/Users/roelschuurkes/rev-2238`, detached, its own `target/`, no
`CARGO_TARGET_DIR` exported. macOS aarch64, `/bin/bash` 3.2.57, `python3` 3.x, `pre-commit` 4.6.0,
`cargo-audit 0.22.0` and `cargo-deny 0.20.2` from `~/.cargo/bin`, `rustc`/`cargo 1.96.0` from
`rust-toolchain.toml`.

### Branch update, and why nothing was carried

The branch was `BEHIND` and the ruleset sets `strict_required_status_checks_policy: true`, so an
update was mandatory, not optional. I updated it with `gh pr update-branch` (a merge — a rebase would
rewrite history, which `AGENTS.md` says never carries a review).

Recording the two equivalence conditions explicitly:

1. **The new head introduces no change, to any file, that is not already on `main`. SATISFIED, by
   whole-tree identity rather than by a file list.**
   `git merge-tree --write-tree 802a033 2dd803def` → `a90fa3a2997442784e3b15debaeb9056540a2aad`;
   `git rev-parse d95ae972d^{tree}` → `a90fa3a2997442784e3b15debaeb9056540a2aad`. Identical, so the
   pushed head's tree is exactly what merging `main` into the reviewed head produces without conflicts.
   Corroborated: `git diff 802a033 d95ae972d` is the PR's own diff and nothing else — same five files,
   `159 insertions(+), 5 deletions(-)`, differing from `git diff cd8dc5d6f 2dd803def` only in blob
   hashes and in hunk offsets shifted by the 16 lines #2236 inserted earlier in
   `.pre-commit-config.yaml`. No `+`/`-` line differs.

2. **The advance touched no file the review covered. NOT SATISFIED.** #2236 changed
   `.github/workflows/ci.yml` *and* `.pre-commit-config.yaml`; this PR changes both. So **nothing
   carries**, and I did not carry it. Every result below was re-measured on `d95ae972d` directly.

### Angle 1 — the `main` overlap on `ci.yml`: clean, and the guard passes on the combination

The merge was clean (`Auto-merging` both files, no conflict markers, no manual resolution). The two
changes are independently coherent because they occupy disjoint structural regions:

| file | #2236 | #2238 |
|---|---|---|
| `ci.yml` | the `ci` rollup's `needs:`, its decision table, a `# ci-gate: not-required` marker on `ebpf-smoke-self-hosted` (~756, ~927–1010) | two `run:` invocations and three comments *inside* the existing `deps-security` job (~178–218) |
| `.pre-commit-config.yaml` | new `ci-gate-coverage` hook (~262) | comment on `cargo-audit`, new `install-cargo-deny-self-test` hook (~324–347) |

Neither undoes any part of the other. The decisive point: **#2238 adds no job**, so it cannot change
what the rollup must wait on. #2236's guard, run on the combination rather than on either change alone:

```
$ python3 scripts/ci/check-ci-gate-coverage.py --self-test
ci-gate-coverage self-test=passed                          → exit 0
$ python3 scripts/ci/check-ci-gate-coverage.py
ci-gate-coverage=passed                                    → exit 0
$ bash scripts/ci/test-ci-gate-expectations.sh
PASS: CI gate expectation contract                         → exit 0
$ pre-commit run ci-gate-coverage --all-files
Required CI gate waits on every gating job.......Passed
```

The self-test is run first, so the guard is shown able to fail before it is believed when it passes.
#2236's sibling gate (`test-ci-gate-expectations.sh`) also passes on the combination.

### Angle 2 — the pin: **exactly one literal**, and the echo's reach is stated

```
$ git grep -c -- '0\.20\.2' d95ae972d -- . ':!*.lock' ':!Cargo.toml'
d95ae972d:scripts/ci/install-cargo-deny.sh:1
```

One occurrence, in one file, repo-wide. The new self-test does **not** restate it — it reads the value
back out of the single place with
`PIN="$(sed -n 's/^CARGO_DENY_VERSION="\(.*\)"$/\1/p' "${UNDER_TEST}")"`, and exits 1 with a named
reason if that assignment's shape moves. So install, assertion and test all read one value. This is
one-rule-one-function done the way the rule prefers — one function, not the parity-test fallback — and
the PR **removes** an instance of the two-literal defect rather than adding one.

**What the echo proves, and what it does not.** `install-cargo-deny.sh:52` now prints
`cargo-deny resolved: ${resolved} at $(command -v cargo-deny)`. Under the `if false;` mutation the log
read:

```
cargo-deny resolved: cargo-deny 0.0.0-not-the-pin at /var/folders/…/mismatch/cargo-deny
```

— a stub reported both a version *and* a path, and only the assertion caught it. So the echo makes a
version claim checkable against the log and names which `PATH` entry answered; it does **not** prove
that `cargo install` wrote that binary. The script's own comment at lines 45–50 states exactly that
reach, including why asserting the path was declined. On `AGENTS.md`'s alias caveat: `command -v`
would expose PATH shadowing, and would not expose a shell function of the same name — an exposure the
comment's scope covers in substance.

### Angle 3 — the bypass claim: reproduced, with a control that shows the test can fail

`rustup` does not own these binaries, which is *why* the direct form works:

```
$ rustup which cargo-fmt     → …/.rustup/toolchains/1.96.0-aarch64-apple-darwin/bin/cargo-fmt
$ rustup which cargo-audit   → error: unknown binary 'cargo-audit' in toolchain '1.96.0-…'
$ rustup which cargo-deny    → error: unknown binary 'cargo-deny' in toolchain '1.96.0-…'
```

`~/.cargo/bin/cargo-audit` is the real binary; `~/.cargo/bin/cargo-fmt` is a rustup shim. Both are
`Mach-O` files in the same directory, so file type does not distinguish them — hence the control:

```
$ RUSTUP_TOOLCHAIN=9.9.9-bogus cargo-fmt --version        → exit 1 (shimmed CONTROL: the test can fail)
$ RUSTUP_TOOLCHAIN=9.9.9-bogus cargo audit --version      → exit 1 (through cargo)
$ RUSTUP_TOOLCHAIN=9.9.9-bogus cargo-audit audit --version → exit 0, cargo-audit-audit 0.22.0
```

**The pre-commit site, the one with no `env:` cover, measured against the hook's exact `entry:` string
with `RUSTUP_TOOLCHAIN` unset** (mutation/restore table at the end):

| form | result |
|---|---|
| `cargo audit --ignore RUSTSEC-2023-0071` (before this PR) | **exit 1** — `error: custom toolchain '1.94.0-not-installed' specified in override file '…/rust-toolchain.toml' is not installed` |
| `cargo-audit audit --ignore RUSTSEC-2023-0071` (this PR, `:339`) | **exit 0** — `warning: 3 allowed warnings found` |
| `cargo-deny --version` (the neighbour #2218 already fixed) | **exit 0** — `cargo-deny 0.20.2` |

The defect is real, the fix removes it, and it does not rest on an `env:` block surviving an edit.
The CI fuzz-step form was checked separately, because a security gate's argv is worth measuring rather
than inferring: `( cd /tmp && cargo-audit audit --file …/fuzz/Cargo.lock )` →
`Scanning …/fuzz/Cargo.lock for vulnerabilities (336 crate dependencies)`, exit 0. `cargo-audit audit
<flags>` is argv-identical to what `cargo` itself execs for `cargo audit`, which is why the flags carry.

Nothing on `PATH` in CI resolves differently: `ubuntu-latest` puts `~/.cargo/bin` on `PATH`, which is
where `cargo install --locked cargo-audit` writes, and rustup adds no `cargo-audit` shim to compete.

### Angle 4 — symmetry with the neighbour: achieved on the axes #2237 was about

| | `cargo-deny` hook (`:311`) | `cargo-audit` hook (`:328`) |
|---|---|---|
| invocation | binary — `cargo-deny check …` | binary — `cargo-audit audit …` ✅ now |
| rationale comment | present | present ✅ now, and cites the neighbour by name |
| version assertion in-hook | none, by stated design | none — no pin exists |
| stage | `[pre-commit]` | `[pre-push]` |

The two axes that constituted the defect — invocation shape and stated rationale — are now genuinely
symmetric, and the new comment cross-references the neighbour, which is what makes them resist drifting
apart again rather than merely matching today. The residual asymmetries (stage, and the absent pin) are
pre-existing, deliberate, and now documented.

**Is a third tool of the same class left alone?** In the files this PR touches, **no**, and I checked
rather than assumed. `.pre-commit-config.yaml` has two other cargo invocations — `cargo fmt --all --
--check` (`:303`) and `scripts/precommit/cargo-clippy.sh` (`:358`) — but `rustfmt` and `clippy` **are**
rustup toolchain components (`rustup which cargo-fmt` resolves inside the toolchain, above), so they
*must* resolve through `rust-toolchain.toml`. That is the opposite requirement, not the same defect
left unfixed. In `ci.yml`, `cargo-deny` and `cargo-audit` are the only `cargo install`ed third-party
tools. Other workflows do carry same-class unpinned installs (`cargo-nextest`, `cargo-hack`,
`cargo-semver-checks` in `split-wave0-gates.yml`; `cargo-mutants` in a script) and two inline-pinned
ones (`cargo-cyclonedx 0.5.9`, `cargo-fuzz 0.13.2`); the PR's "Not claimed" section discloses these as
untouched, and that disclosure holds.

### Angle 5 — blast radius, and no gate weakened

**No required check is weakened, relabelled or bypassed.** Mechanically checked, not eyeballed: across
the whole `.github/` portion of the diff there is not one `+` line matching
`continue-on-error`, `|| true`, `exit 0`, or any change to an `if:`, `needs:`, `name:` or `required`
line. The change is two `run:` invocations plus comments. Required contexts (`CI`,
`lane-check/proof`, `host-capability-check`) are untouched, and the PR adds a gate rather than removing
one.

**Developer, locally:** the `cargo-audit` hook stops depending on `rust-toolchain.toml` naming an
installed toolchain — measured above as exit 1 → exit 0. For a developer who has never installed the
tool, the failure stays equally loud: `bash: cargo: command not found` becomes
`bash: cargo-audit: command not found`. No silent pass is introduced in either direction. New cost: the
`install-cargo-deny-self-test` hook, ~0.4s, and only when one of three named files changes.

**CI:** the two `deps-security` steps stop depending on the job-level `RUSTUP_TOOLCHAIN: stable`
surviving a later edit. That `env:` block stays, so this is defence in depth, not a swap.

**The new hook is a real gate, not a script someone must remember.** It carries no `stages:`, so it
runs at the `pre-commit` stage that `kernel-matrix.yml:86`'s `pre-commit run --all-files` uses, and its
`files:` pattern matches a file that exists. Verified rather than reasoned:
`pre-commit run install-cargo-deny-self-test --all-files` → `Passed`. #2237's fourth acceptance box is
genuinely ticked.

### Angle 6 — the bots: neither made a single point to adjudicate

I read what they say rather than that they exist, and both are the exact pattern `AGENTS.md` names as
"unavailable infrastructure wearing the shape of a verdict":

- **Copilot** — `COMMENTED` on `2dd803def`, body in full: *"Copilot was unable to review this pull
  request because the user who requested the review has reached their quota limit."*
- **CodeRabbit** — *"Review limit reached … you've reached your PR review limit, so we couldn't start
  this review."* Its comment lists the five files it *would* have read and reports run configuration;
  it contains no finding.

**Points to adjudicate: zero.** There is nothing to mark valid, invalid or follow-up, because neither
produced a technical claim. Neither counts toward quorum, and per `AGENTS.md` their absence needs no
substitute — but it does mean this record is the only review this PR gets, which is how I treated it.

### Findings

**Blocking: none.**

**Follow-up — Medium — `scripts/ci/test-install-cargo-deny.sh:89` — filed as #2250.**
`echo "FAIL ${what}: ${needle@Q} not in"` uses `${var@Q}`, a bash **4.4** parameter transformation.
macOS ships `/bin/bash` 3.2.57, the hook entry is `bash scripts/ci/test-install-cargo-deny.sh` which
resolves through `PATH`, and this machine has no other bash. So on a default macOS developer machine —
the environment this hook exists to protect, and the platform the PR was authored on — the gate's
**failure-reporting path** is itself a fatal expansion error:

```
$ bash scripts/ci/test-install-cargo-deny.sh          # with --version deleted from the install
ok   pinned (exit 0)
scripts/ci/test-install-cargo-deny.sh: line 89: FAIL ${what}: ${needle@Q} not in: bad substitution
                                                                                        → exit 1
```

*Why non-blocking, adjudicated rather than waved through:* the exit code stays **1** under all three
mutations, so this is **not** a recurrence of the fail-open defect the PR body describes fixing — the
gate refuses correctly, and `AGENTS.md`'s "never turn a failure into a clean result" is not crossed.
`Lint (pre-commit)` runs on `ubuntu-latest` with bash 5.x, where the intended message is produced, so
no required check is affected. *What is lost:* on bash 3.2 the run aborts at the first failing
assertion, so the `failures` counter and summary never print and the operator sees the test's own line
89 instead of the assertion that failed — #2237's "the failure would point at the install rather than
at the deletion that caused it", reappearing one layer up.

I confirmed the assertion logic underneath is sound, by running a copy patched **only** at that token
(`${needle@Q}` → `${needle}`) against the same mutation:

```
FAIL argv carries --version 0.20.2: argv: install --locked --version 0.20.2 cargo-deny not in
      argv: install --locked cargo-deny
install-cargo-deny self-test: 1 failure(s)                → exit 1
```

It names the deletion, completes the remaining cases, counts them, exits 1. One token stands between
this gate and its intended report. `shellcheck -x -S warning` returns 0 on the file and will not catch
it. Fixing it requires editing this branch, which would make me its builder and destroy the quorum
this review exists to supply, so it is recorded in #2250 with its measurement instead.

**Observation, not a finding against this PR — `.pre-commit-config.yaml:2` vs `kernel-matrix.yml:77`.**
While counting literals I found `minimum_pre_commit_version: "4.4.0"` and `pip install
pre-commit==4.4.0` — two literals in a config file and a workflow, the shape `AGENTS.md` names. It is
weaker than it looks (one is a floor, the other an equality, so they need not be equal, only ordered),
it is pre-existing, and this PR touches neither line. Not attributable here; noted so the observation
is not lost.

### Verification

All at `d95ae972d` in `/Users/roelschuurkes/rev-2238` unless stated.

| command | result |
|---|---|
| `gh pr view 2238 --json headRefOid` | `d95ae972df1c77b4565e1c1dc4eb6583b25435ee` |
| `git merge-tree --write-tree 802a033 2dd803def` vs `git rev-parse d95ae972d^{tree}` | both `a90fa3a2997442784e3b15debaeb9056540a2aad` — identical |
| `git diff 802a033 d95ae972d --stat` | 5 files, `159 insertions(+), 5 deletions(-)` — the PR's diff, nothing else |
| `python3 scripts/ci/check-ci-gate-coverage.py --self-test` | `ci-gate-coverage self-test=passed`, exit 0 |
| `python3 scripts/ci/check-ci-gate-coverage.py` | `ci-gate-coverage=passed`, exit 0 |
| `bash scripts/ci/test-ci-gate-expectations.sh` | `PASS: CI gate expectation contract`, exit 0 |
| `bash scripts/ci/test-install-cargo-deny.sh` | 10 cases pass, `all cases passed (pin 0.20.2)`, exit 0 |
| `pre-commit run install-cargo-deny-self-test --all-files` | `Passed` |
| `pre-commit run ci-gate-coverage --all-files` | `Passed` |
| `pre-commit run cargo-deny --all-files` | `Passed` |
| `pre-commit validate-config .pre-commit-config.yaml` | exit 0 |
| `shellcheck -x -S warning scripts/ci/{install,test-install}-cargo-deny.sh` | exit 0 both |
| `git grep -c -- '0\.20\.2' … ':!*.lock' ':!Cargo.toml'` | 1 occurrence, 1 file |
| `( cd /tmp && cargo-audit audit --file …/fuzz/Cargo.lock )` | `Scanning … (336 crate dependencies)`, exit 0 |
| `gh pr checks 2238` on `2dd803def` | 53 pass, 10 skipping, 0 non-pass |
| `.github/` diff grep for `continue-on-error\|if:\|needs:\|required\|\|\| true\|exit 0\|name:` | no matching `+`/`-` line |

**Mutation / restore table.** Every mutation is in my own review worktree, on a file I first copied
aside; every restore names the exact path and is followed by a hash check. **No bare-tree command was
used at any point** — no `git checkout --`, no `git restore .`, no `git clean`. The one `git merge
--abort` was in my own detached worktree, undoing only the trial merge I had just made there.

| # | mutation | file | restore command | result |
|---|---|---|---|---|
| 1 | `channel = "1.96.0"` → `"1.94.0-not-installed"` | `/Users/roelschuurkes/rev-2238/rust-toolchain.toml` | `cp /tmp/h2-toolchain.bak rust-toolchain.toml` | old form exit **1** (toolchain error), new form exit **0** — the fix bites |
| 2 | drop `--version "${CARGO_DENY_VERSION}"` from the install argv | `/Users/roelschuurkes/rev-2238/scripts/ci/install-cargo-deny.sh` | `cp /tmp/h2-install.bak scripts/ci/install-cargo-deny.sh` | gate exit **1** — caught (diagnostic lost to `@Q`, see #2250) |
| 3 | `export RUSTUP_TOOLCHAIN=…` → plain assignment | same | `cp /tmp/rev2238-install-cargo-deny.sh.bak scripts/ci/install-cargo-deny.sh` | gate exit **1** — caught; probe shows `toolchain: <unset>` |
| 4 | version assertion → `if false; then` | same | same | gate exit **1** — caught; `FAIL mismatch: exit 0, expected 1` counted, 3 failures |
| 5 | `${needle@Q}` → `${needle}` in a **copy at `/tmp`**, never the tracked file | `/tmp/rev2238-probe.sh` | copy deleted; `ls scripts/ci/.probe.sh` → `No such file` | proves the assertion logic is sound |

Post-restore state: `shasum -a 256` → `96af3a6c4b49ae3409f3700176586caba86b215d4dfdc3ab6de03d58dc2ce6ad`
(`install-cargo-deny.sh`) and `fa0addad5676243d360310f981a69528b0341d295cb5012253d9721aba4d30c0`
(`rust-toolchain.toml`), both matching pre-mutation; `git status --porcelain` empty; the self-test green
again. Mutations 2–4 were each re-run at `d95ae972d` after the branch update so the evidence binds to
the merged head and not only to `2dd803def`.

Mutations 3 and 4 reproduce the PR body's table; **mutation 4 also demonstrates the fail-open fix the
PR body claims** — `run_case` reports `exit 0, expected 1`, counts it, and the run still exits 1, where
the first draft exited 0.

### Non-claims

- **I did not verify any of this on a GitHub runner.** Every measurement is macOS aarch64 with
  `/bin/bash` 3.2.57. GitHub Actions is the integration proof; my merge is gated on its result, not on
  my local runs. The bash-version finding specifically depends on the local bash and I state the CI
  bash (5.x on `ubuntu-latest`) as inference from the runner image, not as something I measured.
- **The `cargo-audit` hook is exercised by no CI check.** It is `stages: [pre-push]`, and
  `kernel-matrix.yml` runs `pre-commit run --all-files` at the pre-commit stage, so that hook does not
  run there. Its changed `entry:` rests on my local reproduction and the author's, not on CI.
- **I did not measure that `cargo install --locked cargo-audit` places a binary on a runner**, nor that
  `~/.cargo/bin` precedes any competing entry on a runner's `PATH`. Both are inferences from the image's
  documented layout.
- **I did not re-litigate the decision to leave `cargo-audit` unpinned.** The PR documents the reasoning
  in `ci.yml`, #2237 holds the open question, and I confirmed only that the asymmetry is now *stated*
  rather than silent — which is what #2237's second acceptance box asks for. Whether an advisory scanner
  should float is a judgement I am not settling.
- **My advisory-database snapshot is not CI's.** I measure `3 allowed warnings found` where the PR body
  measured `5`; the database is live. Nothing follows from the difference about this change.
- **The `@Q` finding is a diagnostic defect, not a fail-open one.** I claim the exit code is correct
  under the three mutations I ran; I do not claim to have enumerated every path through that script.
- **Six other `cargo install` sites remain unpinned** across other workflows. Out of scope, disclosed by
  the author, unverified by me beyond confirming they exist.
- Nothing here touches evidence ingestion, evidence verification or MCP authorization, so ADR-042/043
  boundaries are not engaged and the stop list is not approached. No public-facing string in the diff
  makes a compliance, certification or whole-action claim.

**Merge intent.** Squash-merging once all required checks are green on `d95ae972d`, with the quorum
satisfied by this record, the one Medium finding dispositioned to #2250, and the equivalence conditions
recorded above — condition 1 satisfied by tree identity, condition 2 **not** satisfied and therefore
not relied on.

Rul1an:
## Merged

Squash-merged as **`bc07ffb0e8f47313b4ce7353c63d3e22c6299514`** on `main`, from reviewed head
`d95ae972df1c77b4565e1c1dc4eb6583b25435ee`, with `--match-head-commit` so the merge could not have
taken a head other than the one reviewed above. #2237 closed as `COMPLETED`.

Preconditions at merge, each checked rather than assumed:

- **All required checks green on `d95ae972d`**: `CI` pass, `lane-check/proof` pass,
  `host-capability-check` pass; 53 pass / 10 skipping / 0 failures across 63 contexts — the same counts
  the pre-update head carried, now re-earned on the merged head.
- `mergeStateStatus: CLEAN`, head unchanged from the head this review was measured on.
- **Quorum**: one non-building review, recorded above and bound to `d95ae972d`. Neither bot contributed
  — both were rate-limit non-reviews, so nothing was substituted for their absence.
- **Findings**: no blocking finding; the one Medium finding dispositioned to #2250 with its measurement.

Post-merge confirmation: `git diff bc07ffb0e d95ae972d` is **empty across the whole tree**, so the
squash preserved the reviewed tree exactly — what landed is what was reviewed, not a neighbouring
version of it. Re-run on `bc07ffb0e`: `check-ci-gate-coverage.py` → `ci-gate-coverage=passed`;
`test-install-cargo-deny.sh` → `all cases passed (pin 0.20.2)`; the pin is still one literal in one file.

No programme is active per `AGENTS.md`, so there is no ledger to record this to and none was invented.
The branch and the builder's worktree at `/Users/roelschuurkes/wt-2214` are left in place; only my own
review worktree was removed.

PR #2240, opened 2026-08-10
feat(cli): publish assay init as a machine-readable report (#2161)

### LINKED ISSUE #2161: Init failure is not legible from stdout and exit code alone
Parent: #2177 · Origin: #2154

## Driven observation

Measured on `18abbebf025e4a9243b1ec80f798db5ccad2afed`:

```text
assay init --preset not-a-preset
exit: 2
stdout: partial human progress output
stderr: fatal: unknown preset ...
```

`assay init` has no structured output mode. A stdout-and-exit-only caller sees partial success-looking prose but not the actual failure, registered reason, or next command.

## Acceptance

- Define a machine-readable output path for `init` without changing the existing default human output.
- On failure, stdout carries a registered `reason_code` and non-empty `next_step`.
- A binary-level test covers successful initialization and an invalid preset.

Rul1an:
Closes #2161.

## What this changes

`assay init` had no machine channel. A caller reading only stdout and the exit status saw partial human progress prose and never the reason a run failed or the command to run next; the actionable diagnosis went to stderr behind `fatal:`.

`--format json` now replaces the progress stream with exactly one `assay.init_report.v0` document on stdout, carrying the registered `reason_code`, a non-empty `next_step`, and the files `init` created and skipped on a successful run.

**Correction (review of `8f9b6ef6`).** This paragraph previously claimed the document also carries the files a *failing* run had already written. No reachable path does that: both `InitReport::fail` sites sit upstream of every record, so a classified failure's `created` and `skipped` are always empty, and a failure after a write leaves via `?` and emits no document at all. The Non-claims section below always said so; this sentence contradicted it and has been corrected rather than left to be the part people read.

## Acceptance, restated from the issue

- **A machine-readable output path for `init` without changing the existing default human output.** `--format text` stays the default. Its stdout is byte-identical, measured (below), not inspected.
- **On failure, stdout carries a registered `reason_code` and non-empty `next_step`.** An unusable `--preset` is `E_INVALID_ARGS`, exit 2.
- **A binary-level test covers successful initialization and an invalid preset.** `crates/assay-cli/tests/init_stdout_contract.rs` drives the real binary and reads only stdout and the exit code.

## One rule, one function

- **Reason identity, remediation and exit class** all come from the reason-code registry via `RunOutcome::from_reason`, the same way `doctor` does since #2207. No literal reason string is invented in `init`.
- **The next step has one source.** Both channels read the same argv: the text channel joins it the way a user types it, which is the line `init` has always printed, and the machine channel publishes that same argv through the registry's `format_recovery_argv`, because `--config` carries a caller-controlled value and a shell string and a JSON argv disagree about what a value containing a space or a quote means. `format_recovery_argv` became `pub(crate)` rather than being reimplemented.
- **What `init` did is recorded once** in `InitReport` and rendered by whichever channel was asked for, so the two cannot drift about which files were written.

## Non-claims

Failures the registry does not name — a filesystem write error, an empty `--from-trace` file — still produce **no** document rather than a fabricated code. Under `--format json` their stdout is empty, the exit is 2, and the diagnosis stays on stderr. The generated golden-path contract states that limit in `failure_summary` rather than claiming coverage `init` does not have. A follow-up issue for those paths is worth filing; they need a registered code first, which is a registry decision and not this slice's.

This adds no new reason code, so the `reason_code_version: 1` registry is untouched.

## Verification

Provenance: worktree `/Users/roelschuurkes/wt-2161`, exact head `0fc682d2702c37cd7f336431f08905b3d0ecd35f`, rustc/cargo 1.96.0 (`ac68faa20`, pinned), macOS aarch64. Earlier revisions of this section named `0194ee6dc0f3bfd9678af1b6e55bf33359606258`, which review correctly flagged as not the head under review; every number below is re-measured at the head named here.

**This is not an all-pass claim for the PR.** `Build + Test (windows-latest)` is red on this branch, in `bounded_process::tests::kills_quiet_descendant_after_normal_parent_exit` and `kills_descendant_that_holds_inherited_output_open`. That is [#2249](https://github.com/Rul1an/assay/issues/2249), owned elsewhere; this PR touches no file in that failure and merge stays blocked until #2249 lands.

- `cargo test -p assay-cli --no-fail-fast` — 64 test binaries, 897 passed, 0 failed
- `cargo clippy -p assay-cli --all-targets -- -D warnings` — pass
- `git diff --check` — pass
- `scripts/ci/test-agent-golden-path-skill-hardening.sh` — 86 cases, 22 structural probes, pass
- `scripts/docs/generate-agent-golden-path.py` re-run leaves `git status --porcelain` empty — no generated-doc drift
- `git status --porcelain` empty at measurement time

### RED before GREEN

`init_stdout_contract.rs` was written and run first. It failed for the intended missing behavior: `--format` was not a recognized argument, so stdout was empty and the success case exited 2 instead of 0.

### Negative control 1 — the "one document" gate bites

The claim that JSON stdout is *exactly one document* is the one most easily satisfied by a neighbouring property: a `contains("{")` or a parse-from-the-first-brace check passes with the human stream still printed above it. So the rule was broken deliberately — the `is_json()` guard was removed from `InitReport::progress`, letting progress text print in JSON mode:

| | result |
|---|---|
| all three JSON tests | **FAILED** — `stdout is not one JSON document: expected value at line 1 column 1` |
| `text_is_still_the_default_...` | `ok` (correctly unaffected) |

Guard restored; all pass.

### Negative control 2 — "default output unchanged" is now pinned, not described

This started as a one-time measurement: the binary built at this branch's tree and again with the change stashed, run over six variants, byte-identical. Review reproduced it over twenty-one variants — and then showed it had no regression guard, because swapping the `Note:` and `Next:` lines in the closing block passed every test in the crate.

The part of the stream a run controls is now compared byte for byte over three shapes: a fresh `--hello-trace` run, the same directory again so every line is `Skipped … (exists)`, and a plain run whose closing block is one line rather than two. The lines above `Generating Assay Policy & Config` report which MCP configuration the host happens to have, so they are not the same bytes on a laptop and on a runner and are deliberately outside the comparison. With the `Note:`/`Next:` swap applied, the new test fails and the other twelve in the file, plus all twenty-five golden-path contract tests, pass.

## Contract change

`starter-files` loses `gap_issue: 2161` and gains two outcomes, `success-json` and `unknown-preset-json`. The contract's own coverage gate (`every_cli_contract_outcome_is_executed_once`) requires both to be exercised, and they are. The contract test reads each expected value with `.as_str().expect(...)` rather than comparing two `Value`s, so a generator that dropped a field back to `null` while the binary stopped emitting it cannot pass on `Null == Null`.

## Review

Please attack, in order:

1. **Is the default output claim real?** Try to find an `init` path whose text stdout or stderr changed — the `--from-trace` closing block and the `Note:`/`Next:` ordering under `--hello-trace` are where I'd look first.
2. **Can the JSON document be corrupted?** Find any `println!` still reachable under `--format json`, including from `write_file_if_missing`, the CI scaffolding branch, and `--list-presets`.
3. **Is `E_INVALID_ARGS` honest for an unknown preset,** and does exit 2 still match what the binary did before?
4. **Does any unclassified failure produce a success-shaped document,** or a fabricated reason code? Absence of a document must stay absence, never a clean result.


Rul1an:
## Independent adversarial review — INTERIM record (attack 1 complete)

**Reviewed head: `8f9b6ef6614fc2dee4d83c983f2a53f19803747d`** (`fix(cli): exclude the init stdout contract from the published package`). Worktree `/Users/roelschuurkes/wt-2161`, macOS aarch64, `cargo build -p assay-cli --bin assay` at that head, binary `target/debug/assay` sha256 prefix `6036ae96532f5d48`, mtime 17:48:48. I did not write this change.

Posting early and partial by design: this is an honest interim record with real reproductions, not a placeholder. Attacks 2–6 follow in a second comment.

### Interim verdict: CHANGES-REQUESTED (P2 only so far — no P0, no P1)

The headline non-claim survived every attack I made on it. One overclaim in the PR body / doc does not.

### Worktree hygiene note (not a finding against the PR)

`git -C /Users/roelschuurkes/wt-2161 status --porcelain` at the start of my review was **not** empty:

```
 M crates/assay-cli/src/cli/commands/init_report.rs
```

The diff was an `if false { ... }` wrapped around the `message` insertion in `InitReport::emit` — a leftover negative-control mutation from an earlier interrupted review, not authored work. I backed it up to `/tmp/leftover-init_report.rs.bak` and restored the file by name (`git checkout -- crates/assay-cli/src/cli/commands/init_report.rs`), then rebuilt. Anyone reading a CI or local result measured in that worktree before 17:48 today should re-measure.

### Attack 1 — the unclassified-failure boundary: REPRODUCED AS CLAIMED

The PR states that init failures the reason-code registry does not name emit **no** document: stdout empty, exit 2, diagnosis on stderr. I tried four independent ways to make that boundary emit a success-shaped document or a fabricated code. All four behaved as stated.

| forced failure | how | exit | stdout bytes | stderr |
|---|---|---|---|---|
| empty `--from-trace` | `: > empty.jsonl` | 2 | **0** | `fatal: no events found in trace file: empty.jsonl` |
| `create_dir_all` collision | `: > traces` then `--hello-trace` | 2 | **0** | `fatal: File exists (os error 17)` |
| unwritable target | `chmod 0500 .` | 2 | **0** | `fatal: failed to write policy.yaml: Permission denied (os error 13)` |
| unparsable trace line | `echo garbage > bad.jsonl` | 2 | **0** | `fatal: no valid events found (1 lines skipped, 0 ok)` |

Each run was `assay init --format json <args>` in a fresh directory with stdout and stderr captured to separate files. In all four: exit is non-zero; stdout is **exactly zero bytes**, so there is no `"exit_code": 0`, no `"reason_code": ""`, and no document at all; and no reason code the registry does not define is invented. **No P0.**

Positive control that the machine channel is actually wired in the same conditions — `--format json --preset nope`, exit 2, stderr empty, whole stdout parses as one value via `python3 -c "json.load(...)"`:

```json
{ "schema": "assay.init_report.v0", "exit_code": 2, "reason_code": "E_INVALID_ARGS",
  "created": [], "skipped": [],
  "message": "unknown preset 'nope'. Use --list-presets.",
  "next_step": "Run: assay --help for usage" }
```

### Finding 1 (P2) — `created` on a failing document is unreachable, so the PR's headline benefit is not yet real

`crates/assay-cli/src/cli/commands/init.rs:61` and `init.rs:212` are the only two `report.fail(...)` call sites, and **both are positioned before `init` writes anything**: the unknown-preset branch is upstream of the `policy.yaml` write at `init.rs:76`, and the trace-not-found branch is upstream of the write at `init.rs:259`. Therefore `created` and `skipped` are necessarily `[]` on **every** failing document the current code can emit — as both rows above show.

The PR body claims the document carries "the files `init` created and skipped — *including the files a failing run had already written, which is the difference between retrying and cleaning up first*". That capability exists in `InitReport` but no reachable failure exercises it. The `traces`-as-a-file case above is precisely the scenario the sentence describes — `policy.yaml` and `eval.yaml` were on disk when it failed (`ls` confirms) — and because that failure is unclassified, the caller receives **zero bytes**, not a partial record. So the stated benefit is unreachable in both directions at once: classified failures write nothing, and failures that write something are unclassified.

This is P2, not P1: nothing lies, and the "no document" behaviour is the documented, safe choice. What needs to change is the claim, not the code — the PR body (and, if it repeats it, the generated `failure_summary`) should not advertise partial-write reporting as a delivered property of this slice.

### Provenance correction on my own measurement

My first build of the worktree printed `Blocking waiting for file lock on build directory`, i.e. a cargo process I did not control was building. Runs I took against that binary showed `"created": []` on *successful* fresh-directory inits. That reading did **not** reproduce: on the clean rebuild at this head, `created` populates correctly, verified three consecutive fresh-directory runs plus an instrumented build (`eprintln!` inside `emit` printing `created=["policy.yaml", "eval.yaml", "traces/hello.jsonl"]`, matching the emitted JSON exactly). The instrumentation was reverted by name and the tree is clean. I am recording this because the earlier numbers are unattributable to any tree, not because they stand.

### Explicit non-claims at this point

I have not yet verified: byte-identity of the default text channel (attack 2), that removing the `is_json()` guard fails the three JSON tests (attack 3), that no other writer can corrupt stdout under `--format json` (attack 4), whether each new test bites (attack 5), or generated-doc idempotency and the contract's argv (attack 6). Absence of a finding in those areas so far means I have not looked, not that they are fine.

Rul1an:
## Independent adversarial review — FINAL verdict (attacks 1–6 complete)

**Reviewed head: `8f9b6ef6614fc2dee4d83c983f2a53f19803747d`.** Worktree `/Users/roelschuurkes/wt-2161`, macOS aarch64, cargo/rustc 1.96.0. All numbers below come from builds of that exact tree (`git status --porcelain` empty at measurement time); the binary is `target/debug/assay` built by me at that head. I did not write this change. Continues [my interim record](https://github.com/Rul1an/assay/pull/2240#issuecomment-5242610599), which stands.

# Verdict: CHANGES-REQUESTED

**No P0.** I could not make the machine channel lie in any way I tried. The behavioural core of this slice is sound and the hardest claim in it — that an unclassified failure stays absent rather than becoming a clean result — survived four independent attacks. What blocks it is one test whose name and doc-comment claim coverage the test does not have and the code cannot reach, and a red required check.

## Findings

### P1-1 — a required check is red at this head; merge is BLOCKED

`gh pr checks 2240` at `8f9b6ef6`: `Build + Test (windows-latest)` **fail**, and `CI` **fail** solely because of it (`Required CI dependency test ended with failure`, `TEST_RESULT: failure`). `gh pr view 2240 --json mergeStateStatus` → `BLOCKED`.

The two failing tests are `bounded_process::tests::kills_quiet_descendant_after_normal_parent_exit` and `bounded_process::tests::kills_descendant_that_holds_inherited_output_open` in `tests/support/bounded_process.rs:676`/`:690`, failing inside `assay-mcp-server`'s `agent_golden_path_contract` on a PowerShell `Start-Process` harness step (`deadline of 5s expired; status=exit code: 1`). **This PR does not touch that file** — `gh pr view 2240 --json files` lists 15 paths, none of them `tests/support/bounded_process.rs`, and none in `assay-mcp-server`. So the failure is not attributable to `init` or to `--format json`.

That does not make it disposable. Per `AGENTS.md`, GitHub Actions is the final integration proof and auto-merge may be enabled only when all required checks are green. Two things are needed before merge: the check has to go green (or carry a recorded technical disposition), and the PR body's Verification section has to stop reading as all-pass while a required check is red.

Related provenance gap: the Verification block names head `0194ee6dc0f3bfd9678af1b6e55bf33359606258`, which is **not** the head under review. `8f9b6ef6` adds a `Cargo.toml` exclude and is unverified by that block. I closed that gap myself at `8f9b6ef6` — `cargo fmt --all -- --check` pass, `git diff --check` pass, `cargo clippy -p assay-cli --all-targets -- -D warnings` pass, `cargo test -p assay-cli --test init_stdout_contract --test agent_golden_path_contract` → 12 passed / 0 failed and 25 passed / 0 failed.

### P1-2 — `a_failure_after_partial_work_still_reports_what_it_wrote` pins none of what it is named for

`crates/assay-cli/tests/init_stdout_contract.rs:128-151`. The doc-comment says "The report describes the directory it left behind, including a failure that wrote files first… a caller that reads only the failure envelope cannot tell an `init` that touched nothing from one that wrote `policy.yaml` and then failed". The test asserts exit 2, `reason_code == "E_INVALID_ARGS"`, one-JSON-document, and that a seeded `policy.yaml` is not overwritten. It never reads `created` or `skipped`.

Falsified by mutation. I deleted both keys from the emitted document entirely (`init_report.rs:134-135`, removing `"created": self.created, "skipped": self.skipped`) and ran `cargo test -p assay-cli --test init_stdout_contract`:

```
test a_successful_init_publishes_the_files_it_created_and_an_empty_reason_code ... FAILED
test a_failure_after_partial_work_still_reports_what_it_wrote ... ok      <-- still green
test an_unknown_preset_publishes_a_registered_reason_and_a_next_step_on_stdout ... ok
test text_is_still_the_default_and_still_a_human_progress_stream ... ok
test result: FAILED. 11 passed; 1 failed
```

It stays green with the field it is named after removed from the product. And it cannot do otherwise: `init.rs:61` and `init.rs:212` are the only two `report.fail(...)` sites and both sit upstream of every write (`init.rs:76` and `init.rs:259`), so `created` and `skipped` are necessarily `[]` on every failing document the code can produce. What the test actually pins — a failing `init` does not clobber an existing `policy.yaml`, and its failure envelope is one document with a registered code — is worth pinning. Please rename it to that and drop the partial-write claim from the doc-comment, or leave a `// TODO(#issue)` naming the unregistered-failure work the claim depends on. As written it will tell the next reviewer this boundary is covered.

### P2-1 — the PR body advertises partial-write reporting as delivered

Same root cause as P1-2, restated because it is reader-facing: "including the files a *failing* run had already written, which is the difference between retrying and cleaning up first." The capability is in `InitReport` but no reachable failure exercises it. I forced exactly the scenario that sentence describes — `: > traces` then `assay init --format json --hello-trace`, which fails in `create_dir_all` *after* `policy.yaml` and `eval.yaml` are on disk (`ls` confirms both) — and the caller receives **0 bytes** on stdout with exit 2, because that failure is unclassified. Both directions are closed at once: classified failures write nothing, and failures that write something are unclassified.

### P2-2 — the generated contract records `next_step: null` for `success-json`, but the binary emits one

`docs/generated/agent-golden-path.json`, `starter-files/success-json`, and the agent-facing copy at `packaging/claude-plugin/skills/assay-golden-path/references/agent-golden-path.json`. Running the contract's own argv:

```
$ assay init --preset dev --hello-trace --format json     # exit 0
next_step = Run argv: ["assay","validate","--config","eval.yaml","--trace-file","traces/hello.jsonl"]
```

`null` is only the `outcome()` default in `scripts/docs/generate-agent-golden-path.py:94`, and its meaning is not stated anywhere — but the pre-existing `policy-validation/valid` outcome carries the same `null` and its test reads it as *asserted absent* (`assert!(success_json.get("next_step").is_none())`, `agent_golden_path_contract.rs:276`). So one field now means "asserted absent" for one `kind=json` success and "unspecified" for another, in the same generated file, and for `init` the recorded value contradicts the binary. The same file's `stdout_summary` says the document names `next_step`, so the prose and the machine field disagree with each other. Set `next_step=` explicitly for `success-json` and assert it in `init_json_publishes_the_registered_reason_and_next_step`; that also removes the ambiguity for the next outcome added.

### P2-3 — the byte-identity claim has no regression guard

I reproduced the claim (see below) but it is protected by nothing in CI. I swapped the `Note:`/`Next:` order in the text closing block (`init.rs:159-162`) and ran both suites:

```
cargo test -p assay-cli --test init_stdout_contract --test agent_golden_path_contract
test result: ok. 12 passed; 0 failed
test result: ok. 25 passed; 0 failed
```

Nothing bites — exactly the neighbouring-property gap the PR body itself identifies. `rg "hello trace uses demo|Note: hello"` outside `init.rs` returns nothing repo-wide, and the only assertions on that stream are two `contains("Next: assay validate")`. The PR answered this honestly with a one-time measurement, which is the right call for the claim but does not survive into the next change to that block. A stdout snapshot over two or three variants would make it durable.

## Claims I reproduced, and how I tried to break them

**Unclassified failures stay absent (attack 1, the headline non-claim) — TRUE.** Four forced failures, each `assay init --format json` in a fresh directory, stdout and stderr captured separately:

| forced failure | how | exit | stdout | stderr |
|---|---|---|---|---|
| empty `--from-trace` | `: > empty.jsonl` | 2 | **0 bytes** | `fatal: no events found in trace file: empty.jsonl` |
| `create_dir_all` collision | `: > traces`, `--hello-trace` | 2 | **0 bytes** | `fatal: File exists (os error 17)` |
| unwritable target | `chmod 0500 .` | 2 | **0 bytes** | `fatal: failed to write policy.yaml: Permission denied (os error 13)` |
| unparsable trace line | `echo garbage > bad.jsonl` | 2 | **0 bytes** | `fatal: no valid events found (1 lines skipped, 0 ok)` |

No `"exit_code": 0`, no `"reason_code": ""`, no document, and no code the registry does not define.

**Default text output is byte-identical — TRUE, measured independently.** I did not trust the source reading. I built the binary with `git show 1a58c933e4f7abc42ab5a65a0116882ae93c7ae3:crates/assay-cli/src/cli/commands/init.rs` in place (it compiles at this tree with dead-code warnings only), captured **21 variants** × stdout + stderr + exit code, restored `init.rs` by name, rebuilt, and captured again:

```
diff -r /tmp/cap-old /tmp/cap-new
IDENTICAL: no byte difference in stdout, stderr, or exit code across all 21 variants
aggregate sha256 both sides: 145e0bbfde7075dff6b2376f9e81298ecf637d106afbdb615d34be79508d97d3
```

Variants beyond the six the PR measured: `--ci gitlab`, `--ci` with an empty value, `--preset hardened`, `--preset dev`, a four-flag combo, `--config sub/eval.yaml`, `--config "my dir/eval.yaml"`, `--from-trace --heuristics`, `--from-trace --ci gitlab --gitignore`, and **second runs** of plain / `--hello-trace` / `--ci github` / `--from-trace` to exercise every `Skipped … (exists)` line — including the two sites where the old code hard-coded the literal `policy.yaml` and the new code renders `path.display()`. The `--from-trace` closing block with its embedded `\n` prefixes, the `Note:`-before-`Next:` order, `--list-presets`, and the unknown-preset stderr (`fatal: unknown preset 'not-a-preset'. Use --list-presets.`, exit 2) are all byte-identical. I found no seventh differing invocation.

**Removing the `is_json()` guard fails exactly the three JSON tests — TRUE.** With `if !self.is_json()` deleted from `init_report.rs:52`, all three fail on whole-stdout parse (`stdout is not one JSON document: expected value at line 1 column 1`, with the progress stream visible above the document in the panic output) and `text_is_still_the_default_…` stays `ok`. `test result: FAILED. 9 passed; 3 failed`. Guard restored, 12/12 green.

**The document cannot be corrupted by another writer — TRUE.** Structurally, the only stdout writers reachable from `init` are `init_report.rs:53` (guarded) and `:149` (the document itself); everything in `generate/` writes to stderr, and the one `println!` in `generate.rs:111` is inside `generate::run`, which `init` does not call. Empirically I ran **23** `--format json` invocations — every CI provider, `--gitignore`, `--list-presets`, `--from-trace`, nested/spaced/quoted `--config`, and re-runs — and parsed the *whole* stdout with a strict decoder that also rejects trailing bytes. All 23 were either exactly one `assay.init_report.v0` value with empty trailing, or (for the two unclassified failures) empty.

**Other mutations that correctly bite:** omitting `next_step` fails the unknown-preset and success tests; swapping the unknown-preset reason code to `E_CFG_PARSE` fails both `init_stdout_contract` tests *and* the generated-contract test with a clear message (`init failure reason must match the generated contract, left: "E_CFG_PARSE", right: "E_INVALID_ARGS"`); leaking the text diagnosis to stdout in `InitReport::fail` fails `init_text_stays_the_human_progress_stream_it_has_always_been`.

**One rule, one function — TRUE and load-bearing.** `format_recovery_argv` (`exit_codes.rs:117`) is genuinely shared, and the rationale is demonstrable rather than decorative: with `--config 'we"ird dir/eval.yaml'` the machine channel emits `Run argv: ["assay","validate","--config","we\"ird dir/eval.yaml",…]`, which round-trips back to the exact argv, while the text line it preserves is not shell-safe. Both channels read the same `argv` slice.

**Contract and packaging — TRUE.** `python3 scripts/docs/generate-agent-golden-path.py` re-run leaves `git status --porcelain` empty. Both new contract argv match observation: `success-json` exit 0 / one `assay.init_report.v0` / `reason_code ""`; `unknown-preset-json` exit 2 / `E_INVALID_ARGS` / `next_step` exactly `Run: assay --help for usage` — with the one discrepancy in P2-2. The `8f9b6ef6` packaging fix is correct and complete: `rg -l '#\[path = "\.\./\.\./\.\./'` finds exactly five test files and all five are in `exclude`, and `cargo package -p assay-cli --list` confirms none of them ships.

## Explicit non-claims

- I did **not** run `cargo test --workspace` or the full `cargo test -p assay-cli` (64 binaries) at this head; disk was constrained. I ran the two focused suites, fmt, clippy `-D warnings`, and `git diff --check`. The PR's "64 test binaries, 0 failures" is unverified by me at `8f9b6ef6`.
- All local measurements are macOS aarch64. I verified nothing on Linux or Windows.
- I could **not** force the `emit` error path (no `/dev/full` on macOS), so the comment at `init_report.rs:98-99` about a report that cannot be rendered exiting through the error path is unverified empirically. By reading, a stdout write failure panics rather than returning `Err`, which is still a non-zero exit and not a clean result.
- A consumer that truncates the stream (`… --format json | head -c 20`) sees exit 0 with partial bytes. That is the consumer's truncation, not a partial write: the document is a few hundred bytes and never exceeds the pipe buffer, so I could not construct a real `EPIPE` during the document write. Not a finding, stated so it is not mistaken for one.
- I did not assess whether the Windows failure in P1-1 is a genuine flake. The merge base `c2c41b03` passed that job on `main`, but two earlier `main` runs today concluded failure; one data point either way is not a flakiness verdict.
- I reviewed the `init` rows of the golden-path docs and SKILL prose, not the rest of those files.

## Worktree state

Recorded at start: **not clean** — ` M crates/assay-cli/src/cli/commands/init_report.rs`, an `if false { … }` around the `message` insert, a leftover negative control from the earlier interrupted review. Backed up to `/tmp/leftover-init_report.rs.bak` and restored by name before any measurement.

Every mutation above was reverted with `git checkout -- <exact path>` (only `init.rs` and `init_report.rs` were ever touched); no bare checkout, reset, clean, or stash was run. Final `git -C /Users/roelschuurkes/wt-2161 status --porcelain` is **empty**, `HEAD` is `8f9b6ef6614fc2dee4d83c983f2a53f19803747d`, the binary was rebuilt at that tree, and both suites are green (12/12 and 25/25).

Rul1an:
## Disposition of review findings — new head `e4ec65d9c9705ccd06e2f7ac60ac04f696d005b1`

Two independent adversarial reviews landed on `0194ee6d`/`8f9b6ef6`. Both confirmed the headline
non-claim survives attack: every unclassified failure is exit 2 with **zero** stdout bytes, no
success-shaped document and no fabricated reason code. One re-measured byte-identity of the default
text channel across 21 variants, the other across 44, both against a rebuilt pre-change binary —
identical stdout, stderr and exit code. I did not ask for two; the pair was an accident of timing,
and their agreement is evidence rather than quorum.

Fixed in `e4ec65d9c`:

- **P1, the misnamed test.** `a_failure_after_partial_work_still_reports_what_it_wrote` stayed green
  when `created`/`skipped` were deleted from the product. It could not have failed: both
  `InitReport::fail` sites are upstream of every record, so those lists are structurally always
  empty on a classified failure. Renamed to
  `a_classified_failure_reports_no_files_because_it_fails_before_the_first_write`, and it now pins
  that boundary in the direction that can regress. Verified by moving a `record_created` above the
  preset `fail` site: `FAILED`. Mutation reverted by exact path and confirmed byte-identical
  against a pre-mutation copy.
- **P1, the PR body.** The "What this changes" paragraph advertised partial-write reporting that no
  path delivers, contradicting this PR's own Non-claims section. Corrected in the body, with the
  correction left visible rather than silently rewritten.
- **P2, contract drift.** `success-json` recorded `next_step: null` — which in this contract's
  vocabulary means the key is absent — while the binary emits
  `Run argv: ["assay","validate","--config","eval.yaml","--trace-file","traces/hello.jsonl"]`. The
  generator now carries the real value and `agent_golden_path_contract.rs` compares it as a string,
  the same way the failure side already did.

Accepted, not fixed here:

- **P2, no regression guard on byte-identity.** Reordering `Note:`/`Next:` still passes every test.
  Both reviewers are right that the claim rests on a one-off local measurement. A golden-file test
  over a few `init` variants is the fix; it is a separate concern from this PR's subject and I would
  rather file it than bolt a fixture on at review time.
- **P1, `Build + Test (windows-latest)`.** Red at this head, on `bounded_process` harness self-tests
  in a file this PR does not touch. It is red on #2241 too, which shares nothing with this change
  but that helper, and a re-run reproduced it. Not attributable here; tracking separately rather
  than calling it flaky without evidence.

Verification on `e4ec65d9c`: `cargo test -p assay-cli` no failures; the two contract suites 25 and
12 passed; `cargo fmt --all -- --check`, `cargo clippy -p assay-cli --all-targets -D warnings`,
`git diff --check`, and the publish-shape guardrail all clean.

**This push invalidates the reviews measured on `8f9b6ef6`.** No review record exists on
`e4ec65d9c` yet.

Rul1an:
## Builder response to CHANGES-REQUESTED — head `0fc682d2702c37cd7f336431f08905b3d0ecd35f`

Every finding is addressed in code or in the body; one is dispositioned rather than fixed, and it is the red check. Provenance for everything below: worktree `/Users/roelschuurkes/wt-2161`, head `0fc682d2702c37cd7f336431f08905b3d0ecd35f`, which is `e4ec65d9c` plus this push plus a merge of `origin/main` `bc07ffb0e`. `cargo 1.96.0 (30a34c682)` / `rustc 1.96.0 (ac68faa20)`, pinned, macOS aarch64. Mutations were reverted from a byte copy of the named file; no bare-tree command was run.

### P1-1 — required check red, and a Verification block that read as all-pass on a head that was not under review

**Dispositioned, not fixed.** The two failing tests are [#2249](https://github.com/Rul1an/assay/issues/2249) and another agent owns it. This PR touches no file in that failure, and per the contract I did not investigate it, did not modify `tests/support/bounded_process.rs`, and did not treat its red as feedback here. `mergeStateStatus` is `BLOCKED` and stays that way until #2249 lands.

The provenance half of the finding *is* fixed. The Verification section now names this head, re-measures every number on it, and opens by stating that a required check is red rather than reading as an all-pass. Your own re-measurement at `8f9b6ef6` closed a gap I had left; thank you, and the section no longer depends on it.

### P1-2 — `a_failure_after_partial_work_still_reports_what_it_wrote` pinned none of what it was named for

Fixed in `e4ec65d9c`, which was already on the branch when I picked this up: renamed to `a_classified_failure_reports_no_files_because_it_fails_before_the_first_write`, the doc-comment states the structural reason instead of the capability, and the test now reads `created` and `skipped` and asserts both are empty across two classified failures (`E_INVALID_ARGS`, `E_TRACE_NOT_FOUND`). Your mutation — deleting both keys from the emitted document — now fails it instead of leaving it green, and so does the opposite direction: a future `fail` site landing after a write makes those lists non-empty and fires the test, which forces the choice between reporting the partial work and keeping the limit honest. It still pins what you said was worth pinning: the existing `policy.yaml` is not clobbered, and the failure envelope is one document with a registered code.

### P2-1 — the body advertised partial-write reporting as delivered

Fixed. The sentence is gone and the paragraph now carries an explicit correction naming the review of `8f9b6ef6`, with the structural reason (both `fail` sites upstream of every write; a post-write failure leaves via `?` and emits no document). The generated `failure_summary` never made that claim — I checked it and the three SKILL copies rather than assuming — so nothing else needed editing.

### P2-2 — `next_step: null` for `success-json` while the binary emits one

Fixed in `e4ec65d9c`. The generator carries the real argv for `success-json`, the generated JSON and the packaged copy carry it, and `init_json_publishes_the_registered_reason_and_next_step` now reads it with `.as_str().expect(...)` and compares it — the same shape as the failure side, so a regenerated `null` cannot pass on `Null == Null`. That also settles the ambiguity you named: in this file `null` keeps its one meaning, "asserted absent", and `init`'s success no longer borrows it to mean "unspecified".

### P2-3 — the byte-identity claim had no regression guard

Fixed in this push, and this is the only new code here. `the_default_text_stream_is_pinned_where_the_run_controls_it` compares stdout byte for byte over three shapes: a fresh `--hello-trace` run, the same directory again so every line is `Skipped … (exists)`, and a plain run whose closing block is one line rather than two.

The comparison starts at `🏗️  Generating Assay Policy & Config...` on purpose. Everything above it reports which MCP configuration the host happens to have — on this laptop it prints `✨ Detected Claude Desktop config (global)`, which a runner will not — so pinning it would be pinning the machine. Everything from that marker down is a function of the arguments and the directory, and it is the region your finding is about: the closing-block ordering, the `Created`/`Skipped` rendering, and both shapes of the next-step line.

Verified in the failing direction with your exact mutation. With `Note:` and `Next:` swapped in `init.rs`:

```
test the_default_text_stream_is_pinned_where_the_run_controls_it ... FAILED
test result: FAILED. 12 passed; 1 failed          (init_stdout_contract)
test result: ok. 25 passed; 0 failed              (agent_golden_path_contract)
```

Only the new test moves, which is what makes it the guard rather than a coincidence. `init.rs` was then restored from a byte copy.

## Verification at `0fc682d2702c37cd7f336431f08905b3d0ecd35f`

- `cargo test -p assay-cli --no-fail-fast` — 64 test binaries, **897 passed, 0 failed**
- `cargo test -p assay-cli --test init_stdout_contract` — 13 passed, 0 failed
- `cargo clippy -p assay-cli --all-targets -- -D warnings` — pass
- `git diff --check` — pass
- `scripts/ci/test-agent-golden-path-skill-hardening.sh` — 86 cases, 22 structural probes, pass
- `python3 scripts/docs/generate-agent-golden-path.py` re-run leaves `git status --porcelain` empty
- `git status --porcelain` empty at measurement time

## Two things I did not do

I did not add a guard for the *host-dependent* prefix of the text stream, so a change to the MCP-scanning lines is still unpinned; it is outside this slice and cannot be pinned hermetically in this test. And your non-claim about `emit`'s error path stands — I did not construct a stdout write failure either.

Rul1an:
## Interim review record — head `0fc682d2702c37cd7f336431f08905b3d0ecd35f`

**Non-building reviewer.** I authored and edited neither this change nor any specification or plan governing it. Read-only worktree at `/tmp/rv2240`, detached at `0fc682d2702c37cd7f336431f08905b3d0ecd35f`, `git status --porcelain` empty at checkout. Toolchain: `cargo 1.96.0`, macOS aarch64. Binary: `/tmp/rv2240/target/debug/assay`, built from that tree.

This is published now rather than at the end, so that a record exists on this exact head. **The full adjudicated review follows in a second comment**; treat that one as the verdict.

### First reproductions (all at the head above)

**The honest limit the prior review confirmed reproduces exactly.** Four failure paths under `--format json`, exit 2, zero stdout bytes:

```
C) --from-trace <empty file>   exit=2  stdout_bytes=0   stderr: fatal: no events found in trace file: empty.jsonl
D) read-only cwd (write fail)  exit=2  stdout_bytes=0   stderr: fatal: failed to write policy.yaml: Permission denied (os error 13)
F) --from-trace <malformed>    exit=2  stdout_bytes=0   stderr: fatal: no valid events found (1 lines skipped, 0 ok)
   (line-level "warning: skipping line 1" also goes to stderr, not stdout — no document contamination)
```

Absence stays absence: no document, no fabricated reason code, no success shape. Confirmed as stated.

**Both classified paths publish what #2161 asked for.** Verbatim stdout:

```
$ assay init --preset not-a-preset --format json      # exit 2, 239 bytes
{"schema":"assay.init_report.v0","exit_code":2,"reason_code":"E_INVALID_ARGS",
 "created":[],"skipped":[],"message":"unknown preset 'not-a-preset'. Use --list-presets.",
 "next_step":"Run: assay --help for usage"}

$ assay init --from-trace absent.jsonl --format json  # exit 2, 236 bytes
{"schema":"assay.init_report.v0","exit_code":2,"reason_code":"E_TRACE_NOT_FOUND", ...}
```

### First substantive finding, under investigation

`own_output()` (`crates/assay-cli/tests/init_stdout_contract.rs:246`) starts the byte comparison at `🏗️  Generating Assay Policy & Config...`. Measured on this host, the excluded region is three lines plus a blank:

```
🔍 Scanning project for MCP configurations...                              <- unconditional, NOT host-dependent
   ✨ Detected Claude Desktop config (global)                              <- genuinely host-dependent
   ℹ️  No specific MCP config found, initializing generic project.         <- cwd-determined; fixed in a tempdir
                                                                           <- blank line
```

Only the middle line is host-dependent. The first line is emitted unconditionally on every non-`--from-trace` run (`init.rs:20`), and `rg` finds no other test in the tree that pins it. Mutation results and adjudication in the follow-up comment.

Rul1an:
## Independent review — FINAL verdict, head `0fc682d2702c37cd7f336431f08905b3d0ecd35f`

**Non-building reviewer.** I authored and edited neither this change nor any normative specification or implementation plan governing it. Continues [my interim record](https://github.com/Rul1an/assay/pull/2240#issuecomment-5243575383), which stands. This is the only review on this head.

**Provenance for every number below.** Worktree `/tmp/rv2240`, detached at `0fc682d2702c37cd7f336431f08905b3d0ecd35f`, `git status --porcelain` empty before and after all work. `cargo 1.96.0` / `rustc 1.96.0`, macOS aarch64. Binary rebuilt from that clean tree at `/tmp/rv2240/target/debug/assay`. One self-correction: an early `--from-trace` capture was taken against a binary still linked from the M6-mutated tree; I rebuilt and re-measured, and only the rebuilt reading is reported.

# Verdict: APPROVE

No P0, no P1. Three P2 findings, all follow-up, all dispositioned below; none of them makes this change unsafe to land and none of them is a behavioural defect. The hardest claim in the slice — that an unclassified failure stays absent rather than becoming a clean result — survived every attack. The byte-identity guard the last round asked for is real and bites harder than the builder demonstrated. It also has two holes, one disclosed with an inaccurate reason and one not disclosed.

---

## Result per angle

### Angle 1 — what the byte-identity guard's starting point excludes: **a real hole, partly disclosed, wrong reason**

`own_output()` (`crates/assay-cli/tests/init_stdout_contract.rs:246-251`) slices from `🏗️  Generating Assay Policy & Config...`. Measured, the excluded region is:

| excluded line | source | genuinely host-dependent? |
|---|---|---|
| `🔍 Scanning project for MCP configurations...` | `init.rs:20` | **No — unconditional on every non-`--from-trace` run** |
| `   ✨ Detected Claude Desktop config (global)` | `init.rs:32` | Yes (`dirs::home_dir()` probe) |
| `   ℹ️  No specific MCP config found, initializing generic project.` | `init.rs:52` | No — decided by cwd alone, therefore fixed in a `tempdir` |
| blank line | leading `\n` at `init.rs:55` | No |

So of four excluded lines exactly one is host-dependent. The PR body says "The lines above `Generating Assay Policy & Config` report which MCP configuration the host happens to have", and the delta comment says the prefix "cannot be pinned hermetically in this test". **Both are inaccurate for the first line**, which is a fixed literal that a `starts_with` would pin hermetically. The conclusion that the *block as a whole* can't be byte-compared across hosts is correct; the generalisation to every line in it is not.

**Verified as a hole, not merely argued** (M5 below): replacing that first line with `MUTATED-BANNER` leaves the entire crate green — **897 passed, 0 failed, 64 test binaries**. Nothing anywhere in `assay-cli` pins the first line of `assay init`'s default stdout.

**Can the marker move or vanish silently? No.** Renaming it (M7) fails loudly at `init_stdout_contract.rs:249` with `stdout never reaches "🏗️  Generating Assay Policy & Config..."`. A marker moved later shortens the compared region and a marker moved earlier lengthens it; either way the constant no longer matches. The `find` is first-occurrence, and the string is not reachable twice. **No silent-comparison failure mode.**

### Angle 2 — does the guard bite? **Yes, including on three mutations the builder did not try**

Builder's mutation reproduced exactly. My three inventions — trailing whitespace, indent, and line reorder inside the compared region — are each caught, and each caught by that test *alone*.

### Angle 3 — is the contract genuinely machine-readable? **Yes, and the limit is stated where a consumer reads it**

- **Shape is stable and pinned.** `schema: "assay.init_report.v0"` is compared against the generated contract's `stdout.document` by `assert_document()` (`agent_golden_path_contract.rs:191-201`), so the identity cannot drift from the published contract silently.
- **Success is distinguishable from failure without parsing prose.** Both `exit_code` and `reason_code` are top-level, and the in-document `exit_code` equals the process exit status on every path I ran (0/`""` on success, 2/`E_INVALID_ARGS` and 2/`E_TRACE_NOT_FOUND` on the classified failures). `message` is the only prose and is not load-bearing.
- **`--format` is closed.** `OutputFormat` is `{Text, Json}` only, so there is no third value that silently degrades to text; clap rejects it.
- **The limit is stated outside the PR body.** The four unclassified-failure paths reproduce exactly as the prior review found (exit 2, **zero** stdout bytes — see my interim comment for the runs). That limit is published in `docs/generated/agent-golden-path.json` `failure_summary` *and* in all three `SKILL.md` copies, which is what an agent consumer actually reads: "Failures the reason-code registry does not name, such as a filesystem write error, still produce no document: stdout is empty and the diagnosis stays on stderr." **This is not a PR-body-only limit.** Absence stays absence; no fabricated code, no success-shaped document. Line-level `warning:` output on a malformed trace also goes to stderr, so it cannot corrupt the document.

The one place the limit is *not* stated is `assay init --help`, whose `--format` text describes the JSON channel without its boundary. Minor; noted as follow-up.

### Angle 4 — two channels, one rule: **mostly one function, one thing decided twice**

Decided once and read by both channels, verified by reading the code: reason identity, `message`, `next_step` and the exit class all come from `RunOutcome::from_reason`; the next step has a single `argv` slice that the text channel joins via `next_line()` and the machine channel publishes via the shared `format_recovery_argv`; `created`/`skipped` are recorded once in `InitReport`. That is the right shape and it is genuinely shared, not mirrored.

**One exception (P2-3 below).** On a classified failure the text channel throws `outcome.exit_code` away — `InitReport::fail` returns `Err(anyhow!(message))` (`init_report.rs:121-126`), which lands in `main.rs:44-49` and exits a **hardcoded `2`**, whose own comment concedes it ("hardcoded 2 is safe here"). The JSON channel returns `outcome.exit_code` from the registry. So "what exit code does this reason code produce" now has two answers for `init`. They agree today only because both reachable codes map to `EXIT_CONFIG_ERROR`. This is the PR #2247 shape in a narrow form; see disposition.

### Angle 5 — claims and the ADR-042 stop list: **clean for what this PR introduces**

The document carries `schema`, `exit_code`, `reason_code`, `created`, `skipped`, `message`, `next_step`, `presets`. **No scalar trust score. No whole-action verdict** — `reason_code: ""` plus `exit_code: 0` is the existing registry's run-outcome vocabulary describing whether `init` wrote its files, not a judgement about an agent action. Success language does not overstate: the machine channel makes no success claim in prose, and `next_step` points at `assay validate` rather than implying anything was validated. `✅  Initialization complete.` is pre-existing and byte-unchanged.

One pre-existing string sits on the stop list's "no compliance claim" edge: `   Tip: For EU AI Act compliance scanning, add: --pack eu-ai-act-baseline` (`init.rs:329`, and the sibling at `ci.rs:164`). **It is not introduced or altered by this PR** — the diff converts the `println!` to a `format!` with identical bytes, and it is on `main` already. Recording it as an observation for a separate issue, explicitly *not* as a finding against this change.

---

## Findings

### Blocking: none.

### Follow-up

**P2-1 — the excluded prefix has no guard anywhere, and its stated justification is inaccurate for the lines that are not host-dependent.**
`crates/assay-cli/tests/init_stdout_contract.rs:213-251`, against `crates/assay-cli/src/cli/commands/init.rs:20`.
The gap was disclosed by the builder, which is why this is not P1 — but the reason given ("cannot be pinned hermetically") does not hold for `🔍 Scanning project for MCP configurations...`, which is a fixed literal. M5 proves nothing in the crate catches replacing it.
*Disposition:* follow-up, non-blocking. This is a missing regression guard on unchanged bytes, not a behavioural change; the PR does not claim coverage of this region and the test's doc comment names its boundary. The cheap fix is an assertion that stdout *starts with* the banner line before slicing, plus correcting the rationale in the test doc comment and PR body to say "one host-dependent line sits in this block" rather than "these lines report the host". I am not making that edit — I am the quorum and cannot also be the builder.

**P2-2 — the `--from-trace` closing block is entirely unpinned, including its `Next:` line.**
`crates/assay-cli/src/cli/commands/init.rs:315-331`.
This gap was **not** disclosed, and unlike P2-1 nothing about this block is host-dependent — it is a pure function of the arguments. It is also the exact place the PR body's own review request points at first ("the `--from-trace` closing block … are where I'd look first"). M6 reorders all three lines (`Tip:` first, `Next:` last, newline placement changed) and the full crate suite is **897 passed, 0 failed**. The observable output really does change; I captured the mutant's stream before rebuilding.
*Disposition:* follow-up, non-blocking. I verified the bytes are **unchanged** by this PR, both by reading the diff (`\n{next_line}` reproduces `"\n   Next: assay validate --config X --trace-file Y"` exactly; `CI:` and `Tip:` are literal copies) and by running the path. So this is the same missing-guard class as P2-1 on a sibling path, one that the new golden test could cover in three more lines by adding a `--from-trace` shape.

**P2-3 — the exit code for a classified failure is answered in two places.**
`crates/assay-cli/src/cli/commands/init_report.rs:114-127` (registry value on the JSON channel, discarded on the text channel) versus `crates/assay-cli/src/main.rs:44-49` (literal `2`).
*Disposition:* follow-up, non-blocking. No divergence is reachable at this head: both `EInvalidArgs` and `ETraceNotFound` map to `EXIT_CONFIG_ERROR`, and the tests assert exit 2 on both channels. The trigger is specific and foreseeable — a future `fail` site with a reason outside exit class 2 (e.g. any `E_*` infra code, exit 3) makes the JSON channel exit 3 while the text channel still exits 2, and the text channel would be the wrong one. `AGENTS.md`'s fallback where code cannot be shared is a parity test over a table; there is no test asserting the two channels agree on exit code for the same reason code. The structural fix is to route `init`'s text failure through `CliFailure` so both channels read `outcome.exit_code`. Worth attaching to #2247 rather than filing fresh, since it is the same defect shape.

### Observations, not findings

- `assay init --help`'s `--format` text describes the JSON channel without stating the zero-stdout limit that the SKILL and generated contract do state.
- `presets` is inserted only `if !self.presets.is_empty()` (`init_report.rs:146`), so a consumer cannot rely on the key's presence, and `--list-presets --format json` has no row in the generated contract. It does emit a well-formed document (captured below).
- The `created` array's order is unpinned; M4's reorder changed it and the 25 golden-path contract tests stayed green. Only the text channel caught it.
- The PR body's "This is not an all-pass claim … `Build + Test (windows-latest)` is red on this branch" is now **stale in the safe direction**: it passed on this head. I am not treating one green Windows run as proof #2249 is resolved — see Non-claims.

---

## Verification

All commands run in `/tmp/rv2240` at `0fc682d2702c37cd7f336431f08905b3d0ecd35f`.

```
cargo test -p assay-cli --no-fail-fast   ->  64 test binaries, 897 passed, 0 failed
cargo test -p assay-cli --test init_stdout_contract        ->  13 passed, 0 failed
cargo test -p assay-cli --test agent_golden_path_contract  ->  25 passed, 0 failed
cargo fmt --all -- --check               ->  pass
cargo clippy -p assay-cli --all-targets -- -D warnings     ->  pass (Finished dev profile in 1m 26s)
git diff --check                         ->  pass
python3 scripts/docs/generate-agent-golden-path.py; git status --porcelain  ->  empty (no generated-doc drift)
```

The 897/0 independently reproduces the builder's headline number, binary count included.

### Behaviour captured directly from the rebuilt binary

Default text, `--preset dev --hello-trace` (excluded prefix above the marker, compared region below it):

```
🔍 Scanning project for MCP configurations...          <- unconditional, unguarded (P2-1)
   ✨ Detected Claude Desktop config (global)          <- the one genuinely host-dependent line
   ℹ️  No specific MCP config found, initializing generic project.
                                                       <- marker follows; comparison starts here
🏗️  Generating Assay Policy & Config...
   Created policy.yaml (preset: dev)
   Created eval.yaml
   Created traces/hello.jsonl
✅  Initialization complete.
   Note: hello trace uses demo prompt/response text only; treat real traces as potentially sensitive.
   Next: assay validate --config eval.yaml --trace-file traces/hello.jsonl
```

`--from-trace` success, whose closing block no test pins (P2-2):

```
🔍 Generating policy from trace: e.jsonl
   Aggregated 1 unique entries from 1 events
   Created policy.yaml (1 allow, 0 needs_review, 0 deny)
   Created eval.yaml

✅  Initialization complete.

   Next: assay validate --config eval.yaml --trace-file e.jsonl
   CI:   assay ci --config eval.yaml --trace-file e.jsonl

   Tip: For EU AI Act compliance scanning, add: --pack eu-ai-act-baseline
```

`--list-presets --format json`, undocumented in the contract but well-formed:

```json
{"schema":"assay.init_report.v0","exit_code":0,"reason_code":"","created":[],"skipped":[],
 "next_step":"Run argv: [\"assay\",\"init\"]",
 "presets":[{"name":"default","description":"Balanced defaults (current protections)"},
            {"name":"hardened","description":"No shell, no network, read-only FS (strict)"},
            {"name":"dev","description":"Permissive, logging-first (for local iteration)"}]}
```

### Mutation / restore table

Every mutation touched exactly one named file and was restored from a byte copy taken before any edit (`/tmp/rv2240-init.rs.orig`, `/tmp/rv2240-init_report.rs.orig`). **No bare-tree command was run at any point** — no `git checkout --`, no `git restore .`, no `git clean`, no `reset`, no `stash`. After each restore I confirmed `git status --porcelain -- <that exact path>` was empty.

| # | mutation | file:line | expectation | result | restore command |
|---|---|---|---|---|---|
| M1 | swap the `Note:` and `Next:` lines (builder's) | `init.rs:159-162` | caught | **caught** — `the_default_text_stream_is_pinned_where_the_run_controls_it` FAILED; `init_stdout_contract` 12 passed / 1 failed; `agent_golden_path_contract` **25 passed / 0 failed**. Reproduces the builder's report exactly. | `cp /tmp/rv2240-init.rs.orig crates/assay-cli/src/cli/commands/init.rs` |
| M2 | **mine** — trailing space appended to `   Skipped {} (exists)` | `init_report.rs:67` | caught | **caught** — 12 passed / 1 failed, same single test | `cp /tmp/rv2240-init_report.rs.orig crates/assay-cli/src/cli/commands/init_report.rs` |
| M3 | **mine** — indent 3→2 spaces on `   Created {}` | `init_report.rs:61` | caught | **caught** — 12 passed / 1 failed, same single test | `cp /tmp/rv2240-init_report.rs.orig crates/assay-cli/src/cli/commands/init_report.rs` |
| M4 | **mine** — reorder two lines *inside* the compared region (write `traces/hello.jsonl` before `eval.yaml`) | `init.rs:86-93` | caught | **caught** by the text guard — 12 passed / 1 failed. Note the JSON contract did **not** notice: `agent_golden_path_contract` 25 passed / 0 failed. | `cp /tmp/rv2240-init.rs.orig crates/assay-cli/src/cli/commands/init.rs` |
| M5 | **mine** — replace the excluded banner `🔍 Scanning project for MCP configurations...` with `MUTATED-BANNER` | `init.rs:20` | hole → not caught | **NOT caught — full crate suite 897 passed, 0 failed.** This is P2-1. | `cp /tmp/rv2240-init.rs.orig crates/assay-cli/src/cli/commands/init.rs` |
| M6 | **mine** — reorder the `--from-trace` closing block (`Tip:` first, `CI:` second, `Next:` last) | `init.rs:326-330` | hole → not caught | **NOT caught — full crate suite 897 passed, 0 failed**, with the reordered stream verified as really emitted. This is P2-2. | `cp /tmp/rv2240-init.rs.orig crates/assay-cli/src/cli/commands/init.rs` |
| M7 | **mine** — rename the marker itself, simulating it moving or vanishing | `init.rs:55` | must fail loudly, never silently compare a different region | **caught loudly** — panic at `init_stdout_contract.rs:249`: `stdout never reaches "🏗️  Generating Assay Policy & Config..."`; 12 passed / 1 failed | `cp /tmp/rv2240-init.rs.orig crates/assay-cli/src/cli/commands/init.rs` |

After M6 I restored the source but read the binary before relinking it, which produced one capture attributable to no tree; I rebuilt (`cargo build -p assay-cli --bin assay`) and re-measured, and only the rebuilt reading appears above. Final state: `git status --porcelain` empty, `HEAD` `0fc682d2702c37cd7f336431f08905b3d0ecd35f`, binary rebuilt from that tree, full suite 897/0.

### Required checks on this exact head

```
CI                     pass   .../job/93534935794
host-capability-check  pass   .../job/93530212470
lane-check/proof       pass   no delegated runner proof required
```

`mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`, head unchanged at `0fc682d2702c37cd7f336431f08905b3d0ecd35f`. `host-capability-check` being green is what answers the diagnostics-path classification question; I did not compute that classification myself.

---

## Non-claims

- **Windows is one sample, not a soundness proof.** `Build + Test (windows-latest)` passed on this head, and #2249 is red on sibling PR #2241 in `bounded_process` self-tests neither PR touches. One green run is one observation; I am not claiming #2249 is resolved, nor that it is flaky. I did not investigate it.
- **P2-1 and P2-2 are missing guards, not regressions.** I verified the excluded prefix's and the `--from-trace` block's bytes are unchanged by this PR from the diff and by running both paths. I did **not** independently re-run the prior reviewers' 21- and 44-variant byte-diff against a pre-change binary; I relied on the diff plus my own mutation evidence.
- I did not construct a stdout write failure, so `InitReport::emit`'s error path is unexercised here, as the prior review also recorded.
- I did not review the non-`init` rows of the golden-path docs, the `doctor` changes visible in the generated contract, or `assay-mcp-server`.
- The ADR-042 assessment covers what this PR's output introduces. The pre-existing `EU AI Act compliance scanning` string is outside that scope and I make no claim about whether it is compliant with the stop list.
- CodeRabbit reports `pass` with "Review rate limited" and Copilot's only review record sits on the superseded head `0194ee6d`. Per `AGENTS.md` neither is a review and neither counts toward quorum; I am not substituting for their absence.
- Every number here is from a debug build in `/tmp/rv2240`; none is a performance measurement.

**All three findings are dispositioned as follow-up with the technical basis recorded above. No actionable finding is left open. Proceeding to squash-merge (no `--admin`).**

PR #2241, opened 2026-08-10
test(cli): assert the whole doc block transports the integrity boundary

Rul1an:
Follow-up to #2213 (merged as `c2c41b038e89ff2a8029c84334261126fdade987`). Test-only; no production code changes.

## The gap

`the_doc_comment_transports_the_boundary_rather_than_restating_it` checked only the line immediately above the `EEvidenceIntegrity` variant:

```rust
let last = head.lines().next_back().unwrap_or_default().trim();
assert_eq!(last, "#[doc = include_str!(\"exit_codes/evidence_integrity_boundary.md\")]", ...);
```

rustc concatenates all `///` lines and `#[doc]` attributes on one item in source order. A `///` line placed *above* the include therefore also documents the variant, and renders **ahead of** the boundary text in the shipped rustdoc, while leaving the checked line untouched.

So the test asserted a neighbouring property — *"the include is present, adjacent to the variant"* — rather than the transport claim its name makes. Asserting the whole block is exactly the include says *"the include is the only thing that documents this variant"*, which is the property the transport claim actually needs.

## What this does not claim

**There is no live resurrected claim on `main` today.** The lines above the include are plain `//` comments, which are not doc text, so nothing is currently mis-documented. This is not a security fix and should not be merged as one. It closes the gap for the next edit.

**Explicit non-claims:** enum-level docs and rendered rustdoc are not covered.

## Current head (reconciliation)

- Base (`origin/main`): `c9b036de8198d3bfa7f9752d414863b20c9b7c18`
- Head: `2fc43bb91b24d294b5faecbd78378aa70bf1f991`
- Merge of `origin/main` was conflict-free (merge commit; no rebase)
- Net PR diff vs main: only `crates/assay-cli/tests/spec_reason_code_registry.rs`
- `Cargo.toml` / `Cargo.lock` untouched in the PR diff
- Fresh mutation matrix + verification commands: see the reconciliation comment on this head
- Did **not** mark ready / merge; fresh final-head review required

## Review ask

The one thing worth attacking: **find a way to document `EEvidenceIntegrity` that this assertion still misses**, without counting enum-level docs or rustdoc rendering (named non-claims).


Rul1an:
## Independent adversarial review — interim record (head `c520deac06c39fe61f4727fda0e3840f36a79418`)

Reviewer did not author this change. Read-only worktree `/Users/roelschuurkes/wt-2199-patch` at `c520deac06c39fe61f4727fda0e3840f36a79418`, base `origin/main` `c2c41b038e89ff2a8029c84334261126fdade987`, macOS aarch64. Posting the primary reproduction now; a second comment will carry the bypass matrix and final verdict.

**Interim verdict: the PR's central claim reproduces as stated.**

### What I ran

Injected one line above the include in `crates/assay-cli/src/exit_codes.rs` (new line 85), production file otherwise untouched:

```rust
    /// The verifier checks the Merkle root and its inclusion proofs.
    #[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]
    EEvidenceIntegrity,
```

Then ran the same command against the two test versions, same injected source:

`cargo test -p assay-cli --test spec_reason_code_registry the_doc_comment_transports`

| test file | result |
|---|---|
| this PR's (`c520deac06c39fe61f4727fda0e3840f36a79418`) | **FAILED** |
| base's (`git show c2c41b038e89ff2a8029c84334261126fdade987:crates/assay-cli/tests/spec_reason_code_registry.rs`) | **ok** — `1 passed; 0 failed` |

Observed failure output on the PR head, verbatim:

```
thread 'the_doc_comment_transports_the_boundary_rather_than_restating_it' panicked at crates/assay-cli/tests/spec_reason_code_registry.rs:416:5:
assertion `left == right` failed: `EEvidenceIntegrity` is documented by ["/// The verifier checks the Merkle root and its inclusion proofs.", "#[doc = include_str!(\"exit_codes/evidence_integrity_boundary.md\")]"] rather than by including crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md and nothing else. ...
  left: ["/// The verifier checks the Merkle root and its inclusion proofs.", "#[doc = include_str!(\"exit_codes/evidence_integrity_boundary.md\")]"]
 right: ["#[doc = include_str!(\"exit_codes/evidence_integrity_boundary.md\")]"]
```

So the PR body's opposite-verdicts table is reproduced independently. The gap it closes is real, not theoretical.

### Sibling test is confirmed blind

Full-file run with the injection still in place, PR test file restored:

```
test the_evidence_integrity_boundary_reads_the_same_in_the_spec_and_the_code ... ok
test the_doc_comment_transports_the_boundary_rather_than_restating_it ... FAILED
test result: FAILED. 8 passed; 1 failed
```

The PR's statement that `the_evidence_integrity_boundary_reads_the_same_in_the_spec_and_the_code` does **not** catch this is confirmed, and for the stated reason: it compares the §5.1 row against the included fragment, and the injected claim is in neither.

### Methodology note against a false negative

My first run of the PR test reported `ok` with what I believed was the injection in place. It was not: the edit had not reached disk, and `git diff` on the file was empty. I now verify the injected bytes with `rg` in the same shell invocation as every `cargo test`. Any reviewer repeating this should do the same — an unwritten injection reads exactly like a missed bypass.

### Still to come (second comment)

Attempting bypasses the new assertion may still miss: multi-line `#[doc = "..."]`, a non-doc attribute between the doc block and the variant (`take_while` stopping early), `cfg_attr`-wrapped doc, `#[doc(hidden)]`, leading whitespace, enum-level doc, and whether the `split_once("    EEvidenceIntegrity,")` anchor can be made to match an earlier point in the file so the assertion reads a location other than the one it names.

Rul1an:
# Independent adversarial review — CHANGES-REQUESTED

Reviewed head `c520deac06c39fe61f4727fda0e3840f36a79418`. Worktree `/Users/roelschuurkes/wt-2199-patch`, base `c2c41b038e89ff2a8029c84334261126fdade987`, rustc/cargo 1.96.0 (`ac68faa20`, pinned), macOS aarch64. All attack edits were reverted; the worktree was verified clean (`git status --porcelain` empty, `9 passed`) as the last action.

I did not write this change. The PR asked for one thing — find a way to document `EEvidenceIntegrity` that the new assertion misses. **There are five.** Four are single-line edits.

## The claim does not hold

The claim under review is *"the include is the only thing that documents this variant, and any second source of doc text for it fails the test."* The new assertion actually checks a narrower property: *"no line prefixed `///` or `#[doc` sits in the unbroken run of lines immediately above the declaration."* Those are not the same property, and the gap between them is reachable in one edit.

## Attack matrix

Every attack: edit `crates/assay-cli/src/exit_codes.rs`, then
`cargo test -p assay-cli --test spec_reason_code_registry the_doc_comment_transports`, then revert.

| # | Shape | Verdict |
|---|---|---|
| 1 | `#[doc = concat!(...)]` spanning multiple source lines, above the include | **MISSED** |
| 2a | Non-doc attribute directly above the variant (below the include) | CAUGHT — but a false failure, see F5 |
| 2b | `///` claim above one benign non-doc attribute above the include | **MISSED** |
| 3 | `#[cfg_attr(doc, doc = "...")]` above the include | **MISSED** |
| 4 | `#[doc(hidden)]` | CAUGHT |
| 5a | `/** ... */` block doc comment above the include | **MISSED** |
| 5b | Tab-indented `///` line | CAUGHT (`.map(str::trim)`) |
| 5c | `//!` at variant position | N/A — rustc `E0753`, not expressible |
| 6 | Doc on the enum / bleed from the neighbouring variant | CAUGHT (no bleed; the walk stops at the `EReplayLimitExceeded,` line) |
| 7 | Second, earlier match for the `split_once` anchor | **MISSED** |

## F1 (P1) — Four single-edit shapes of doc text pass the assertion

All four inserted immediately above `crates/assay-cli/src/exit_codes.rs:85` (the include line).

**Attack 1 — multi-line doc attribute:**

```rust
#[doc = concat!(
    "The verifier checks the Merkle root ",
    "and its inclusion proofs."
)]
#[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]
EEvidenceIntegrity,
```

Observed: `test the_doc_comment_transports_the_boundary_rather_than_restating_it ... ok`.
Also `cargo fmt --all -- --check` → exit 0, so rustfmt preserves the wrapping. The `take_while` walks up, matches the include, then hits `)]` and stops.

**Attack 2b — this PR's own reproduction line, with one benign attribute between it and the include:**

```rust
/// The verifier checks the Merkle root and its inclusion proofs.
#[serde(alias = "E_EVIDENCE_INTEGRITY_V1")]
#[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]
EEvidenceIntegrity,
```

Observed: `... ok`. rustc collects doc attributes on an item regardless of interleaved non-doc attributes; the line-based walk does not.

**Attack 3 — `cfg_attr`-wrapped doc:**

```rust
#[cfg_attr(doc, doc = "The verifier checks the Merkle root and its inclusion proofs.")]
#[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]
EEvidenceIntegrity,
```

Observed: `... ok`. This one is worse than the others because the claim appears *only* in the shipped rustdoc.

**Attack 5a — block doc comment:**

```rust
/** The verifier checks the Merkle root and its inclusion proofs. */
#[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]
EEvidenceIntegrity,
```

Observed: `... ok`, and `cargo fmt --all -- --check` → exit 0 (rustfmt does not normalize block doc comments by default).

**These are real doc text, not cosmetic.** Confirmed with a standalone `rustdoc` 1.96.0 run on an equivalent enum outside the repo, grepping the rendered `enum.E.html` and comparing offsets:

```
MULTILINE-CONCAT-CLAIM   PRESENT in rendered rustdoc, renders AHEAD OF the boundary text
INTERLEAVED-CLAIM        PRESENT in rendered rustdoc, renders AHEAD OF the boundary text
CFGATTR-CLAIM            PRESENT in rendered rustdoc, renders AHEAD OF the boundary text
BLOCKDOC-CLAIM           PRESENT in rendered rustdoc, renders AHEAD OF the boundary text
```

That is exactly the failure this PR's body describes: the claim renders ahead of the boundary text while the checked region stays untouched.

## F2 (P1) — The `split_once` anchor is not location-safe; the test can read a different variant entirely

`crates/assay-cli/tests/spec_reason_code_registry.rs:406`. `split_once` takes the *first* substring match of `"    EEvidenceIntegrity,"`, which any indentation of four or more spaces satisfies. Nothing asserts the anchor is unique, or that it landed inside `pub enum ReasonCode`.

I added a plausible legacy enum above `ReasonCode`:

```rust
/// Legacy v1 view of the registry, retained for `--exit-codes v1` consumers.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
pub enum ReasonCodeV1 {
    #[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]
    EEvidenceIntegrity,
}
```

and, at the same time, injected the withdrawn claim on the **real** variant at line 85:

```rust
/// The verifier checks the Merkle root and its inclusion proofs.
#[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]
EEvidenceIntegrity,
```

`cargo test -p assay-cli --test spec_reason_code_registry` observed:

```
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

All nine tests green with the withdrawn ADR-043 claim live and documenting the shipped variant. The test reports on a location it does not pin.

## F3 (P2) — The test hand-rolls a second implementation of "what documents this item"

This is the `AGENTS.md` "one rule, one function" failure in its literal form. The rule *which text documents `EEvidenceIntegrity`* is owned by rustc; the test re-answers it with a line-prefix scan. F1 is the drift — present on the first edit rather than later. The old test had the same defect in a smaller form, so this PR narrows the gap without changing its kind.

The robust form is to compare the variant's **rendered** doc text against the boundary file, which is what `the_evidence_integrity_boundary_reads_the_same_in_the_spec_and_the_code` already does for the §5.1 row — the pattern exists in this same file. Reading the `docs` field for the variant out of rustdoc JSON and asserting `flowed(docs) == flowed(read(BOUNDARY))` would state the transport claim directly and close attacks 1, 2b, 3, 5a and 7 at once. That is a larger change than this PR; a cheaper interim would at minimum assert the anchor occurs exactly once and anchor it inside the `pub enum ReasonCode {` body.

## F4 (P2) — The failure message is wrong about plain `//` comments

`crates/assay-cli/tests/spec_reason_code_registry.rs:421` says "Plain `//` comments are not doc text and are fine." True above the include, false below it:

```rust
#[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]
// Exit code 2: an operator decision, not a producer defect.
EEvidenceIntegrity,
```

Observed: `FAILED`, with `left: []`. The message then tells the reader that what they just did is fine. The next person gets a diagnostic that contradicts the failure in front of them.

## F5 (P2) — Benign non-doc attributes below the include break the test (attack 2a)

```rust
#[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]
#[serde(alias = "E_EVIDENCE_INTEGRITY_V1")]
EEvidenceIntegrity,
```

Observed: `FAILED` — ``` `EEvidenceIntegrity` is documented by [] rather than by including crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md and nothing else ```. The variant *is* documented by the include and by nothing else; the assertion is false and the diagnostic points nowhere useful. Same root cause as F4.

## Secondary checks

**The PR's own reproduction — confirmed, both directions.** With `/// The verifier checks the Merkle root and its inclusion proofs.` injected above the include:

- New test at head `c520deac06c39fe61f4727fda0e3840f36a79418`: `FAILED`
  `left: ["/// The verifier checks the Merkle root and its inclusion proofs.", "#[doc = include_str!(\"exit_codes/evidence_integrity_boundary.md\")]"]`
  `right: ["#[doc = include_str!(\"exit_codes/evidence_integrity_boundary.md\")]"]`
- Old test, same injection still in place: `ok. 1 passed`.
  I restored the old test with `git checkout HEAD~1 -- crates/assay-cli/tests/spec_reason_code_registry.rs` rather than `git stash`, because the test change is committed at head, not a working-tree modification. `HEAD~1` resolves to `c2c41b038e89ff2a8029c84334261126fdade987`, the base named in the PR body.

Same source, opposite verdicts, as claimed.

**The `main` non-claim — I agree.** `crates/assay-cli/src/exit_codes.rs:81-84` are plain `//` comments, not doc text, and this PR does not touch that file. Nothing is currently mis-documented and this is correctly not a security fix. I also confirmed `the_boundary_claims_no_structure_the_verifier_does_not_build` reads only `BOUNDARY` (line 430), so it cannot catch a claim injected into `exit_codes.rs` — the PR's statement about that is accurate.

**Whole file passes on the clean tree at `c520deac06c39fe61f4727fda0e3840f36a79418`:** `cargo test -p assay-cli --test spec_reason_code_registry` → `9 passed; 0 failed`.

## Not verified

- I did not run clippy, and did not run the wider `assay-cli` suite. `cargo fmt --all -- --check` was run only on the attack-1 and attack-5a states (both exit 0), not on every attack state.
- rustdoc rendering was confirmed with a standalone probe crate on rustc 1.96.0, not by building `assay-cli`'s own docs.
- I did not check CI status on this PR.
- I did not assess whether the attack-7 shape would survive human review as a real code change; it is a two-step edit, which is why it is graded P1 rather than P0. The four F1 shapes are single edits.

## Bottom line

The new assertion is a better neighbouring property than the old one, but it is still a neighbouring property, and it fails the standard this PR's own body sets for it. I would not merge as-is. Either narrow the test name and its comment to the property actually checked, or move the check onto rendered doc text so the assertion and the claim are the same statement.

Rul1an:
## Independent adversarial review — final verdict (head `c520deac06c39fe61f4727fda0e3840f36a79418`)

**Verdict: CHANGES-REQUESTED.** The central claim reproduces (see [interim record](https://github.com/Rul1an/assay/pull/2241#issuecomment-5242594121)) and the PR is strictly stronger than `main`. But you asked to be told if the new assertion is itself a neighbouring property, and it is: **five ways of putting live doc text on `EEvidenceIntegrity` pass the new test**, four of them unconditional plain Rust, and one of them makes the test read a location other than the one it names.

Provenance: reviewer did not author this change. Worktree `/Users/roelschuurkes/wt-2199-patch` at `c520deac06c39fe61f4727fda0e3840f36a79418`, base `c2c41b038e89ff2a8029c84334261126fdade987`, `cargo 1.96.0 (30a34c682 2026-05-25)` / `rustc 1.96.0 (ac68faa20 2026-05-25)` (pinned by `rust-toolchain.toml`), macOS aarch64. Every case below was applied to `crates/assay-cli/src/exit_codes.rs` around the include at line 85 / variant at line 86, run with `cargo test -p assay-cli --test spec_reason_code_registry the_doc_comment_transports`, and reverted. Final `git status --porcelain` is empty.

### Attack B — CAUGHT / MISSED

| # | edit (relative to the include at `exit_codes.rs:85`) | result |
|---|---|---|
| B0 | `/// <claim>` line above the include | **CAUGHT** |
| B1 | `#[doc = "<claim>` … newline … `and more."]` — one attribute whose string literal spans source lines, above the include | **MISSED** |
| B1b | `#[doc = concat!(` / `"<claim>"` / `)]` above the include | **MISSED** |
| B2 | `/// <claim>`, include, then `#[allow(dead_code)]` directly above the variant | CAUGHT (`left: []`) |
| B3 | `#[cfg_attr(all(), doc = "<claim>")]` above the include — `all()` is unconditionally true | **MISSED** |
| B4 | `#[doc(hidden)]` above the include | CAUGHT |
| B5 | tab-indented `\t/// <claim>` above the include | CAUGHT (`str::trim` covers it) |
| B6 | `/** <claim> */` block doc comment above the include | **MISSED** |
| B7 | `/* … */` comment at line 56 containing the include line followed by `    EEvidenceIntegrity,`, plus `/// <claim>` on the real variant | **MISSED** |
| B8 | `/// <claim>` appended to the enum-level doc | MISSED (documents the enum, not the variant — I do not count this against the PR) |
| B9 | `/// <claim>` below the include, above the variant | CAUGHT |

Why each miss slips through: the collector at `spec_reason_code_registry.rs:409-414` walks upward line by line and stops at the first line that does not `starts_with("///")` or `starts_with("#[doc")`. A continuation line (`and more."]`, `)]`), a `/**` opener, and a `#[cfg_attr(` opener are all such lines, so the walk stops *below* the offending doc text and reports a clean `doc_block` of exactly the include.

**These are real doc text, not lint curiosities.** Proof, isolated crate, same pinned `rustdoc 1.96.0 (ac68faa20)`: an enum whose variants carry each of the four forms renders all four in `enum.E.html`.

```
MULTILINE_CLAIM_ONE  -> RENDERED (1.96.0)
CONCAT_CLAIM_TWO     -> RENDERED (1.96.0)
CFGATTR_CLAIM_THREE  -> RENDERED (1.96.0)
BLOCKDOC_CLAIM_FOUR  -> RENDERED (1.96.0)
```

So B1, B1b, B3 and B6 each put a withdrawn claim ahead of the boundary text in the shipped rustdoc — the exact outcome the test's own comment says the new assertion forbids — while the test prints `ok`.

### Findings

**P1-1 — the assertion misses four unconditional doc forms.** `crates/assay-cli/tests/spec_reason_code_registry.rs:413`. The comment at :391-403 concludes "Asserting the block is exactly the include says *the include is the only thing that documents this variant*". As implemented it does not say that; it says "no `///`-prefixed or `#[doc`-prefixed line sits immediately above the include". B1/B1b/B3/B6 are counterexamples. Commands and outputs above.

**P1-2 — `split_once("    EEvidenceIntegrity,")` can be made to read the wrong slice (B7).** `spec_reason_code_registry.rs:406`. The anchor is the first textual occurrence anywhere in the file, including inside comments and string literals. Inserting this before the enum at line 56:

```rust
/* Layout of the integrity variant, for reference:
    #[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]
    EEvidenceIntegrity,
*/
```

leaves `head` ending at the include line *inside the comment*. The real variant at line 91 then carried `/// The verifier checks the Merkle root and its inclusion proofs.` and the test reported:

```
test the_doc_comment_transports_the_boundary_rather_than_restating_it ... ok
```

This is worse than the individual misses: the test silently asserts about a different location than its name and message claim, and nothing in the output says so. Anchoring inside the `pub enum ReasonCode {` … `}` body slice, and asserting the anchor occurs exactly once there, would close it.

**P2-3 — the failure message misdirects, and the assertion is stricter than the message admits.** `spec_reason_code_registry.rs:421` says "Plain `//` comments are not doc text and are fine." That is true only *above* the include, which is the shape on `main` today. Measured:

| edit | result |
|---|---|
| plain `//` above the include | PASSES (message correct) |
| plain `//` between include and variant | **FAILS**, `left: []` |
| blank line between include and variant | **FAILS**, `left: []` |

Both are legal Rust that leaves the doc attribute attached, and both produce `left: []` under a message that says plain `//` is fine and names "a second copy of the rule" as the cause. B2 (`#[allow(dead_code)]` between include and variant) fails the same way. A reader hitting `left: []` is told the opposite of what happened. Either accept intervening non-doc lines, or say that the include must be the line immediately above the variant.

### Claims I reproduced as true

- **The gap is real.** Same injected `///` line: base test `ok`, PR test `FAILED`. Verbatim outputs in the interim comment.
- **The sibling test is blind.** With the injection in place: `the_evidence_integrity_boundary_reads_the_same_in_the_spec_and_the_code ... ok`, `the_doc_comment_transports... FAILED`, `8 passed; 1 failed`. Confirmed, and for the stated reason.
- **The non-claim holds; I agree with it.** `exit_codes.rs:81-84` are plain `//`, not doc text, and `rg -c Merkle crates/assay-cli/src/exit_codes.rs` finds nothing. There is no live resurrected claim on this head. This is correctly not a security fix.
- **Clean head is green.** `cargo test -p assay-cli --test spec_reason_code_registry` → `9 passed; 0 failed`.

### Non-claims (not verified by me)

- I did not run `cargo clippy`, `cargo fmt --check`, the wider `assay-cli` suite, or inspect CI on this head.
- I did not verify that any missed form would survive this repo's clippy/rustdoc lint gates — a gate that rejects `/**` or bare `#[doc = "..."]` would narrow B1/B6 in practice, and I did not look.
- I did not render `assay-cli`'s own rustdoc; the rendering proof is an isolated crate under the same pinned toolchain.
- I did not verify the PR body's provenance narrative about #2213.

### Suggested disposition

The change is worth landing — it closes the shape that actually occurred. What I would not land is the comment at :391-403 asserting a property the code does not hold. Minimum: fix the anchor (P1-2), and either strengthen the collector to stop only at a line that is neither doc-ish nor a continuation, or narrow the comment and message to what is pinned. Per AGENTS.md's one-rule-one-function note, this collector is a line-based approximation of rustc's "what documents this item"; if it stays an approximation, it should enumerate the forms it does not reject rather than claim exhaustiveness.

### Method note

My first run of the PR's test reported `ok` with what I believed was the injection applied. It was not on disk — `git diff` was empty. Every result above was taken from a run whose same shell invocation first printed the injected bytes. An unwritten injection is indistinguishable from a missed bypass in the test output alone.

Rul1an:
## Disposition of review findings — new head `3dee8fd89f107f257957d79f1067ceb3d545c0e1`

Two independent adversarial reviews landed on `c520deac`. They agreed, having attacked separately:
the PR's own reproduction holds in both directions, and the `main` non-claim is accurate — but the
assertion was still a neighbouring property. Between them they found six shapes of live doc text
that passed it (`#[doc = concat!(...)]` and a `#[doc = "..."]` whose string spans lines, a `///`
separated from the include by one benign attribute, `#[cfg_attr(..., doc = "...")]`, a `/** */`
block) plus a `split_once` anchor hijack that let the withdrawn Merkle claim sit live on the variant
with all nine tests green.

They were right, and the root cause is the one they named: the test answered, in a second
vocabulary, a question the Rust grammar already answers. So the question now goes to `syn`. The
variant is located in `enum ReasonCode` by name rather than by first textual match, its doc
attributes are read from the parsed attribute list, and `cfg_attr` counts as doc text because it
expands to some.

I re-ran two of the reported attacks against the new implementation rather than assuming the
rewrite covered them:

| attack | on `c520deac` | on `3dee8fd8` |
|---|---|---|
| `#[doc = concat!(...)]` spanning source lines | `ok` (MISSED) | **FAILED** (caught) |
| `#[cfg_attr(all(), doc = "...")]` | `ok` (MISSED) | **FAILED** (caught) |

Both mutations were applied to `crates/assay-cli/src/exit_codes.rs` and reverted with
`git checkout -- crates/assay-cli/src/exit_codes.rs`, then confirmed byte-identical against a
pre-mutation copy; `git status --porcelain` is empty.

The P2s are addressed by the same change: the failure message no longer claims plain `//` comments
are fine while failing on one, because an ordinary comment or a benign non-doc attribute between
the include and the declaration is no longer read as doc text at all.

What stays unpinned is named in the test rather than implied away: doc text on the enum itself, and
the rendering step — this reads the parsed attributes, not rustdoc's output.

Verification on `3dee8fd89`: `cargo test -p assay-cli` no failures, `spec_reason_code_registry`
9 passed, `cargo fmt --all -- --check`, `cargo clippy -p assay-cli --all-targets -D warnings`, and
the publish-shape guardrail all clean. `main` is merged in at `795ee909`.

**This push invalidates the reviews measured on `c520deac`.** No review record exists on
`3dee8fd89` yet.

Rul1an:
## Builder response to CHANGES-REQUESTED — head `6cc52f1913db74dcd5845ae579683073d530b68d`

Thank you for the five shapes. All five reproduced here, along with a sixth (your `B1`, one `#[doc]` attribute whose string literal spans source lines) and an in-body variant of the decoy anchor. All eight now fail, and the three legal shapes that used to fail now pass.

Provenance: worktree `/Users/roelschuurkes/wt-2199-patch`. The **before** column was measured at `c520deac06c39fe61f4727fda0e3840f36a79418`; the **after** column at `59d84bb62599e04961a786ab64b67bc9ab4e130f`, whose test file is byte-identical to this head — `6cc52f191` is `59d84bb62` plus a merge of `origin/main` `bc07ffb0e` (CI files only; net PR diff against `main` is one file, +104/−8). `cargo 1.96.0 (30a34c682)` / `rustc 1.96.0 (ac68faa20)`, pinned by `rust-toolchain.toml`, macOS aarch64. Each row is one edit to `crates/assay-cli/src/exit_codes.rs`, one `cargo test -p assay-cli --test spec_reason_code_registry <name>`, then a restore from a byte copy taken before the first edit — no bare-tree command was run at any point. Following your method note, each run printed `git diff --stat` for the edited file in the same shell invocation and the harness refuses a `NO-DIFF` row.

## What changed, and why it is not a longer prefix list

`the_doc_comment_transports_the_boundary_rather_than_restating_it` →
`nothing_but_the_boundary_include_documents_the_integrity_variant`.

Your F3 is the finding the others follow from, and its operative part is the *direction the walk fails in*. Walking up while a line **looks like** documentation stops at the first line without a `///` or `#[doc` prefix — and a continuation line, a `/**` opener and a `#[cfg_attr(` opener are all such lines, so doc text above any of them reads as absent. Adding `/**` and `#[cfg_attr` to that list would only move the boundary; the next shape is still admitted by default.

So the walk is inverted. It now runs on lines that **cannot** carry doc text and stops everywhere else:

- blank;
- a comment that is `//` and not `///` or `//!`;
- an attribute that opens and closes on one line and mentions `doc` nowhere.

Everything else stops it, and the line it stops on must declare the previous variant. Any doc text between the two variants is therefore a failure by default, including shapes I did not think of: an unrecognised line is refused, not admitted. The comment states this and names what it still does not pin.

The anchor is scoped to the `pub enum ReasonCode` body, that header is asserted unique, and `    EEvidenceIntegrity,` is asserted unique **inside the body** — so your F2/P1-2 is closed at the root rather than by hoping the first textual match is the right one.

## The five (plus three) shapes, before and after

| # | shape | at `c520deac` | at `59d84bb6` |
|---|---|---|---|
| 1 | `#[doc = concat!(…)]` spanning source lines, above the include | PASSES | **FAILS** |
| 2 | `///` claim above one `#[serde(alias = …)]` above the include | PASSES | **FAILS** |
| 3 | `#[cfg_attr(all(), doc = "…")]` above the include | PASSES | **FAILS** |
| 4 | `/** … */` above the include | PASSES | **FAILS** |
| 5 | decoy `    EEvidenceIntegrity,` in a `/* */` comment before the enum, claim live on the real variant | PASSES | **FAILS** |
| 5b | the same decoy moved *inside* the enum body, where scoping alone cannot exclude it | (not tried) | **FAILS** |
| 6 | one `#[doc = "…"]` whose literal spans two source lines (your `B1`) | PASSES | **FAILS** |
| — | control: `///` claim above the include (the shape this PR was written for) | FAILS | **FAILS** |
| — | control: `#[doc(hidden)]` above the include | FAILS | **FAILS** |

Each failure names the line that caused it. Verbatim, rows 4 and 5b:

```
`EEvidenceIntegrity`'s own text reaches back to "    /** The verifier checks the Merkle root
and its inclusion proofs. */", which does not declare the previous variant. …
```
```
assertion `left == right` failed: `EEvidenceIntegrity,` occurs 2 times in the `ReasonCode` body;
the declaration this reads is then not necessarily the one it names
```

Row 5 is worth checking twice: the decoy is still in the file and the assertion no longer reads it, because the search is scoped to the enum body. Row 5b is the follow-up you did not ask for — the decoy moved inside the body, where scoping cannot help — and the uniqueness assertion catches it.

## F4 / F5 / P2-3 — the false failures

The message said plain `//` comments are fine while three shapes of legal Rust failed with `left: []`. All three now pass, and the message now describes the rule that is actually applied:

| shape | at `c520deac` | at `59d84bb6` |
|---|---|---|
| `#[serde(alias = …)]` between the include and the variant | FAILS (`left: []`) | **passes** |
| `//` comment between the include and the variant | FAILS (`left: []`) | **passes** |
| blank line between the include and the variant | FAILS (`left: []`) | **passes** |

One deliberate narrowing you should weigh: an attribute broken across source lines is refused even when it is plainly not documentation, because accepting it means reasoning about where an attribute ends, which is the open question again. The message says so in those words rather than leaving the next author with `left: []`.

## Does the name still overclaim

The old name did, so it is gone. The new one is the property the code holds: nothing but the boundary include documents this variant. Two things it does **not** pin are now named in the comment instead of implied away — doc text on `enum ReasonCode` itself, which documents the enum and not this variant (your `B8`, which you correctly did not count), and the rendering step, since it reads the source rustdoc is given rather than rustdoc's output.

## The parser answer, and why it is not here

Your F3 suggested reading the variant's rendered doc text. Rustdoc JSON is nightly-only under the pinned 1.96.0 toolchain, so the nearest available form was `syn`: parse the file, find the variant in `enum ReasonCode` by name, read its attribute list. I built that first and it produced exactly the verdict table above — commit `3dee8fd89f107f257957d79f1067ceb3d545c0e1`, briefly this PR's head, then replaced.

It is not what shipped, for a reason worth recording because neither of us raised it: **a dev-dependency moves `Cargo.lock`, and `Cargo.lock` in a PR's diff routes `lane-check` to `gates=all`.** Measured at `3dee8fd8`:

```
lane-check/proof   fail   delegated proof required: gates=all
- Cargo.lock: runner workflow, CLI, cgroup, or workspace dependency surface requires gates=all
```

`lane-check/proof` is a **required** check (`gh pr checks 2241 --required`), so a test-only change would have needed a delegated runner proof pack — or a recorded `override_reason` dispatch — to become mergeable. That is a real cost for a `syn` dev-dependency, and dispatching an override to clear a gate my own change tripped is not a builder's call. The shipped version keeps `Cargo.lock` untouched and `lane-check/proof` green. If you would rather have the parser and think the delegated proof is worth obtaining, `3dee8fd8` is on the remote and I will restore it.

## Verification

At `59d84bb62599e04961a786ab64b67bc9ab4e130f`:

- `cargo test -p assay-cli --test spec_reason_code_registry` — 9 passed, 0 failed
- `cargo test -p assay-cli --no-fail-fast` — 63 test binaries, **883 passed, 0 failed**
- `cargo clippy -p assay-cli --all-targets -- -D warnings` — pass
- `git diff --check` — pass
- `cargo deny (advisories, licenses, bans, sources)` — pass, via pre-commit
- `git status --porcelain` empty at measurement time

Re-run at this head, `6cc52f1913db74dcd5845ae579683073d530b68d`, after merging `origin/main` `bc07ffb0e`: `cargo test -p assay-cli --test spec_reason_code_registry` — 9 passed, 0 failed; `cargo fmt --all -- --check` pass; `git diff --check` pass. The merge added CI files only and no Rust, which is why the 883-test number is attributed to `59d84bb6` rather than restated here.

## Still red, and not mine

`Build + Test (windows-latest)` fails on this branch in `bounded_process::tests::kills_quiet_descendant_after_normal_parent_exit` and `kills_descendant_that_holds_inherited_output_open`. That is [#2249](https://github.com/Rul1an/assay/issues/2249), owned by another agent. This PR does not touch `tests/support/bounded_process.rs`, and I did not investigate it. Nothing above should be read as a green required-check claim for the PR as a whole.

Rul1an:
## Independent non-building review — interim record (head `6cc52f1913db74dcd5845ae579683073d530b68d`)

I did not author this change and have not pushed to this branch. Read-only worktree `/Users/roelschuurkes/rev-2241`, detached at `6cc52f1913db74dcd5845ae579683073d530b68d` (confirmed with `gh pr view 2241 --json headRefOid`), merge-base `origin/main` `bc07ffb0e8f47313b4ce7353c63d3e22c6299514`, `cargo 1.96.0 (30a34c682 2026-05-25)` / `rustc 1.96.0 (ac68faa20 2026-05-25)`, macOS aarch64. Posting the reproductions now rather than at the end; the verdict and full record follow in one comment.

**Interim: all seven evasions reproduce as caught, each for the right reason, and the honest form still passes. Five legitimate shapes are refused — that is the finding.**

### The rewrite holds where it was attacked

Baseline on the pristine head: `cargo test -p assay-cli --test spec_reason_code_registry` → `9 passed; 0 failed`. So the test still discriminates; it has not become a test that fails on everything.

Each row below is one injection into `crates/assay-cli/src/exit_codes.rs`, one `cargo test -p assay-cli --test spec_reason_code_registry nothing_but_the_boundary_include_documents_the_integrity_variant -- --exact`, then a restore. My harness refuses any row whose `git diff --stat` for that file is empty, which is why the pristine row shows as `NO-DIFF` rather than as a pass — the pristine baseline was measured separately, above.

| # | shape | result | caught by |
|---|---|---|---|
| E1 | `#[doc = concat!(…)]` spanning source lines above the include | **FAIL** | reaches back to `"    )]"` |
| E2 | `///` claim above one `#[serde(alias = …)]` above the include | **FAIL** | reaches back to the `///` line |
| E3 | `#[cfg_attr(all(), doc = "…")]` above the include | **FAIL** | reaches back to the `cfg_attr` line |
| E4 | `/** … */` above the include | **FAIL** | reaches back to the `/**` line |
| E5 | decoy `    EEvidenceIntegrity,` in a `/* */` before the enum + live `///` claim | **FAIL** | reaches back to the `///` line (body scoping made the decoy irrelevant) |
| E5b | the same decoy moved **inside** the enum body | **FAIL** | ``occurs 2 times in the `ReasonCode` body`` |
| E6 | one `#[doc = "…"]` whose literal spans two source lines | **FAIL** | reaches back to `"and its inclusion proofs.\"]"` |
| C1 | control: `///` claim above the include | **FAIL** | as designed |
| C2 | control: `#[doc(hidden)]` above the include | **FAIL** | as designed |
| L1 | `#[serde(alias = …)]` between include and variant | **passes** | the F5 false failure is gone |
| L2 | `//` comment between include and variant | **passes** | the F4 false failure is gone |
| L3 | blank line between include and variant | **passes** | the P2-3 false failure is gone |

E5b is the one worth stating plainly: scoping to the enum body cannot exclude a decoy that is inside the body, and what catches it is the uniqueness count on `INTEGRITY_DECL`, not the scoping. Both are needed and both are present.

### F1 (follow-up) — five legitimate shapes are refused

The inversion traded failure directions, and this is the cost. Measured the same way:

| # | shape | result | why |
|---|---|---|---|
| O1 | `#[doc(alias = "E_EVIDENCE_INTEGRITY")]` above the include | **FAIL** | `cannot_document` refuses any single-line attribute containing `doc`; `doc(alias)` adds a search alias, not prose |
| O2 | `#[allow(rustdoc::bare_urls)]` above the include | **FAIL** | `"rustdoc"` contains `"doc"` as a substring; this is a lint allow, not documentation |
| O5 | a non-doc attribute wrapped across source lines (`#[cfg_attr(\n  feature = "serde",\n  serde(alias = …)\n)]`) | **FAIL** | disclosed by the builder as a deliberate narrowing |
| O6 | previous variant carries a trailing comment: `EReplayLimitExceeded, // kept for --exit-codes v1 consumers` | **FAIL** | `declares_a_variant` requires the line to end in `,` |
| O7 | the integrity variant moved to **first** position in the enum, documented honestly by the include alone | **FAIL** | `.expect("the `ReasonCode` body opens on the integrity variant")` — the message states the condition that caused it |

O7 verbatim, at `spec_reason_code_registry.rs:478`:

```
thread 'nothing_but_the_boundary_include_documents_the_integrity_variant' panicked at
crates/assay-cli/tests/spec_reason_code_registry.rs:478:10:
the `ReasonCode` body opens on the integrity variant
```

The `expect` message asserts as impossible the exact state that reached it. A reader who moves the variant to the top of the enum and nothing else gets told that the body opens on the integrity variant — which is true, and is the reason for the panic.

O1, O2 and O6 are the ones I would ask about, because they are correct code with no second doc source, and the next author who trips over one weakens the guard rather than reading it. Whether any of them is worth a code change, and the rest of the record, follow in one comment.

### Cargo.lock and the delegated lane — clean

- `git diff --name-status origin/main...HEAD` → one file, `crates/assay-cli/tests/spec_reason_code_registry.rs`.
- `git diff origin/main -- Cargo.lock | wc -l` → `0`. `Cargo.lock` is byte-identical to `origin/main`'s tip, not merely unchanged in the three-dot diff.
- `crates/assay-cli/Cargo.toml` is unchanged vs `main`; `rg '\bsyn\b|\bquote\b|proc-macro2'` on it finds nothing; `[dev-dependencies]` is `tempfile, regex, assert_cmd, predicates, proptest, serial_test, criterion`.
- `lane-check/proof` on this head reports `pass — no delegated runner proof required`, which is the consequence that matters: the `syn` version at `3dee8fd8` routed it to `gates=all`.

### Mutation and restore, for the record

Every mutation above was applied to `crates/assay-cli/src/exit_codes.rs` and restored with `cp /tmp/rev2241-exit_codes.rs.orig crates/assay-cli/src/exit_codes.rs`, a byte copy taken before the first edit and verified by SHA-256 (`8da15397f6ada3665cae7b0eb0ceca39ee8b90c0ff955ad89c494aab0d6dca58`) after each restore. **No bare-tree command was run at any point** — no `git checkout --`, no `git restore .`, no `git clean`. `git status --porcelain` in the review worktree is empty as of this comment.

Rul1an:
# Independent non-building review — final record (head `6cc52f1913db74dcd5845ae579683073d530b68d`)

**Verdict: APPROVE.** No blocking findings. Six follow-up findings, all dispositioned below. The five shapes the prior review found are closed, and closed at the root rather than by a longer prefix list. Merge is blocked by a red required check that this PR does not cause and does not touch — see *Merge state*.

**Non-building status.** I did not author this change, have not pushed to this branch, and have edited nothing on it. I authored no specification or implementation plan governing it. My mutations were made in my own detached worktree and reverted; the branch is untouched.

**Provenance.** Head `6cc52f1913db74dcd5845ae579683073d530b68d`, confirmed with `gh pr view 2241 --json headRefOid`. Read-only worktree `/Users/roelschuurkes/rev-2241`, detached at that SHA, merge-base `origin/main` `bc07ffb0e8f47313b4ce7353c63d3e22c6299514`. `cargo 1.96.0 (30a34c682 2026-05-25)` / `rustc 1.96.0 (ac68faa20 2026-05-25)`, pinned by `rust-toolchain.toml`. macOS aarch64. Rendering evidence comes from a separate probe crate at `/tmp/rev2241-probe` under the same toolchain, named as such below. Reproductions were published mid-review as an [interim record](https://github.com/Rul1an/assay/pull/2241#issuecomment-5243619511); this comment supersedes it and repeats what matters.

---

## Findings

### Blocking

**None.** Every finding below fails loudly on correct code rather than passing silently on wrong code, which is the direction the rewrite deliberately chose. None is live on this head. I would not hold a test-only hardening PR for any of them.

### Follow-up

**F1 — `cannot_document` matches `"doc"` as a substring, so `#[allow(rustdoc::bare_urls)]` is refused.** `crates/assay-cli/tests/spec_reason_code_registry.rs:411`. `!line.contains("doc")` also rejects `rustdoc::`, so any single-line `#[allow(rustdoc::…)]` / `#[expect(rustdoc::…)]` on this variant fails the guard. Measured: **FAIL**, and `cargo fmt --all -- --check` exit 0, so rustfmt does not rewrite it away. Rendering probe: the attribute adds nothing to the rendered variant and the boundary text renders intact. This is the one I would most expect a real author to hit, because a lint allow is what you reach for next to an included markdown file in a crate that runs `Rustdoc (deny warnings)`. *Disposition: follow-up.* Matching on the attribute's name rather than on `contains("doc")` would fix it, and the safe direction is preserved either way.

**F2 — `#[doc(alias = "…")]` is refused although it renders no prose.** Same line. Measured **FAIL**, fmt-clean. Probe evidence: `ALIAS_PROBE_VALUE` occurs **0** times in the rendered `enum.E.html` (it lands in the search index and the source listing only) while the boundary text renders unchanged. So this is correct code with no second doc source, refused. *Disposition: follow-up.* Unlike F1 this one is at least honest about itself — the assertion message says any `#[doc …]` is refused rather than guessed at.

**F3 — a non-doc attribute wrapped across source lines is refused.** Same line. Measured **FAIL** for `#[cfg_attr(\n feature = "serde",\n serde(alias = …)\n)]`. The builder disclosed this and asked it to be weighed: accepting it means deciding where an attribute ends, which is the open question again. *Disposition: accepted narrowing, no change.* It also has a second-order effect I would keep, not remove — see F6.

**F4 — `declares_a_variant` refuses a trailing comment on the previous variant's line.** `spec_reason_code_registry.rs:416-421`. `strip_suffix(',')` fails on `EReplayLimitExceeded, // kept for --exit-codes v1 consumers`, so the walk's stop line is not recognised as a declaration and the guard fails, naming a line the author has no reason to connect to `EEvidenceIntegrity`. Measured **FAIL**, and `cargo fmt --all -- --check` exit 0 — rustfmt preserves the trailing comment, so this is reachable in a formatted tree. *Disposition: follow-up.* Trimming a trailing `// …` before the suffix check would close it.

**F5 — the `expect` message states the condition that caused it.** `spec_reason_code_registry.rs:478`. Move the integrity variant to the top of the enum body, documented honestly by the include and nothing else, and the test panics with `the ReasonCode body opens on the integrity variant` — which is exactly what the author just did, presented as an impossibility. This is the same class of defect as the prior review's P2-3, in the one branch that was not reworked. *Disposition: follow-up.* The other diagnostics on this path are good; this one is not.

**F6 — a single-line `#[deprecated(note = "…")]` puts rendered text on the variant and is admitted.** `spec_reason_code_registry.rs:406-413`. This is a **third** thing the test does not pin, beyond the two the comment names. Measured: `#[deprecated(note = "Merkle root and inclusion proofs are checked")]` above the include is fmt-clean (72 columns) and the guard **PASSES** it. Probe evidence — the note renders on the variant, markdown-processed, in the `item-info` stab **ahead of** the docblock carrying the boundary text:

```html
<span class="item-info"><div class="stab deprecated"><span class="emoji">👎</span><span>Deprecated
since 5.1.0: <p>MERKLE-CLAIM-PROBE: the verifier checks the Merkle root and its inclusion
proofs</p></span></div></span><div class="docblock"><p>The verifier compares recorded hashes.
BOUNDARY-TEXT-PROBE.</p></div>
```

That is the PR body's own failure shape — a withdrawn claim rendering ahead of the boundary text while the checked region stays clean — reached through a channel the rule admits by design, since a deprecation note is not in the `doc` namespace.

Two measurements bound it, which is why it is follow-up and not blocking. First, a note long enough to matter does not survive formatting: at 105 columns rustfmt wraps the attribute across lines, and F3's narrowing then catches it (`reaches back to "    )]"`). Second, and decisively, the shape trips a different gate — `cargo check -p assay-cli --bins` reports `warning: use of deprecated unit variant 'exit_codes::ReasonCode::EEvidenceIntegrity': Merkle root and inclusion proofs are checked` at three use sites under `#[warn(deprecated)]` on by default, so `Clippy (deny warnings)` turns it into an error. *Disposition: follow-up, and the honest fix is a sentence, not code* — the comment at `:445-447` should name this third gap alongside the other two rather than leave the enumeration reading as complete.

**F7 (minor) — the boundary path is written twice.** `spec_reason_code_registry.rs:343` and `:392-393`: `BOUNDARY` carries `crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md` and `BOUNDARY_INCLUDE` embeds `exit_codes/evidence_integrity_boundary.md`. Two literals that must agree, which is the AGENTS.md "one rule, one function" shape. It is minor because the drift direction is refusal — renaming the file and updating one of them fails the test rather than passing it — and `BOUNDARY_INCLUDE` could be derived from `BOUNDARY`. *Disposition: follow-up, pre-existing (`BOUNDARY` is not introduced by this PR).*

---

## Result per angle

### Angle 1 — does it now over-refuse? **Yes, in five shapes. This is the review's finding, and none of it blocks.**

F1–F5 above. Three legitimate shapes I expected to be refused are not: `#[cfg(not(target_os = "windows"))]`, a single-line `#[deprecated(since, note)]`, and doc text on the enum itself all pass. What makes the over-refusal survivable rather than corrosive is that four of the five failures name the offending line and the message at `:482-489` states the admitted set in the author's own vocabulary ("blank lines, `//` comments and single-line attributes that mention `doc` nowhere … are fine"). A reader who trips over F1 or F2 is told the rule they broke. F5 is the exception and is the one I would fix first.

### Angle 2 — do the seven evasions really fail now? **All seven fail, each for the right reason, and the test still discriminates.**

Baseline on the pristine head: `cargo test -p assay-cli --test spec_reason_code_registry` → `9 passed; 0 failed`. I reproduced each shape myself rather than reading the table.

| # | shape | result | what caught it |
|---|---|---|---|
| E1 | `#[doc = concat!(…)]` spanning source lines, above the include | **FAIL** | reaches back to `"    )]"` |
| E2 | `///` claim above one `#[serde(alias = …)]` above the include | **FAIL** | reaches back to the `///` line |
| E3 | `#[cfg_attr(all(), doc = "…")]` above the include | **FAIL** | reaches back to the `cfg_attr` line |
| E4 | `/** … */` above the include | **FAIL** | reaches back to the `/**` line |
| E5 | decoy `    EEvidenceIntegrity,` in a `/* */` before the enum, claim live on the real variant | **FAIL** | reaches back to the `///` line — the decoy is out of scope, so the real claim is what fails |
| E5b | the same decoy moved **inside** the enum body | **FAIL** | ``occurs 2 times in the `ReasonCode` body`` |
| E6 | one `#[doc = "…"]` whose literal spans two source lines | **FAIL** | reaches back to `"and its inclusion proofs.\"]"` |
| C1 | control: `///` claim above the include | **FAIL** | as designed |
| C2 | control: `#[doc(hidden)]` above the include | **FAIL** | as designed |
| C0 | control: honest form, unmodified | **passes** | 9/9, measured separately |
| L1 | `#[serde(alias = …)]` between include and variant | **passes** | the prior F5 false failure is gone |
| L2 | `//` comment between include and variant | **passes** | the prior F4 false failure is gone |
| L3 | blank line between include and variant | **passes** | the prior P2-3 false failure is gone |

E5b is the interesting one, and the answer to what catches it is not the scoping. Scoping to the enum body cannot exclude a decoy that is inside the body; what catches it is the uniqueness count on `INTEGRITY_DECL` at `:463-469`. Both mechanisms are load-bearing and both are present, so the pair is right — but the PR's framing ("the anchor is scoped … and asserted unique") should be read as two independent defences, not one.

Two things I checked that the table does not cover. A decoy at an indentation other than four spaces is not counted by `INTEGRITY_DECL` (`"\n    EEvidenceIntegrity,"`), and it is also not what `split_once` lands on, so the slice read is still the real declaration — the uniqueness count and the anchor agree on which shapes matter. And doc text placed on the same line as the previous variant's declaration, or on the same line as the include, is refused: neither line is `cannot_document` and neither declares a variant.

### Angle 3 — does the name match what it checks? **Yes, literally. The list of gaps beside it is one short.**

`nothing_but_the_boundary_include_documents_the_integrity_variant`. Read against the implementation, the conjunction asserted is: the enum header occurs exactly once in the file (`:449-454`); `    EEvidenceIntegrity,` occurs exactly once inside that enum's body (`:463-469`); every line between the previous variant's declaration and this one is blank, a `//` comment, a single-line attribute naming `doc` nowhere, or the boundary include (`:475-490`); and the include occurs there exactly once (`:491-499`). Because rustc's doc for an item comes from `///` lines and `#[doc]` attributes in that item's attribute list, and everything between two variants belongs to the later one, that conjunction *is* "nothing but the include documents this variant" within the `doc` namespace — not an approximation of it. The previous name asserted a property of one adjacent line; this one asserts a property of a region. The difference is real.

The two gaps the builder names are gaps, and I verified the first rather than granting it: `///` appended to the enum's own doc comment **passes** (it documents the enum, not the variant — correctly not counted). The rendering step is likewise not read.

They are not the only two. F6 is a third: rendered text attached to this variant that does not live in the `doc` namespace. The name survives it — a deprecation note is not documentation — but the comment's "what this does not pin, so that it is not read as more than it is" reads as exhaustive and is not.

### Angle 4 — `Cargo.lock` and the delegated lane. **Clean.**

- `git diff --name-status origin/main...HEAD` → exactly one file: `crates/assay-cli/tests/spec_reason_code_registry.rs`, +104/−8.
- `git diff origin/main -- Cargo.lock | wc -l` → `0`. `Cargo.lock` is byte-identical to `origin/main`'s tip, which is stronger than unchanged in the three-dot diff.
- `crates/assay-cli/Cargo.toml` is unchanged vs `main`. `rg '\bsyn\b|\bquote\b|proc-macro2'` on it finds nothing. `[dev-dependencies]` is `tempfile, regex, assert_cmd, predicates, proptest, serial_test, criterion` — no addition.
- `lane-check/proof` on this head: `pass — no delegated runner proof required`. `host-capability-check`: `pass`.

The builder's account of why the `syn` version was abandoned is therefore consistent with what this head carries, and the decision was the right one on the merits they give: a dev-dependency that moves `Cargo.lock` routes a required check to `gates=all`, and dispatching an override to clear a gate your own change tripped is not a builder's call.

### Angle 5 — what the test cannot see. **Three boundaries; the PR's claims stay inside all three.**

It reads source text, so: (1) it does not read rustdoc's output, which is where F6 lives and is the only place I found that the source-text reading and the rendered result disagree; (2) it does not pin the *content* of the included file — `the_evidence_integrity_boundary_reads_the_same_in_the_spec_and_the_code` pins it against the §5.1 row and `the_boundary_claims_no_structure_the_verifier_does_not_build` forbids `Merkle` in it, both green here, so the content is covered but by siblings and not by this test; (3) it reads `crates/assay-cli/src/exit_codes.rs` from the workspace root **at test runtime**, so its subject is the file on disk when the test runs, not the bytes compiled into the binary running it. That last one is theoretical in CI and was a practical hazard for me — see the method note.

The PR stays inside this. The body says test-only and claims no security fix; the comment at `:445-447` says in as many words that it reads "the source rustdoc is given and not rustdoc's output". The `main` non-claim also still holds on this head: `exit_codes.rs:81-84` are plain `//` comments, and nothing is mis-documented today.

---

## Verification

On the pristine head, first-hand:

```
$ git rev-parse HEAD
6cc52f1913db74dcd5845ae579683073d530b68d
$ cargo test -p assay-cli --test spec_reason_code_registry
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
$ cargo fmt --all -- --check          # exit 0
$ git diff --check                    # clean
$ git status --porcelain              # empty
```

Every mutation row: inject into `crates/assay-cli/src/exit_codes.rs`, prove the injection reached disk (`git diff --stat` for that path non-empty — the harness refuses a `NO-DIFF` row, which is why the honest control shows as refused there and was measured separately), then `cargo test -p assay-cli --test spec_reason_code_registry nothing_but_the_boundary_include_documents_the_integrity_variant -- --exact`, then restore. Verbatim, the F5 panic:

```
thread 'nothing_but_the_boundary_include_documents_the_integrity_variant' panicked at
crates/assay-cli/tests/spec_reason_code_registry.rs:478:10:
the `ReasonCode` body opens on the integrity variant
```

and the shape that F6 admits:

```
$ cargo fmt --all -- --check   # exit 0, with #[deprecated(note = "Merkle root ...")] above the include
$ cargo test ... nothing_but_the_boundary_include_documents_the_integrity_variant -- --exact
test result: ok. 1 passed; 0 failed
$ cargo check -p assay-cli --bins
warning: use of deprecated unit variant `exit_codes::ReasonCode::EEvidenceIntegrity`:
         Merkle root and inclusion proofs are checked
= note: `#[warn(deprecated)]` on by default          # x3 use sites
```

### Mutation and restore

Every mutation touched exactly one file, `crates/assay-cli/src/exit_codes.rs`. Every restore was `cp /tmp/rev2241-exit_codes.rs.orig crates/assay-cli/src/exit_codes.rs` — a byte copy taken before the first edit, verified after each restore against SHA-256 `8da15397f6ada3665cae7b0eb0ceca39ee8b90c0ff955ad89c494aab0d6dca58`. **No bare-tree command was run at any point in this review**: no `git checkout --`, no `git restore`, no `git stash`, no `git clean`.

| mutation | insertion point | restored by |
|---|---|---|
| E1–E6, C1, C2, F1, F2, F3, F6 (12 shapes) | above the include, `exit_codes.rs:85` | `cp /tmp/rev2241-exit_codes.rs.orig crates/assay-cli/src/exit_codes.rs` |
| L1, L2, L3 | between the include and the declaration, `exit_codes.rs:86` | same |
| E5 decoy | before the enum, `exit_codes.rs:56` | same |
| E5b decoy | inside the enum body, `exit_codes.rs:60` | same |
| F4 | replaced `exit_codes.rs:80` in place | same |
| F5 | moved lines 85–86 to the top of the enum body | same |
| enum-level doc (angle 3) | replaced `exit_codes.rs:56` in place | same |

**One integrity incident, recorded because it bears on the numbers.** Partway through the rustfmt reachability pass the machine's data volume hit 100% (`117Mi` available), and the restore `cp` itself failed with `No space left on device`, leaving the file holding the `#[deprecated]` mutation — 26702 bytes, SHA-256 `8ad297f08227f0fd99db3d9321c547379e2e37d1f83ff78094dba578bb875b3e`. Nothing was lost: the pristine copy in `/tmp` was intact and the file was truncated by nothing. I recovered it by re-running the same named-path `cp` once space was free, and re-verified `8da15397…`. The two rows measured before the failure (F1, F2) were re-derived from output captured before it; the two after it (F6, F4) were re-run from scratch afterwards. The general point is the one the prior reviewer made about unwritten injections: a restore is a command that can fail, so its success has to be asserted rather than assumed, which is why the SHA check is in the loop.

---

## Merge state

`mergeStateStatus: BLOCKED`, `mergeable: MERGEABLE`. Of the three required checks on this head, `host-capability-check` and `lane-check/proof` pass and **`CI` is red**. It is red only through `Build + Test (windows-latest)`:

```
bounded_process::tests::kills_descendant_that_holds_inherited_output_open ... FAILED
bounded_process::tests::kills_quiet_descendant_after_normal_parent_exit ... FAILED
test result: FAILED. 15 passed; 2 failed
```

That is [#2249](https://github.com/Rul1an/assay/issues/2249), open and owned by another agent, and this PR does not touch `tests/support/bounded_process.rs` — confirmed against the changed-file list, which is one test file. **I am not merging and I did not attempt to.** Nothing in this review should be read as a green required-check claim: a required check is red, and per AGENTS.md the auto-merge preconditions are not met until it is green. This record exists so that the review quorum is not what is missing when it clears.

## Non-claims

- **I did not merge, enable auto-merge, or use `--admin`, and I did not compute which open PRs need a host-capability proof** — that is another agent's in-flight work and I left it alone.
- **I did not re-run clippy or the wider `assay-cli` suite locally.** I rely on `Clippy (deny warnings)` and `Build + Test (macos-latest)` being green on this exact head in CI, and on the builder's `883 passed` at `59d84bb62599e04961a786ab64b67bc9ab4e130f`, which I did not reproduce. My own first-hand runs are the ones listed under *Verification*.
- **The rendering evidence is from an isolated probe crate**, `/tmp/rev2241-probe`, under the same pinned `rustdoc 1.96.0 (ac68faa20)`. I did **not** build `assay-cli`'s own rustdoc. So F2 and F6 are claims about how rustdoc 1.96.0 treats those attributes on an enum variant, not about the shipped `assay-cli` documentation.
- **F6's mitigation is measured, not argued, but it is measured on one command.** `cargo check -p assay-cli --bins` emits the deprecation warnings; I did not run CI's clippy invocation to watch `-D warnings` convert them to errors.
- **I did not verify the PR body's provenance narrative** about #2213, the recovered worktree modification, or the `3dee8fd8` history beyond confirming that this head carries no dev-dependency and no `Cargo.lock` movement.
- **I did not investigate #2249** and take no position on whether it is intermittent or deterministic beyond confirming which two tests fail and that this PR does not touch that helper.
- **I did not assess whether any over-refused shape in F1–F5 would survive human review as a real change.** They are shapes a Rust author could plausibly write; whether one would land here is a judgement I did not make for them.
- **The findings are exhaustive over the attack surface I searched**, which is source-text shapes around this one variant plus rustdoc rendering of the attributes I named. I did not search macro expansion, `build.rs` code generation, or `--cfg doc`-conditional items.

Rul1an:
## Prepared update, deliberately not pushed

Head is unchanged at `6cc52f1913db74dcd5845ae579683073d530b68d` and the approval on it stands. A follow-up fix was finished and verified but **withheld on purpose**: every push voids the approval, and this branch is blocked by #2249 regardless, so it gets one update rather than two.

**What the prepared change does.** Fixes the substring match in `cannot_document` (`!line.contains("doc")` refuses `#[allow(rustdoc::bare_urls)]` because "rustdoc" contains "doc") and refuses the `#[deprecated(note = …)]` shape. Both are recorded independently in #2260, so nothing depends on the artifacts below surviving.

**Where it is**, for whoever performs the single update after #2249 lands:

- pinned ref `refs/wip/2241-guard` → `0dfd54969582ca04a4616c1941a811e57def205a` (branch head plus a merge of `origin/main` `ea161accd`)
- bundle and patches under `~/pr2241-guard-preserved/`, with a README carrying the full 18-row before/after matrix
- worktree `~/wt-2199-patch`

**Verified on `0dfd54969`:** 9/9 focused, `cargo test -p assay-cli` 897 passed / 0 failed, fmt, clippy `-D warnings`, `git diff --check` clean, `Cargo.lock` byte-identical to `origin/main`. All five over-refusals pass, all seven evasions and both controls still fail, `#[deprecated]` flips to refused.

Whoever updates this branch should re-verify rather than trust that summary — `main` will have moved again by then, and these numbers were measured against `ea161accd`.

Rul1an:
## Reconciliation update — head `2fc43bb91b24d294b5faecbd78378aa70bf1f991`

Sole Cursor writer pass on `cursor/2199-doc-block-transport`. Did **not** mark ready / merge. Did not touch #2231 / rsa-advisory.

### Provenance
- Worktree: `/Users/roelschuurkes/wt-2199-patch` (existing isolated checkout for this branch)
- Base (`origin/main`): `c9b036de8198d3bfa7f9752d414863b20c9b7c18` (verified tip match before merge)
- Prior remote head: `6cc52f1913db74dcd5845ae579683073d530b68d`
- New head: `2fc43bb91b24d294b5faecbd78378aa70bf1f991`
- Merge: `origin/main` → branch, **conflict-free** merge commit (`ort`), no rebase / no history rewrite
- PR net diff vs `origin/main`: only `crates/assay-cli/tests/spec_reason_code_registry.rs` (+227/−8)
- `Cargo.toml` / `Cargo.lock`: **untouched** in the PR diff
- Toolchain: `cargo 1.96.0 (30a34c682)` / `rustc 1.96.0 (ac68faa20)`, pinned by `rust-toolchain.toml`, macOS aarch64

### Code beyond the merge
No new edits in this pass. The merge head also carries the previously prepared (local, unpushed) guard commit `d514b09f3` (*match the attribute's path, and refuse the note that renders ahead of it*), which was already on this worktree ahead of remote before the main merge. That commit is test-only in the same hard-scoped file.

### Attack shapes re-verified on `2fc43bb91`
Each row: edit `crates/assay-cli/src/exit_codes.rs`, run  
`cargo test -p assay-cli --test spec_reason_code_registry nothing_but_the_boundary_include_documents_the_integrity_variant -- --exact`, then restore from a pre-mutation byte copy. Harness refused any `NO-DIFF` row. Final `exit_codes.rs` sha256 matched the backup; focused suite re-ran green after restore.

| # | shape | result |
|---|---|---|
| 1 | extra `///` claim above include | **FAIL** — reaches back to the `///` line |
| 2 | `/** … */` block doc above include | **FAIL** — reaches back to the `/**` line |
| 3 | multiline `#[doc = "…"]` above include | **FAIL** — reaches back to the multiline `#[doc]` unit |
| 4 | `#[cfg_attr(all(), doc = "…")]` above include | **FAIL** — reaches back to the `cfg_attr` line |
| 5 | `#[doc(hidden)]` above include | **FAIL** — reaches back to `#[doc(hidden)]` |
| 6 | benign `#[allow(dead_code)]` between include and variant | **PASS** (expected; non-doc) |
| 7 | comment containing `    EEvidenceIntegrity,` before enum + live `///` claim | **FAIL** — reaches back to the live `///` (body scoping) |
| 8 | duplicate/ambiguous `    EEvidenceIntegrity,` inside enum body (`/* */`) | **FAIL** — `occurs 2 times in the ReasonCode body` |
| 9 | `///` above `#[serde(alias=…)]` above include | **FAIL** — reaches back to the `///` line |

Representative three called out by the reconciliation brief:
- multiline / `cfg_attr`: rows 3 and 4
- anchor ambiguity: row 8

Explicit non-claims preserved in the test comment: **enum-level docs** and **rendered rustdoc** are not covered.

### Verification on exact head `2fc43bb91` (exit codes)
- `cargo test -p assay-cli --test spec_reason_code_registry` → **0** (9 passed)
- `cargo test -p assay-cli` → **0** (all suites green; unit suite 547 passed / 0 failed among others)
- `cargo fmt --all -- --check` → **0**
- `cargo clippy -p assay-cli --all-targets -- -D warnings` → **0**
- `git diff --check` (`origin/main...HEAD`) → **0**

This push invalidates any review measured on a prior head. Fresh final-head review still required. PR left draft/non-ready.

Rul1an (commented):
## Quorum review — READY

**Bound head:** `2fc43bb91b24d294b5faecbd78378aa70bf1f991` (re-checked via `gh pr view 2241 --json headRefOid` immediately before posting; match).

**Identity / process:** Non-building, read-only review of this head only. Detached worktree `/tmp/assay-pr2241-quorum-review` at that SHA. No edits, push, ready-toggle, or merge. Same GitHub identity as the PR author (`Rul1an`), so this is **COMMENT + READY** rather than APPROVE. Builder self-review and prior-head reviews are not relied on.

**Scope:** Net `origin/main...HEAD` is only `crates/assay-cli/tests/spec_reason_code_registry.rs` (+227/−8). `Cargo.toml` / `Cargo.lock` untouched.

### Verdict on the asked properties

1. **Non-tautological / fail-closed over the whole variant doc block — holds.** The walk admits only units that cannot put text on the variant (blank, `//` non-doc, non-`doc`/`deprecated` attribute paths, exact `BOUNDARY_INCLUDE`) and otherwise requires the stop line to be the previous variant declaration; unrecognized shapes fail closed. The include must appear exactly once in the variant’s own text.
2. **Required mutation matrix — all match.** Extra `///`, `/** */`, multiline `#[doc]`, `cfg_attr` doc, `doc(hidden)`, decoy/comment anchor, and in-body duplicate all **FAIL**; benign non-doc attribute **PASS**.
3. **Explicit non-claims confirmed:** enum-level docs are outside the enum-body slice (injected enum doc still **PASS**); rendered rustdoc is not asserted (stated in the test comment and PR body).

### Mutation evidence (this head)

Each row: edit `crates/assay-cli/src/exit_codes.rs`, run
`cargo test -p assay-cli --test spec_reason_code_registry nothing_but_the_boundary_include_documents_the_integrity_variant -- --exact`, restore from pre-mutation byte copy. Harness refused empty diffs; final `exit_codes.rs` sha256 matched backup (`1ee7a75d…`).

| # | shape | result |
|---|---|---|
| 1 | extra `///` claim above include | **FAIL** — reaches back to the `///` line |
| 2 | `/** … */` block doc above include | **FAIL** — reaches back to the `/**` line |
| 3 | multiline `#[doc = "…"]` above include | **FAIL** — reaches back to the multiline `#[doc]` unit |
| 4 | `#[cfg_attr(all(), doc = "…")]` above include | **FAIL** — reaches back to the `cfg_attr` line |
| 5 | `#[doc(hidden)]` above include | **FAIL** — reaches back to `#[doc(hidden)]` |
| 6 | benign `#[allow(dead_code)]` between include and variant | **PASS** |
| 7 | comment containing `    EEvidenceIntegrity,` before enum + live `///` | **FAIL** — reaches back to live `///` (body scoping) |
| 8 | duplicate `    EEvidenceIntegrity,` inside enum body (`/* */`) | **FAIL** — `occurs 2 times in the ReasonCode body` |
| 9 | `///` above `#[serde(alias=…)]` above include | **FAIL** — reaches back to the `///` line |

Additional probes (not blockers): `#[doc = concat!(…)]` multiline **FAIL**; multiline `cfg_attr`+`doc` **FAIL**; `#[deprecated(note=…)]` **FAIL**; `#[allow(rustdoc::bare_urls)]` **PASS** (path-based, not substring); `#[doc(alias=…)]` **PASS** (stated exception; search-index only); enum-level Merkle line **PASS** (named non-claim). `#[must_use = "…"]` is admitted by the guard but is a rustc future-compat warning on enum variants (`cannot be used on enum variants`) — not a live documentation channel; confidence below quorum threshold.

### Verification

- Baseline focused test: **ok**
- Full `spec_reason_code_registry`: **9 passed; 0 failed**
- Toolchain: `cargo 1.96.0 (30a34c682)` / `rustc 1.96.0 (ac68faa20)`, pinned by `rust-toolchain.toml`, macOS aarch64

### Findings ≥80

**None.**

**READY** for `2fc43bb91b24d294b5faecbd78378aa70bf1f991`.

Issue #2246, opened 2026-08-10
Build + Test (windows-latest) failed and passed on unchanged trees four times today


Rul1an:
`CI / Build + Test (windows-latest)` failed and then passed on an unchanged tree at least four times today, across three different pull requests and on `main`. Each occurrence cost a re-run, and — the part that matters more — each one made a red Windows result uninformative until someone spent time deciding whether it meant anything.

## Occurrences, 2026-08-10

| where | head | observation |
|---|---|---|
| `main` | `4125d81b`, `5ddc9010` red; `cd8dc5d6` green | reported while investigating an unrelated gate |
| PR #2236 | pre-merge head | failed, took the rollup red, passed on re-run |
| PR #2239 | `4ca3b122` | **failed twice, then passed on a third run with no change to the tree** |
| PR #2241 | `c520deac` | failed in `bounded_process::tests::kills_quiet_descendant_after_normal_parent_exit` and `kills_descendant_that_holds_inherited_output_open`, in a helper that PR does not touch |

The named tests in the #2241 occurrence are the strongest lead: both are in `bounded_process` and both concern killing descendant processes — one after the parent exits normally, one that holds an inherited output handle open. Process-tree teardown and inherited-handle lifetime are timing-dependent on Windows in ways they are not on Unix, which is consistent with a race rather than with a defect in any of the changes that observed it.

## Why this is worth fixing rather than re-running

`AGENTS.md` treats GitHub Actions as the final integration proof and forbids weakening a required check to make a branch mergeable. A check that fails on an unchanged tree undermines that from the other direction: it does not weaken the gate on paper, it teaches everyone reading it to discount a red result. Once a red Windows run is routinely re-run rather than investigated, a genuine Windows regression arrives looking exactly like the noise.

The agents working today handled it correctly — one recorded a disposition on the PR rather than quietly re-running, another dispatched a re-run and explicitly declined to call it a flake until it came back. That is the right behaviour and it is also the evidence that this is costing real attention on every PR.

## What would close this

Determine whether these tests are racy or whether they are detecting something real about process teardown on Windows. The distinction is the whole issue, and it is answerable: run the `bounded_process` tests in a loop on a Windows runner and measure the failure rate on an unchanged tree. A test that fails at a measurable rate with nothing changing is racy by definition; one that never reproduces under repetition is something else and needs a different investigation.

If they are racy, fix the race rather than adding a retry, since a retried assertion about process teardown asserts less than it appears to.

## Non-claims

[...]

PR #2251, opened 2026-08-10
fix(test-support): name the ending a bounded run reached, and stop timing PowerShell startup

### LINKED ISSUE #2249: windows-latest: two bounded_process descendant tests fail intermittently

## The failure

```
```

```
```

```
```

## The datum worth explaining

## What would close this

## Non-claims

Rul1an:
**Does not close #2249 yet.** The closing keyword is deliberately absent: this branch may not be shown to fix an intermittent defect until rate evidence at *this* head says so. See "What is still unproven".

Bound to head **`5ca462fc18e108035b6e61262f721e5dc60f5420`** (branch `cursor/fix-2249-bounded-process-windows`, base `main` at `bc07ffb0e`). Earlier evidence in this body referred to `870a02c9` and has been removed: that head's Windows contract and its control channel are both gone.

## What this branch now contains

**1. The harness says which ending it reached.** `deadline of 5s expired` and `status=exit code: 1` were both printed for one run and the pairing was meaningless: `process_wrap`'s Windows `start_kill` is `TerminateJobObject(job, 1)`, so `exit code: 1` was the harness's own kill code rendered exactly like a child's choice. The deadline text now names what it was still waiting for — `outstanding=[child still running]` versus `outstanding=[stdout handle still open, stderr handle still open]` — and a status reaped after termination is labelled as such instead of as `(child exit)`.

**2. PowerShell startup is out of the bounded window.** On Windows the bounded child is this test binary re-executed in a helper mode. One `powershell.exe` start was observed at ~18s inside a 5s bound (run 31404812176), so interpreter startup, not process-tree teardown, was deciding the result.

**3. The descendant pid no longer travels through a path from the environment.** CodeQL's high on `870a02c9` and `8b232621` was a real sink: the helper read `PID_FILE_ENV` into a `PathBuf` and wrote to it. Sanitising the path would narrow the sink; removing the channel removes it. The pid is now one `READY pid=<u32>` record on both captured channels, parsed strictly — digits only, non-zero `u32`, and every record present must agree. Identical repetition is admitted because the record goes to both channels and the diagnostic quotes both; a disagreement is an error, since two pids cannot say which descendant a teardown claim is about.

**4. Each platform asserts the ending it can reach.** Job 93527135001 showed a `Stdio::null()` descendant keeping the run's handles open after the child exited 0 with all of its bytes read. On Windows `CreateProcess` runs with `bInheritHandles = TRUE` and an inherited handle stays inheritable, so every process the bounded child spawns holds the run's stdout and stderr. A live descendant plus a successful bounded run is therefore reachable on Unix and not on Windows. Unix keeps `kills_quiet_descendant_after_normal_parent_exit`; Windows asserts the deadline ending with the descendant still killed, and **fails if the inheritance ever stops**, which is the signal to make the test cross-platform again.

## Evidence

GREEN, macOS `darwin 25.6.0`, cargo/clippy 1.96.0, head `5ca462fc1`, worktree `/tmp/assay-2249`:

- `cargo test -p assay-mcp-server --test agent_golden_path_contract bounded_process` — 14 passed, 1 ignored (the re-executed helper);
- `cargo clippy -p assay-mcp-server --tests -- -D warnings`, `cargo fmt --all -- --check`, pre-commit and pre-push hooks: pass.

The new assertions were checked against mutations rather than assumed to bite: dropping the diagnostic's `outstanding=` detail, and reverting the pid parser to a best-effort scan, each fail the tests that guard them. The parser's table caught the diagnostic's repeated record before Windows could, and showed an explicit digit-length check to be redundant with the overflow check.

## What is still unproven

- **Windows at this head.** No Windows evidence is bound to `5ca462fc1` yet. #2249 is intermittent, so one green run would not settle it; the rate instrument in #2253 is being used instead, and its numbers will be reported here with the exact head they were taken on.
- **Which process holds the open handles.** The claim that the *descendant* holds them is inferred from the child having exited with its output fully read and only that descendant alive. The decisive control — the same helper spawning no descendant at all, which should then reach EOF — has not been run on Windows.
- Review quorum: unsatisfied. No non-building review exists on this head.

### DIFF
--- a/tests/support/bounded_process.rs
+++ b/tests/support/bounded_process.rs
@@ -556,7 +556,8 @@ mod tests {
         /// Keeps stdout and stderr open after its parent exits, so the reader
         /// threads never reach EOF.
         Inherited,
-        /// Detached from both, so the readers finish when the parent exits.
+        /// Asks for neither handle. Whether asking is enough is a platform
+        /// question, answered where the two descendant tests are cfg-split.
         Detached,
     }
 
@@ -882,6 +883,24 @@ mod tests {
         }
     }
 
+    /// A descendant that asks for no output handles releases the bounded run's
+    /// on Unix, and cannot on Windows. `CreateProcess` is called with
+    /// `bInheritHandles = TRUE`, and a handle the bounded child inherited stays
+    /// inheritable inside it, so every process it spawns receives the run's
+    /// stdout and stderr write handles whatever its own stdio is set to.
+    ///
+    /// Job 93527135001 measured exactly that: the child exited 0, every byte it
+    /// wrote had been read, and the readers still had no EOF with only a
+    /// `Stdio::null()` descendant alive. The spawner this replaced escaped it
+    /// through `Start-Process`, which goes via `ShellExecuteEx` and inherits no
+    /// handles -- at the price of a `powershell.exe` start inside the bounded
+    /// window, which is the defect in #2249.
+    ///
+    /// So the success path with a live descendant is reachable on Unix and not
+    /// on Windows, and each platform asserts the ending it can actually reach.
+    /// The Windows test fails if that ever stops being true, which is the signal
+    /// to restore one cross-platform test here.
+    #[cfg(unix)]
     #[test]
     fn kills_quiet_descendant_after_normal_parent_exit() {
         let temp = tempfile::tempdir().expect("temporary pid directory");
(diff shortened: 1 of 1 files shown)


Rul1an:
## Correction required — current Windows evidence falsifies the proposed fix

Measured on PR head `870a02c9fb10f8296f90ab2b0c32da8cc3478dd0`.

The first Windows run does not support the PR's current root-cause or completion claim:

- the re-executed Rust helper starts, writes `READY`, and exits successfully in `0.00s`;
- the bounded runner still reaches its deadline because the stdout and stderr handles remain open;
- `kills_quiet_descendant_after_normal_parent_exit` still fails;
- the new helper self-test `the_re_executed_helper_publishes_a_pid_and_reports_readiness` also fails;
- CodeQL reports a high-severity uncontrolled path use at `tests/support/bounded_process.rs:649`: `PID_FILE_ENV` is converted directly to a `PathBuf` and written.

This proves PowerShell startup overhead may have explained an earlier observation, but it is not the complete cause of #2249. The replacement helper still permits a descendant/output-handle condition that the implementation does not account for.

Required before another review:

1. Update the PR/issue claim: this head does not fix #2249.
2. Do not rerun unchanged code; first identify which process owns the open capture handles on Windows.
3. Redesign the helper/control channel so the test does not write to an arbitrary environment-controlled path. If a file is retained, construct it under a trusted test root from an allowlisted opaque component before filesystem access; a non-filesystem control channel is preferable.
4. Keep diagnostic improvements only if they remain independently useful and tested.
5. Provide Windows evidence on the new exact head, including a negative control or repeated run sufficient to distinguish a fix from the intermittent behavior already recorded in #2249.
6. Obtain one independent non-building review on that final head after CI is green.

No merge on this head.

Rul1an:
## Follow-up on current head `8b23262140d91cdb5c96a094a21448fc795200df`

The platform split now reflects the Windows observation instead of continuing to claim that a detached descendant reaches normal completion. That addresses the semantic mismatch found on `870a02c9`.

Two blockers remain:

1. The CodeQL source is still present. `descendant_spawner_helper_process` reads `PID_FILE_ENV` into a `PathBuf` and writes it at current lines 638–651. Splitting the tests does not remove the uncontrolled-path finding. The cleanest design is to publish the descendant PID in the already bounded/captured readiness record (for example one parseable `READY pid=<u32>` line) and parse it from the bounded result/error. That removes the filesystem control channel and the arbitrary-path sink entirely.
2. The PR body and verification are still bound to old head `870a02c9`, while the current head changes the Windows contract and includes a later `main` merge. Replace the completion claim with evidence from the current final head. Because #2249 is intermittent, one green Windows CI run is insufficient: use the hardened rate instrument from #2253, bound to this exact head, and report observed rates without causal overclaim.

Until both are resolved, `Closes #2249` is premature and the review quorum remains unsatisfied.

Rul1an:
## Windows rate evidence, and what it does not settle

Taken with the hardened instrument from #2253 (`919f1b1f0`), on `windows-latest` image `win25-vs2026 20260803.193.1`, rustc 1.96.0, 20 iterations per arm, 300s per-iteration ceiling. Selection: `-p assay-mcp-server --test agent_golden_path_contract bounded_process`.

| tree | measured head | subject blob | load=none | load=workspace-tests |
|---|---|---|---|---|
| `main` @ `bc07ffb0e` | `919f1b1f04d045ac70ef854713a339051ec41431` ([run 31417348260](https://github.com/Rul1an/assay/actions/runs/31417348260)) | `0db763f87636` | **0 / 20** | **2 / 20** |
| this branch @ `5ca462fc1` | `42ac2a841c8ea00331a0401caad24937ca30ad0d` ([run 31418630120](https://github.com/Rul1an/assay/actions/runs/31418630120)) | `aad18d216bd8` | **0 / 20** | **1 / 20** |

Every iteration was measured; no iteration executed nothing; no arm lost its load (minimum per-iteration load coverage 1.0 and 0.9565).

**Four reasons not to read this as "fixed".**

1. **The measured heads are not this PR's head.** Both proofs were taken on proof branches carrying this tree plus the instrument, because `workflow_dispatch` resolves from the default branch and #2253 has not landed. The instrument's own reuse rule requires an identical `head_sha`, so by its own rule these rates do not transfer to `5ca462fc1`. The subject blob is identical; that is provenance, not the rule. Once #2253 lands, `-f ref=5ca462fc1` measures this head exactly.
2. **The remaining failures are not the #2249 pair, in either tree.** Baseline: one 300s ceiling breach and one `stderr_flood_uses_its_own_ceiling_and_bounded_diagnostic`. Candidate: one 300s ceiling breach, no test-level failure. `stderr_flood` is untouched by this branch, so its failure is load sensitivity elsewhere in the same file — recorded on #2246's question rather than this one.
3. **Both ceiling breaches were the instrument's own load build, not either tree.** Iteration 2 breached and iteration 3 took 243s in *both* runs. #2253 `37ead31ce` now builds the load selection before measuring; every rate above predates that and carries the confound.
4. **20 iterations cannot separate this fix from luck.** The #2249 pair failed once in twenty in the earlier baseline arm and not at all here. At a rate near 5%, twenty iterations distinguish very little; the interval around 1/20 and around 0/20 overlap almost entirely. What would settle it is a much larger iteration count at this exact head, and a control that isolates the mechanism.

**The control still owed.** The claim that the *descendant* holds the run's output handles on Windows is inferred, not measured: the child exited 0 with all its bytes read and only a `Stdio::null()` descendant alive. The decisive run is the same helper spawning **no** descendant, which should then reach EOF and complete. That is one dispatch of the instrument against a helper mode this branch does not yet have.

No causal claim is made about load from these numbers. `Closes #2249` remains absent from the body.

Rul1an:
Work on this PR is stopped; #2249 is now `horizon:post-mvp`. Leaving it open and unmerged.

For whoever returns to it: **the merit that stands on its own is removing the environment-controlled write path CodeQL flagged**, replaced by a strictly parsed `READY pid=<u32>` record in bounded output. It should never carry `Closes #2249`. The control recorded on that issue establishes the mechanism — descendants inherit the run's write handles — but the measured rates (2/20 against 1/20 under load) separate nothing, so this PR has never been shown to change the failure rate.

Rul1an:
Closed as a merge vehicle by the CI-1 architecture decision recorded in #2302. The useful diagnostics and fixture evidence will be reimplemented as separate RED-first slices from current `main`; this stale combined branch will not be rebased or merged.

PR #2253, opened 2026-08-10
ci(proof): measure a Windows flake rate instead of asserting one

Rul1an:
Bound to head **`919f1b1f04d0`** (branch `cursor/windows-flake-rate-proof`, base `main` at `bc07ffb0e`).

## Why

#2249 and #2246 both turn on a question about a **rate**, not an outcome, and an author without a Windows host cannot take that measurement locally. This adds the instrument that can, modelled on the repository's delegated-proof pattern. Contract: `docs/reference/ci/windows-flake-rate-proof.md`.

`.github/workflows/windows-flake-rate-proof.yml` runs one `cargo test` selection N times on `windows-latest` and uploads a JSON artifact recording every iteration. The manifest binds the numbers to head SHA, workflow SHA, repository, ref, run and attempt, and digests the subject and the instrument.

## The six blockers from the review of `33a374f74`

Every one of them was a way the instrument could report a clean measurement of nothing, so each is now a way it fails instead.

| # | Blocker | Change |
|---|---|---|
| 1 | No per-iteration ceiling; one iteration was observed at 462s and could have consumed the job and taken the artifact with it | `--iteration-ceiling-s` (default 300, an input), a bounded `taskkill /T /F` process-tree kill on breach, and the breach recorded as data — a failed run, not a lost one |
| 2 | Reuse key incomplete: one test file's blob OID could carry a rate across changed test code or dependencies | Reuse is pinned to an **identical `head_sha`** plus instrument digests, toolchain and image version. The subject closure is not something this instrument can honestly enumerate, so it does not pretend to |
| 3 | `max_starts=20` silently stopped restarting the load while `iterations` is unbounded | The cap is gone; the supervisor restarts as often as needed and records `load.starts` |
| 4 | Load was checked only between iterations, so it could vanish mid-test while the iteration stayed labelled loaded | A supervisor thread samples liveness every 250ms **during** each iteration, samples before restarting so a gap is visible rather than papered over, and records `load.alive_fraction` per iteration. Coverage below 0.95 — or no sample at all — is a gap |
| 5 | "Can never attach a check to a PR" was inaccurate | Stated as a repository convention, in the workflow and the contract: no PR is opened from `proof/windows-flake-rate/**`, and if one were, its push checks would appear on it |
| 6 | No tests for the 305-line parser | `scripts/ci/windows_flake_rate_proof_selftest.py` pins pass, failure, timeout, executed-nothing, the rate arithmetic, load absence, and the fail-closed verdict each produces. A pre-commit hook runs it, and so does the workflow on the host about to measure |

Exit codes now carry the stance: `2` the subject would not build, `40` an iteration executed no test, **`41` a load arm lost its load**. The last one is the delegated lane's exit-40 rule applied where it matters most here — if load can vanish without the run failing, a clean load arm reads as "no defect" when it means "no load".

The self-test was verified to fail, not just to pass: inverting the executed-nothing rule breaks four assertions, and neutralising the load-gap rule breaks the one that exists for it.

## Correlation, not causation

`0/20` versus `2/20` supports the hypothesis that load affects #2249. It does not establish it, and that is now said in the artifact's `non_claims`, in the contract, and in the step summary: the samples are small and their intervals overlap, the arms ran on two separate hosted machines, and they were neither interleaved nor randomised. Claiming causation needs both arms on one runner, interleaved in randomised order, a pre-registered iteration count and effect size, and a further manipulation of the suspected mechanism that moves the rate as predicted.

Per the review, no rate is offered here as transferable evidence. Fresh measurements with this hardened version are running; they will be reported with their exact heads and read as observed rates only.

## Boundaries

Instrument, not a gate: not in the `CI` rollup, not in any required context, discharges no check. Claim ceiling stated in the artifact — `hosted_windows_flake_rate_measurement_only_not_a_verdict` — with no privileged-host or attestation language borrowed from the delegated lane, because this runs on a GitHub-hosted image.

## Verification

At `919f1b1f0`: self-test passes; `cargo fmt --all -- --check`, `actionlint`, `typos`, yaml checks and pre-push clippy pass. The workflow's own dispatch path is exercised by the two runs in flight.

## Summary by CodeRabbit

* **New Features**
  * Added an on-demand Windows test flake-rate measurement workflow with configurable iterations, timeouts, and load conditions.
  * Produces detailed JSON results, job summaries, provenance metadata, and downloadable artifacts.
  * Detects incomplete or unmeasurable runs and reports test failures without treating measurements as pass/fail gates.

* **Documentation**
  * Added guidance covering measurement claims, reuse rules, diagnostics, artifacts, failure semantics, and reporting requirements.

* **Tests**
  * Added behavioral self-tests and a pre-commit check to validate classifications, timeouts, load coverage, and failure-rate calculations.

### DIFF
--- a/.github/workflows/windows-flake-rate-proof.yml
+++ b/.github/workflows/windows-flake-rate-proof.yml
@@ -42,6 +42,11 @@ on:
         required: true
         default: "20"
         type: string
+      iteration_ceiling_s:
+        description: "Per-iteration ceiling in seconds; a breach is recorded as data"
+        required: true
+        default: "300"
+        type: string
       load:
         description: "Concurrent load arm(s) to measure"
         required: true
@@ -54,8 +59,12 @@ on:
   # Bootstrap lane. `workflow_dispatch` resolves the workflow from the default
   # branch, so an unmerged instrument cannot be dispatched — and measuring an
   # unmerged fix is the whole purpose. Pushing to this reserved namespace runs
-  # the definition from that branch. The namespace carries no pull request, so
-  # this can never attach a check to one, and it never matches `main`.
+  # the definition from that branch, and it never matches `main`.
+  #
+  # By convention no pull request is opened from this namespace, which is why a
+  # push here does not put a check on one. That is a convention and not a
+  # guarantee: open a PR from such a branch and its push-triggered checks would
+  # appear on it like any other.
   push:
     branches:
       - "proof/windows-flake-rate/**"
@@ -84,6 +93,14 @@ jobs:
           ref: ${{ inputs.ref || github.sha }}
           persist-credentials: false
 
+      # Run the instrument's own behavioural self-test on the host that will
+      # take the measurement, before it takes it. A broken classifier reports a
+      # confident rate for something else.
+      - name: Self-test the instrument
+        shell: bash
+        working-directory: scripts/ci
+        run: python3 windows_flake_rate_proof_selftest.py
+
       - name: Measure
         shell: bash
         env:
@@ -93,6 +110,7 @@ jobs:
           PROOF_TEST_TARGET: ${{ inputs.test_target || 'agent_golden_path_contract' }}
           PROOF_FILTER: ${{ inputs.filter || 'bounded_process' }}
           PROOF_ITERATIONS: ${{ inputs.iterations || '20' }}
+          PROOF_ITERATION_CEILING_S: ${{ inputs.iteration_ceiling_s || '300' }}
           PROOF_REF_INPUT: ${{ inputs.ref || github.ref }}
         run: |
           set -euo pipefail
@@ -104,6 +122,7 @@ jobs:
             --test-target "$PROOF_TEST_TARGET" \
             --filter "$PROOF_FILTER" \
             --iterations "$PROOF_ITERATIONS" \
+            --iteration-ceiling-s "$PROOF_ITERATION_CEILING_S" \
             --load "${{ matrix.load }}" \
             --out "proof-out/windows-flake-rate-${{ matrix.load }}.json"
 
@@ -129,9 +148,14 @@ jobs:
           print(f"- runner image: {runner['image_os']} {runner['image_version']}")
           print(f"- toolchain: {proof['toolchain']['rustc']}")
           print(f"- failed runs: {summary['failed_runs']} / {summary['measured']} measured"
-                f" (rate {summary['failure_rate']})")
+                f" (observed rate {summary['observed_failure_rate']}, not a causal claim)")
+          print(f"- reuse rule: {manifest['reuse_rule']}")
+          if summary["timed_out"]:
+              print(f"- iterations that breached the ceiling: {summary['timed_out']}")
           if summary["could_not_measure"]:
               print(f"- iterations that measured nothing: {summary['could_not_measure']}")
+          if summary["load_gaps"]:
+              print(f"- iterations that lost their load: {summary['load_gaps']}")
           for name, count in sorted(summary["failures_by_test"].items()):
               print(f"  - {name}: {count}")
           EOF
--- a/.github/workflows/windows-flake-rate-proof.yml
+++ b/.github/workflows/windows-flake-rate-proof.yml
@@ -1,0 +1,146 @@
+# Windows flake-rate proof producer.
+#
+# Contract: docs/reference/ci/windows-flake-rate-proof.md. An instrument, not a
+# gate. It mirrors host-capability-proof.yml: dispatch-only, produces a measurement
+# on a host the developer cannot reach, and uploads an artifact that *is* the proof.
+#
+# It makes no pass/fail claim about any change. A red required check is never
+# discharged by a green proof here, and this workflow is deliberately absent from
+# the `CI` rollup and from every required context.
+#
+# Why a rate and not an outcome: one green run cannot separate "fixed" from "won
+# the race this time", so every iteration is recorded and the artifact reports the
+# rate behind them. The `load` arms make the load hypothesis directly testable —
+# the same SHA measured alone and under the concurrency a full `cargo test`
+# produces — rather than inferred from which branches happen to be red.
+name: windows-flake-rate-proof
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Ref or SHA whose tree is measured"
+        required: true
+        type: string
+      package:
+        description: "cargo package (-p)"
+        required: true
+        default: assay-mcp-server
+        type: string
+      test_target:
+        description: "Integration test target (--test)"
+        required: true
+        default: agent_golden_path_contract
+        type: string
+      filter:
+        description: "libtest name filter"
+        required: false
+        default: bounded_process
+        type: string
+      iterations:
+        description: "How many times to run the selection"
+        required: true
+        default: "20"
+        type: string
+      load:
+        description: "Concurrent load arm(s) to measure"
+        required: true
+        default: both
+        type: choice
+        options:
+          - none
+          - workspace-tests
+          - both
+  # Bootstrap lane. `workflow_dispatch` resolves the workflow from the default
+  # branch, so an unmerged instrument cannot be dispatched — and measuring an
+  # unmerged fix is the whole purpose. Pushing to this reserved namespace runs
+  # the definition from that branch. The namespace carries no pull request, so
+  # this can never attach a check to one, and it never matches `main`.
+  push:
+    branches:
+      - "proof/windows-flake-rate/**"
+
+permissions:
+  contents: read
+
+concurrency:
+  # Serialize arms so a load arm never becomes another arm's ambient load.
+  group: windows-flake-rate-proof
+  cancel-in-progress: false
+
+jobs:
+  arms:
+    name: "Measure (load=${{ matrix.load }})"
+    runs-on: windows-latest
+    timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      matrix:
+        load: ${{ (github.event_name == 'push' || inputs.load == 'both') && fromJSON('["none","workspace-tests"]') || fromJSON(format('["{0}"]', inputs.load)) }}
+    steps:
+      - name: Checkout the tree under measurement
+        uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
+        with:
+          ref: ${{ inputs.ref || github.sha }}
+          persist-credentials: false
+
+      - name: Measure
+        shell: bash
+        env:
+          # A push-triggered bootstrap run has no inputs, so the defaults live
+          # here once and both triggers read them from this one place.
+          PROOF_PACKAGE: ${{ inputs.package || 'assay-mcp-server' }}
+          PROOF_TEST_TARGET: ${{ inputs.test_target || 'agent_golden_path_contract' }}
+          PROOF_FILTER: ${{ inputs.filter || 'bounded_process' }}
+          PROOF_ITERATIONS: ${{ inputs.iterations || '20' }}
+          PROOF_REF_INPUT: ${{ inputs.ref || github.ref }}
+        run: |
+          set -euo pipefail
+          mkdir -p proof-out
+          python3 scripts/ci/windows_flake_rate_proof.py \
+            --ref-input "$PROOF_REF_INPUT" \
+            --workflow-path ".github/workflows/windows-flake-rate-proof.yml" \
+            --package "$PROOF_PACKAGE" \
+            --test-target "$PROOF_TEST_TARGET" \
+            --filter "$PROOF_FILTER" \
+            --iterations "$PROOF_ITERATIONS" \
+            --load "${{ matrix.load }}" \
+            --out "proof-out/windows-flake-rate-${{ matrix.load }}.json"
+
+      - name: Summarize
+        if: always()
+        shell: bash
+        run: |
+          set -euo pipefail
+          file="proof-out/windows-flake-rate-${{ matrix.load }}.json"
+          if [ ! -f "$file" ]; then
+            echo "- no proof produced for load=${{ matrix.load }}" >> "$GITHUB_STEP_SUMMARY"
+            exit 0
+          fi
+          python3 - "$file" >> "$GITHUB_STEP_SUMMARY" <<'EOF'
+          import json, sys
+          proof = json.load(open(sys.argv[1], encoding="utf-8"))
+          manifest, runner, summary = proof["manifest"], proof["runner"], proof["summary"]
+          print(f"### load={proof['load']['mode']}")
+          print(f"- claim ceiling: `{manifest['claim_ceiling']}`")
+          print(f"- head_sha: `{manifest['head_sha']}`")
+          for path, digest in manifest["subject_digests"].items():
+              print(f"- `{path}` blob: `{digest.get('blob_oid','')}`")
+          print(f"- runner image: {runner['image_os']} {runner['image_version']}")
+          print(f"- toolchain: {proof['toolchain']['rustc']}")
+          print(f"- failed runs: {summary['failed_runs']} / {summary['measured']} measured"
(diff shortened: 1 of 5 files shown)


Rul1an:
## Correction required — useful instrument, but its current proof/reuse claims are too strong

Reviewed head `33a374f7418e82415878b16da24b1d2eaa42a791`.

The observed rates are useful evidence: `0/20` without the load arm and `2/20` with it support the hypothesis that load affects #2249. They do **not** establish “the split is load” or a causal rate difference. The samples are small, the arms run on separate hosted machines, and the uncertainty intervals overlap. Report the two observed rates and provenance; describe causality as a hypothesis unless the design and sample support that inference.

Blocking correctness issues in the instrument:

1. **No per-iteration ceiling.** `subprocess.run(measured)` has no timeout. One recorded iteration already took 462 seconds; a hung iteration can consume the 120-minute job and prevent the proof artifact from being produced. Add an explicit per-iteration ceiling and record a timed-out iteration as data, with bounded process-tree cleanup.
2. **The reuse key is incomplete.** The result depends on more than `tests/support/bounded_process.rs`: at least the integration-test target, package/workspace manifests, lockfile, and invoked test/build behavior contribute. The documented carry rule can currently reuse a rate across changed test code or dependencies. Either require the exact `head_sha` for reuse, or define and fail-closed over a defensible subject closure.
3. **The load arm silently weakens.** `BackgroundLoad(max_starts=20)` stops restarting after twenty starts while `iterations` is user-configurable above twenty. Later iterations can be labelled `workspace-tests` without that load. Remove the hidden fixed cap or make failure to maintain load explicit in each iteration.
4. **“Keeps load alive for the whole measurement” is not guaranteed.** The load process is checked only before an iteration; if it exits during the measured test, the rest of that iteration is unloaded. Record load liveness/overlap per iteration or supervise it continuously.
5. **The bootstrap non-claim is inaccurate.** A `proof/windows-flake-rate/**` branch can have a PR, and push-triggered checks on its head can appear on that PR. State this as a repository convention, not “can never attach a check”.
6. **No behavioral tests accompany the 305-line proof parser.** Pin, at minimum, pass/fail/no-test/timeout parsing, summary-rate calculation, and the fail-closed digest/reuse behavior with mutation-sensitive tests.

Keep this as an instrument, not a gate. Do not use its current content-addressed carry rule or causal wording until these points are fixed and a non-building reviewer has reviewed the final head.

Rul1an:
## Review — changes requested at `37ead31ce35d99860bd4c733897c6960f112abd4`

Independent read-only review, reviewing this head only. I am the non-building reviewer for it.

**Provenance for everything measured below.** Detached worktree at `/tmp/rev-2253` checked out at `37ead31ce35d99860bd4c733897c6960f112abd4`; host macOS 26.6, `arm64` (darwin); CPython 3.14.3. No `cargo` invocation and no Rust build, so no build profile applies — every number here is Python-level and none of it is a Windows observation. Mutation and probe work was done on **copies of the two scripts under `/tmp/rev2253-mut/`**, never on a repository file; no tracked file was modified, so no restore command is needed.

The instrument is well built and the six blockers are largely closed. Two findings are blocking, because both are the specific failure this PR exists to prevent: a broken instrument whose self-test stays green.

---

### Blocking

**B1 — The exit-code contract is not pinned, and breaking it is silent.** `check_fail_closed` asserts `instrument_verdict(...)[0] == proof.EXIT_LOAD_ABSENT` and `== proof.EXIT_COULD_NOT_MEASURE` — against the constants themselves, not against `41` and `40`. The literal `41` appears nowhere in the self-test and nowhere in the workflow; only `docs/reference/ci/windows-flake-rate-proof.md` states it. So the rule "load absence fails the run" is written in two places that cannot check each other.

Measured: setting `EXIT_LOAD_ABSENT = 0` leaves the self-test **green** (exit 0, zero failed assertions). `EXIT_COULD_NOT_MEASURE = 0` likewise. With either, `main()` returns 0, the step goes green, the artifact uploads — and because the message is printed only `if code != 0`, not even the stderr line appears. A load arm that lost its load would report success. That is exactly claim 3's stance, unpinned.

Suggested: assert the literals (`== 41`, `== 40`, `== 2`) and assert `!= 0` for each, in the self-test.

**B2 — The 250 ms sampling interval is unpinned, and a low sample count reads as full coverage.** Two halves:

- `check_load_supervision` always passes `interval_s=0.05` explicitly, so `LOAD_SAMPLE_INTERVAL_S` is never exercised. Measured: changing it to `600.0` leaves the self-test green.
- `iteration_coverage()` divides `alive` by `samples` with no expected-sample-count check. Measured against the unmodified instrument: `samples=1, alive=1` → `{'alive_fraction': 1.0, 'gap': False}`; same at 2 and 3 samples. The only zero-evidence guard is `samples == 0`.

Together: any path that starves the supervisor thread — a 600 s interval, or GIL/scheduler starvation on the saturated 4-core Windows box the loaded arm deliberately creates — yields one or two "alive" samples per iteration and labels a 300-second iteration fully loaded. The load could die at t=1 s in every iteration and `load_gaps` would be empty. Blocker #4's claim ("sampled every 250 ms *during* each iteration") is the one claim whose violation is invisible in the artifact.

Suggested: compare `samples` against `duration_s / interval_s` (say, require ≥ 50 % of expected samples) and treat a shortfall as a gap; and exercise the production default in the self-test.

---

### Major

**M1 — `instrument_verdict` passes a zero-iteration run.** Measured on the unmodified instrument: `instrument_verdict(summarize([], 0), 0)` → `(0, '')`, with `observed_failure_rate: None`. The only thing preventing a green measurement of nothing is the `args.iterations < 1` guard in `main()`, and that guard is untested — removing it leaves the self-test green. The docstring says the fail-closed rule is kept in `instrument_verdict` "so the fail-closed rule has one home"; the rule actually has two homes, and the one that is tested is not the one that enforces it. This is `AGENTS.md` one-rule-one-function on the load-bearing rule.

**M2 — Nothing bounds `iterations × ceiling` against the job timeout, and the artifact is all-or-nothing.** Defaults are 20 iterations × 300 s = 100 minutes for the loop alone, against `timeout-minutes: 120`, with **no cargo cache step in the workflow** — so both the subject build and the new load pre-build are cold. `iterations` and `iteration_ceiling_s` are free string inputs with no upper bound; `-f iterations=50` guarantees overrun. `Path(args.out).write_text(...)` runs once, after the loop, so a job timeout or a lost runner discards every completed iteration and `if-no-files-found: error` fails the upload. Blocker #1 bounded the hung iteration but not the job: "one hung iteration cannot consume the job and take the artifact with it" is true; "a hung *run* cannot" is not. Suggested: write the artifact incrementally after each iteration, and refuse at startup when `iterations × ceiling` plus build headroom exceeds the job budget.

**M3 — Nothing checks that the two arms are comparable.** The arms are two matrix jobs on two separate hosted machines, uploaded as two separate artifacts, with no step that combines or compares them. `image_version`, `processor_count` and the toolchain are recorded per arm but never diffed, so an image rollout between arms produces a difference in the numbers and no signal anywhere — a wrong number, not a refusal. The `non_claims` list names the machine and scheduling confound, which is the honest part; it does not name a *material* divergence that would make the comparison meaningless rather than merely noisy.

A second uncontrolled difference is not in `non_claims` at all: with no cache, the loaded arm enters its loop having just built the whole workspace, so its `target/` is many GB and its filesystem cache is warm; the baseline arm built one test target. The arms therefore differ in disk pressure and prior I/O as well as in concurrent load.

**M4 — `load.prebuilt` is a tautology that reads like a provenance marker.** `"prebuilt": load_command is not None` is identical to `mode == "workspace-tests"`; it carries no information about whether a pre-build happened or succeeded, and it would still report `true` if the pre-build block were deleted. Claim 4's "rates measured before that change are marked as such" is satisfied only indirectly, through `instrument_digests`. A field that appears to mark the fix but cannot is worse than no field. Suggested: record the pre-build's duration and exit status, or drop the key.

**M5 — The load selection is a second copy of `ci.yml`'s cross-platform selection.** `scripts/ci/windows_flake_rate_proof.py:360-368` restates `.github/workflows/ci.yml:692` (`--workspace --exclude assay-ebpf --exclude assay-it --exclude assay-monitor --exclude assay-cli`) literal for literal. I verified they currently agree. The comment above it asserts the equality — "The Windows leg of CI runs this selection" — which is precisely the `AGENTS.md` defect of two literals that must agree and are free to drift. When `ci.yml` changes, the load arm silently stops reproducing CI concurrency while still claiming to, and the instrument's whole validity rests on that claim. Suggested: derive the excludes from one place, or add a parity test asserting the two lists match.

---

### Minor

**m1 — The PR body is bound to a stale head.** It opens "Bound to head **`919f1b1f04d0`**" and its Verification section reads "At `919f1b1f0`: self-test passes; `cargo fmt`, `actionlint`, `typos`, yaml checks and pre-push clippy pass." `37ead31ce` changed the script's control flow (the load pre-build) and has no recorded verification in the body. Per `AGENTS.md` measurement provenance, the reported SHA is not the tree under review. It also means claim 4's "rates measured before that change are marked as such" could not have been written for a change that did not yet exist.

**m2 — "Proof" is the delegated lane's word, and it is the noun everywhere here.** The workflow name, `SCHEMA = "assay.windows-flake-rate-proof/v2"`, the artifact name `windows-flake-rate-proof-<run>-<arm>` and the document title all say *proof*, while the document's first line says "It measures; it never decides" and its Reuse section says nothing enforces the rule. A reader who receives `windows-flake-rate-proof-…-workspace-tests.json` has been handed a proof of a flake rate by name. Every string that states a rate is careful — "observed rate …, not a causal claim" in the step summary, the `non_claims` list, the docs' Correlation section — so this is the one place the vocabulary outruns the claim. Suggested: `windows-flake-rate-measurement`.

**m3 — `concurrency` does not serialize the arms.** The comment at `.github/workflows/windows-flake-rate-proof.yml:76` reads "Serialize arms so a load arm never becomes another arm's ambient load", but workflow-level `concurrency` scopes a *run*: the two matrix entries are jobs inside one run and start in parallel. The protection the comment names comes from the arms landing on separate runners — which is also what makes them non-comparable (M3). The group is still useful against overlapping dispatches; the sentence is wrong about arms.

**m4 — The `none` arm has no quiet check.** For `command is None`, `iteration_coverage()` returns `gap: False` unconditionally. The loaded arm fails closed when its load is missing; the baseline arm cannot fail closed when load is *present*, and nothing in the artifact records processes left behind by a ceiling breach. I did not establish a leak: the subject wraps children in a `process_wrap` `JobObject`, whose kill-on-close semantics probably bound orphans when `taskkill /F` reaches the test binary. The finding is that the artifact could not tell you either way, on the arm where contamination would understate the effect.

**m5 — Two more restatements of one rule.** (a) The self-test's `iteration()` fixture hand-writes the load record — including `"gap": False` — so `summarize`'s reaction to a gap is tested against a hand-written flag, never against `iteration_coverage()`'s output; neutralising the real gap rule breaks exactly one assertion, and it is in `check_load_supervision`, not in `check_fail_closed`. (b) The 300 s ceiling exists as four literals: the input default, the push fallback, the argparse default, and the docs. CodeRabbit reached (b) and the record-shape half of (a) independently at this head.

**m6 — `pkill -KILL -P` is not a tree kill.** `kill_tree`'s docstring promises the whole tree including what the test spawned; the POSIX branch kills direct children only. Windows-only in effect, so cosmetic — CodeRabbit flagged this too. `subprocess.run(["pkill", ...])` also raises `FileNotFoundError` uncaught where `pkill` is absent.

**m7 — `samples == 0 or` in the gap expression is redundant.** `fraction` is already set to `0.0` when `samples` is zero, so `fraction < MIN_LOAD_ALIVE_FRACTION` already returns `True`; measured, removing the clause changes nothing and the self-test stays green. Same redundancy the PR reports having removed elsewhere.

**m8 — The self-test cannot always report its own findings.** With a genuinely broken `PANIC` regex, it exits 1 (fail-closed, good) but via an uncaught `IndexError` at line 82 rather than through `check()`, so the operator sees a traceback instead of `FAIL fail captures one panic`. Guard the index.

**m9 — No coverage for exit 2.** `EXIT_INSTRUMENT_BROKEN` appears nowhere in the self-test; the argument guard and both build-failure paths are untested (see M1). The two `--no-run` builds also have no timeout of their own, so the phase that caused the observed 462 s breach is the one phase the ceiling does not cover.

**m10 — The pre-commit hook's `files:` omits the workflow.** `^scripts/ci/windows_flake_rate_proof(_selftest)?\.py$` does not fire when `.github/workflows/windows-flake-rate-proof.yml` changes, although that file is in `INSTRUMENT_PATHS` and holds two of the four ceiling literals.

**m11 — Two expressions for one question.** Checkout uses `${{ inputs.ref || github.sha }}`; the recorded `ref_input` uses `${{ inputs.ref || github.ref }}`. Harmless today because `manifest.head_sha` comes from `git rev-parse HEAD` and is authoritative.

---

### Claims: reproduced versus taken on trust

**Reproduced.** Claim 1's ceiling-as-failed-run: a timed-out record is never `could_not_measure`, always enters `measured`, and always counts in `failed` — pinned by the self-test, and my mutation removing `record["timed_out"]` from `failed` goes red. Restart cap removed and `starts` recorded: confirmed, and a load that keeps dying is caught. Reuse pinned to `head_sha`: confirmed as text, and the docs are candid that nothing enforces it. Bootstrap stated as convention: confirmed in both the workflow and the contract.

Claim 2, the builder's own two mutations: reproduced, both red. Inverting the executed-nothing rule breaks 12 assertions in my run, not four — the count depends on the mutation's shape, so it is not a property of the self-test worth quoting. I ran 18 further mutations; 8 survived, and B1, B2, M1 and m7 are what the survivors mean. The "digit-length check proved redundant" half of claim 2 I could **not** verify: no such check exists anywhere in this branch's history (`33a374f74`, `919f1b1f0`, `37ead31ce`), and the PR body does not mention it. It appears to belong to different work.

Claim 3: exits 2, 40 and 41 are each reachable in code (2 from both `--no-run` failures and the argument guard). But "no path returns success when the measurement did not happen" is **false as it stands** — B1 and M1 are both such paths.

Claim 4: the fix is a pre-build and not a raised ceiling — confirmed, the default is still 300 s and `--iteration-ceiling-s` is unchanged. Marking pre-fix rates: only indirectly, see M4 and m1.

**Taken on trust.** Everything Windows: `taskkill /T /F` behaviour, `python3` on the hosted image, whether the two selections' feature resolution produces separate cached artifacts (I believe it does, which is why the pre-build is the right fix rather than a source of rebuild thrash), and the real magnitude of cargo build-directory lock waiting inside `duration_s` when the restarting load arm re-checks freshness. I have no Windows host, which is the reason this instrument exists.

### Final state of the checks at this head

56 pass, 10 skipping, 0 pending, 0 failing; `mergeStateStatus: BEHIND`. `Wave 0` gates, `Clippy`, `Build + Test` on all three OSes, `Criterion benches`, `Dependency Security` and `Lint (pre-commit)` were pending when this review began and have all since passed. Two review records exist for this head-lineage that are not reviews: Copilot returned `COMMENTED` at `33a374f74` with "unable to review … quota limit", and `github-advanced-security` returned an empty body. Correcting one premise I was handed: CodeRabbit is **not** rate-limited here — it posted a real review at `37ead31ce` with 4 nitpicks, three of which overlap m5, m6 and part of m2's neighbourhood. It remains an optional reviewer under `AGENTS.md` and is not part of the quorum.

### Verdict

**Changes requested. Do not merge at `37ead31ce`.** B1 and B2 each let the instrument report a confident clean number while its self-test stays green, which is the one defect class this PR was built to exclude; M1 and M2 are the fail-closed rule having a second, untested home and the run being able to lose its whole measurement. None of this is a redesign — B1 and B2 are assertions and one arithmetic guard. Nothing in this review argues against the instrument's design or its correlation-not-causation discipline, which is the most careful part of it.

Rul1an:
Closed as an experimental measurement vehicle by the CI-1 architecture decision in #2302. It has unresolved exact-head review blockers and is not required for diagnostics-only CI-1A. Existing run provenance remains recorded in #2249/#2246; no rate or causal claim is withdrawn.

PR #2255, opened 2026-08-10
docs(runner): state the host-capability proof gate as implemented, not planned

Rul1an:
## What

`docs/reference/runner/host-capability-proof.md` opened with:

> Status: accepted design, not yet implemented. This document is the contract; the checker and workflows land in a follow-up PR and must match what is written here.

That is false, and it is the kind of false that costs review attention: it tells a reader the gate is aspirational, so evidence the gate actually requires looks optional. The checker, both workflows, and the required branch-protection context are all live.

Documentation only. No code, no workflow, no behaviour change.

## Corrections, each checked against the thing it describes

Measured on head `0296b9c63a765ea0f31d78abe1aee980256f55c8`.

| Statement now in the doc | Verified against |
|---|---|
| Status implemented and active; names checker + both workflows | `scripts/ci/assay_host_capability_check.py`, `.github/workflows/host-capability-check.yml`, `.github/workflows/host-capability-proof.yml` all present |
| `host-capability-check` is required on `main` alongside `CI` and `lane-check/proof` | branch protection reports contexts `CI`, `host-capability-check`, `lane-check/proof` |
| `runs-on: [self-hosted, linux, assay-bpf-runner]` | `host-capability-proof.yml` L36 — the doc previously omitted `linux` |
| `concurrency: group assay-bpf-runner-host`, `cancel-in-progress: false`, `queue: max` | `host-capability-proof.yml` L25-31 |
| group is shared with `Runner Spike Delegated`; host is a singleton | `runner-spike-delegated.yml` L29 uses the same group |
| a run is proof only for the exact dispatched head, so never dispatch on `main` | checker returns `head_sha_mismatch` (L284) and `meta_sha_mismatch` (L343) |
| check workflow triggers are `pull_request` + `workflow_dispatch` with `pr_number` / `expected_head_sha` | `host-capability-check.yml` L11-22 — the doc previously said `on: pull_request` only |
| there is no path filter, because a required check that does not run blocks as pending | `host-capability-check.yml` has no `paths:` under `pull_request` |

The "never dispatch on `main`" line is not a style note. A proof run was dispatched on `main` today and could not have counted for any PR regardless of whether it finished.

## Verification

- `git diff --check` clean
- `python3 scripts/ci/assay_host_capability_check.py --self-test` → `self-test: all cases passed`
- pre-commit hooks passed on commit (fmt, typos, lane-check self-test, gating-map drift)
- This file is on the gate's own exemption list (`docs/**`, and the doc names itself as `not required`), so this PR does not itself need host proof

## Non-claims

- Does not claim the gate is correctly *configured*, only that it exists and is required.
- Does not claim the gate is currently *satisfiable*. It is not, at present, for PRs that trigger it; that is a separate finding filed against the timeout, not fixed here.
- Does not change any trigger path, field list, or failure reason.
- Reviewer note: I am this PR's builder, so my own reading is not the quorum review.

### DIFF
--- a/docs/reference/runner/host-capability-proof.md
+++ b/docs/reference/runner/host-capability-proof.md
@@ -1,7 +1,15 @@
-# Host-Capability Proof Gate (RFC, v0)
+# Host-Capability Proof Gate (v0)
 
-Status: accepted design, not yet implemented. This document is the contract; the checker and
-workflows land in a follow-up PR and must match what is written here.
+Status: implemented and active. This document remains the contract, and the implementation must
+match what is written here:
+
+- checker: `scripts/ci/assay_host_capability_check.py`
+- gate workflow: `.github/workflows/host-capability-check.yml`
+- proof producer: `.github/workflows/host-capability-proof.yml`
+
+`host-capability-check` is one of the contexts branch protection on `main` requires, alongside
+`CI` and `lane-check/proof`. A change that trips the trigger paths below cannot merge without a
+host proof bound to the exact PR head.
 
 ## Problem
 
@@ -53,11 +61,19 @@ suspicion as a change to the lane-check classifier.
 
 ```yaml
 on: workflow_dispatch
-runs-on: [self-hosted, assay-bpf-runner]
+runs-on: [self-hosted, linux, assay-bpf-runner]
 permissions:
   contents: read
   actions: read
+concurrency:
+  group: assay-bpf-runner-host
+  cancel-in-progress: false
+  queue: max
 ```
+
+The concurrency group is shared with `Runner Spike Delegated`: the host is a singleton, so
+dispatches queue rather than run in parallel. A run is usable as proof only for a PR whose head is
+exactly the dispatched SHA, so dispatch against the PR head, never against `main`.
 
 The run builds `assay-cli` from the dispatched ref and uploads an artifact containing:
 
@@ -77,12 +93,18 @@ operated infrastructure, not a safe arbitrary-fork runner.
 ## Check workflow (`host-capability-check.yml`)
 
 ```yaml
-on: pull_request
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs: [pr_number, expected_head_sha]
 permissions:
   contents: read
   actions: read
   pull-requests: read
 ```
+
+There is no path filter. A required check that does not run blocks merges as pending, so the
+checker runs on every PR and passes immediately when no trigger path changed.
 
 The job runs the checker and passes or fails; it posts no comment (a fork PR's `GITHUB_TOKEN` is
 read-only and must stay that way). No `pull_request_target` in v0: a write-token checker is exactly


Rul1an:
## Non-building review record — head `0296b9c63a765ea0f31d78abe1aee980256f55c8`

**Verdict: APPROVE.** Reviewer: Cursor (non-building; authored neither this change nor the document it corrects). Measured on head `0296b9c63a765ea0f31d78abe1aee980256f55c8` (fetched via `refs/pull/2255/head`, `git rev-parse` confirms), worktree `/Users/roelschuurkes/assay`, checker run under `python3` from that tree. Diff is one file, `docs/reference/runner/host-capability-proof.md`, +27/−5.

Documentation change, so the only defect class that matters is a statement untrue of the thing it describes. Each claim checked against its source at that head, not against the PR body.

| Claim in the new text | Verified against | Result |
|---|---|---|
| Checker + both workflows exist at the named paths | `git ls-tree -r` at head lists `scripts/ci/assay_host_capability_check.py`, `.github/workflows/host-capability-check.yml`, `.github/workflows/host-capability-proof.yml` | true |
| `host-capability-check` required on `main` alongside `CI`, `lane-check/proof` | `gh api repos/Rul1an/assay/branches/main/protection` → `["CI","host-capability-check","lane-check/proof"]`; job `name: host-capability-check` (check yml L31) | true, and the context name matches character for character — no drift |
| `runs-on: [self-hosted, linux, assay-bpf-runner]` | `host-capability-proof.yml` L36, exact | true (`linux` was genuinely missing before) |
| `group: assay-bpf-runner-host`, `cancel-in-progress: false`, `queue: max` | `host-capability-proof.yml` L25–31 | true |
| Group shared with `Runner Spike Delegated`; host is a singleton, so dispatches queue | `runner-spike-delegated.yml` L25–31 uses the identical group with the same two settings; both jobs target the same `[self-hosted, linux, assay-bpf-runner]` label set | true — a real shared singleton, not a coincidence of naming |
| Exact-head rule is enforced, not merely intended | checker L285 returns `head_sha_mismatch` against the run's immutable `head_sha`; L346 returns `meta_sha_mismatch`; self-test case L451 pins the first | enforced |
| Check triggers are `pull_request` + `workflow_dispatch(pr_number, expected_head_sha)` | `host-capability-check.yml` L11–22, exact input names | true |
| No path filter, because a required check that does not run blocks as pending | no `paths:` under `pull_request` in that file | true |

`python3 scripts/ci/assay_host_capability_check.py --self-test` at this head → `self-test: all cases passed`.

### Overcorrection check — the failure mode a corrected document invites

Clean. The new header stops at existence and requiredness; it does not migrate into capability. The three places that could have drifted all still say the narrow thing: L33–34 ("presence and JSON type … never their values. A red host is also evidence"), L169–171 (`"abi_probe_status": "unsupported"` passes; the gate "proves the diagnostics were executed on the host for the PR head, not that the host is eligible"), L197 ("evidence gate, not quality gate"). That matches the implementation exactly — `check_landlock_fields` (L110–144) is presence plus `isinstance`, with no comparison operator anywhere in it, and the doc's nine-row field table is field-for-field and type-for-type identical to `REQUIRED_FIELDS` (7) and `OPTIONAL_FIELDS` (2). The sentence "cannot merge without a host proof bound to the exact PR head" is a statement about the gate, not about the host, and is true.

### The two missing facts

The first is **not** missing: L75–76 states a run is proof only for a PR whose head is exactly the dispatched SHA, and "never against `main`" is explicit. Good — that is the fact that cost a run today.

The second, #2257 (the producer exceeds `timeout-minutes: 30` from a cold workspace, so the gate is presently unproducible), **is** absent, and I went the way of *should be mentioned, but not blocking*. My reasoning, stated so it can be disagreed with: the document is a contract, and a contract legitimately describes the obligation rather than today's outage, so its absence is an omission and not a false statement — nothing in the file asserts satisfiability. But the failure this PR exists to prevent is a reader planning around a wrong picture of the gate, and a reader who plans a `diagnostics/**` change on this text will plan to dispatch a proof that currently cannot finish. That is the same planning cost in a different dimension. One line under "Proof workflow" pointing at #2257, removable when it closes, would close it. I am a non-building reviewer and will not edit the file to add it; recorded as a follow-up instead of a change request, since holding a true document hostage to an operational note it does not contradict would be the worse trade.

### Non-claims

- I did not verify the gate is correctly *configured* beyond the statements the file makes, and I did not attempt to produce a proof run.
- I did not re-derive the docs-only exemption; per the checks' own status, `host-capability-check` passed on its own classification.

### Merge preconditions (AGENTS.md)

All three required contexts green on this head: `CI` pass, `host-capability-check` pass (13s), `lane-check/proof` pass ("no delegated runner proof required"). No delegated runner proof targets a different SHA — none is required. Review quorum satisfied by this record. No actionable finding is unresolved; the single recommendation above is dispositioned as a follow-up. `Build + Test (windows-latest)` is `skipping` on this PR, not red — CI scope detection excluded it for a docs-only change — so this PR yields **no datum** for the #2249 flake rate, and I am deliberately not posting a non-observation there, since a skipped job would dilute the rate that issue needs. `Build CLI Artifact` is pending and is not a required context.

Merging with `gh pr merge --squash` (no `--admin`).

Issue #2257, opened 2026-08-10
ci: host-capability-proof cannot finish inside its 30m timeout on a cold workspace, so a required gate is currently unproducible


Rul1an:
Verified against `origin/main` at `ea161accd1db0921c0f938a455d2eafa7b3fddd2`.

## The consequence first

`host-capability-check` is a required status check on `main` (branch protection reports contexts `CI`, `host-capability-check`, `lane-check/proof`, `strict: true`). It demands a `host-capability-proof` run bound to the exact PR head. That proof currently cannot be produced. So **any PR touching `crates/assay-cli/src/diagnostics/**` cannot satisfy a required merge gate right now.**

## Evidence

`.github/workflows/host-capability-proof.yml` L37: `timeout-minutes: 30`.

Run [31412409798](https://github.com/Rul1an/assay/actions/runs/31412409798):

| Field | Value |
|---|---|
| event | `workflow_dispatch` |
| head SHA | `bc07ffb0e8f47313b4ce7353c63d3e22c6299514` (branch `main`) |
| started / completed | 17:06:57Z → 17:37:51Z = **30m54s** |
| conclusion | `cancelled` |
| artifacts | `total_count: 0` |

Step-level detail confirms where it died:

```
3  Checkout repository              success
4  Build assay CLI                  cancelled   <-- timeout landed inside cargo build
5  Run doctor and collect provenance skipped
6  Upload proof artifact            skipped
```

The upload step declares `if-no-files-found: error`, but it never ran, so the run produced no artifact at all rather than a failed upload.

## Cause: the workspace went cold across a major version jump

The workflow has **no caching whatsoever** — no `Swatinem/rust-cache`, no `sccache`, no `actions/cache`. The only thing that ever made this fast was `target/` persisting in the self-hosted runner's workspace between runs. The complete run history of this workflow:

| Run | Head | Version at that SHA | Duration | Conclusion |
|---|---|---|---|---|
| 27238068622 | `a0ba3967` | 3.20.0 | 3m36s | success |
| 27238396301 | `be61ac4c` | 3.20.0 | 2m34s | success |
| 31412409798 | `bc07ffb0` | 5.0.0 | 30m54s | cancelled |
| 31412444715 | `bc07ffb0` | 5.0.0 | 28s | cancelled |

[...]

PR #2262, opened 2026-08-10
fix(cli): publish typed evidence failures on stdout

### LINKED ISSUE #2164: Evidence show JSON failure writes no report to stdout
Parent: #2177 · Origin: #2154

## Driven observation

Measured on `18abbebf025e4a9243b1ec80f798db5ccad2afed` against the committed tampered corpus vector:

```text
assay evidence show <tampered.bundle.tar.gz> --format json
exit: 2
stdout: empty
stderr: bundle verification / content hash mismatch
```

The command promises JSON bundle inspection on stdout, but emits no machine report on the integrity-failure path.

## Acceptance

- `--format json` emits a parseable failure report on stdout for integrity/open failures.
- The report carries a registered `reason_code` and non-empty `next_step`.
- A binary-level test drives the committed tampered vector and reads stdout plus exit code only.

Rul1an:
## ADR Boundary Slice

- Canonical ledger: none; this is a standalone MVP fix
- Slice: #2164, machine-readable `evidence show` open/read and integrity failures
- Builder: Codex, `/Users/roelschuurkes/wt-2164`
- Reviewers: pending one independent non-building exact-head review
- Final head SHA: `74f5379657537a3a9026ead81d7227a93b27dc01`
- ADR invariants affected: ADR-042 evidence-first positioning; ADR-043 evidence verification boundary

## Behavior

`assay evidence show <bundle> --format json` now emits one `assay.run_summary.v1` document on stdout for two typed failure families:

- exit `2`, `E_EVIDENCE_INTEGRITY` when verification establishes a recorded-value mismatch;
- exit `2`, `E_EVIDENCE_UNREADABLE` when the bundle cannot be opened or read to completion.

Classification is deliberately narrow and typed. Only `IntegrityManifestHash`, `IntegrityEventHash`, `IntegrityFileSizeMismatch`, and `IntegrityRunRootMismatch` map to integrity. Direct filesystem I/O and verifier codes `IntegrityIo`, `IntegrityGzip`, and `IntegrityTar` map to unreadable. One production mapping serves both classification sites and is pinned to the normative boundary and registry.

Successful JSON inspection now carries `verify_mode: "enabled"` or `"disabled"`; a tampered bundle opened with `--no-verify` therefore cannot look verified. The generated golden-path contract explicitly retains format-contract failures as exit `2` with empty stdout and `gap_issue: 2219` instead of claiming that deferred path is solved.

Both typed failure envelopes carry non-empty shell-free recovery. Caller-controlled bundle paths are encoded as JSON argv, including paths containing spaces. The generated golden-path skill, plugin copy, JSON contract, guide, normative output registry, and reason-code implementation describe the same behavior.

Closes #2164.

## Non-Claims

- [x] Unreadable establishes no content mismatch, format defect, tampering, intent, or cause
- [x] Integrity establishes a mismatch only, with no tampering, intent, provenance, or cause attribution
- [x] Format-contract failures remain outside both reasons; #2219 tracks that separate post-MVP gap

## Test Evidence

All final-head measurements below target `74f5379657537a3a9026ead81d7227a93b27dc01` in `/Users/roelschuurkes/wt-2164` with Rust/Cargo 1.96.0.

### RED

- Typed integrity and unreadable tests initially failed because non-zero JSON paths wrote zero stdout.
- Removing `IntegrityFileSizeMismatch` from the production set made the normative parity test fail.
- Before the review-blocker fix, verified and `--no-verify` JSON successes both lacked `verify_mode`.
- Before restoring the residual disclosure, the golden-path test failed because step 7 no longer named #2219; the public-vocabulary hook independently rejected the unstructured issue reference.

### GREEN

```text
cargo test -p assay-cli
# 547 unit tests passed; all integration suites passed

cargo test -p assay-cli --test agent_golden_path_contract
# 26 passed, including valid, tampered, unreadable, format-contract, and --no-verify paths

python3 scripts/ci/test-agent-golden-path-skill.py
# portable, byte-identical, and contract-complete

cargo fmt --all -- --check
cargo clippy -p assay-core -p assay-cli --all-targets -- -D warnings
git diff --check
# all passed
```

CLI probes:

```json
{"verify_mode":"enabled","event_count":2}
{"verify_mode":"disabled","event_count":2}
```

The pre-commit and pre-push hook suites also passed, including generated-doc mutation checks, workspace clippy, and the Linux cross-target gate.

## Review Quorum

- [ ] One non-building agent review on `74f5379657537a3a9026ead81d7227a93b27dc01`
- [x] No delegated host proof is triggered: this PR does not touch `crates/assay-cli/src/diagnostics/**`

Optional automated reviews can add evidence, but do not count toward or block quorum.

## Review Round on `17d0d714d`

The independent reviewer found three related blockers. All are fixed on the final head:

- `verify_mode` now uses the registered `enabled|disabled` vocabulary through one shared `assay_core::report::summary::verification_mode` function;
- the top-level failure funnel receives the actual verification setting, so `--no-verify` failures report `provenance.verify_mode: "disabled"` instead of the false `"enabled"`;
- the generated golden-path contract adds a driven `verification-disabled` outcome and publishes the field and vocabulary in every generated skill, guide, and plugin copy.

RED was observed independently for the old `"skipped"` value and the missing generated-contract disclosure. GREEN includes all 26 exact contract outcomes, the full CLI suite, affected core tests, deny-warnings clippy, generator drift/mutation checks, portable skill simulation, workspace pre-push clippy, and Linux cross-target compilation.


Rul1an:
## Independent exact-head review — `1452b047f229754801d3387df7974e44f4e2551d`

Verdict: **changes requested.** The typed boundary itself is well built and the non-claims are honest, but two items are actionable: `Closes #2164` claims an acceptance clause this change deliberately does not meet, and the emitter's admit set is a third unpinned copy of a rule the repository otherwise pins carefully. Everything else below is minor; the owner decides what is actionable.

I am the whole quorum for this head. Copilot returned `COMMENTED` with "unable to review … quota limit" and CodeRabbit reports `pass / Review rate limited`; per `AGENTS.md` neither is a review, and neither substitutes for one.

**Measurement provenance.** All runtime measurements are of `1452b047f229754801d3387df7974e44f4e2551d`, built and run in a detached worktree at `/tmp/rev-2262` (not the shared checkout), `cargo build -p assay-cli`, **debug** profile, binary `target/debug/assay` reporting `assay_version 5.0.0`. Platform macOS 26.6, arm64 (Apple Silicon), `rustc 1.96.0 (ac68faa20 2026-05-25)` / `cargo 1.96.0`. Bundles used are the committed corpus vectors `conformance/privileged-mcp-action-v0/vectors/ok-001-deny-bound-observation.bundle.tar.gz` and `bad-101-tampered-bundle.bundle.tar.gz`. The tampered vector's dominant verifier code is `IntegrityEventHash` (`assay evidence verify` on it prints `Content hash mismatch at seq 0 (IntegrityEventHash)`).

**Mutation disclosure.** One file was mutated, only inside the disposable `/tmp/rev-2262` worktree, for the negative control in finding 4: `crates/assay-cli/src/cli_failure.rs`. It was restored by name with `cp /tmp/rev2262-scratch/cli_failure.rs.orig /tmp/rev-2262/crates/assay-cli/src/cli_failure.rs`, verified by `git status --porcelain -- crates/assay-cli/src/cli_failure.rs` returning empty. No bare `git checkout --` and no whole-tree restore was run anywhere. The shared checkout at `/Users/roelschuurkes/assay` was never written to; its `git status --porcelain` and `HEAD` (`cd8dc5d6`) are unchanged from my first command to my last. Unrelated observation, recorded because it is destructive if real and it was **already true before my first command**: the conversation-start snapshot of that checkout listed ` M docs/reference/runner/host-capability-proof.md`, and my first `git status --porcelain` showed only the untracked `docs/audit/REPORT-2026-08-10-deep-risk-verification.md`. That uncommitted modification disappeared before I touched anything, and `HEAD` did not move, so it was not committed. Someone should check whether another agent lost work.

---

## Findings

### 1. Moderate — every unclassified failure still produces zero stdout bytes with the *same* exit code as the classified one, and `Closes #2164` claims otherwise

`crates/assay-cli/src/cli/commands/evidence/mod.rs:301` and `:367-372`.

Measured, `assay evidence show <input> --format json`:

| input | exit | stdout bytes | stderr |
|---|---|---|---|
| valid vector | 0 | 4395 | empty |
| tampered vector | 2 | 631 | `[E_EVIDENCE_INTEGRITY] …` |
| `./does-not-exist.tar.gz` | 2 | **0** | `fatal: failed to open bundle …` |
| first 400 bytes of the valid vector | 2 | **0** | `IntegrityIo` |
| `not a gzip at all` | 2 | **0** | `IntegrityTar` (invalid gzip header) |
| zero-length file | 2 | **0** | `IntegrityTar` |
| a directory | 2 | **0** | `Is a directory (os error 21)` |

Exit `2` therefore means both "stdout holds one `assay.run_summary.v1`" and "stdout is empty", with nothing to discriminate them. A machine consumer must treat an empty read as normal, which is exactly the shape that lets a parse failure be mistaken for a clean absence.

Three consequences worth separating:

- Issue #2164's acceptance reads "`--format json` emits a parseable failure report on stdout for integrity/**open** failures". Open failures are deliberately excluded here, correctly per the boundary — but the PR body says `Closes #2164`, which will close an issue with an unmet clause. #2219 covers registering a code for `Contract*` defects; nothing tracks the empty-stdout half.
- The generated contract now records `gap_issue: null` for this step, so the marker that previously flagged the empty-stdout path is gone. By the boundary file's own measurement, 3 of 496 detections in `docs/experiments/evidence-mutation-cost-2026-06` carry a code this rule admits and 479 carry `ContractInvalidJson`, so the newly machine-readable path covers a small minority of measured detections and the majority is now untracked.
- `SPEC-PR-Gate-Outputs-v1` §7 is a MUST: "When exiting with exit_code ≠ 0, the process MUST print at least one line that is a concrete command or hint." The five rows above print a bare `fatal:` chain with no hint. That MUST predates this PR, but this PR is what pulls `assay evidence show` into §5.1's registry scope (its own changelog row names the command), so the command is now inside the spec's vocabulary while most of its non-zero exits are outside its §7 obligation.

Falsified by: any of those five inputs producing a parseable stdout document, or a distinct exit code, or a recorded disposition narrowing #2164 rather than closing it.

### 2. Minor — the success and failure documents are two schemas and only one self-identifies

`evidence/mod.rs:309-313`. Success is a bare `{"manifest": …, "events": …}` with no `schema` key; failure carries `"schema": "assay.run_summary.v1"`. A consumer cannot parse the union without knowing which path produced it, which is the property this PR set out to give it. The docs describe both shapes, so this is a shape complaint, not a disclosure one. Falsified by a `schema` key on the success document.

### 3. Moderate — `--no-verify --format json` returns exit 0 and a clean-looking document for a tampered bundle, with nothing recording that verification was skipped

`evidence/mod.rs:303-315`. Measured on the tampered vector: `exit=0`, 4365 stdout bytes, empty stderr, and the document contains no `verified`, `verify_mode`, or equivalent field. The table format prints `Verified: ⚠️ SKIPPED`; the JSON format prints nothing at all, so `verified` at `:303` is computed and used only by the branch this PR does not touch.

That was survivable while this command had no machine contract. This PR makes `assay evidence show … --format json` the advertised integrity surface in `.agents/skills/`, `.claude/skills/` and the packaged plugin copy, read by agents. An agent that adds `--no-verify`, or a wrapper that already sets it, gets exit 0 and a well-formed document from a bundle whose content does not verify. Nothing in the changed docs discloses this. Falsified by a `verified`/`verify_mode` field in the JSON success document, or a sentence in the skill saying so.

### 4. Moderate — the emitter's admit set is a third copy of the boundary rule and nothing pins it (negative control run)

`crates/assay-cli/src/cli_failure.rs:47-53`.

The repository pins this rule unusually well elsewhere: `evidence_integrity_boundary.md` is the single statement, `#[doc = include_str!(…)]` transports it to the variant, `the_evidence_integrity_boundary_reads_the_same_in_the_spec_and_the_code` pins the §5.1 row to the file by flowed text, and `the_doc_comment_transports_the_boundary_rather_than_restating_it` stops the doc comment from becoming a copy. This PR adds the first *executable* statement of the same rule and leaves it outside that mechanism.

Negative control, run at this head in the `/tmp` worktree: I removed `| ErrorCode::IntegrityFileSizeMismatch` from the `matches!` at `:51` **and** `ErrorCode::IntegrityFileSizeMismatch,` from the positive array in the same file's test — what one developer's single edit would look like. Result:

```
test cli_failure::tests::evidence_integrity_classification_matches_the_normative_code_boundary ... ok
test result: ok. 3 passed; 0 failed; ... 541 filtered out
cargo test -p assay-cli --test spec_reason_code_registry
test result: ok. 9 passed; 0 failed
```

Green, with the emitter now violating a normative MUST. The test named for the boundary asserts agreement with a hand-written copy of the list sitting eleven lines above it, so both sides of the "parity" move in the same commit. `the_boundary_names_error_codes_that_still_exist_in_assay_evidence` pins the boundary to `assay-evidence`, not to the emitter. The doc comment at `:39-40` says "Keep this list aligned with the normative boundary" — an instruction, which is what `AGENTS.md` says to replace with a function or a real pin.

Falsified by a test that extracts the four codes from `exit_codes/evidence_integrity_boundary.md` and compares that set to the emitter's admitted set, such that the mutation above fails.

### 5. Minor (latent) — `IntegrityEventHash` is also raised for a computation failure, so keying on the code alone can assert a mismatch nobody measured

`crates/assay-evidence/src/bundle/writer_next/verify.rs:445-451`: a `compute_content_hash` error is mapped to `ErrorClass::Integrity` / `ErrorCode::IntegrityEventHash`. That is a failure to compute, not a recorded value disagreeing. Reaching `cli_failure.rs:47-53` it becomes `E_EVIDENCE_INTEGRITY` with `next_step` "the content this bundle carries does not match what it records" — the precise claim the boundary file forbids for anything that is not a mismatch. Only the mismatch at `:453-459` is the fact the code names. I did not find a reachable input (the event is already deserialized, so canonicalisation rarely fails), so I record this as latent, not as a demonstrated defect. Falsified by showing `compute_content_hash` cannot fail for an event that reached that line; confirmed by an input where it does.

### 6. Minor — the machine document carries strictly less than the human channel, and the mismatch location is lost from both

`cli_failure.rs:59` formats `{error}`, which is anyhow's `Display` — the outermost context only. Measured, the classified `message` is `evidence bundle <path> failed content verification: Bundle verification failed`, while the same bundle through `assay evidence verify` yields `Content hash mismatch at seq 0 (IntegrityEventHash)`. So `seq 0` is dropped everywhere, and `verifier_code: "IntegrityEventHash"` survives only in the stderr diagnostic's context object — the stdout `assay.run_summary.v1` has no field for which of the four codes fired. The typed distinction the PR builds is therefore not observable by the machine consumer it is built for; it collapses to one string. Falsified by finding the verifier code or seq in the stdout document.

### 7. Minor — hostile path strings are escaped on stdout and emitted raw on stderr

Measured with a bundle copied to a path containing `ESC ] 0 ; PWNED BEL` and ANSI SGR, and to a path containing a newline.

Stdout is correct: `"message": "evidence bundle hostile/a\u001b]0;PWNED\u0007b\u001b[31mred\u001b[0m.tar.gz …"`, and the newline as `\n`. Stderr is not: the headline is emitted byte-for-byte, so the OSC sequence set my terminal's title, and the embedded newline split one diagnostic across two lines in a way a line-oriented log reader will mis-attribute. Path is `cli_failure.rs:36` → `pipeline_error.rs:94-103`, where both `format_terminal` and `format_plain` write unsanitised. The repository already owns `assay_evidence::sanitize` for exactly this class and `CLAUDE.md` states the invariant for rendered strings.

The value is caller-controlled argv, which lowers severity, and the sink predates this PR (`CliFailure::policy_parse` reaches it too). Recording it because this PR is the second caller and the first whose value names an evidence artifact, so the argument that "the operator typed it themselves" is weaker here. Falsified by sanitisation at the emit site.

Also checked, no finding: the ingest ceiling is applied before materialisation — `reader.rs:118-127` wraps the source in `LimitReader(max_bundle_bytes, SourceBytes)` *before* `read_to_end`, and `open_unverified` keeps the full limit set. Traversal and absolute-path escape are not reachable here: the argument names a local file to open, not a path resolved inside an archive. Non-UTF-8 paths I could not test — APFS rejected the filename (`Illegal byte sequence`), so `path.display()`'s lossy conversion is unexercised on this platform; on a filesystem that permits it, `:57` would silently substitute U+FFFD and the emitted `path` would not round-trip to the file.

### 8. Minor — two sites answer "is this `evidence show` invocation JSON"

`crates/assay-cli/src/cli/args/mod.rs:59` (`args.format == ShowFormat::Json`) and `crates/assay-cli/src/cli/commands/evidence/mod.rs:306` (the same comparison, fully qualified). The sibling policy arm avoids this with the accessor `args.is_json()`. Nothing pins the two, and disagreement is exactly the failure this feature is about: a summary emitted for a table run, or suppressed for a JSON run. Low drift risk today because both are equality on one field; it stops being low the moment a second way to ask for JSON exists. Falsified by one accessor used at both sites.

### 9. Trivial — inline fully-qualified path where the sibling arm uses an import

`args/mod.rs:58` writes `crate::cli::commands::evidence::EvidenceCmd::Show` inside the `matches!` while the arm above it uses the imported `PolicyCommand`. `EvidenceCmd` lives in `cli/commands/evidence/mod.rs:39`, so it is genuinely not in `args`' glob scope; an explicit `use` would still be the consistent form and matches the workspace no-inline-imports rule.

### 10. Trivial — `docs/reference/cli/evidence.md` documents no `evidence show` at all

Zero occurrences of `evidence show ` in that file, so this command's exit codes exist only in the golden-path guide, the generated JSON and the three `SKILL.md` copies. Neither location records that exit `2` has two different stdout behaviours (finding 1). Whatever is decided about finding 1, the reference page is where a consumer would look.

### 11. Trivial — `verify_mode: "enabled"` is asserted by construction

`cli_failure.rs:38` passes `true` unconditionally to `summary_from_outcome`, so the emitted `provenance.verify_mode` is a constant, not a measurement. It happens to be true today, because only the verifying branch can produce a classified failure. It becomes false the first time anything on a `--no-verify` path is classified.

---

## Non-claims

The ADR-042 stop list holds. No scalar trust score, no whole-action verdict, no identity/delegation/federation surface, no provider-outcome verification, no compliance, certification or partnership claim. The `next_step` is deliberately prose rather than a command, and the reasoning at `exit_codes.rs:244-246` for why re-verifying is not a remedy is right. `E_EVIDENCE_INTEGRITY` asserts non-verification and not tampering, intent or cause, in the code, the boundary file and the emitted strings alike; the injection-shaped context test at `exit_codes.rs:474` is a good pin. The PR body's non-claims checklist is accurate as written.

The inverse check is where the gaps are, and they are findings 1, 3 and 6: a real limitation this change introduces or leaves in place, which nothing user-facing discloses. The body's "no claim that an unreadable or malformed bundle has an integrity mismatch" discloses the *classification* boundary honestly; it does not disclose that those bundles still yield exit 2 with an empty stdout under the same flag, nor that `--no-verify` returns a clean document for a bundle that does not verify.

---

## `host-capability-check`: green **by classification**, not by proof — and correctly so for this PR

It reported success because the classifier decided this PR does not need a proof. The check's own output says so, in run [31419286045](https://github.com/Rul1an/assay/actions/runs/31419286045/job/93555862328), 6 seconds:

```
Checker self-test:              self-test: all cases passed
Validate host-capability proof: PASS: no host-capability proof required for this PR.
```

No proof artifact was fetched or validated; the job ran the checker's self-test and the classifier, nothing else. `.github/workflows/host-capability-check.yml` is deliberately built this way (no path filter, so a required check cannot hang as "pending"; the checker itself decides).

The classifier's input is correct here. `scripts/ci/assay_host_capability_check.py:47` sets `TRIGGER_PREFIX = "crates/assay-cli/src/diagnostics/"` and requires a proof for any non-Markdown change beneath it. This PR changes 13 files — 5 under `crates/assay-cli/src/` and `crates/assay-cli/tests/`, 8 under `docs/`, `scripts/docs/`, `.agents/`, `.claude/`, `packaging/` — and none is under `crates/assay-cli/src/diagnostics/`. So the classification is right, and the PR body's "No delegated host proof is triggered" is accurate.

This PR is **not** in the #2257 position. I verified #2257's substance and it holds: `host-capability-proof.yml` has `timeout-minutes: 30` and no caching of any kind, run 31412409798 was cancelled at 30m54s inside `Build assay CLI` with `total_count: 0` artifacts, and the last successful runs were 2026-06-09 at workspace version 3.20.0 before the 5.0.0 jump invalidated the warm `target/`. So a PR that genuinely needs the proof cannot currently obtain one — but this is not such a PR, and its green tick is not standing in for a proof it needed.

One nuance for whoever relies on this gate: in the `pull_request` trigger `EXPECTED_HEAD_SHA` is empty (confirmed in the job log), so the SHA-pinning branch at `:52-61` is skipped and the classification is made against whatever the API returns for PR #2262 at run time rather than against a pinned head. Here that is the head under review — the 13-file list I checked is the list at `1452b047` — so the conclusion binds correctly, by coincidence of timing rather than by construction. `lane-check/proof` is green on the same basis ("no delegated runner proof required").

## Things I could not review

- Non-UTF-8 path handling (finding 7): APFS refuses such filenames, so `path.display()`'s lossy branch is untested on darwin/arm64. It needs a Linux run.
- Reachability of finding 5: I did not construct an input where `compute_content_hash` fails, so that finding is latent rather than demonstrated.
- Everything measured here is a **debug** build on macOS/arm64. The repository's CI builds on `ubuntu-latest`; I did not re-measure the stdout envelope there, and nothing in the paths involved looks platform-dependent apart from the `Is a directory (os error 21)` text.

Rul1an:
## Independent exact-head review — `c9c949412ccc912f0523ca0900961ceb94b410c8`

Verdict: **changes requested**, on disclosure only. All four prior blockers are genuinely resolved, not reworded, and I verified each by deliberate mutation rather than by reading the code. Two blocking items remain and both are disclosure defects with cheap fixes: the generated contract now declares **no gap** on step 7 while a reachable exit-2 path still writes zero stdout, and `--no-verify --format json` still returns exit 0 with a clean-looking document for a bundle whose content does not verify, with nothing recording that verification was skipped.

I am the whole quorum for this head. Copilot returned `COMMENTED` with "unable to review this pull request because the user who requested the review has reached their quota limit", and CodeRabbit reports `pass` alongside "Review limit reached … we couldn't start this review". Per `AGENTS.md` neither is a review; both are unavailable infrastructure wearing the shape of a verdict. The `1452b047f` review is void: three commits landed after it.

**Measurement provenance.** Every runtime number below is of `c9c949412ccc912f0523ca0900961ceb94b410c8`, built and run in a detached worktree at `/tmp/rev2262-final/wt` (never the shared checkout), `cargo build -p assay-cli`, **debug** profile, binary `target/debug/assay` reporting `assay_version 5.0.0`. Platform macOS 26.6, arm64 (Apple Silicon), `rustc 1.96.0 (ac68faa20 2026-05-25)` / `cargo 1.96.0`. Bundles are the committed corpus vectors `conformance/privileged-mcp-action-v0/vectors/ok-001-deny-bound-observation.bundle.tar.gz` and `bad-101-tampered-bundle.bundle.tar.gz`; the tampered vector's dominant verifier code is `IntegrityEventHash`.

Independently reproduced at this head: `cargo fmt --all -- --check` clean, `cargo clippy -p assay-cli --all-targets -- -D warnings` clean, `cargo test -p assay-cli --test agent_golden_path_contract` 25 passed, `cargo test -p assay-cli --test spec_reason_code_registry` 9 passed, `cargo test -p assay-cli --bin assay cli_failure::` 4 passed, `python3 scripts/docs/generate-agent-golden-path.py --check` exit 0, and `cmp` byte-identical for both generated-artifact copy pairs. No CI check has failed at this head; 4 were still running when I finished.

**Mutation disclosure.** Three files were mutated, only inside the disposable `/tmp/rev2262-final/wt` worktree, for the negative controls below: `crates/assay-cli/src/cli_failure.rs`, `crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md`, `docs/architecture/SPEC-PR-Gate-Outputs-v1.md`. Each was restored by name from a pre-mutation copy:

```
cp /tmp/rev2262-final/cli_failure.rs.orig /tmp/rev2262-final/wt/crates/assay-cli/src/cli_failure.rs
cp /tmp/rev2262-final/boundary.md.orig   /tmp/rev2262-final/wt/crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md
cp /tmp/rev2262-final/SPEC.md.orig       /tmp/rev2262-final/wt/docs/architecture/SPEC-PR-Gate-Outputs-v1.md
```

verified by `git status --porcelain` returning empty in that worktree with `HEAD` still at `c9c94941`. No bare `git checkout --`, no `restore`, no `clean`, no whole-tree operation was run anywhere. The shared checkout at `/Users/roelschuurkes/assay` was never written to; `git status --porcelain` there returned exactly `?? docs/audit/REPORT-2026-08-10-deep-risk-verification.md` and `HEAD` `cd8dc5d6f` on my first command and on my last. Recorded because it is someone's lost work if real and it was **already true before my first command**: the conversation-start snapshot of that checkout listed ` M docs/reference/runner/host-capability-proof.md`, and my first `git status` showed no such modification, with `HEAD` unmoved so it was not committed. Same observation as the prior review; still worth someone checking.

---

## The four prior blockers

### 1. Open and read failures emit `E_EVIDENCE_UNREADABLE` with parseable stdout — **resolved**

All five inputs, `assay evidence show <input> --format json`:

| input | exit | stdout bytes | stdout `reason_code` | stderr |
|---|---|---|---|---|
| valid vector | 0 | 4392 | — (bare `{manifest,events}`) | empty |
| tampered vector | 2 | 549 | `E_EVIDENCE_INTEGRITY` | `[E_EVIDENCE_INTEGRITY] …` + `verifier_code: IntegrityEventHash` |
| `./does-not-exist.tar.gz` | 2 | 555 | `E_EVIDENCE_UNREADABLE` | `[E_EVIDENCE_UNREADABLE] … (os error 2)` |
| first 400 bytes of the valid vector | 2 | 529 | `E_EVIDENCE_UNREADABLE` | `[E_EVIDENCE_UNREADABLE] …` |
| `not a gzip at all` | 2 | 525 | `E_EVIDENCE_UNREADABLE` | `[E_EVIDENCE_UNREADABLE] …` |
| zero-length file | 2 | 521 | `E_EVIDENCE_UNREADABLE` | `[E_EVIDENCE_UNREADABLE] …` |
| a directory | 2 | 521 | `E_EVIDENCE_UNREADABLE` | `[E_EVIDENCE_UNREADABLE] … (os error 21)` |

Every one parses as `assay.run_summary.v1` with a non-empty `next_step`. The five zero-stdout rows the prior head produced are gone.

Precedence is decided in one place, twice over. `evidence_reason_for_verifier_code` (`cli_failure.rs:31-39`) is the single verifier-code mapping and returns at most one reason per code, so the two families cannot both claim a code; `classify_show_error` (`evidence/mod.rs:374-386`) is the single ordering site and is what both call sites in `cmd_show` use. That is the right shape. See blocking finding **B1** for the failure path in `cmd_show` that does *not* route through it.

### 2. One shared production mapping, pinned to the normative boundary — **resolved**

There is exactly one production statement per family (`EVIDENCE_INTEGRITY_CODES`, `EVIDENCE_UNREADABLE_CODES` at `cli_failure.rs:10-22`) behind one function, and both classifiers key on that function rather than on `ErrorClass` or the `Integrity*` prefix. Nothing restates it: the positive test cases now *iterate* the consts instead of carrying a parallel array, which is what made the prior head's mutation survivable.

The one hand-written list of the seven codes (`spec_reason_code_registry.rs:293-301`) is not a second copy of the mapping — it does not distinguish required from forbidden and so cannot answer the question the mapping answers. It encodes a different property (the boundary still names each code, and `assay-evidence` still declares it), and it works *for* the pin rather than against it, as blocker 4's escalation shows.

### 3. The derived parsers fail when they find zero codes — **resolved, verified red three ways**

I broke the parse deliberately in three ways. Each went red; none went green-on-nothing.

- Renamed the clause anchor `An emitter MUST key on` → `Emitters are required to key on` in the boundary document: `evidence_integrity_code_set_matches_the_normative_boundary` FAILED at `cli_failure.rs:208` with "normative boundary must retain its required/forbidden code clauses".
- Kept the anchors and removed the backticks around the four codes (the exact green-on-nothing shape): FAILED at `cli_failure.rs:213` with "the normative integrity boundary parser must find verifier codes".
- Same trap on the other parser — removed the backticks from `IntegrityIo`/`IntegrityGzip`/`IntegrityTar` in the `| E_EVIDENCE_UNREADABLE |` registry row: `evidence_unreadable_code_set_matches_the_normative_registry` FAILED at `cli_failure.rs:239` with "the unreadable reason registry parser must find verifier codes".

Both comparisons are `assert_eq` on sets, so drift in either direction is caught, and the row/clause lookups are `.expect(...)` rather than defaults, so a moved section fails rather than yielding nothing.

### 4. Removing `IntegrityFileSizeMismatch` fails — **resolved, with defence in depth**

Deleting `assay_evidence::ErrorCode::IntegrityFileSizeMismatch,` from `EVIDENCE_INTEGRITY_CODES`, nothing else:

```
test cli_failure::tests::evidence_integrity_code_set_matches_the_normative_boundary ... FAILED
  left:  {"IntegrityEventHash", "IntegrityManifestHash", "IntegrityRunRootMismatch"}
  right: {"IntegrityEventHash", "IntegrityFileSizeMismatch", "IntegrityManifestHash", "IntegrityRunRootMismatch"}
test result: FAILED. 3 passed; 1 failed; 543 filtered out
```

The nine registry tests stayed green on that mutation, as designed — they pin a different property. So I ran the escalation a determined developer would reach for, editing the normative document to match the code (`the four` → `the three`, code line dropped). The unit tests then went green, and two registry tests caught it instead:

```
test the_boundary_names_error_codes_that_still_exist_in_assay_evidence ... FAILED
  IntegrityFileSizeMismatch is checked here and the boundary no longer names it
test the_evidence_integrity_boundary_reads_the_same_in_the_spec_and_the_code ... FAILED
  the §5.1 registry row is no longer the text of …/evidence_integrity_boundary.md
test result: FAILED. 7 passed; 2 failed
```

Three edits in three files are now needed before the emitter can silently drop a code, and the third one still fails. This is the property the prior head lacked.

---

## Blocking findings

### B1. The generated contract now declares **no gap** on step 7, while a reachable exit-2 path still writes zero stdout

`scripts/docs/generate-agent-golden-path.py:376-406`, and the artifacts it writes.

Measured at this head. A bundle whose `manifest.json` is not well-formed JSON (`tar czf` of a two-file directory, `manifest.json` containing `this is not json`):

```
assay evidence show contract.tar.gz --format json
exit: 2
stdout: 0 bytes
stderr: fatal: failed to open bundle reader
        Caused by:
            0: Bundle verification failed
            1: Contract: Manifest JSON is not well-formed (ContractInvalidJson)
```

No `reason_code`, no `next_step`, no document. The whole `Contract*`/`Limit*`/`Security*` family reaches this: `classify_show_error` correctly declines to classify it and falls through to `error.context(fallback)`, which the top-level funnel renders as an untyped `fatal:` chain (`main.rs:46-49`). That fall-through is the right *classification* decision — the objection is only to what the contract says about it.

Before this PR, `docs/generated/agent-golden-path.json` carried `gap_issue: 2164` on step 7's failure outcome. At this head **all three** of step 7's outcomes carry `gap_issue: null`, so the machine-readable artifact — the one agents read, and the one this PR ships into `.agents/skills/`, `.claude/skills/` and the packaged plugin copy — now says step 7 has no known gap. The skill's own prose reinforces it: "Exit: Valid `0`; integrity failure `2`; unreadable bundle `2`", which reads as an exhaustive enumeration of exit 2.

Two things make this more than pedantic. First, the size: the boundary file's own measurement is that of 496 detections in `docs/experiments/evidence-mutation-cost-2026-06`, 3 carry a code this rule admits and 479 carry `ContractInvalidJson` — so the outcome that is now documented as machine-readable is a small minority of measured detections, and the majority is the one with no gap marker. Second, `failure_summary` lumps the four excluded things into one clause: "I/O, gzip, tar, and format-contract failures do not [map to `E_EVIDENCE_INTEGRITY`]". Three of those four now get `E_EVIDENCE_UNREADABLE`; the fourth gets nothing at all. Read as an agent would read it, the sentence says all four are handled alike.

For completeness, #2219 carries `horizon:post-mvp` at review time (labels: `area:core`, `cluster:reason-codes`, `size:M`, `horizon:post-mvp`), not `horizon:mvp`. So the missing *code* is a deliberate deferral and I am not asking for it. The defect is the removed disclosure, which is why the fix is cheap: restore a `gap_issue` on step 7 pointing at #2219, or state in `failure_summary` that a format-contract defect exits `2` with no stdout document. Falsified by either.

### B2. `--no-verify --format json` returns exit 0 and a clean-looking document for the tampered bundle, with nothing recording that verification was skipped

`evidence/mod.rs:301-322`. Measured on the tampered vector: `exit 0`, 4362 stdout bytes, empty stderr, and the document's top-level keys are exactly `["manifest", "events"]` — no `verified`, no `verify_mode`, no equivalent. The table format prints `Verified: ⚠️ SKIPPED` and `⚠️ Verification Skipped`; the JSON format prints nothing, so `verified` at `:310` is computed and consumed only by the branch this PR does not touch.

The machinery to say it already exists and this PR is already using it: the *failure* document carries `provenance.verify_mode: "enabled"`, which I read on every failing run above. So the success document is the only place the mode is unobservable.

This behaviour predates the PR, and it survived while the command had no machine contract. This PR is what makes `assay evidence show … --format json` the advertised integrity surface in three shipped agent-facing artifacts. An agent that passes `--no-verify`, or a wrapper that already sets it, now gets exit 0 and a well-formed document from a bundle whose content does not verify, and nothing in the changed docs, the skill, or the PR's non-claims discloses it. Falsified by a `verify_mode` field on the JSON success document, or one sentence in the skill saying so.

---

## Polish notes — owner's disposition, no action needed now

**P1. Hostile path bytes are correctly escaped on stdout and still emitted raw on stderr.** Thirteen hostile paths tested (space, single and double quote, backtick, `$( )`, `;`, embedded newline, backslash, `ESC ] 0 ; PWNED BEL` + ANSI SGR, U+202E). stdout was correct in every case. stderr was not: the headline is written byte-for-byte, so the OSC sequence reaches the terminal verbatim and an embedded newline splits one diagnostic across two lines in a way a line-oriented log reader mis-attributes. Sink is `pipeline_error.rs:94-103` → `Diagnostic::render` (`crates/assay-core/src/errors/diagnostic.rs:122-143`), which writes `self.message` unsanitised while the `context` object beside it is correctly JSON-escaped. Pre-existing and shared with `CliFailure::policy_parse`, and the value is caller-controlled argv, which is why this is a note. The repository already owns `assay_evidence::sanitize` and `CLAUDE.md` states the invariant for rendered strings.

**P1b. Shell-free argv recovery is correct — verified by execution, not by inspection.** For each hostile path I parsed `next_step`, stripped `Run argv: `, `json.loads`'d the array, asserted `argv[3]` equals the original path byte-for-byte, and executed `[binary] + argv[1:]` as argv with no shell. All seven round-tripped exactly and replayed to the same `reason_code`; no shell metacharacter expanded (no `PWNED` artifacts were created). Nothing to fix here — recording it because it is the property the recovery format exists for.

**P2. The success and failure documents are two schemas and only one self-identifies.** Failure carries `"schema": "assay.run_summary.v1"`; success is a bare `{manifest, events}` with no `schema` key. A consumer still cannot parse the union without knowing which path produced it.

**P3. The typed four-way distinction is not observable on the machine channel.** stdout's `message` is `evidence bundle <path> failed content verification: Bundle verification failed`; `assay evidence verify` on the same bundle says `Content hash mismatch at seq 0 (IntegrityEventHash)`. So `seq` is dropped from both channels, and `verifier_code` survives only in the stderr context object — stdout has no field for which of the four codes fired. On the unreadable path it is worse: the stderr context is `{"path": …}` only, so `IntegrityTar` versus `IntegrityGzip` versus a direct `os error 2` is unrecoverable from either channel except by reading English.

**P4. `EEvidenceUnreadable`'s doc comment is a hand-written restatement of its §5.1 row, unpinned.** `exit_codes.rs:87-89` is a two-line `///` summary; the registry row is six sentences. This is precisely the shape `the_doc_comment_transports_the_boundary_rather_than_restating_it` exists to prevent for `EEvidenceIntegrity`, which gets `#[doc = include_str!(…)]` plus a flowed-text equality test. An asymmetry rather than a defect today, and the new code's rule is the weaker of the two, so the cost of the missing transport is lower.

**P5. Two sites still answer "is this `evidence show` invocation JSON".** `args/mod.rs:57` and `evidence/mod.rs:313` both compare `format == ShowFormat::Json`; the sibling policy arm uses the accessor `args.is_json()`. Low drift risk while both are equality on one field; it stops being low as soon as a second way to request JSON exists, and disagreement is exactly this feature's failure mode (a summary emitted for a table run, or suppressed for a JSON run).

**P6. U+202E passes through raw in the stdout JSON string.** Valid JSON and a parser is unaffected; a human reading the document sees reordered text. Mentioned only because P1's sanitiser would cover it.

**P7. SPEC §7's MUST is met on the two classified paths and not on the residual one.** "When exiting with exit_code ≠ 0, the process MUST print at least one line that is a concrete command or hint" — the `Contract*`/`Limit*`/`Security*` path in B1 prints only a `fatal:` chain. Same root cause as B1.

**Checked, no finding.** `E_EVIDENCE_UNREADABLE` is registered in the one exit-code table the repository has (`SPEC-PR-Gate-Outputs-v1.md` §5.1) and maps to `EXIT_CONFIG_ERROR` under both `ExitCodeVersion::V1` and `V2`; the only reachable exits for this command are `0` and `2` and both are registered; `E_EVIDENCE_INTEGRITY` is correctly *removed* from §5.4 now that something constructs the variant, which `the_reserved_section_claims_nothing_a_variant_construction_contradicts` enforces. Generated-artifact parity is one source, not a copy set: `scripts/docs/generate-agent-golden-path.py` writes all five outputs from one `STEPS` table, `--check` is the drift gate, and it is wired into `.pre-commit-config.yaml` and `scripts/ci/check-docs-generated-drift.sh`, so a hand-edited `SKILL.md` or plugin copy fails `Lint (pre-commit)`. I confirmed `--check` exits 0 and both copy pairs are byte-identical at this head. ADR-042's stop list: I found no scalar score, whole-action verdict, identity/federation layer, provider-outcome claim, detector catalogue, or compliance/certification language introduced; the new reason string's own text disclaims tampering, intent and cause, and the failure documents I read carry no verdict.

**Angle I could not exercise.** Non-UTF-8 paths: APFS rejects the filename (`[Errno 92] Illegal byte sequence`), so `path.display()`'s lossy conversion is unexercised on this platform. On a filesystem that permits it, `cli_failure.rs:69` and `:110` would substitute U+FFFD and the emitted `path` — and therefore the published recovery argv — would not round-trip to the file.

---

## Host-capability proof: not triggered, and this was a classifier decision rather than a produced proof

Reported separately as asked, because a gate that ran and validated a proof and a gate that decided none was needed are indistinguishable in the checks list.

`host-capability-check` is `completed / success` for head `c9c949412ccc912f0523ca0900961ceb94b410c8` ([job 93566407265](https://github.com/Rul1an/assay/actions/runs/31422514696/job/93566407265)). Its own log ends:

```
self-test: all cases passed
PASS: no host-capability proof required for this PR.
```

That is `assay_host_capability_check.py:490-491`, the early return taken when the path classifier finds nothing requiring proof — not `:501`, the line that reports a validated proof for a head. No proof was produced, fetched or checked.

I verified the classifier's inputs independently rather than trusting the conclusion: I imported `proof_required` from the script at this head and ran it over the PR's actual 14-file list from the API. Result `False` with an empty reasons list. `TRIGGER_PREFIX` is `crates/assay-cli/src/diagnostics/`, and none of the 14 files is under it — the four Rust files are `cli/args/mod.rs`, `cli/args/tests.rs`, `cli/commands/evidence/mod.rs`, `cli_failure.rs`, `exit_codes.rs` and `tests/agent_golden_path_contract.rs`. Neither exemption did any work here: no path is in `SELF_EXEMPT`, and while four changed files are `.md` and would have been exempt anyway, none of them is under the trigger prefix either, so the answer does not depend on the markdown carve-out. `python3 scripts/ci/assay_host_capability_check.py --self-test` passes at this head. Corroborating from the other side: the `Host Capability Proof` workflow has **no run** at this head — its most recent runs are `bc07ffb0e` (cancelled, twice) and two successes from 2026-06-09 — and the workflow names that did run at this head do not include it.

So the PR body's claim "No delegated host proof is triggered: this PR does not touch `crates/assay-cli/src/diagnostics/**`" is accurate, and the green check is a correct *not-required* answer rather than evidence of host capability.

One property of that answer worth knowing, not a finding on this PR: the not-required branch returns before any head comparison, and the head-pinning block in the workflow only runs when `expected_head_sha` is set, which happens on `workflow_dispatch` and not on `pull_request`. So a "no proof required" pass is bound to a SHA only by the check-run's own head association and by the file list the API returned at run time. Harmless for a not-required answer on a settled head; it would matter for a PR whose head moves while the classifier is reading its files.

---

Reviewer worktree removed after review, including its `target/`.

Rul1an:
## Independent review — bound to `17d0d714d0d1302d0bb2084579ec1e2ec2b12da5`

**Verdict: request changes.** Three blocking findings, all in the same family: the disclosure field this push adds is
correct on the success path but forks a registered vocabulary, is contradicted by the failure envelope of the same
command, and is named in none of the shipped artifacts.

The typed-boundary work is good and the mutation pins are real. All five requested reproductions hold; the blocking
findings are about `verify_mode` specifically, not about the classification this PR was opened for.

**Measurement provenance.** All measurements below are of `17d0d714d0d1302d0bb2084579ec1e2ec2b12da5`, built and run in
my own detached worktree `/tmp/rev-2262` (removed, with its `target/`, after review). Platform Darwin 25.6.0 arm64
(macOS), `rustc`/`cargo` 1.96.0, **debug** profile (`cargo build -p assay-cli`, binary `target/debug/assay`). No shared
checkout was built in or modified.

---

## Blocking

### B1. `verify_mode: "skipped"` is outside the registered vocabulary and contradicts a normative MUST

`docs/architecture/SPEC-PR-Gate-Outputs-v1.md` §3.2 registers `verify_mode` as `"enabled"` or `"disabled"`, and states
normatively: *"If the run was executed with `--no-verify`, `verify_mode` MUST be `"disabled"`."* That rule already has
exactly one implementation, `Provenance::new` at `crates/assay-core/src/report/summary/types.rs:290`.

This PR adds a second, hand-written implementation at `crates/assay-cli/src/cli/commands/evidence/mod.rs:319` that
emits `"skipped"` for the exact condition the spec names. The diff amends neither §3.2 nor the §9 change log for it.

Measured at this head:

| invocation | field | value |
|---|---|---|
| `evidence show <valid> --format json` | top-level `verify_mode` | `"enabled"` |
| `evidence show <tampered> --format json --no-verify` | top-level `verify_mode` | `"skipped"` |
| `evidence show <tampered> --format json` | `provenance.verify_mode` | `"enabled"` |

Two producers of one field name with disjoint domains, in one CLI. This is the AGENTS.md *one rule, one function* case
directly: the rule is written twice and the second copy already disagrees. The consumer consequence is the failure the
field was added to prevent — a consumer branching per the registry on `{enabled, disabled}` meets `"skipped"` and
matches neither branch, so it cannot tell a skipped verification from an unknown value.

Fix: emit `"disabled"`, or register `"skipped"` in §3.2 and §9 and reconcile `Provenance::new`. Preferably let the one
function answer it rather than adding a third site.

### B2. The `--no-verify` failure envelope affirmatively reports that verification was enabled

```
$ assay evidence show "$W/nope.tar.gz" --format json --no-verify ; echo $?
2
  reason_code            = E_EVIDENCE_UNREADABLE
  provenance.verify_mode = "enabled"     <-- verification was skipped
```

Cause: `CliFailure::emit` calls `summary_from_outcome(&self.outcome, true)` — `verify_enabled` hardcoded `true` at
`crates/assay-cli/src/cli_failure.rs:133`.

The hardcode pre-exists on the merge base, but it was vacuous there: the only `CliFailure` producer was `policy_parse`,
and `assay policy validate` has no `--no-verify`. This PR routes the first `--no-verify`-bearing command into that
funnel, so the pre-existing hardcode starts emitting a false statement. Newly reachable, therefore in scope for this
head.

This violates the §3.2 normative MUST, and it is the same defect class the previous head was refused on: a
machine-readable document positively asserting a verification posture that did not hold.

### B3. The new disclosure is named in none of the shipped artifacts, and the PR body claims otherwise

`rg verify_mode` across `docs/generated/agent-golden-path.json`, `.agents/skills/assay-golden-path/SKILL.md`,
`.claude/skills/assay-golden-path/SKILL.md`, `packaging/claude-plugin/skills/assay-golden-path/SKILL.md`, and
`docs/guides/agent-golden-path.md` returns **zero hits**. Step 7's generated `stdout_summary` still reads *"Success
parses as an object containing `manifest` and `events`"*, and the generated contract registers no `--no-verify`
outcome, so nothing regenerates or pins the field. The only pin is a hand-written assertion at
`crates/assay-cli/tests/agent_golden_path_contract.rs:429` and `:447`; `assert_document` checks schema identity only,
so the generated contract and the binary can drift on step 7's stdout shape with nothing going red.

The PR body states: *"The generated golden-path skill, plugin copy, JSON contract, guide, normative output registry,
and reason-code implementation describe the same behavior."* For `verify_mode`, none of them describes it at all. That
claim is not established.

Answering the review question directly: the field **is** present, **is** correctly valued on the success path, and
**is** observable in the bytes by a consumer who does not know which flag was passed. It is not *discoverable* — the
three shipped agent artifacts that advertise this command as the integrity surface do not mention it, so a consumer has
no reason to read it.

---

## Post-MVP notes (with dispositions, non-blocking)

No issues were opened, per the freeze until the five release-blockers land.

### N1. `Security`- and `Limits`-class failures also exit 2 with empty stdout, and nothing discloses them

Measured: a bundle carrying a `../escape.json` member exits 2 with **0 bytes** on stdout and
`Security: Invalid path component (SecurityPathTraversal)` on stderr. `LimitBundleBytes` takes the same untyped path by
code — the PR's own unit test at `cli_failure.rs:279-280` asserts that both return `None` from both classifiers.

Step 7's `failure_summary` names only format-contract as the residual empty-stdout family and attributes it to #2219,
which owns `Contract*` codes only. As written the sentence reads exhaustive and is not.

*Disposition:* accept for MVP. The repair does what was asked — I/O, gzip and tar now receive a code and
format-contract does not, so the families that behave differently are no longer lumped together — and these two
families were never in #2164's scope. Recommend the wording generalise to *non-integrity, non-I/O verifier failures*
so the disclosure does not read as a closed list.

### N2. stderr still emits caller-controlled paths raw; stdout is correct

Eight hostile paths measured. stdout JSON-escapes `ESC` to `\u001b` and parses as
`assay.run_summary.v1`/`E_EVIDENCE_UNREADABLE` in **8 of 8**. stderr carries raw `ESC` in 4 (ANSI SGR, OSC 8 hyperlink,
OSC 52 clipboard write, CSI `2J`/`H`) and raw `BEL` in 2; confirmed by `od -c`, e.g.
`033 ] 8 ; ; h t t p : / / e v i l 033 \`. OSC 52 is the notable one — a clipboard write driven by a filename.

Not made worse by this push: the merge base already printed the raw path to stderr via
`.with_context(|| format!("failed to open bundle {}", ...))`. This PR changes the wrapper text, not the sanitization
property. That comparison is code-level; I did not build a merge-base binary.

*Disposition:* post-MVP, severity unchanged by this head. stderr is the human channel and the machine channel is
clean, which is the boundary that matters for the contract. Worth fixing when the sanitizer already used for TUI
rendering is extended to operator diagnostics.

### N3. The success document has no schema identity or version

The generated contract records `stdout.document: null` for step 7 `valid`, and `assert_document` asserts the *absence*
of a `schema` key. `verify_mode` was therefore added to an unidentified, unversioned ad-hoc object, so a consumer has
no identity to branch on and no version bump to detect the field's arrival.

*Disposition:* post-MVP. This is the pre-existing shape of the command's success output; giving it an identity is
larger than this slice and would be a contract change in its own right.

---

## ADR-042 / ADR-043 non-claims, both directions

- **A claim the change does not establish.** The PR body's *"a tampered bundle opened with `--no-verify` therefore
  cannot look verified"* holds for the success document only; on the same flag the failure envelope of the same command
  asserts `verify_mode: "enabled"` (B2). The *"describe the same behavior"* claim is likewise unestablished (B3).
- **A real limitation it introduces that nothing discloses.** The second `verify_mode` vocabulary (B1). Nothing in the
  diff tells a consumer that this CLI now emits two disjoint value sets under one field name, and the normative
  registry still says the field is two-valued.
- **Stop list: clean.** No scalar trust score, no whole-action verdict, no identity/delegation/federation layer, no
  provider-outcome verification, no detector catalogue, no compliance/certification/partnership claim.
  `E_EVIDENCE_UNREADABLE` establishes only that a readable bundle could not be obtained; `E_EVIDENCE_INTEGRITY` stays
  mismatch-only with no tampering, intent, or cause attribution. Reserved-section bookkeeping is consistent: the §5.4
  entry was removed as the code went live, and `every_registered_code_is_emittable_or_reserved` and
  `every_emittable_code_is_registered` both pass.

---

## The five reproductions — 5 of 5 genuinely hold

1. **Verified JSON success says `verify_mode=enabled`. HOLDS.** Exit 0; keys `["events","manifest","verify_mode"]`;
   `verify_mode="enabled"`.
2. **Tampered + `--no-verify` exits 0 and says `verify_mode=skipped`. HOLDS as specified**, subject to B1/B3. Exit 0,
   4393-byte document, keys now `["events","manifest","verify_mode"]` with `verify_mode="skipped"`, stderr empty. The
   field is present, correctly signals that verification did not run, and is observable without knowing the flag. It is
   genuinely fixed, not reworded — the previous head's 4362-byte `["manifest","events"]` document is gone. Two caveats
   stand: the value is outside the registered vocabulary (B1) and the field is documented nowhere (B3).
3. **Malformed `manifest.json` exits 2 with empty stdout; contract carries `gap_issue: 2219` and a failure summary.
   HOLDS.** Exit 2, 0 bytes stdout, stderr `Contract: Manifest JSON is not well-formed (ContractInvalidJson)`.
   Generated contract: `format-contract-failure` outcome has `gap_issue: 2219`, and `failure_summary` contains both
   `gap #2219` and `empty stdout`, pinned by
   `evidence_inspection_contract_discloses_the_deferred_format_contract_gap`. The number is right: #2219 is *"Evidence
   content defects in ErrorClass::Contract have no registered reason code"*, OPEN, labelled `horizon:post-mvp`. And the
   summary no longer lumps the families together — I/O, gzip and tar receive `E_EVIDENCE_UNREADABLE`, format-contract
   receives no code. See N1 for the two families it omits. Confirmed as characterised: the zero-stdout behaviour
   remains and the repair is that it is disclosed again rather than recorded as `gap_issue: null`.
4. **Generated skill and plugin copies remain byte-identical. HOLDS — regenerated, not read.**
   `python3 scripts/docs/generate-agent-golden-path.py` then `git status --porcelain` came back **empty**, so all six
   in-tree artifacts are byte-exact reproductions of the generator at this head. `.agents/…/SKILL.md` ==
   `.claude/…/SKILL.md` byte-identical (`cmp`); `docs/generated/agent-golden-path.json` == plugin
   `references/agent-golden-path.json` byte-identical. `packaging/claude-plugin/…/SKILL.md` differs at line 12 by
   design (plugin-root contract path). `python3 scripts/ci/test-agent-golden-path-skill.py` →
   *"portable, byte-identical, and contract-complete"*.
5. **Prior unreadable/integrity mappings and mutation controls still hold. HOLDS, every part.** All five unreadable
   inputs — missing file, truncation to 100 bytes, non-gzip, empty file, directory — exit 2 with a parseable
   `assay.run_summary.v1` carrying `E_EVIDENCE_UNREADABLE`. Precedence resolves in one place:
   `evidence_reason_for_verifier_code` is the single mapping and both `evidence_integrity` and `evidence_unreadable`
   consult it. Mutation controls re-run at this head, each restored by name:

   | mutation | result |
   |---|---|
   | redact the four code names from the boundary's required clause | **red** — `evidence_integrity_code_set_matches_the_normative_boundary`: *"the normative integrity boundary parser must find verifier codes"* (not a vacuous pass) |
   | remove `IntegrityFileSizeMismatch` from the production set only | **red** — *"the executable integrity classifier drifted from the one normative boundary"* |
   | remove it **and** edit the normative boundary document to match | **still red** — `spec_reason_code_registry::the_evidence_integrity_boundary_reads_the_same_in_the_spec_and_the_code` and `the_boundary_names_error_codes_that_still_exist_in_assay_evidence` both fail |

   Three independent pins, and the third survives the coordinated edit. A repair elsewhere in the file did not stop
   them working.

---

## Host-capability gate

**Not triggered.** `host-capability-check` logged `PASS: no host-capability proof required for this PR.`, and no
`host-capability-proof.yml` run exists at `17d0d714d0d1302d0bb2084579ec1e2ec2b12da5`. The classifier decided this PR
needs no proof. That green tick records the classifier's decision; it is **not** positive evidence about the host, and
no proof artifact was produced. `lane-check/proof` likewise reports *"no delegated runner proof required"*. Consistent
with the PR body's own statement, and with the diff touching no `crates/assay-cli/src/diagnostics/**` path.

---

## Quorum and bot records

This review is the entire quorum for `17d0d714d0d1302d0bb2084579ec1e2ec2b12da5`. No bot review counts at this head:

- `copilot-pull-request-reviewer[bot]` — `COMMENTED`, body *"Copilot was unable to review this pull request because
  the user who requested the review has reached their quota limit"*, and submitted against `1452b047`, not this head.
  No reviewer, no findings, not a verdict.
- `CodeRabbit` — check state `pass`, text *"Review rate limited"*. Unavailable infrastructure wearing the shape of a
  verdict; it is not a pass on the change.

## Verification I ran

`cargo test -p assay-cli --test agent_golden_path_contract` → 26 passed.
`cargo test -p assay-cli --bin assay cli_failure` → 4 passed.
`cargo test -p assay-cli --test spec_reason_code_registry` → 9 passed.
`cargo fmt --all -- --check` → clean.
CI at time of review: 55 pass, 9 skipping, 0 pending, 0 failing.

## Mutations and restores

Every mutation was made in my own detached worktree `/tmp/rev-2262`, never in a shared checkout, and restored by
naming the file individually:

- `crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md` →
  `git checkout -- crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md`
- `crates/assay-cli/src/cli_failure.rs` → `git checkout -- crates/assay-cli/src/cli_failure.rs`
- Regeneration rewrote the six generated artifacts; output was byte-identical, so `git status --porcelain` was empty
  and no restore was needed.

After restore, `git status --porcelain` was empty and `HEAD` was `17d0d714d0d1302d0bb2084579ec1e2ec2b12da5`. The
worktree and its `target/` are removed.

Nothing blocked an angle. Head did not move during review.

Rul1an:
## READY — bound to `74f5379657537a3a9026ead81d7227a93b27dc01`

Independent read-only review, sole quorum member. No blocker found.

Reviewer did not author or edit this change, nor SPEC-PR-Gate-Outputs-v1 or any plan governing it.

**Provenance of everything measured below:** head `74f5379657537a3a9026ead81d7227a93b27dc01`, merge base `4aa86feeec2014c8b8240d2089738b103c307b60`; detached worktree `/tmp/rev2262` (removed after review, incl. `target/`); macOS 26.6 / Darwin 25.6.0 arm64; `rustc 1.96.0 (ac68faa2 2026-05-25)`, `cargo 1.96.0`; **dev/debug profile** (`cargo build -p assay-cli`, binary `target/debug/assay`, version string `5.0.0`). No performance number is claimed.

### Seven reproductions — all seven hold, run against the built binary

| # | Claim | Result |
|---|---|---|
| 1 | verified JSON success emits `verify_mode=enabled` | **holds** — exit 0, top-level keys exactly `manifest`, `events`, `verify_mode`; `verify_mode="enabled"` |
| 2 | tampered bundle + `--no-verify` → exit 0, `verify_mode=disabled` | **holds** — exit 0, `"disabled"`. `"skipped"` no longer appears as a verify-mode value anywhere in `crates/` |
| 3 | missing bundle + `--no-verify` → exit 2, `provenance.verify_mode=disabled` | **holds** — exit 2, `E_EVIDENCE_UNREADABLE`, `provenance.verify_mode="disabled"`. Control without `--no-verify` on the same path reports `"enabled"`, so the field now tracks the flag rather than a constant. The former hardcoded `true` is gone: `CliFailure::emit` takes `Option<bool>` and `Cli::machine_output_verify_enabled()` resolves it in one place. `Some(true)` for `policy validate --json` is not a residual hardcode — that command has no verification toggle. |
| 4 | generated step 7 carries the driven outcome; every shipped copy names `verify_mode` with `enabled\|disabled` | **holds** — step 7 now has five outcomes incl. `verification-disabled` (exit 0, `--no-verify` in argv). `verify_mode`, `` `enabled` `` and `` `disabled` `` each present in all six shipped artifacts: the guide, `docs/generated/agent-golden-path.json`, all three `SKILL.md` copies, and the plugin `references/agent-golden-path.json` |
| 5 | `assay_core::report::summary::verification_mode` is shared | **holds — genuinely one function with two readers.** See below |
| 6 | prior mapping + three mutation controls still bite | **holds** — see below |
| 7 | generated copies byte-identical | **holds** — re-ran `python3 scripts/docs/generate-agent-golden-path.py`; `git status --porcelain` empty afterwards |

### On #5, the crux

One producer, two readers, and nothing else constructs the value.

- `verification_mode(bool) -> &'static str` at `crates/assay-core/src/report/summary/types.rs:303` is the sole site holding the two literals.
- Reader 1: `Provenance::new` (`types.rs:290`), reached only from `Summary::success`/`Summary::failure`.
- Reader 2: `crates/assay-cli/src/cli/commands/evidence/mod.rs:319`.
- A repo-wide search for `verify_mode` and for the literals `"enabled"`/`"disabled"`/`"skipped"` in `crates/` finds no third producer. The one other `"enabled"`/`"disabled"` pair (`assay-evidence/src/store/mod.rs:50`) is Object Lock detection, a different domain. This is not a helper standing beside two survivors.

**The code was brought to the spec, not the spec to the code.** SPEC §3.2 already required `"enabled"` or `"disabled"` and its normative "`--no-verify` ⇒ `disabled`" line; the diff does not touch §3.2 or anything else in §3. The spec's only changes are additive registration of `E_EVIDENCE_UNREADABLE` in §5.1, removal of the now-live `E_EVIDENCE_INTEGRITY` reservation from §5.4, and two §5.5 changelog rows.

Both readers are pinned by tests (`types.rs:454` for provenance; `agent_golden_path_contract.rs:429/444/461` for the CLI), so a future divergence is caught.

### On #6, the mapping and the three controls

All five unreadable inputs typed correctly by the binary, exit 2 / `E_EVIDENCE_UNREADABLE` for each: missing path, permission-denied (`chmod 000`), non-gzip bytes, truncated gzip, and — via the unit test — a typed `IntegrityTar`. Precedence is resolved in exactly one place: `classify_show_error` (`evidence/mod.rs:376`) tries integrity then unreadable, and both answers come from the single `evidence_reason_for_verifier_code`. A typed `ContractInvalidJson` carrying a nested `io::Error` is not reclassified.

Baseline green before mutating: `cargo test -p assay-cli --bin assay` 546 passed / 0 failed / 1 ignored; `--test agent_golden_path_contract` 26 passed; `--test spec_reason_code_registry` 9 passed. `cargo fmt --all -- --check` clean; `cargo clippy -p assay-cli -p assay-core --all-targets` clean.

Mutations, each in the throwaway worktree only:

1. **Dropped `IntegrityFileSizeMismatch` from `EVIDENCE_INTEGRITY_CODES`** (code only) → `cli_failure::tests::evidence_integrity_code_set_matches_the_normative_boundary` **FAILED** as intended. Worth recording precisely: the `agent_golden_path_contract` integration test still passed, so the tampered corpus vector does not trip that code and the parity test is the thing doing the work here.
2. **Coordinated edit — same removal from `crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md` as well** → the four `cli_failure` unit tests and all 26 golden-path tests went green, i.e. neither of those suites catches it. It is caught, by two tests in a third suite: `spec_reason_code_registry::the_boundary_names_error_codes_that_still_exist_in_assay_evidence` and `::the_evidence_integrity_boundary_reads_the_same_in_the_spec_and_the_code`, both **FAILED**. The pin holds; the reader should know it lives one suite over from the classifier.
3. **Zero-code parse** — removed the backticks around the four codes in the boundary document and around the three in the SPEC `E_EVIDENCE_UNREADABLE` row, so both derived parsers find nothing → `evidence_integrity_code_set_matches_the_normative_boundary` and `evidence_unreadable_code_set_matches_the_normative_registry` both **FAILED** on their explicit "parser must find verifier codes" assertions, rather than passing on an empty set.

Restores, each naming its files individually:
`git checkout -- crates/assay-cli/src/cli_failure.rs crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md`
`git checkout -- crates/assay-cli/src/exit_codes/evidence_integrity_boundary.md docs/architecture/SPEC-PR-Gate-Outputs-v1.md`
`git status --porcelain` empty after each.

### CI on this exact head — not yet settled

At time of review: **52 pass, 3 pending, 7 skipping**. Pending, and therefore not a pass: `Build + Test (ubuntu-latest)`, `Build + Test (windows-latest)`, `Criterion benches (store + suite)`. `Wave 0 feature matrix`, `Wave 0 quality gates`, `Wave 0 semver checks`, `S5 contract and determinism`, `Lint (pre-commit)`, `bundle_reader (evidence chain)`, `Public crates MSRV`, `Dependency Security` and `Criterion compare` have since gone green. `mergeStateStatus` is `BLOCKED`. This `READY` is a verdict on the change, not a substitute for the remaining three checks — auto-merge should wait for them.

No bot review counts here. `copilot-pull-request-reviewer[bot]` returned `COMMENTED` on the earlier head `1452b047` saying it "was unable to review this pull request because the user who requested the review has reached their quota limit"; `CodeRabbit` reports `pass` beside "Review rate limited". Both are unavailable infrastructure in the shape of a verdict, and neither leaves a reviewer or a finding.

### Host-capability gate — not triggered

The gate was **not triggered**: the classifier decided no proof was needed. `scripts/ci/assay_host_capability_check.py` logged `PASS: no host-capability proof required for this PR.`, and `lane-check/proof` reports `no delegated runner proof required`. The green tick is therefore an absence of a requirement and carries **no** positive statement about any host's capabilities; no proof was produced or delegated on this head.

### Accepted post-MVP notes — disposition recorded, not raised as blockers, no issue filed

1. **Security and Limits verifier families still exit 2 with empty stdout.** Confirmed present: both classifiers return `None` for `SecurityPathTraversal` and `LimitBundleBytes`, so those fall through to the untyped `fatal:` stderr path with no machine-readable document. Same shape as the `Contract*` family. *Disposition:* accepted post-MVP; the `Contract*` case is disclosed in the shipped contract and tracked by [#2219](https://github.com/Rul1an/assay/issues/2219), and the golden-path test pins that disclosure.
2. **The `evidence show` success document carries no schema identity.** Confirmed: top-level keys are exactly `manifest`, `events`, `verify_mode` — no `schema` field, unlike the failure document which is a proper `assay.run_summary.v1`. Relatedly, SPEC §3.2 governs `summary.json` provenance, so the success document borrows the §3.2 vocabulary without the spec registering that field on this document; the shipped golden-path contract is what pins it, executably. *Disposition:* accepted post-MVP.

### Non-claims

I did not measure: the pending three CI jobs, any release or cross-platform build beyond this arm64 macOS debug build, Linux-only paths, the eBPF/monitor surface, or any performance property.

Issue #2157, opened 2026-08-08
MCP 2026-07-28: we advertise the revision and omit three of its result requirements


Rul1an:
Measured against the [2026-07-28 changelog](https://modelcontextprotocol.io/specification/2026-07-28/changelog) by driving the server, not by reading the source.

## The good half, so this is not read as worse than it is

`assay-mcp-server` already carries the revision. `server/discover` is implemented, which the spec makes a **MUST**, and it advertises three versions:

```
$ echo '{"jsonrpc":"2.0","id":1,"method":"server/discover","params":{}}' | assay-mcp-server --policy-root .
versions: ['2024-11-05', '2025-11-25', '2026-07-28']
keys:     ['capabilities', 'protocolVersions', 'serverInfo']
```

The legacy `initialize` path is still accepted for older clients, which is the right shape for a stdio server that three different editors will speak to.

## Three requirements of that revision we do not meet

**1. `resultType` is missing from every result.** The changelog: *"All results now carry a required `resultType` field: `"complete"` for ordinary results and `"input_required"`."* Both `server/discover` and `tools/list` return `resultType: None`.

The compatibility rule saves us for now — clients **MUST** treat a missing field from earlier-protocol servers as `"complete"` — but we are not an earlier-protocol server. We advertise `2026-07-28` in `server/discover` and then omit a field that revision requires. That is the specific shape this repository objects to elsewhere: declaring a capability level and then not meeting it.

**2. `tools/list` omits `ttlMs` and `cacheScope`.** The revision adds a `CacheableResult` interface and requires both on `tools/list`, `prompts/list`, `resources/list`, `resources/read` and `resources/templates/list`. We return neither, so a client cannot cache and must poll.

**3. Tool order is not deterministic-by-contract.** The revision: servers **SHOULD** return tools in a deterministic order for client-side caching and prompt-cache hits. Ours comes out `check_args, check_sequence, policy_decide, check_coverage, explain_trace` — declaration order, which is stable today by accident rather than by test. One reordered `vec!` entry changes it silently.

## Why this matters now rather than later

#2152 ships manifests that install this server into Claude Code, Cursor and Codex. Whatever we advertise in `server/discover` is what those clients will hold us to, and the first two gaps are in the results they read on every session.

## Also worth recording from the same revision

- **Roots, Sampling and Logging are deprecated**, with a twelve-month window. Suggested migration for logging on stdio is `stderr` or OpenTelemetry, which is where our logging already goes.
[...]

PR #2267, opened 2026-08-10
fix(mcp): stop claiming unsupported modern protocol

Rul1an:
## Summary
- stop advertising or accepting MCP `2026-07-28` before the server implements its final wire contract
- restore `server/discover` as the standard `-32601` legacy fallback signal
- return typed `-32022` refusal data for modern or unknown request revisions
- preserve both legacy revisions and the exact five-tool editor install surface

## Why
The server claimed the final modern revision while omitting required modern request/result fields. A bounded MVP must be honest about supported capability rather than ship a partial protocol implementation.

This is the release-blocking slice of #2157. Full modern protocol support remains on #2157 as post-MVP work; this PR intentionally does not add `resultType`, cache metadata, structured output, or output schemas.

## TDD
RED on `95950a45f6505df03a388bacd60b0ef00a7b5999` before production changes:
- discovery still returned success instead of `-32601`
- a `2026-07-28` request was still accepted
- unsupported-version data still advertised `2026-07-28`
- dual-era fallback assertion failed

GREEN on the committed head:
- `cargo test -p assay-mcp-server --test protocol_negotiation_e2e --test project_install_surfaces`
- `cargo test -p assay-mcp-server`
- `cargo clippy -p assay-mcp-server --all-targets -- -D warnings`
- `git diff --check`

## Driven proof
Using `target/debug/assay-mcp-server --policy-root .`:
- `server/discover` -> `-32601 Method not found: server/discover`
- `_meta.protocolVersion=2026-07-28` -> `-32022`, supported `[2024-11-05, 2025-11-25]`
- legacy initialize negotiates `2025-11-25`
- `tools/list` names exactly `assay_check_args`, `assay_check_sequence`, `assay_policy_decide`, `assay_check_coverage`, `assay_explain_trace`

## Non-claims
- no final MCP `2026-07-28` support
- no change to tool behavior or editor manifests
- no proxy protocol change


Rul1an:
## Independent non-building review — PR #2267

**READY** — bound to head SHA `95950a45f6505df03a388bacd60b0ef00a7b5999`.

No blocking findings. Everything below was reproduced by running the binary or the tests at that
exact SHA in a throwaway detached worktree, not read off the diff.

### Environment provenance

- OS: `Darwin 25.6.0 arm64` (macOS, Apple Silicon)
- Toolchain: `rustc 1.96.0 (ac68faa20 2026-05-25)`, `cargo 1.96.0 (30a34c682 2026-05-25)`
- Profile: `dev` (unoptimized + debuginfo); binary `target/debug/assay-mcp-server`, `CARGO_PKG_VERSION` = `5.0.0`
- Worktree: detached `/tmp/rev-2267`, created from and verified at `95950a45f6505df03a388bacd60b0ef00a7b5999`,
  `git status --porcelain` empty before and after the mutation work. Worktree removed at the end of the review.
- I authored neither this change nor any specification or plan governing it, and I made no commit,
  push, or branch mutation anywhere. All mutation testing happened in `/tmp/rev-2267` only, restored
  from file copies (never `git restore`/`git checkout --`/`git clean`).

### Delegated host-capability proof: not required

Quoting the classifier's own output at the reviewed SHA rather than my reading of the file list:

```
$ gh api repos/Rul1an/assay/commits/95950a45f6505df03a388bacd60b0ef00a7b5999/status
lane-check/proof   success   no delegated runner proof required
```

So for this PR there is **no requirement** for a delegated host-capability proof. That is an absence
of a requirement, and nothing more: it is not evidence that any host has, or lacks, any capability,
and it must not be read as a substitute for a proof that was never demanded.

Related, and in the same spirit: two reviewer records on this head say they did not review, and
neither is a review under the AGENTS.md quorum rule. `CodeRabbit` reports `pass` with
"Review rate limited", and `copilot-pull-request-reviewer[bot]` submitted `COMMENTED` on
`95950a45f650` saying it "was unable to review this pull request because the user who requested the
review has reached" its quota. Both are unavailable infrastructure wearing the shape of a verdict.

### The seven verification points

**1. Discovery fallback returns `-32601`. Confirmed.**

```
$ echo '{"jsonrpc":"2.0","id":1,"method":"server/discover","params":{}}' | assay-mcp-server --policy-root tests/fixtures/mcp
{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found: server/discover"},"id":1}
```

The method reaches the catch-all dispatch arm; there is no `server/discover` handler and no
`protocolVersions` advertisement anywhere in the tree.

**2. Typed `-32022` for unsupported revisions. Confirmed for every case asked about.** Each row is the
literal response from the binary; the message is `"unsupported protocol version"` in every refusal.

| `_meta` protocol revision | code | `data` |
| --- | --- | --- |
| `"2026-07-28"` | `-32022` | `{"requested":"2026-07-28","supported":["2024-11-05","2025-11-25"]}` |
| `"$$$not-a-version$$$"` | `-32022` | `{"requested":"$$$not-a-version$$$","supported":["2024-11-05","2025-11-25"]}` |
| `""` (empty string) | `-32022` | `{"requested":"","supported":["2024-11-05","2025-11-25"]}` |
| `"2099-01-01"` (well-formed future) | `-32022` | `{"requested":"2099-01-01","supported":["2024-11-05","2025-11-25"]}` |
| `_meta` absent | no error | served normally as legacy — correct, since an absent claim asserts nothing |

The advertised and accepted set is exactly `["2024-11-05","2025-11-25"]` in every refusal payload. I
also checked a near-miss: `" 2025-11-25"` with a leading space is refused rather than trimmed into a
match, which is the right way round.

**3. Both legacy revisions and the legacy `initialize` handshake still work. Confirmed.** This was the
regression risk and it did not materialise.

- `_meta` = `2024-11-05` → no error, `tools` array of 5. `_meta` = `2025-11-25` → same.
- `initialize` with `2024-11-05` → `{"protocolVersion":"2024-11-05","capabilities":{"tools":{}},"serverInfo":{"name":"assay-mcp-server","version":"5.0.0"}}`; with `2025-11-25` → the same shape echoing `2025-11-25`. Each supported revision is echoed exactly, not silently upgraded.
- End to end: `initialize` → `tools/list` → `tools/call assay_check_args` returns `isError: false`.
- `cargo test -p assay-mcp-server`: **359 tests passed, 0 failed** across 32 test binaries, including all 10 in `protocol_negotiation_e2e` and the 4 in `project_install_surfaces`.
- `cargo fmt --all -- --check` clean; `cargo clippy -p assay-mcp-server --all-targets -- -D warnings` clean; `git diff --check` clean.

**4. Exactly five production tools. Confirmed — counted from the live `tools/list` response, not a doc.**

1. `assay_check_args`
2. `assay_check_sequence`
3. `assay_policy_decide`
4. `assay_check_coverage`
5. `assay_explain_trace`

A sixth, `assay_test_outbound`, exists only behind the `test-outbound` cargo feature and is absent
from the default build's surface, so the production count is five. The count is also pinned by
`project_surfaces_launch_the_five_production_tools`, which passes.

**5. No public source claim still advertises modern support. Confirmed, repository-wide.**

`MODERN_PROTOCOL_VERSION`, `SUPPORTED_PROTOCOL_VERSIONS`, `discover_result`, `RequestEra`,
`protocolVersions`, and "dual-era" have **zero** occurrences anywhere in the repository, including
hidden files, generated artifacts, packaging manifests, and skills. Nothing dangles.

I did not treat the small diff as evidence, and I checked the places a stale claim would hide:

- `CHANGELOG.md` never claimed modern support — it already states "This does not claim MCP `2026-07-28` support: the modern `server/discover` and per-request metadata contract remains separate." The post-state of this PR is what the changelog already describes.
- The remaining `2026-07-28` mentions are a different subject and are not support claims for this server: `assay-core`'s MCP era-parity fixtures and parser (*ingesting* modern transcripts as evidence), spec citations in ADR-036/ADR-043, and `examples/mcp-jsonrpc-id-conformance`, which documents a contradiction *in* the spec.
- `docs/guides/editor-mcp-recipe.md` §"Remote servers (provisional, MCP 2026-07-28)" is guidance addressed to the *wrapped upstream* server's OAuth flow, is explicitly marked provisional against the release candidate, and says the local stdio wrap is what is stable today. It does not assert that `assay-mcp-server` implements the revision.
- Packaging and skill mock servers (`packaging/claude-plugin/.../mock_github_mcp.py`, `crates/assay-mcp-server/tests/fixtures/proxy/mock_upstream.py`, `scripts/ci/claude_plugin_install_workflow.py`) all speak `2024-11-05`.

**6. Refusal happens before dispatch. Confirmed in the code path and at runtime.**

The gate is `crates/assay-mcp-server/src/server.rs:265`, immediately after request parsing; the
dispatch `match req.method.as_str()` begins at line 279. The refusal arm writes the response and
`continue`s the read loop, so no handler is reached.

Runtime proof, because "before dispatch" and "after dispatch with the result discarded" produce an
identical wire response and only differ in side effects. The same `tools/call assay_check_args` was
sent twice, once with a refused `_meta` revision and once without:

| | client sees | side-effect events on stderr |
| --- | --- | --- |
| refused revision | `-32022` | **0** — no `tool_call_start`, no `assay.usage.metered`, no `tool_call_done`, no `tool_decision` |
| control (no `_meta`) | `isError: false` | 4 |

No handler ran, nothing was metered, and no decision record was emitted for the refused request.

**7. Mutation sensitivity. The three named mutations are all caught. A fourth of my own is not.**

Each mutation was applied in `/tmp/rev-2267` only, then reverted from a file copy verified by
sha256.

| Mutation | Caught? | By which test |
| --- | --- | --- |
| **M1** — remove the `-32022` refusal (`check_request_version` always returns `Ok`) | **yes** | `a_modern_revision_claim_is_refused_with_the_legacy_set`, `an_unimplemented_revision_is_refused_with_the_supported_set` (2 failed, 8 passed) |
| **M2** — remove the `-32601` fallback by restoring a `server/discover` handler | **yes** | `discovery_is_an_explicit_legacy_fallback_signal`, `a_dual_era_client_can_fall_back_to_the_legacy_handshake` (2 failed, 8 passed) |
| **M3** — widen the supported set back to include `2026-07-28` | **yes** | `a_modern_revision_claim_is_refused_with_the_legacy_set`, `an_unimplemented_revision_is_refused_with_the_supported_set` (2 failed, 8 passed) |
| **M4** *(mine, not on the brief)* — keep the refusal response but let dispatch run first and discard its result | **no** | none — all **359** tests in the crate stayed green |

M3 is worth calling out as a deliberate improvement in this PR: it is caught partly by the new
negative assertion `assert!(!supported.iter().any(|v| v == "2026-07-28"))` at
`protocol_negotiation_e2e.rs:263`. That line is load-bearing, not decoration.

M4 is the one gap. Under it the client still receives the correct `-32022`, but the handler executes
first:

```
=== M4 stdout (what the client sees) ===
{"jsonrpc":"2.0","error":{"code":-32022,...,"data":{"requested":"2026-07-28",...}},"id":1}
=== M4 side-effect events on stderr ===
"event":"tool_call_start"  "event":"assay.usage.metered"  "event":"tool_call_done"  "event":"tool_decision"
```

The policy check ran and usage was metered for a request that was refused. The production code at
this SHA does **not** do this — point 6 proves it does the right thing — so this is a missing
invariant rather than a defect, and it is why I did not block on it. But nothing in the suite pins
the ordering, so a later refactor could move the gate below dispatch and stay green.

### Non-blocking observations, recorded as post-MVP advice

No issues opened for any of these, per the freeze. Each carries my technical disposition.

**A. The "before dispatch" ordering is correct but unpinned (from M4).** *Disposition: worth fixing,
cheap, not a blocker.* The suite asserts the refusal's shape, never its position. A test that sends
`tools/call` with a refused `_meta` and asserts the absence of `tool_call_start` / `assay.usage.metered`
on stderr would close it, and the harness already captures stderr for the leak assertions in
`missing_or_non_string_protocol_version_is_value_free_invalid_params`, so the machinery exists. The
billing signal makes this more than hygiene: under M4 a refused request is a metered request.

**B. The supported set is now written twice, with nothing pinning the two together (AGENTS.md
"One rule, one function").** *Disposition: real drift risk; the fix is a one-liner, so I would take it.*
Two sites independently encode which revisions exist: `SUPPORTED_LEGACY_PROTOCOL_VERSIONS`
(`server.rs:24`) and `LegacyProtocolVersion`'s `negotiate` arms plus `as_str` (`server.rs:81-93`).
This duplication predates the PR, but the PR removed the one test that drove two sites through the
wire and compared them (`every_advertised_revision_is_accepted` read the discover payload and fed it
back through the version check); its replacement,
`every_supported_legacy_revision_is_accepted_in_request_metadata`, hardcodes the list instead. Net
effect: adding a revision to the const would not be noticed by any test, and `initialize` would not
handle it. Because `as_str` is already a `const fn`, the const can simply be derived from the enum:

```rust
const SUPPORTED_LEGACY_PROTOCOL_VERSIONS: &[&str] = &[
    LegacyProtocolVersion::V2024_11_05.as_str(),
    LegacyProtocolVersion::V2025_11_25.as_str(),
];
```

That makes one edit change both, which the contract prefers over a parity test.

**C. The same revision string gets two different answers depending on where it appears.**
*Disposition: honest and spec-shaped in both paths; document, do not change behaviour.*
`_meta` = `2026-07-28` is refused with `-32022`, while
`initialize {"protocolVersion":"2026-07-28"}` succeeds and answers `2025-11-25`:

```
{"jsonrpc":"2.0","result":{"protocolVersion":"2025-11-25",...},"id":6}
```

This does not claim `2026-07-28`, so there is no ADR-042 problem — the legacy handshake is supposed
to answer with a revision the server does support and let the client decide. It is still two answers
to one string, it is pinned by `unsupported_versions_negotiate_to_the_latest_legacy_revision`, and a
one-line comment at the `_ => Self::LATEST` arm saying why the handshake downgrades where the `_meta`
path refuses would save the next reader the trip.

**D. A non-string `_meta` revision is served rather than refused.** *Disposition: pre-existing,
unchanged by this PR, low severity; tighten only if the shape is ever seen in the wild.* The extractor
ends in `.and_then(Value::as_str)`, so a numeric `20260728`, a `null`, or an array-shaped `_meta`
falls through to `Ok(())` and is served as legacy. A malformed version claim is arguably closer to a
refusable claim than to no claim at all, but the protocol types that field as a string, so treating a
non-string as "no revision declared" is defensible and it is not what this PR changed.

**E. A refused revision on a notification produces a response.** *Disposition: pre-existing,
unchanged by this PR; a JSON-RPC conformance wart, worth a follow-up.* The gate is deliberately
method-agnostic, which it must be to precede dispatch, so it also fires for notifications:

```
$ echo '{"jsonrpc":"2.0","method":"notifications/initialized","params":{"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28"}}}' | assay-mcp-server ...
{"jsonrpc":"2.0","error":{"code":-32022,...},"id":null}
```

JSON-RPC 2.0 says a server must not reply to a notification. The gate's position is not the thing to
change; skipping the write when `req.id` is absent would be.

**F. `pub const MODERN_PROTOCOL_VERSION` was publicly reachable and is being removed at an unchanged
crate version.** *Disposition: not a blocker — keeping it would be the worse outcome — but it needs a
recorded release-time decision, and the green semver check is not that decision.*

The facts I verified: `lib.rs` declares `pub mod server`, so
`assay_mcp_server::server::MODERN_PROTOCOL_VERSION` was part of the public surface;
`crates/assay-mcp-server/Cargo.toml` has no `publish = false`, and the crate is in the documented
publish order; and the released tag `v5.0.0` (commit `c68ea4c4`, 2026-08-08) contains that `pub const`
— its `server.rs` is byte-identical to `origin/main`'s (both sha256 `07f9fe76c4…`). The workspace
version is still `5.0.0`. So this removes a public item from a published crate without a version bump.

Two things stop me from blocking. First, the constant's only purpose was to name a revision this
server refuses, so leaving it in place *is* the ADR-042 unearned claim; blocking would preserve the
false claim to protect its symbol. Second, the remedy — a major bump, or an explicit reconciliation —
belongs to the release-line process, not to this diff.

What must not happen is reading the green tick as clearance. `Wave 0 semver checks (public crates)`
runs `cargo semver-checks` over an allowlist of exactly `assay-common`, `assay-policy`,
`assay-metrics`, `assay-core`, `assay-registry`, and `assay-evidence`
(`.github/workflows/split-wave0-gates.yml:470-487`). **`assay-mcp-server` is not in that list**, so
the job's success says nothing whatever about this crate's public API. I also could not establish
whether `5.0.0` is actually live on crates.io: the crates.io API refused the request from this
environment with a data-access error. That is an unverified point, not a clean one, and I am not
resolving it in either direction. The allowlist gap is itself the more durable finding here — a
published crate outside the semver gate.

### On the net −83 lines

Nothing load-bearing was removed. The deletions are `MODERN_PROTOCOL_VERSION`,
`SUPPORTED_PROTOCOL_VERSIONS`, the `RequestEra` enum, `discover_result()`, the `server/discover`
dispatch arm, a `debug_assert!` that only asserted an enum was one of its own variants, and the four
tests that pinned the removed behaviour. There are zero remaining references to any removed item,
the whole crate suite passes, and CI is green on `ubuntu-latest`, `macos-latest`, and
`windows-latest` at this SHA. The `-32022` machinery that is genuinely useful — `error_with_data` and
the `requested`/`supported` payload — is kept and is what now carries the refusal. The removal of the
`RequestEra` enum is a straight improvement: the previous code bound an era that nothing branched on
and said so in its own comment.

The change does what its title says, and it does it in the honest direction: it stops advertising a
revision the server does not implement, and it says so through a typed refusal that hands the client
the set it can actually use.

Rul1an:
READY

Independent read-only review, bound to head `2e79204bd6da3c1bfde2b84def36398270983375`. Non-building reviewer: I authored neither this change nor any specification or plan governing it, and I made no edit, commit, push, or amend on any branch. Every number below was measured on a throwaway detached worktree created from that exact SHA (`/tmp/rev-2267b`), never in the builder's worktree and never in the main checkout. The reviews at `95950a45f` and `b0d0b88b5` are superseded and I relied on neither.

No blocking findings.

---

## Head movement during measurement, and the AGENTS.md equivalence check

While I was measuring, the branch advanced to `0710813d24995ea6a7f795950154c500f6ee3f44`, a merge of `origin/main` into the reviewed head (parents `2e79204bd` and `367ccfc61`, where `367ccfc61` is `origin/main` exactly). Under AGENTS.md a merge commit is a new head like any other, so I ran the recorded equivalence check rather than assuming the review carries:

- **Condition 1 — the new head introduces no change, to any file, that is not already on `main`.** `git merge-tree --write-tree 2e79204bd6da3c1bfde2b84def36398270983375 367ccfc61e8e0fe00d57fcb3db736f6408b15297` exits `0` (no conflicts) and writes tree `7c34b2af51573f8f781eb9801b6cca9400f55f26`. `git rev-parse 0710813d24995ea6a7f795950154c500f6ee3f44^{tree}` is `7c34b2af51573f8f781eb9801b6cca9400f55f26`. Identical, whole tree. **Holds.**
- **Condition 2 — the advance touched no file the review covered.** The PR's changed files as of `2e79204bd` are exactly `crates/assay-mcp-server/src/server.rs` and `crates/assay-mcp-server/tests/protocol_negotiation_e2e.rs`. `git diff --name-only 2e79204bd..0710813d` lists 22 files (`assay-cli` doctor exit-class work, `assay-core` tool-pin validation, golden-path skills and generated docs). `comm -12` of the two sorted lists is empty. **Holds.**

Both conditions are recorded, so this review is revalidated for `0710813d` by the contract's own mechanism. The verdict itself remains bound to `2e79204bd`, which is the tree I actually built and ran.

---

## 1. The deprecation note and doc comment, judged as a public string

Rendered by rustc at a real downstream use site with no `allow`, in a separate crate outside this workspace that depends on `assay-mcp-server` by path:

```
warning: use of deprecated constant `assay_mcp_server::server::MODERN_PROTOCOL_VERSION`:
assay-mcp-server does not implement MCP revision 2026-07-28; this constant is retained only for
API compatibility with 5.0.0
```

Judgement: **honest and self-contained.** The note names the revision, so it resolves standing alone at a use site where nothing else from this crate is visible — which is the defect the previous phrasing ("this revision") had. It states non-implementation as a fact, not a deferral. There is no hedge that could be read as partial support, no capability claim, and no roadmap promise. The phrase "retained only for API compatibility with 5.0.0" tells the reader what the item is for, which is the only reason a downstream would keep naming it.

The doc comment reads:

> MCP revision `2026-07-28`, which `assay-mcp-server` does not implement.
>
> Retained at its 5.0.0 path, type, and value only so that code compiled against `assay-mcp-server` 5.0.0 keeps compiling. It is an API compatibility artifact, not a capability claim: the constant is absent from the accepted version set, is never advertised on the wire, and no accept or dispatch path reads it. A request declaring `2026-07-28` in `_meta` is refused with a typed `-32022` before dispatch, and `server/discover`, which that revision makes mandatory, returns `-32601`.

Every factual assertion in it is one I independently verified below: absent from the accepted set (§3), never advertised (§5 R1), unread by any accept or dispatch path (§3), `-32022` before dispatch (§5 R2, R7), `-32601` for discovery (§5 R1). A doc comment that states checkable facts and is checkable-true is the opposite of the ADR-042 defect. Naming the revision here is not a claim precisely because the sentence it sits in disclaims it.

**The removed roadmap phrasing is gone from every public surface, not just this string.** `git grep "complete modern wire contract"` over the whole tree at `2e79204bd` returns nothing. `git grep -i "until the .*is implemented|not yet implemented|will be implemented|coming soon|planned support|roadmap"` over `crates/assay-mcp-server/` returns nothing.

## 2. Source compatibility against `v5.0.0` is real, not similar

The declaration line is byte-identical to the released tag, not merely equivalent:

- `2e79204bd:crates/assay-mcp-server/src/server.rs:31` → `pub const MODERN_PROTOCOL_VERSION: &str = "2026-07-28";`
- `git show v5.0.0:crates/assay-mcp-server/src/server.rs:25` → `pub const MODERN_PROTOCOL_VERSION: &str = "2026-07-28";`
- String comparison of the two extracted lines: **IDENTICAL**. Visibility `pub`, type `&'static str`, value `"2026-07-28"` all unchanged, and the doc-only delta `b0d0b88b5..2e79204bd` did not disturb any of them.

The module path is unchanged too: `pub mod server;` is line 9 of `crates/assay-mcp-server/src/lib.rs` at both `v5.0.0` and `2e79204bd`.

Proved the way a downstream would, not by reading: a separate crate at `/tmp/downstream-probe`, outside this workspace (`[workspace]` stanza to detach it), depending on `assay-mcp-server` by path, containing

```rust
#[allow(deprecated)]
pub fn probe() -> &'static str { assay_mcp_server::server::MODERN_PROTOCOL_VERSION }
```

`cargo test` → exit `0`, `1 passed; 0 failed`, asserting the value is `"2026-07-28"`.

This matters because nothing else pins it. `assay-mcp-server` has no `publish = false` and takes `version.workspace = true`, still `5.0.0`; `Wave 0 semver checks (public crates)` runs `cargo semver-checks` over exactly six crates (`assay-common`, `assay-policy`, `assay-metrics`, `assay-core`, `assay-registry`, `assay-evidence`) and `assay-mcp-server` is not one of them. Its green on this SHA says nothing about this crate's public API.

## 3. Unreachable from acceptance, discovery, and the supported set

`git grep -n "MODERN_PROTOCOL_VERSION"` over the whole tree at `2e79204bd` returns exactly three lines, and no more:

```
crates/assay-mcp-server/src/server.rs:31            (the definition)
crates/assay-mcp-server/tests/protocol_negotiation_e2e.rs:8   (the doc comment naming it)
crates/assay-mcp-server/tests/protocol_negotiation_e2e.rs:19  (the compatibility assertion)
```

The real accepted set is `SUPPORTED_LEGACY_PROTOCOL_VERSIONS` (renamed from `SUPPORTED_PROTOCOL_VERSIONS`), a private `const` at `server.rs:37` holding `&["2024-11-05", "2025-11-25"]` as two literals. It does not reference the constant. Being private, it is not an API concern. `check_request_version` reads only that list; the dispatch `match` has no `server/discover` arm, so discovery falls through to the `-32601` default. No reference outside the definition and the test can influence acceptance, because there is no such reference.

## 4. The compatibility test, and both compile-failure modes

Test `deprecated_modern_protocol_version_stays_publicly_reachable`. I verified in the throwaway worktree that it goes red on each half of the contract, then restored the file from a checksummed backup each time.

| Mutation | Result | Compiler |
|---|---|---|
| Delete the constant | `cargo test -p assay-mcp-server --test protocol_negotiation_e2e` exit `101` | `error[E0425]: cannot find value MODERN_PROTOCOL_VERSION in module assay_mcp_server::server` |
| Change `pub const` to `const` | exit `101` | `error[E0603]: constant MODERN_PROTOCOL_VERSION is private` |

I also checked the doc comment's stated *reason* rather than taking it on trust. With the same privatisation applied, I added a unit test inside `server.rs` making the identical assertion through `super::`:

```
cargo test -p assay-mcp-server --lib mb_control  → exit 0
test server::mb_control::in_module_test_cannot_detect_lost_pub ... ok
```

It passes while the integration test fails to compile. So the doc comment is accurate: an in-module test cannot catch the visibility half, and the `tests/` location is load-bearing rather than stylistic. The test does not re-assert support — it asserts one value equality behind `#[allow(deprecated)]`, and the `allow` is itself the assertion that the item is deprecated.

Non-blocking observation carried below: nothing mechanical keeps the test in `tests/`.

## 5. Behavioural reproductions, driven against the binary

`./target/debug/assay-mcp-server --policy-root ./tests/fixtures/mcp` built from `2e79204bd`, no `ASSAY_*` variables in the environment.

| # | Input | Observed |
|---|---|---|
| R1 | `server/discover` | `{"error":{"code":-32601,"message":"Method not found: server/discover"}}`, no `result` |
| R2 | `tools/list`, `_meta` = `2026-07-28` | `{"error":{"code":-32022,"message":"unsupported protocol version","data":{"requested":"2026-07-28","supported":["2024-11-05","2025-11-25"]}}}` |
| R3a | `_meta` = `2099-01-01` | `-32022`, `requested` echoed, `supported` = the two legacy revisions only |
| R3b | `_meta` = `@@@garbage@@@` | `-32022`, same shape |
| R3c | `_meta` = `""` | `-32022`, `requested` = `""`, same shape |
| R3d | no `_meta` | no `error`, `result.tools` length `5` |
| R4 | `_meta` = `2024-11-05`, then `2025-11-25` | both served, no `error`, 5 tools each |
| R5 | legacy `initialize` at `2024-11-05` and `2025-11-25` | each echoed exactly, `serverInfo` = `{name: assay-mcp-server, version: 5.0.0}`, no `meta`, no status claim |
| R6 | `tools/list` | exactly `assay_check_args`, `assay_check_sequence`, `assay_policy_decide`, `assay_check_coverage`, `assay_explain_trace` |
| R7a | `tools/call assay_check_args` **with** `_meta` = `2026-07-28` | `-32022` only; stderr carries `server_start` and nothing else — **0** occurrences of `tool_call_start`, `assay.usage.metered`, `tool_decision`, `tool_call_done` |
| R7b | the identical `tools/call` **without** `_meta` (control) | tool runs, `allowed: true`, **4** such events on stderr |
| R7c | `initialize` with a valid `protocolVersion` but `_meta` = `2026-07-28` | `-32022`; refusal precedes negotiation |

R7a against R7b is the side-effect proof: same call, same arguments, and the refused one produces no execution, no metering, and no decision record. The five-tool surface is not only observed but pinned, by `project_surfaces_launch_the_five_production_tools`, which asserts an exact sorted array.

## 6. No stale support claim anywhere in the repository

Searched the whole tree at `2e79204bd`, not only the changed files. `git grep -iE "support[s]?[^.]{0,80}2026-07-28|2026-07-28[^.]{0,80}support"`, after excluding explicit non-claims, spec quotations and `Unsupported…` identifiers, returns **nothing**. The surviving `2026-07-28` mentions fall into four honest groups:

- explicit non-claims — `CHANGELOG.md:270` ("This does not claim MCP `2026-07-28` support"), `ADR-043` §4 row ("does not provide … MCP `2026-07-28` support"), and the `docs/superpowers/` specs, which exclude it by name;
- upstream spec references and pinned schema fixtures under `crates/assay-core/tests/fixtures/` and `examples/mcp-jsonrpc-id-conformance/`;
- the transcript-ingest corpus in `assay-core` (`docs/mcp/protocol-era-parity.md`, `docs/mcp/import-formats.md`), which is about *parsing* third-party transcripts — a different, separately checkable capability from serving the revision;
- unrelated date literals (`ADR-044` acceptance date, a `dependabot.yml` comment, a nightly toolchain pin).

No `SKILL.md`, README, generated artifact, or packaging manifest claims the revision. The one wording I would put in front of a doc owner is pre-existing and out of this PR's scope; see advice A5.

## 7. Mutation table

Each mutation applied only in `/tmp/rev-2267b`, never pushed; `server.rs` restored from a SHA-256-checksummed backup after each and the worktree confirmed clean (`git status --porcelain` empty at `2e79204bd`).

| Mutation | Suite result | Caught by |
|---|---|---|
| Remove the `-32022` refusal | exit `101`, 2 failed | `a_modern_revision_claim_is_refused_with_the_legacy_set`, `an_unimplemented_revision_is_refused_with_the_supported_set` |
| Remove the `-32601` fallback (re-add a `server/discover` arm returning a result) | exit `101`, 2 failed | `discovery_is_an_explicit_legacy_fallback_signal`, `a_dual_era_client_can_fall_back_to_the_legacy_handshake` |
| Widen the supported set back to include `2026-07-28` | exit `101`, 2 failed | `a_modern_revision_claim_is_refused_with_the_legacy_set`, `an_unimplemented_revision_is_refused_with_the_supported_set` |
| Delete the constant | exit `101` (compile) | `deprecated_modern_protocol_version_stays_publicly_reachable` (`E0425`) |
| Make the constant private | exit `101` (compile) | `deprecated_modern_protocol_version_stays_publicly_reachable` (`E0603`) |
| Keep the refusal response but let dispatch run first and discard its result | **exit `0`, all 360 tests pass** | **none** — see advice A1 |

## Checks at `2e79204bd`

| Command | Exit | Counts |
|---|---|---|
| `cargo test -p assay-mcp-server` | `0` | 360 passed, 0 failed, 0 ignored, across 32 test binaries; `protocol_negotiation_e2e` 11 passed |
| `cargo fmt --all -- --check` | `0` | no output |
| `cargo clippy -p assay-mcp-server --all-targets -- -D warnings` | `0` | 0 warning lines |

The deprecation is **not** silenced too broadly. `lib.rs` carries only `#![deny(unsafe_code)]`; there is no crate-wide `#![allow(deprecated)]`. The only `allow` touching this constant is the item-level `#[allow(deprecated)]` on the single use site at `protocol_negotiation_e2e.rs:16`. Three file-scoped `#![allow(deprecated)]` exist in `src/config.rs`, `src/auth/mod.rs` and `tests/auth_integration.rs`; all three are present verbatim at `v5.0.0`, are lexically scoped to those files, and cannot reach `server.rs`.

**On the 137-column `note` literal:** not a finding. `cargo fmt --check` passes because rustfmt does not reflow string literals, and `#[deprecated(note = …)]` takes a literal so `concat!` is unavailable — the builder's account of the mechanism is correct. There is no line-length gate on Rust in CI, and `crates/**/*.rs` already contains 2010 lines over 100 columns, 91 of them in this crate's own `src/`. Splitting the note is not possible without changing what rustc prints, and a self-contained note is worth more than a column count that nothing enforces.

### CI as of my measurement, 2026-08-10T21:30Z

Bound to `2e79204bd`: 53 check-runs — 47 `success`, 5 `skipped` (`Archive closed PR branch with Bencher`, `Integration Parity (Optional)`, `Wave 0.1 nightly safety`, both `eBPF monitor smoke` lanes), 1 `cancelled` (`Wave 0 feature matrix`, cancelled by the superseding push, so neither a pass nor a failure). Nothing failed. CI is re-running on `0710813d` and was still in progress when I measured; I report that as in progress rather than as a result.

### Classifier dispositions, in their own words

- `lane-check/proof` on `2e79204bd`: state `success`, description **"no delegated runner proof required"**. That is the absence of a requirement. It is not positive evidence about any host, and I claim nothing about any host from it.
- `host-capability-check` on `2e79204bd`: check-run, conclusion `success`, with `output.title` and `output.summary` both `null` — so it recorded no words of its own beyond the conclusion.

### Automated reviewers are not reviews here

- `CodeRabbit` commit status on `2e79204bd`: state `success`, description **"Review rate limited"**. Per AGENTS.md, a check reporting `pass` beside a rate-limit message is unavailable infrastructure wearing the shape of a verdict. It leaves no findings and no reviewer, and I have not counted it.
- `copilot-pull-request-reviewer[bot]`: `COMMENTED` on the superseded head `95950a45f`, body "Copilot was unable to review this pull request because the user who requested the review has reached their quota limit." Same category, and bound to a head that no longer exists on this branch.

Neither is part of the quorum, and their absence needs no substitute.

---

## Post-MVP advice, with dispositions. No issues opened — creation is frozen.

**A1 — "refuse before dispatch" is correct but unpinned.** The mutation that keeps the refusal response while letting dispatch run first and discarding its result passes all 360 tests. I confirmed it is genuinely observable and not a no-op: under that mutation the same refused `tools/call` emits 4 `tool_call_start`/`assay.usage.metered`/`tool_decision`/`tool_call_done` events and actually executes the tool, while the wire response stays byte-identical. Production behaviour at this SHA is correct (R7a), so this is a test-coverage gap and not a defect. *Disposition:* not blocking; the ordering is verified correct here by direct observation. If it is pinned later, the cheapest form is an assertion in `protocol_negotiation_e2e.rs` that a refused `tools/call` produces no `tool_call_start` or `assay.usage.metered` on stderr — the `run_session` harness already captures stderr, so it is a few lines rather than new machinery. This is the same gap a prior reviewer flagged; it survives the delta unchanged.

**A2 — only prose keeps the compatibility test in `tests/`.** You asked me to say so if that were the case, and it is. Nothing mechanical prevents a future editor moving `deprecated_modern_protocol_version_stays_publicly_reachable` into `server.rs`, and the move would look harmless because every test would still pass — while silently retiring the `E0603` half of the contract, as my in-module control demonstrates. *Disposition:* accept for MVP. The doc comment states the constraint precisely, names the separate-crate mechanism, and says which half of the contract the move would lose, which is as much as prose can do. Mechanising it would mean a gate asserting the test's file location, which costs more than it saves at this size.

**A3 — the semver gate does not cover this crate.** `Wave 0 semver checks (public crates)` is green on this SHA, but its allowlist is six crates and `assay-mcp-server` is not among them, so that green is not evidence about this constant. The compatibility test is the only thing standing between a published `5.0.0` API item and silent removal. *Disposition:* record only. Widening the allowlist is a CI-scope decision that belongs outside a protocol-honesty PR.

**A4 — behavioural compatibility versus released `v5.0.0` does change, deliberately.** `v5.0.0` shipped a `server/discover` arm returning a result advertising `2026-07-28`, and accepted `_meta: 2026-07-28`; at this SHA both change. That is the point of the PR and is required by ADR-042 — an advertised capability with no implementation behind it is exactly an unearned claim — and source compatibility, which is what the crate version governs, is preserved. *Disposition:* no code change. Worth one CHANGELOG line at release time so the withdrawal of an advertised-but-unimplemented capability is recorded where a consumer of the binary would look; this PR adds none.

**A5 — one pre-existing doc heading is the nearest thing to a residual claim.** `docs/guides/editor-mcp-recipe.md:157`, "## Remote servers (provisional, MCP 2026-07-28)", followed by "This section is provisional against the release candidate and will be finalised once the spec is final". Read closely it is advice about aligning the *wrapped upstream* server to OAuth 2.1, not a claim that `assay-mcp-server` implements the revision, and the section is untouched by this PR. But the heading alone could be read as provisional Assay support, and "will be finalised once the spec is final" is the forward-looking phrasing AGENTS.md discourages in public strings. *Disposition:* out of scope for this PR; for the doc owner. Not blocking, because it does not describe this crate's protocol behaviour and predates this branch.

**A6 — a non-string `_meta` protocol version is served as if unstated.** `{"_meta":{"io.modelcontextprotocol/protocolVersion":2026}}` is accepted and served, because the extraction chain ends in `and_then(Value::as_str)` and a non-string collapses to "no revision declared". Pre-existing, unchanged by this PR, and low risk given `initialize` handles ill-typed input with a value-free `-32602`. *Disposition:* no action; noting it so a future reader does not mistake it for a regression introduced here.

---

## Environment provenance

- Reviewed SHA: `2e79204bd6da3c1bfde2b84def36398270983375` (branch `codex/2157-mcp-version-honesty`, base `main`, merge-base `4a9d4d0a9b1c8ca27324c4b319e1a9fc015587a1`).
- Worktree: `/tmp/rev-2267b`, detached, created from that SHA, its own `target/`, no `CARGO_TARGET_DIR` exported. Removed after review. Downstream probe at `/tmp/downstream-probe`, outside the workspace.
- Host: macOS 26.6 (build 25G72), Darwin 25.6.0, `arm64` / `aarch64-apple-darwin`.
- Toolchain: `rustc 1.96.0 (ac68faa20 2026-05-25)`, `cargo 1.96.0 (30a34c682 2026-05-25)`.
- Profile: `dev` (unoptimized + debuginfo). Binary under test: `/tmp/rev-2267b/target/debug/assay-mcp-server`, built from `2e79204bd`.
- Measurement window: 2026-08-10T21:16Z–21:31Z.
- Neither `/Users/roelschuurkes/wt-2157` nor `/Users/roelschuurkes/assay` was built in or modified.

Rul1an:
## Review carry: recorded equivalence check

The independent READY verdict was measured on `2e79204bd6da3c1bfde2b84def36398270983375`. Branch protection is strict, so the already-merged MVP slices on `main` were brought in without rebasing or conflict resolution, producing `0710813d24995ea6a7f795950154c500f6ee3f44`.

Per `AGENTS.md`, both review-carry conditions were measured and hold.

### Condition 1: whole-tree equivalence

The merge commit has exactly the reviewed head and current `main` as parents:

```
git show -s --format='%P' 0710813d24995ea6a7f795950154c500f6ee3f44
2e79204bd6da3c1bfde2b84def36398270983375 367ccfc61e8e0fe00d57fcb3db736f6408b15297
```

The expected conflict-free merge tree and actual head tree are identical:

```
git merge-tree --write-tree 2e79204bd6da3c1bfde2b84def36398270983375 367ccfc61e8e0fe00d57fcb3db736f6408b15297
7c34b2af51573f8f781eb9801b6cca9400f55f26

git rev-parse 0710813d24995ea6a7f795950154c500f6ee3f44^{tree}
7c34b2af51573f8f781eb9801b6cca9400f55f26
```

### Condition 2: reviewed-file intersection

The merge base before the upstream advance was `4a9d4d0a9b1c8ca27324c4b319e1a9fc015587a1`. The reviewed PR covered exactly:

- `crates/assay-mcp-server/src/server.rs`
- `crates/assay-mcp-server/tests/protocol_negotiation_e2e.rs`

The advance from that base to `367ccfc61e8e0fe00d57fcb3db736f6408b15297` touched 22 files. The sorted intersection with the two reviewed files is empty.

### Disposition

Both conditions hold. The READY verdict on `2e79204bd6da3c1bfde2b84def36398270983375` therefore carries to final PR head `0710813d24995ea6a7f795950154c500f6ee3f44`.

PR #2247, opened 2026-08-10
fix(cli): decide the doctor exit class once, and measure the preflight success row on a config

Rul1an:
Closes #2215. Closes #2210.

Two separable commits on one branch because both land on the same `Ok(0)` in
`doctor/implementation.rs`; as separate PRs they conflict by construction. A
third commit fixes the divergence the first two created, and a merge commit
resolves #2240.

## 1. The exit class depended on a formatting flag (#2215)

Measured on `c2c41b038` (base), debug build, macOS aarch64, in a tree from
`assay init --preset dev --hello-trace`:

```text
$ assay doctor --config eval.yaml --trace-file traces/nope.jsonl
exit: 1
$ assay doctor --format json --config eval.yaml --trace-file traces/nope.jsonl
exit: 0
data_diagnostics[0]: {"severity":"error","code":"E_PATH_NOT_FOUND", ...}
```

The text branch computed `has_errors`; the JSON branch returned `Ok(0)`
unconditionally once the config loaded. One command, two answers about one tree
— and the channel reporting the clean result is the one an agent reads.

Both channels now call `crate::cli::helpers::decide_exit`, the same
registry-backed function `assay validate` and `assay run` use. It keys on the
ADR-046 class table rather than on severity alone, so a config-class code is
`2`. An earlier revision of this branch introduced a new `exit_class` function
for the two channels to share; review found that was a third answer to a
question `decide_exit` already owned, and it was deleted. **This changes the
text channel from `1` to `2`, which is a behaviour change beyond what #2215
asked for** — #2215 treats the text channel's `1` as correct. The alternative
was to keep `1` and document why doctor disagrees with the two commands
answering the same question, which preserves a divergence by adding a rule.

**The negative control is half the change.** "Return non-zero when any
diagnostic exists" would pass the first test and be wrong. A tree whose only
diagnostic is `warn` (`E_TRACE_INVALID`, legacy `function_call` shape) must
still exit `0` on both channels. That test was green before this change and
stays green after, so it constrains the fix rather than following it.

## 2. Then the exit class depended on `--fix` — and that divergence was introduced here

`doctor/fixes.rs` recomputed the error-severity predicate by hand and turned the
count into a literal `1` at three returns. On one tree with one
`E_PATH_NOT_FOUND`:

```text
$ assay doctor                 --config eval.yaml --trace-file traces/absent.jsonl   # 2
$ assay doctor --fix --dry-run --config eval.yaml --trace-file traces/absent.jsonl   # 1
```

[...]

Issue #2215, opened 2026-08-10
doctor --format json exits 0 while publishing error-severity diagnostics


Rul1an:
Parent: #2177 · Found reviewing #2160 (PR #2207)

## Driven observation

`assay doctor --format json` returns exit `0` while publishing an error-severity diagnostic. The text path returns `1` on the same input. Measured on `26f223267a2427017cdf4cf26c455dd0e2811ea7`, debug build, macOS, `ASSAY_EXIT_CODES` unset:

```text
$ assay init --preset dev --hello-trace     # a config that loads
$ assay doctor --config eval.yaml --trace-file traces/nope.jsonl
exit: 1

$ assay doctor --format json --config eval.yaml --trace-file traces/nope.jsonl
exit: 0
data_diagnostics[0]: {"severity":"error","code":"E_PATH_NOT_FOUND",
                      "message":"Trace file not found: traces/nope.jsonl"}
```

## Why this is a defect and not a preference

Two rules in `AGENTS.md` bear on it, and the second is the sharper one:

1. The exit class depends on a *formatting* flag. `crates/assay-cli/src/cli/commands/doctor/implementation.rs:104` returns `Ok(0)` unconditionally once the config loads, while the text branch at `:152-156` computes `has_errors` from `normalize_severity` and returns `1`. One command, two answers about the same tree.
2. **A failed validation is published as a clean result.** That is the rule against turning absence of evidence or a failed check into a clean result, and it bites hardest here because the JSON channel is the one an agent reads. An agent driving the golden path sees exit `0`, treats preflight as passed, and proceeds — while the report it just received says an input it needs does not exist.

The defect is pre-existing: it is not introduced by #2160. It is filed now because #2160 rewrote exactly this contract row — the golden-path artifact publishes `Success 0` for the JSON preflight outcome — so the shipped contract now implies, more clearly than before, that exit `0` means there was nothing wrong. PR #2207 adds a non-claim recording that it does not, which stops the artifact from lying but does not make the exit truthful.

## Why MVP

The agent golden path's whole value is that a machine can read the outcome. A preflight step whose success code does not mean success defeats the step, and it does so silently — there is no error to notice, only a `0`. Everything downstream of preflight in the golden path inherits it.

## Acceptance

- `doctor --format json` and `doctor` (text) return the same exit class for the same tree, derived from one function rather than computed twice. Per one-rule-one-function, the `has_errors` decision moves to a single place both branches call.
[...]

Issue #2210, opened 2026-08-10
doctor --format json reports success for a preflight that examined no config


Rul1an:
Origin: #2160 · Related: #2207

Pre-existing. Not introduced by PR #2207; the PR made it visible by pinning the surrounding
contract.

## Deliberately not filed under #2177

Epic #2177 is about the agent *error* envelope — commands whose failures write `fatal:` to stderr
instead of publishing a `reason_code` and `next_step`. This is the mirror defect: a *success* report
that overclaims. Nothing here fails, and no error envelope is missing. Folding it into #2177 would
widen that epic from "failures are illegible" to "outcomes are inaccurate", so it is filed
standalone under `cluster:stdout-contract`, whose definition it matches literally: "a command's
machine-readable output is missing, empty, or reports success it did not establish".

## Observed

Verified statically, by reading control flow at `e931cb285493d8070120e3080c7eefcb992e70d1`
(branch `cursor/2160-doctor-json-contract`, PR #2207). Nothing was executed, so the JSON document
below is derived from the insertion sites, not captured from a run.

With no `--config` and no `eval.yaml` in the working directory,
`crates/assay-cli/src/cli/commands/doctor/implementation.rs:45-52` takes the `else` arm and yields
`(None, None)` — no config, no config error. The JSON branch at `implementation.rs:54-104` then
inserts `generated_at` and the system probe, skips the `cfg_err` block (nothing to report) and
skips the `cfg` block (nothing to check), and returns `Ok(0)` at `implementation.rs:104`.

The text branch is honest about this. `implementation.rs:159` prints
`Policy Check: SKIPPED (No config found; run inside project or use --config)` before its `Ok(0)`.
The JSON document carries no equivalent marker: no `policy_check`, no `skipped`, no absent-config
field. So `doctor --format json` returns the same exit code and the same document shape for "config
examined and clean" as for "no config examined", and a consumer cannot tell which happened.

This is measured as the contract's *success* outcome. `crates/assay-cli/tests/agent_golden_path_contract.rs:237-241`
establishes the preflight `success` row on exactly that run — a fresh `tempfile::tempdir()` with no
config in it, argv `["assay","doctor","--format","json"]` — and asserts exit and stdout shape. The
published `Success 0` for step 2 of the golden path is therefore established by a run in which zero
config validation occurred.

## Why this matters

`AGENTS.md` states the invariant directly:

> Never turn absence of evidence, failed validation, skipped review, or unavailable infrastructure
> into a clean result.

[...]

PR #2268, opened 2026-08-10
fix(core): reject malformed tool pin hashes at load

### LINKED ISSUE #2248: ToolIdentity::fingerprint slices [0..8] off an unvalidated tool_pins hash, so a truncated pin aborts the proxy while reporting drift
Found while reviewing PR #2239 (non-building review). **Pre-existing and not caused by that PR** — `fingerprint()` is untouched by its diff — so it was dispositioned as a follow-up rather than fixed there.

`ToolIdentity::fingerprint()` slices a fixed byte range off a field that arrives from operator-authored YAML with no validation:

```
crates/assay-core/src/mcp/identity.rs:38-45
    pub fn fingerprint(&self) -> String {
        format!("{}:{}:{}", self.server_id, self.tool_name, &self.schema_hash[0..8])
    }
```

`ToolIdentity` derives `Deserialize`, and `McpPolicy.tool_pins` is `HashMap<String, ToolIdentity>` with `#[serde(default)]` and no validator — the only place `tool_pins` appears in `model/validation.rs` is the `args_valid` metric's UNENFORCED list, which is a different policy shape and bails rather than validating this one. So `schema_hash` can be any string an operator writes, including one shorter than 8 bytes, or one whose byte index 8 falls inside a multi-byte character.

The reachable path is the deny path specifically. `tool_drift_decision` calls `fingerprint()` only *after* it has decided the runtime identity differs from the pin:

```
crates/assay-core/src/mcp/policy/engine_next/fail_closed.rs:11-24
    let pinned = policy.tool_pins.get(tool_name)?;
    let runtime = runtime_identity?;
    if pinned == runtime { return None; }
    ... runtime.fingerprint(), pinned.fingerprint()
```

So a truncated pin does not fail at load, and does not fail while calls are allowed. It aborts the proxy at the moment it is reporting a tool-integrity drift — the worst moment to lose the process, and `panic = "abort"` is set for both `dev` and `release` in the workspace `Cargo.toml`, so there is no unwind.

## Reproduction

Probe placed at `crates/assay-core/tests/`, run, then deleted. Measured on `4ca3b122e819022313562296fe7f6a8e5aa5999c` (PR #2239 head), reviewer worktree `/Users/roelschuurkes/rev-2239`, `cargo 1.96.0`. It parses a policy YAML with a two-character pin and calls `McpPolicy::evaluate` with a runtime identity that does not match:

```yaml
tool_pins:
  read_file:
    server_id: "srv"
    tool_name: "read_file"
    schema_hash: "ab"
    meta_hash: "cd"
```

```
PROBE parsed pin = Some(ToolIdentity { server_id: "srv", tool_name: "read_file", schema_hash: "ab", meta_hash: "cd" })

thread 'probe_short_pin_hash_panics_on_the_deny_path' panicked at crates/assay-core/src/mcp/identity.rs:43:30:
end byte index 8 is out of bounds for string of length 2
```

Run under `CARGO_PROFILE_TEST_PANIC=unwind` only so `catch_unwind` could observe it; under the shipped profile this is an abort.

The documented example makes the mistake easy to reach without malice — `docs/reference/config/policies.md:266` shows `schema_hash: "9f4d4d0f..."`, an elided value. That literal happens to be 11 bytes and survives, but it invites an operator to paste a truncated hash, and a 7-character one would not.

## What I am not claiming

- Not a bypass. It denies nothing that would otherwise be allowed; it aborts on a path that had already decided to deny. The cost is availability and a lost drift report, not an admitted call.
- Not attacker-reachable from the wire. `schema_hash` on the runtime side is always 64 hex characters from `compute_json_hash`; only the operator-authored pin can be malformed.
- I did not survey what other `Deserialize`d fields in `McpPolicy` are sliced or indexed without validation. This is one instance, found while reading one function.

## Direction of a fix (not prescriptive)

Validate the pin at deserialization — a `schema_hash`/`meta_hash` that is not 64 lowercase hex characters is not a hash, and rejecting it at load turns an abort during a drift report into a policy-load error naming the field. A `chars().take(8)` in `fingerprint()` would remove the panic but keep an unvalidated pin that can never match, which is the quieter failure.

Rul1an:
## Summary
- reject tool pin hashes unless they are exactly 64 lowercase hexadecimal characters
- apply one validator to both `schema_hash` and `meta_hash`
- name the exact malformed `tool_pins.<name>.<field>` in the load error
- replace the misleading truncated hashes in the policy reference with valid-length examples

Fixes #2248.

## Why
Operator-authored YAML could carry a short or non-ASCII pin. The policy loaded successfully and later aborted at `schema_hash[0..8]` precisely while reporting tool drift. This moves the failure to policy load as an actionable configuration error.

## TDD
RED before production change:
- `cargo test -p assay-core --test tool_pin_validation`
- failed because `McpPolicy::from_file` accepted `schema_hash: ab`

GREEN on exact head `9a8753cf724c428cb7084f9eca449401db040bd6`:
- `cargo test -p assay-core --test tool_pin_validation --test config_reference_examples_parse`
- `cargo test -p assay-core`
- `cargo clippy -p assay-core --all-targets -- -D warnings`
- `git diff --check`

## Driven operator proof
`cargo run -q -p assay-cli -- policy validate --input <bad-pin.yaml>` now exits `2` before runtime and reports:

```text
fatal: failed to load policy <path>

Caused by:
    tool_pins.read_file.schema_hash must be exactly 64 lowercase hexadecimal characters
```

## Scope / non-claims
- no change to tool drift decisions or wire behavior
- no bypass finding; this is availability and diagnostics hardening
- no broader survey of deserialized policy fields

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool-pin policies now reject invalid schema and metadata hashes, including short or uppercase values.
  * Validation errors identify the specific invalid hash field.

* **Documentation**
  * Updated policy configuration examples to show valid 64-character lowercase SHA-256 hashes.

* **Tests**
  * Added coverage for invalid and valid tool-pin hash formats.


Rul1an:
## Independent non-building review — PR #2268

**READY** — bound to full head SHA `9a8753cf724c428cb7084f9eca449401db040bd6`.

I authored neither this change nor any specification or plan governing it, and I made no
mutation to `codex/2248-validate-tool-pin-hashes` or any other branch. All measurement below
was taken in a throwaway detached worktree created from that exact SHA and removed afterwards.

### Environment provenance

| | |
|---|---|
| Reviewed tree | `9a8753cf724c428cb7084f9eca449401db040bd6` (confirmed against `git ls-remote` before and after) |
| Worktree | `/tmp/review-2268`, detached, own `target/`, removed at end |
| OS | Darwin 25.6.0, `arm64` (Apple M1 Pro, `xnu-12377.161.13~4`) |
| Toolchain | `rustc 1.96.0 (ac68faa20 2026-05-25)` / `cargo 1.96.0 (30a34c682 2026-05-25)` |
| Profile | `test`/`dev` profile, unoptimized + debuginfo. `panic = "abort"` is set on both `[profile.dev]` and `[profile.release]` in the root `Cargo.toml`; Cargo ignores it for the test profile, so the mutation runs below unwound rather than aborted. The shipped binary aborts. Stated because it bounds what a test can observe about the original defect, not because it changes any verdict here. |
| CLI binary | `target/debug/assay` built from that SHA in that worktree |

Tree was pristine at the reviewed SHA when `fmt`/`clippy`/the full suite were measured
(`git status --porcelain` and `git diff --stat HEAD` both empty after every mutation was reverted;
`legacy.rs` restored to `sha256 ec988bdb04491c0d9b5922d32378ced0cff575ab810f62425886fd1dad77b89b`).

### Blocking findings

None. Every one of the seven properties I set out to break held under direct exercise of the
real load path.

---

### 1. Malformed operator-authored hashes are rejected at load, naming the exact field

Driven through `McpPolicy::from_file` — the real load path, not the diff — with 21 hostile values
per field, each run twice (once with the other field held valid). Every rejection named the
failing field by its exact `tool_pins.<name>.<field>` path, and never the other field:

| Input | bytes/chars | `schema_hash` | `meta_hash` |
|---|---|---|---|
| 64 lowercase hex | 64/64 | accepted | accepted |
| all-`f` 64, all-`0` 64 | 64/64 | accepted | accepted |
| 63 chars | 63/63 | rejected, field named | rejected, field named |
| 65 chars | 65/65 | rejected, field named | rejected, field named |
| **7 chars (`abcdef1`) — the original `[0..8]` abort trigger** | 7/7 | rejected, field named | rejected, field named |
| 8 chars exactly | 8/8 | rejected, field named | rejected, field named |
| empty string | 0/0 | rejected, field named | rejected, field named |
| 64 chars, one non-hex `g` | 64/64 | rejected, field named | rejected, field named |
| 64 chars all uppercase | 64/64 | rejected, field named | rejected, field named |
| 64 chars mixed case | 64/64 | rejected, field named | rejected, field named |
| leading space + 63 hex | 64/64 | rejected, field named | rejected, field named |
| 63 hex + trailing space | 64/64 | rejected, field named | rejected, field named |
| valid 64 hex + trailing space | 65/65 | rejected, field named | rejected, field named |
| tab-padded | 64/64 | rejected, field named | rejected, field named |
| Cyrillic а U+0430 ×32 | 64/**32** | rejected, field named | rejected, field named |
| fullwidth digit U+FF10 ×21 + `a` | 64/**22** | rejected, field named | rejected, field named |
| 63 ASCII hex with one U+0430 homoglyph | 64/**63** | rejected, field named | rejected, field named |
| `sha256:`-prefixed valid hex | 71/71 | rejected, field named | rejected, field named |
| `0x`-prefixed | 64/64 | rejected, field named | rejected, field named |
| leading underscore | 64/64 | rejected, field named | rejected, field named |

The Unicode rows are the ones worth calling out: `value.len()` is a **byte** length, so all three
homoglyph payloads are 64 bytes while being 32, 22 and 63 characters. They are rejected by the
`bytes().all(...)` ASCII predicate rather than by the length check, which is the correct order —
a `chars().count() == 64` check would have admitted the 64-character homoglyph string.

The diagnostic is actionable rather than merely correct. Driving the real operator surface:

```
$ ./target/debug/assay policy validate --input /tmp/bad-pin.yaml
fatal: failed to load policy /tmp/bad-pin.yaml

Caused by:
    tool_pins.read_file.schema_hash must be exactly 64 lowercase hexadecimal characters
$ echo $?
2
```

That reproduces the PR body's claim exactly, including the exit code (`2` = config error).
With **both** fields invalid in one pin, `schema_hash` is reported — one field, deterministically,
so an operator with both fields set is never handed an ambiguous "invalid hash".

A missing field is also a clean error rather than a defaulted empty string:
`tool_pins.read_file: missing field 'meta_hash' at line 3 column 5`.

### 2. Valid input still loads

`9f4d4d0f…` (the new docs value), `f`×64 and `0`×64 all load for both fields. A policy with no
`tool_pins` at all still loads. The full `assay-core` suite confirms no fixture regressed.

### 3. One validation rule serves both fields

Satisfied, and by construction rather than by coincidence: `validate_sha256_hex(value, field)` is
a single function, called twice with different labels. There is no second implementation and
therefore no parity test is needed, which is the outcome "One rule, one function" prefers over the
sanctioned fallback.

I also probed the copy-paste failure this shape invites — passing `&pin.schema_hash` while labelling
it `meta_hash`. See mutation H below: the suite catches it.

### 4. Deterministic error order — with the actual collection type

The collection is `HashMap<String, ToolIdentity>` (`crates/assay-core/src/mcp/policy/types.rs:53`),
so the determinism risk was real rather than theoretical. It is closed by `pin_names.sort()` before
iteration.

I did not take one run as evidence. With **ten** invalid pins inserted in an order unrelated to sort
order, over **200** loads in-process: **1 distinct error**, always
`tool_pins.alpha_tool.schema_hash …` — the lexicographically first pin.

The guard is load-bearing, which I confirmed by removing it: the same 200 loads then produced
**10 distinct errors**, cycling through `beta_tool`, `delta_tool`, `cobra_tool`, `yak_tool`,
`alpha_tool`, `mango_tool`, `echo_tool`, `nimbus_tool`, `zebra_tool`, `kiwi_tool`. So `sort()` is
doing real work, not decorating an already-stable iteration. See observation A for the consequence.

### 5. The copied docs example is a valid length

Counted programmatically out of `docs/reference/config/policies.md` at the reviewed SHA rather than
by eye:

- `schema_hash`: **64 characters, 64 bytes, all lowercase hex** — `9f4d4d0f` ×8
- `meta_hash`: **64 characters, 64 bytes, all lowercase hex** — `42f5df3e` ×8

Then verified end-to-end rather than inferred: the example block copied verbatim into a policy file
gives `✔ Policy OK`, exit `0`.

This docs change is load-bearing, not cosmetic. The **pre-PR** text, copied verbatim, is now
rejected:

```
$ ./target/debug/assay policy validate --input /tmp/old-docs-pin.yaml   # schema_hash: "9f4d4d0f..."
Caused by:
    tool_pins.read_file.schema_hash must be exactly 64 lowercase hexadecimal characters
$ echo $?
2
```

Shipping the validator without the docs fix would have made the reference page a trap. Doing both in
one commit is the right coupling.

### 6. No drift-decision or wire behaviour changed

The production diff is 22 lines confined to `legacy::validate()` plus one new private function. It is
a pure pre-check that either returns `Ok(())` or bails, so for any policy that loads, every downstream
decision is bit-identical. `tool_drift_decision` in `engine_next/fail_closed.rs`, the decision
emitters and the wire types are untouched.

Verified rather than reasoned: full `cargo test -p assay-core` is green (40 test binaries,
913 passing in the lib target alone, **0 failed, 0 flaky**), plus `cargo fmt --all -- --check` clean
and `cargo clippy -p assay-core --all-targets -- -D warnings` clean on the pristine tree.

I also checked for a **bypass** of the new validation, since a second load path would leave the abort
reachable. Every `serde_json::from_value::<McpPolicy>` / `serde_yaml::from_str::<McpPolicy>` site in
the workspace is inside `#[cfg(test)]` (`otel/projection.rs:374`, `mcp/proxy/client.rs:345`,
`project_otel.rs:147`, `mcp/tdt.rs:181`, and the `tests/` files). Every operator-facing entry —
`policy validate`, `policy migrate`, `policy fmt`, `mcp`, `monitor_next`, `discover`, and
`assay-mcp-server`'s `tools/check_args.rs` — goes through `McpPolicy::from_file`, which now calls
`validate()`. `model/validation.rs` independently refuses `tool_pins` inside an `args_valid` policy
as unenforced, so that is not a side door either.

One genuine behaviour change, which I read as correct and intended: a policy carrying a malformed pin
for a tool that is never invoked used to load and run. It now refuses at load. Any pin that *was*
exercised already aborted, because a truncated hash can never equal a runtime hash, so drift was
guaranteed and `fingerprint()` was guaranteed to slice out of bounds.

### 7. The test fails if either field's validation is removed — both directions

Two-sided, as required. Each mutation applied in isolation to `legacy.rs` in my `/tmp` worktree, with
the file restored from a checksummed backup between runs, and only the PR's own suite
(`cargo test -p assay-core --test tool_pin_validation`) used as the control:

| # | Mutation | PR suite | Notes |
|---|---|---|---|
| **A** | **`schema_hash` validation removed** (only `meta_hash` validated) | **RED** | `malformed_… ` fails at `tool_pin_validation.rs:18`, dumping the policy that wrongly loaded with `schema_hash: "ab"` |
| **B** | **`meta_hash` validation removed** (only `schema_hash` validated) | **RED** | same test fails, dumping the policy that wrongly loaded with `meta_hash: "AAAA…"` |
| C | Weakened: `is_ascii_hexdigit()` — uppercase hex accepted | **RED** | caught by the uppercase-`meta_hash` case |
| D | Weakened: `!value.is_empty()` only | **RED** | |
| E | Weakened: `len() == 64` only, charset dropped | **RED** | caught by the uppercase case; the length case alone would not have caught this |
| F | Weakened: charset only, length dropped | **RED** | caught by the 2-char case |
| H | Copy-paste: `&pin.schema_hash` passed under the `meta_hash` label | **RED** | the "One rule, one function" failure mode this shape invites is caught |

`valid_tool_pin_hashes_load` stayed green throughout, so the suite is not passing by rejecting
everything. Both removal directions and every weakening I could construct are caught. The two
assertions pull genuinely different weight: the short-`schema_hash` case pins length, the
uppercase-`meta_hash` case pins the charset, and neither substitutes for the other (see D/E/F).

What is **not** caught: mutation I, below.

### Adjacent judgement: does `panic = abort` leave another reachable slice here?

I audited every slice and index expression in the `mcp` module for one reachable from
operator-authored input. Four sites, and after this change none is reachable with a bad index:

- `identity.rs:43` — `&self.schema_hash[0..8]`, the #2248 site. Now guarded on both sides: the pinned
  identity by this PR's load-time check, the runtime identity because `ToolIdentity::new` always
  produces `hex::encode(sha256)`. It is reached only on the drift branch, since `pinned == runtime`
  short-circuits first — which matches the issue's report that the abort fired "precisely while
  reporting tool drift".
- `g3_auth_context.rs:37` — `b[..7]`, guarded by `b.len() >= 7` on the same line.
- `proxy.rs:161` — `&s[..pos]`, `pos` comes from `find("://")` with a `pos == 0` bail.
- `signing.rs:441` — inside `#[cfg(test)]`.

`meta_hash` is never sliced anywhere, so validating it is defence-in-depth rather than
abort-prevention. Validating it anyway is the right call: it is the only way to have one rule for
both fields, and the docs promise the format for both.

### Delegated host-capability proof: not required

Reported from the classifier's own output at the reviewed SHA, not from my reading of the file list:

```
Assay-Runner lane check for PR #2268
Head SHA: 9a8753cf724c428cb7084f9eca449401db040bd6
Changed files: 3
Expected delegated gate: none
```

The `lane-check/proof` status reads `no delegated runner proof required`, and the
`host-capability-check` job log reads `PASS: no host-capability proof required for this PR.`

**This is the absence of a requirement and nothing more.** It is not evidence about any host, any
kernel, any eBPF or LSM capability, and it must not be read or cited as such. Care taken because the
earlier lines in the same `lane-check` log that say a proof pack *is* required belong to the
classifier's `--self-test` against fixture PR #1 at head `aaaa…aaaa`, not to this PR.

All other required checks are green at this SHA, including `Build + Test` on ubuntu, macos and
windows. Recorded for completeness, not as quorum: `CodeRabbit` reports `pass` with
`Review skipped` — per the contract that is no reviewer and no findings, and it needs no substitute.

---

## Post-MVP advice — non-blocking, with technical disposition

Recorded here rather than as issues; new issue creation is frozen. None of these makes the change at
this SHA wrong, and I would not hold the merge for any of them.

**A. The determinism guard is correct but unpinned by any test.** `pin_names.sort()` is production
behaviour added by this PR, and removing it leaves the PR's suite **fully green** (both tests pass)
while my ten-pin probe goes from 1 distinct error to 10. So the guard is free to be deleted by a
future edit — as dead-looking as a sort over keys can be — with CI staying green. This is the one gap
in an otherwise two-sided mutation control, and the only place where AGENTS.md's
"write a failing behavioral test before changing production behavior" is not met for a line that
ships. *Disposition:* the cheapest fix is one test with two invalid pins asserting the error names
the lexicographically-first one; it does not need 200 iterations to be useful, because a single
assertion on `alpha` versus `zebra` kills the unsorted mutant most of the time and always kills the
"iterate in insertion order" variant. Worth doing when the file is next touched. Not blocking:
the shipped behaviour is deterministic and I verified it directly.

**B. Nothing pins the docs example against the validator.** I checked the harness rather than
assuming: `config_reference_examples_parse.rs` only extracts blocks containing `- type:` and
deserialises them as `SequenceRule`, and `config_reference_parity.rs` covers only `sequences.md` and
`eval-yaml.md` headings. Neither reaches the `tool_pins` block in `policies.md`, and no test in the
workspace loads that block through `McpPolicy`. The PR body cites
`--test config_reference_examples_parse` as GREEN, which is true, but that test gives the changed
example **no coverage** — the example is correct today because it was written correctly, and it can
silently regress to the truncated form that I demonstrated exits `2`. *Disposition:* this is exactly
the "validator and the thing it validates against" pair from "One rule, one function", and the fix is
the same shape as the existing docs-parse test — walk `policies.md`'s yaml blocks and push any
carrying `tool_pins` through `McpPolicy::validate`. Post-MVP; the example is verified correct at this
SHA by direct CLI execution.

**C. "64 lowercase hex" is now answered in roughly nine places across the workspace.** Not introduced
by this PR, and the PR's own two call sites correctly share one function, so item 3 is satisfied. But
the predicate now exists as at least three spellings — `is_ascii_digit() || (b'a'..=b'f')` here,
`matches!(b, b'0'..=b'9' | b'a'..=b'f')` in `mcp/tool_decision_truth.rs:31` (same crate, same module
tree), and `is_ascii_hexdigit() && !is_ascii_uppercase()` in
`cli/commands/evidence/mcp_tunnel_observed.rs:497` — and they have **already drifted**:
`assay-evidence/src/trust_basis/classifiers.rs:270` uses bare `is_ascii_hexdigit()` and therefore
accepts uppercase where the others reject it. That is the documented failure mode, live in the tree.
*Disposition:* a shared `is_sha256_lower_hex` is a cross-crate refactor with its own review surface
and belongs in its own change, not smuggled into a one-line abort fix. Adding a tenth local spelling
here is the right trade for this PR; consolidating the set is the follow-up.

**D. No CHANGELOG entry for a config-breaking load refusal in a published crate.** `assay-core` sets
no `publish = false`, `McpPolicy::validate()` is `pub`, `[Unreleased]` is empty, and the header says
"All notable changes to this project will be documented in this file." The 5.0.0 entry documents the
identical shape — "**An assertion that cannot fail is now refused at load**" under `### Breaking`,
with a **Migration** line naming the command that reports the set before upgrading. This change has
the same shape and a concrete affected population, which I demonstrated rather than assumed: anyone
who copied the pre-PR reference example has a policy that stops loading. *Disposition:* one
`[Unreleased]` entry noting the refusal and pointing at `assay policy validate` as the pre-upgrade
check would match the established precedent exactly. Recorded as actionable so it gets a disposition
before auto-merge under the "every actionable finding is fixed or has a recorded technical
disposition" condition; I am not treating it as a defect in the code.

**E. Cosmetic: a pin name containing a dot makes the error path read oddly.** A pin keyed
`read.schema_hash` with a bad `meta_hash` reports
`tool_pins.read.schema_hash.meta_hash must be …`. It still names the right field and is a diagnostic
string, not a boundary. *Disposition:* not worth quoting or escaping the key; noted only so the next
reader of that message is not surprised.

---

### Cleanup

My `/tmp/review-2268` worktree, every mutation in it, and my reviewer-only probe test were removed —
the path no longer exists and it is gone from `git worktree list`. The other worktrees that listing
shows are not mine and I did not touch any of them. I made no commit, push, amend, branch deletion or
issue, did not touch the main checkout at `/Users/roelschuurkes/assay` (still at `cd8dc5d6` with its
one expected untracked audit file), and did not enable auto-merge or approve via the API. This
comment is the review record, bound to `9a8753cf724c428cb7084f9eca449401db040bd6`; a new push
invalidates it.

Rul1an:
## Review carry: recorded equivalence check

The independent READY verdict was measured on `9a8753cf724c428cb7084f9eca449401db040bd6`. Branch protection has `strict: true`, so `main` had to be brought in, producing a new head. Per AGENTS.md ("Review Quorum"), that review carries only with a recorded equivalence check meeting both conditions. Both were computed before the merge and both hold.

**Advance:** `main` moved from `4a9d4d0a9b1c8ca27324c4b319e1a9fc015587a1` to `d2097cd160dada963c5b81faf4864efbfebdfe74` (PR #2247, closing #2210 and #2215), touching 19 files.

**New head:** `32717f4b190a3e52cb8d97a229b003230ecae36d`, a merge commit whose parents are exactly the reviewed head `9a8753cf724c428cb7084f9eca449401db040bd6` and `d2097cd160dada963c5b81faf4864efbfebdfe74`.

### Condition 1 — the new head introduces no change, to any file, that is not already on `main`

Whole-tree comparison, not a file list:

```
git merge-tree --write-tree 9a8753cf72 d2097cd160  ->  36e0aa67b8b6fcc635a2cacd61164f75514d052e
git rev-parse 32717f4b19^{tree}                    ->  36e0aa67b8b6fcc635a2cacd61164f75514d052e
```

Identical. The new head's tree is exactly what merging `main` into the reviewed head produces without conflicts, so nothing rode in on the update.

### Condition 2 — the advance touched no file the review covered

Reviewed files, as of `9a8753cf72` (3):

- `crates/assay-core/src/mcp/policy/legacy.rs`
- `crates/assay-core/tests/tool_pin_validation.rs`
- `docs/reference/config/policies.md`

Intersection with the 19 files changed by the advance: **empty**. #2247 was confined to `assay-cli`, the golden-path generator and its generated artifacts, and `docs/reference/cli/doctor.md`; it did not touch `assay-core` or `docs/reference/config/policies.md`.

### Disposition

Both conditions hold, so the READY verdict on `9a8753cf724c428cb7084f9eca449401db040bd6` carries to `32717f4b190a3e52cb8d97a229b003230ecae36d`. No fresh review is required for this advance.

Recorded by the merge queue. Measured on the SHAs named above; `main` was read at `d2097cd160dada963c5b81faf4864efbfebdfe74`.

PR #2285, opened 2026-08-11
Include otel-mcp fixture lockfile in scheduled OSV targets

Rul1an:
## Summary
- Scheduled OSV discovery found one tracked non-Rust lockfile missing from `.github/osv/non-rust-manifests.txt`: `crates/assay-core/tests/fixtures/otel-mcp-ingest-v0/generator/package-lock.json`.
- Added that path (sorted) so discovery-vs-allowlist coverage is green again.
- One-line allowlist fix only; no exclusion mechanism.

## Non-claim
Coverage-green means every discovered `package-lock.json` / `requirements*.txt` is listed for scheduled scan. It does **not** mean the repository or the fixture is vulnerability-free.

## Base / head
- Base (exact): `161b968fb723d5ea794a9bda40913055567b9391`
- Head: `64d21b4d4a914e9757fdbfa963466075a2827be9`
- Branch: `cursor/fix-osv-fixture-target`
- Worktree: `/tmp/assay-wt-osv-fixture-target`

## RED evidence
Command (at base `161b968fb723d5ea794a9bda40913055567b9391` before the allowlist edit):

```bash
listed=$(grep -Ev '^[[:space:]]*(#|$)' .github/osv/non-rust-manifests.txt | LC_ALL=C sort -u)
discovered=$(git ls-files | grep -E '(^|/)(package-lock\.json|requirements[^/]*\.txt)$' | LC_ALL=C sort -u)
comm -13 <(printf '%s\n' \"$listed\") <(printf '%s\n' \"$discovered\")
```

Output:

```
crates/assay-core/tests/fixtures/otel-mcp-ingest-v0/generator/package-lock.json
```

## GREEN evidence
Same `comm -13` after the edit: empty (coverage green).

Also verified:
- every listed path is git-tracked and exists on disk
- no `Cargo.lock` in the allowlist
- list remains `LC_ALL=C` sorted
- `.github/workflows/osv-scanner-scheduled.yml` parses as YAML (Ruby `YAML.load_file`)
- `git diff --check` clean
- allowlist comments contain no vulnerability-free / safe-agent claim language

Local OSV scanner (v2.3.8 binary `osv-scanner_darwin_arm64`, lockfile-only — no `npm install`, no package scripts):

```
osv-scanner version: 2.3.8
Scanned .../generator/package-lock.json file and found 14 packages
No issues found
osv-scanner_exit=0
```

[...]

PR #2286, opened 2026-08-11
fix(test-support): kill remaining process tree before waiting on pipe EOF

Rul1an:
## Summary
- Once `try_wait` observes the direct child exit, terminate remaining Job Object / process-group members **before** treating reader EOF as the success condition; then drain buffered stdout/stderr to EOF within the same total deadline and return the preserved parent status.
- Replace PowerShell / env-controlled pid-file descendant fixtures with READY-record helpers (`READY pid=<u32>` / `READY solo`), a re-exec helper (Windows bounded child; Unix shell speaking the same protocol), and a solo control.
- Flip the inherited-output contract from `expect_err`/deadline to one cross-platform quiet-success contract for inherited and detached descendants. Addresses [#2249](https://github.com/Rul1an/assay/issues/2249). Does **not** close it (Windows CI evidence still required). Leaves #2253 measurement out of scope.

## Contract change
**Before:** success required `observed_status` **and** stdin/stdout/stderr reader EOF, then `finish_process_tree`. A descendant holding inherited pipes forced the deadline even when the parent had already exited 0 with READY bytes captured.

**After:** on `try_wait → Some(parent_status)`, if any I/O is still outstanding, call `terminate_remaining_tree` (accepts only `process_tree_absent` / ESRCH — **not** macOS EPERM before I/O completion), then continue draining readers within the same deadline, then `finish_process_tree` (post-IO quiescence path unchanged), return saved parent status. Timeout / output-ceiling / stdin-BrokenPipe fail-closed behavior unchanged.

## Base / head
- Base: `161b968fb723d5ea794a9bda40913055567b9391`
- Head: `73ddce963bd08a850fb1fe0122aadb4f12a3a66f`
- Branch: `cursor/fix-bounded-process-kill-before-eof`
- Worktree: `/tmp/assay-wt-bounded-kill-before-eof`

## RED evidence (Unix, production wait loop unchanged)
```
cargo test -p assay-mcp-server --test agent_golden_path_contract bounded_process::tests:: -- --nocapture
```
Failure (expected):
```
parent_exit_with_inherited_output_descendant_returns_ok_before_deadline ... FAILED
inherited output mutation: ... deadline of 250ms expired; status=exit status: 0;
stdout="READY pid=72765" (15 bytes); stderr="READY pid=72765" (15 bytes)
```
Solo control and quiet/detached already passed; only the inherited holder forced the deadline.

## GREEN commands (Unix local)
```
cargo test -p assay-mcp-server --test agent_golden_path_contract bounded_process::tests:: -- --nocapture
cargo test -p assay-cli --test agent_golden_path_contract bounded_process::tests:: -- --nocapture
# also: init_stdout_contract, doctor_config_check_contract, doctor_exit_class_contract,
#       policy_validate_machine_contract, config_recovery_argv_contract,
#       assay-mcp-server enforcement_sarif_cli
cargo test -p assay-mcp-server --test agent_golden_path_contract --test enforcement_sarif_cli
cargo test -p assay-cli --test agent_golden_path_contract --test init_stdout_contract \
  --test doctor_config_check_contract --test doctor_exit_class_contract \
  --test policy_validate_machine_contract --test config_recovery_argv_contract
# 20× descendant + solo filters on assay-mcp-server agent_golden_path_contract: all ok
cargo fmt --all -- --check
cargo clippy -p assay-mcp-server -p assay-cli --all-targets -- -D warnings
git diff --check
```

## Windows non-claim
Unix GREEN is recorded above. Windows Job Object behavior and CI `windows-latest` for this head are **not** claimed until CI runs on this PR. No Windows-expects-deadline split was introduced.

## Test plan
- [x] RED: inherited descendant fails with deadline + exit 0 + READY bytes (Unix)
- [x] GREEN: inherited + quiet + solo Ok across embedding binaries (Unix)
- [x] fmt / clippy `-D warnings` / `git diff --check`
- [ ] CI `windows-latest` on this head
- [ ] Non-building agent review quorum before merge

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when running time-limited processes, including cleaner handling of stalled or unresponsive processes.
  * More consistently captures delayed output from standard input and output streams.
  * Strengthened cleanup of child processes after failures or timeouts.
  * Improved behavior across Unix and Windows environments, including better handling of setup and termination failures.
* **Tests**
  * Expanded coverage for process cleanup, delayed output, readiness detection, and cross-platform behavior.

### DIFF
--- a/tests/support/bounded_process.rs
+++ b/tests/support/bounded_process.rs
@@ -135,7 +135,19 @@ pub fn run_bounded(
 
         if observed_status.is_none() {
             match child.try_wait() {
-                Ok(Some(status)) => observed_status = Some(status),
+                Ok(Some(status)) => {
+                    observed_status = Some(status);
+                    // Descendants that still hold inherited pipes keep readers
+                    // from EOF. Terminate the remaining tree now — before
+                    // treating reader EOF as the success condition — then drain
+                    // buffered output within the same total deadline.
+                    let io_outstanding = !stdin_writer.is_finished()
+                        || !stdout_reader.is_finished()
+                        || !stderr_reader.is_finished();
+                    if io_outstanding {
+                        terminate_remaining_tree(child.as_mut(), context, &command_display)?;
+                    }
+                }
                 Ok(None) => {}
                 Err(error) => {
                     let reap = terminate_and_reap(child.as_mut(), None, context, &command_display);
@@ -151,6 +163,10 @@ pub fn run_bounded(
             && stdout_reader.is_finished()
             && stderr_reader.is_finished()
         {
+            // After I/O completion this may accept post-IO quiescence (including
+            // macOS EPERM on an empty group) that terminate_remaining_tree does
+            // not. When the early path already killed the tree, this is a
+            // no-op or that quiescent acceptance.
             finish_process_tree(child.as_mut(), context, &command_display)?;
             break observed_status
                 .take()
@@ -325,6 +341,27 @@ fn finish_process_tree(
     }
 }
 
+/// Terminates descendants that outlived the direct child so inherited pipes can
+/// reach EOF. Only demonstrably absent trees are accepted here
+/// (`process_tree_absent`); macOS EPERM is not, because I/O may still be
+/// outstanding and that error is reserved for the post-IO finish path.
+fn terminate_remaining_tree(
+    child: &mut dyn ChildWrapper,
+    context: &str,
+    command: &str,
+) -> Result<(), String> {
+    match child.start_kill() {
+        Ok(()) => Ok(()),
+        Err(error) if process_tree_absent(&error) => Ok(()),
+        Err(error) => {
+            let direct_kill_error = child.inner_mut().start_kill().err();
+            Err(format!(
+                "{context}: failed to terminate remaining process tree for {command}: {error}; direct_kill={direct_kill_error:?}"
+            ))
+        }
+    }
+}
+
 #[cfg(unix)]
 fn process_tree_absent(error: &std::io::Error) -> bool {
     error.raw_os_error() == Some(libc::ESRCH)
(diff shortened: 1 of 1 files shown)


Rul1an:
## Findings

### [P1] Do not return while early tree-termination failures still own live I/O workers - `tests/support/bounded_process.rs:148`

When `terminate_remaining_tree` returns any error other than an absent Unix process group, the new `?` returns from `run_bounded` before the joins at lines 190-194. At that point `io_outstanding` is true by construction. Dropping the `JoinHandle`s detaches workers that may still be blocked on stdin or inherited stdout/stderr, and dropping the process wrapper does not provide the missing cleanup: in `process-wrap 9.1.0`, Unix `ProcessGroupChild::start_kill` is only `killpg(SIGKILL)` and has no drop kill, while Windows `JobObjectChild::start_kill` is `TerminateJobObject` and its Job Object is created with `kill_on_drop = false`. The fallback at line 357 calls `start_kill` only on the already-observed direct parent, so it cannot terminate the descendants responsible for inherited-pipe EOF.

This means the explicitly rejected macOS `EPERM` path, a Windows `TerminateJobObject` error, or another non-absent kill failure can return an error while leaking a live descendant and blocked threads, violating this helper's process-tree ownership and resource-bound contract. Preserve the termination error but keep cleanup bounded rather than propagating here immediately, and add an injected/fake tree-kill failure test that proves `run_bounded` cannot return with workers or descendants still live.

## Verdict

**NOT READY** pending resolution of the P1 above.

I am the single independent non-building reviewer for this record. I did not build this change and did not author or edit its specification or implementation plan. This review is bound to exact head **`73ddce963bd08a850fb1fe0122aadb4f12a3a66f`** against base **`161b968fb723d5ea794a9bda40913055567b9391`**. I reviewed the entire one-file diff, not only the PR description.

## Verification

- Read `AGENTS.md` and inspected all `544` changed lines in `tests/support/bounded_process.rs`.
- Read the local pinned `process-wrap 9.1.0` implementation for Unix process groups, Windows Job Objects, `try_wait`, `start_kill`, completion-port waits, and Job Object handle/drop behavior.
- Reconstructed RED outside the repository with base production code and head tests; hashes matched those source portions exactly. The stated MCP command failed only `parent_exit_with_inherited_output_descendant_returns_ok_before_deadline` at `250ms`, preserving parent exit `0` and matching READY records on stdout/stderr.
- Ran the focused MCP suite from an exact-head archive: `13 passed, 1 ignored`.
- Ran the focused CLI suite from the same exact-head archive: `13 passed, 1 ignored`.
- Ran `cargo fmt --all -- --check` and `cargo clippy -p assay-mcp-server -p assay-cli --all-targets -- -D warnings` successfully with `rustc 1.96.0` / `cargo 1.96.0`.
- Confirmed `git diff --check` clean. Package listings for `assay-cli` and `assay-mcp-server` do not include the changed workspace test-support file; publish-shape and distribution checks passed.
- Re-read GitHub state immediately before posting: PR head/base still match the SHAs above. Required checks are green. Exact-head Ubuntu, macOS, and Windows Build + Test jobs passed; Windows logs show inherited, detached, solo, early-stdin, stdout-overflow, and stderr-overflow tests passing in both embedding binaries. macOS logs show the descendant paths and EPERM classification test passing in both embedding binaries.
- CodeRabbit and Copilot records explicitly say they were rate/quota limited; they are not reviews and were not treated as quorum evidence.

Rul1an:
## Findings

### [P1] A persistent tree-kill error still returns before joining workers - `tests/support/bounded_process.rs:237`

The new deferred `tree_termination_error` fixes only the case where the first `terminate_remaining_tree` call fails and the deadline retry succeeds. If the real failure persists, the retry at lines 237-242 calls `terminate_and_reap(...)?`; `terminate_and_reap` returns `Err` when its own `child.start_kill()` fails, so that `?` exits `run_bounded` before the joins at lines 247-251. With the inherited-output descendant, the readers can still be blocked and their dropped `JoinHandle`s detach them.

The process-wrapper ownership semantics do not close this gap: `process-wrap 9.1.0`'s Unix `ProcessGroupChild::start_kill` is a `killpg(SIGKILL)` call with no drop kill, while Windows constructs the Job Object with `kill_on_drop = false`; the direct-child fallback targets an already-exited parent. Thus a persistent macOS `EPERM`, Windows `TerminateJobObject` failure, or another repeatable non-absent error retains the prior leak/early-return behavior.

The injected test is sensitive to restoring the original immediate `?` (it would observe live workers or fail to recover joined output), so it is not wholly tautological. However, `INJECT_TERMINATE_REMAINING_TREE_ERROR.take()` makes the fake one-shot: the deadline retry necessarily uses the real successful kill path. The test therefore cannot fail for the remaining mutation/branch where both termination attempts return errors. Make the failure injectable across the retry and ensure every return path performs bounded cleanup and joins workers before surfacing the preserved first error.

## Verdict

**NOT READY** pending resolution of the P1 above.

I am the single independent non-building reviewer for this record. I did not build this change and did not author or edit its specification or implementation plan. I made no file changes and did not push or merge. This review is bound to exact head **`5dc9a83d53eaec254eb47a074dc1872c1145d04f`**. I reviewed the entire one-file PR diff against its GitHub base **`161b968fb723d5ea794a9bda40913055567b9391`** and checked current `main` **`1a47918f815d3c865bc42fe54a80aff8b2662dbd`**; the only base-to-current-main file change is unrelated (`.github/osv/non-rust-manifests.txt`).

## Verification

- Read `AGENTS.md`, the prior P1 at issue comment `5250649270`, all changed code/tests in `tests/support/bounded_process.rs`, and the pinned `process-wrap 9.1.0` Unix process-group, Windows Job Object, and core child-wrapper implementations.
- Confirmed `cargo fmt --all -- --check` passes locally at the exact head with `rustc 1.96.0` / `cargo 1.96.0`.
- Confirmed `git diff --check 161b968fb723d5ea794a9bda40913055567b9391...5dc9a83d53eaec254eb47a074dc1872c1145d04f` passes and the worktree remains source-clean.
- Attempted `cargo test -p assay-mcp-server -p assay-cli bounded_process::tests -- --nocapture`; compilation reached linking, then local infrastructure failed with `ld: write() failed, errno=28` because the volume had only 159 MiB free. This is not reported as a test pass.
- Local clippy was not rerun after that disk-full failure. Exact-head GitHub CI's `Clippy (deny warnings)` and macOS `Build + Test` jobs passed; Ubuntu and Windows Build + Test jobs were still running when reviewed. Automated quota/rate-limit records were not treated as reviews.
- Re-read GitHub immediately before posting: PR head remained `5dc9a83d53eaec254eb47a074dc1872c1145d04f`.

Rul1an:
## Findings

### [P1] Persistent tree-kill failure still leaves the cleanup joins unbounded - `tests/support/bounded_process.rs:260`

The final fix makes the injected error persist through the early `terminate_remaining_tree` call and the deadline `terminate_and_reap` retry, but recovery then performs a third `child.start_kill()` outside the injection seam and ignores its result before joining all three I/O workers. A real persistent `killpg(SIGKILL)` / `TerminateJobObject` failure also fails this third call. The following direct-child kill targets the already-observed parent, so it cannot close stdout/stderr still inherited by the live descendant. `join_reader` can therefore block forever instead of returning within the helper's bound, and the ignored third kill error hides that cleanup did not establish quiescence.

The pinned `process-wrap 9.1.0` semantics do not supply another owner: Unix `ProcessGroupChild::start_kill` is only `killpg(SIGKILL)` and has no drop kill; Windows `JobObjectChild::start_kill` is `TerminateJobObject`, and its Job Object is created with `kill_on_drop = false`.

The hit-count assertion is genuinely sensitive to the first two branches, but the passing test is not a persistent-failure proof because its seam deliberately stops before the third tree kill required to make the joins complete. In a disposable exact-head copy, I extended the existing injection to that fallback without otherwise changing the path and forced the embedding integration test to rebuild. `early_tree_kill_failure_preserves_error_after_bounded_cleanup` then failed at its 2-second guard with `runner escaped its wall-clock bound: timed out waiting on channel`. This is the production failure shape the current test does not cover. Keep cleanup bounded even when every tree-kill attempt fails, join workers before return without depending on a successful unmodeled retry, and preserve/report the first meaningful termination error plus any cleanup failure rather than discarding it.

## Verdict

**NOT READY** pending resolution of the P1 above.

I am the single independent non-building reviewer for this record. I did not build this change and did not author or edit its specification or implementation plan. I made no repository file changes and did not push or merge. This review is bound to exact head **`42f7b00ea1323a027eba4a2291a1a168ecc00914`**. I reviewed the entire one-file PR diff against current `main` **`1a47918f815d3c865bc42fe54a80aff8b2662dbd`**; the merge base is the PR's GitHub base **`161b968fb723d5ea794a9bda40913055567b9391`**, and the only base-to-current-main change is unrelated.

## Verification

- Read `AGENTS.md`, both prior P1 reviews (`5250649270`, `5250753097`), the full `677`-addition / `120`-deletion diff, and the final-fix delta.
- Inspected pinned `process-wrap 9.1.0` Unix process-group, Windows Job Object, core child-wrapper, wait, kill, and drop/handle semantics.
- Ran both exact-head focused embedded `bounded_process::tests` binaries from `/private/tmp/assay-wt-bounded-kill-before-eof`: each passed `14`, failed `0`, ignored `1`.
- Ran the forced-rebuild external mutation described above: it failed only by the expected 2-second wall-clock timeout, demonstrating the unbounded join when the failure persists through fallback. This is negative mutation evidence, not a passing suite.
- Ran `cargo fmt --all -- --check` successfully at the exact head.
- Ran `cargo clippy -p assay-mcp-server -p assay-cli --all-targets -- -D warnings` successfully with `rustc 1.96.0` / `cargo 1.96.0`.
- Confirmed `git diff --check 1a47918f815d3c865bc42fe54a80aff8b2662dbd...42f7b00ea1323a027eba4a2291a1a168ecc00914` passes and the exact-head worktree remains source-clean.
- Immediately before posting, GitHub still reported head `42f7b00ea1323a027eba4a2291a1a168ecc00914`. Required `host-capability-check` and `lane-check/proof` are green; macOS Build + Test and Clippy are green; Build CLI Artifact is green; Ubuntu Build + Test and Windows Build + Test remain pending. Those pending checks are not reported as passes.
- CodeRabbit says `Review rate limited`; it is unavailable automated review infrastructure and was not treated as quorum evidence.

Rul1an:
## Findings

### [P1] Replace the Windows `PIPE_NOWAIT` setup with a supported capture path - `tests/support/bounded_process.rs:423`

`SetNamedPipeHandleState` is not usable on the `ChildStdout`/`ChildStderr` handles created here. Exact-head Windows CI fails immediately with `stdout capture: Access is denied. (os error 5)` across the bounded-process tests and real MCP command callers; the first embedding binary reports 11 failures. This matches the Win32 access contract: changing state on a read-only pipe requires `FILE_WRITE_ATTRIBUTES`, which these standard anonymous-pipe read handles do not provide.

There is a second Windows correctness problem behind that failure. In `PIPE_NOWAIT`, an empty-but-open pipe makes `ReadFile` return `ERROR_NO_DATA` (232). Rust maps Win32 232 to `ErrorKind::BrokenPipe`, not `WouldBlock`, so `CapturedOutput::drain` would record a terminal read error at lines 464-467 rather than poll again even if mode-setting succeeded. Use a Windows capture mechanism that is actually supported for these handles (and distinguishes no-data from EOF), with a Windows test that starts with an empty open pipe before later output.

### [P1] Capture setup errors abandon the spawned process and stdin worker - `tests/support/bounded_process.rs:175`

Both `CapturedOutput::new(...)?` calls occur after the child is spawned and the stdin writer thread is detached. If either nonblocking setup fails, `run_bounded` returns immediately without tree termination, direct-child termination, reaping, or joining stdin. The exact-head Windows failure takes this path on every invocation. Dropping `process-wrap 9.1.0`'s wrapper is not cleanup: Unix process-group children have no drop kill, and the Windows Job Object is created with `kill_on_drop = false`. A hanging child can therefore survive the reported capture error. Route setup failures through the same bounded cleanup/diagnostic path instead of propagating with `?`.

### [P1] Persistent tree-kill failure can still block forever joining stdin - `tests/support/bounded_process.rs:310`

The new spoolers make stdout/stderr abandonment bounded, but stdin remains a blocking `write_all` worker and is unconditionally joined. If the direct parent exits while a descendant retains the stdin read end without consuming it, a payload larger than the pipe buffer leaves this worker blocked. When every tree-kill attempt fails, the deadline path abandons only stdout/stderr and then hangs at this join. The persistent-failure test passes `b""` and its descendants do not retain a consuming stdin shape, so `LIVE_IO_WORKERS == 0` does not cover this branch. Make stdin cancellation/ownership bounded too and extend the injected persistent-failure wall-clock test with blocked stdin.

## Verdict

**NOT READY** pending the three P1 fixes above. Exact-head Windows Build + Test is failed; Ubuntu Build + Test is still pending and is not reported as a pass.

I am the single independent non-building reviewer for this record. I did not build this change and did not author or edit its specification or implementation plan. I made no PR worktree file changes and did not push or merge. This review is bound to exact head **`a8b22505fba146c4a3cdc4956dee272ef5d448ef`**. I reviewed the entire one-file diff against current `main` **`1a47918f815d3c865bc42fe54a80aff8b2662dbd`**, including all three prior P1 comments `5250649270`, `5250753097`, and `5250856226`.

## Verification

- Inspected all `880` additions / `161` deletions in `tests/support/bounded_process.rs`, the final-head delta, and pinned `process-wrap 9.1.0` ownership/kill behavior.
- Confirmed every process-tree termination attempt now calls the injectable `request_tree_kill` seam; direct-child fallback kills remain separate and do not establish descendant cleanup.
- Ran `cargo test -p assay-mcp-server -p assay-cli bounded_process::tests -- --nocapture` locally at the exact head: every embedding suite passed its `14` active bounded-process tests with `1` ignored helper.
- Ran `cargo fmt --all -- --check` and `cargo clippy -p assay-mcp-server -p assay-cli --all-targets -- -D warnings` successfully with `rustc 1.96.0` / `cargo 1.96.0`.
- Confirmed `git diff --check 1a47918f815d3c865bc42fe54a80aff8b2662dbd...a8b22505fba146c4a3cdc4956dee272ef5d448ef` clean and the exact-head PR worktree source-clean.
- Verified the persistent-failure wall-clock test is mutation-sensitive in a disposable exact-head archive: replacing Unix `set_nonblocking` with a no-op made `early_tree_kill_failure_preserves_error_after_bounded_cleanup` fail at its 2-second channel guard. This is negative mutation evidence, not a passing suite.
- Inspected exact-head CI: macOS Build + Test passed and its bounded descendant/flood/deadline/persistent-failure tests are green; Windows Build + Test failed with `os error 5` at stdout capture; Ubuntu Build + Test remained pending immediately before posting. Other green checks do not override the failed/pending platform evidence.
- Re-checked GitHub immediately before posting: PR head remained `a8b22505fba146c4a3cdc4956dee272ef5d448ef`. Automated quota/rate-limit records were not treated as quorum reviews.

Rul1an:
## Findings

### [P1] Preserve the first bounded-run failure when cleanup also fails - `tests/support/bounded_process.rs:305`

`early_failure` records an output-ceiling violation before cleanup, but any later `tree_termination_error` is returned first at line 305 and the primary failure at line 313 is skipped entirely. Thus a hostile stdout flood that exceeds its cap can be reported only as a tree-cleanup failure if process-tree termination also fails; the diagnostic no longer says that the output ceiling was crossed.

I confirmed this with a review-only test in a disposable exact-head archive: combine `stdout_flood_command()` at a 1,024-byte cap with the persistent tree-kill injection and assert that both the cap failure and cleanup failure survive. The test failed: captured stdout was 1,025 bytes, but the returned error contained only `terminate remaining process tree ... injected cleanup failure` and omitted `stdout exceeded its 1024-byte ceiling`. Use one ordered primary-failure accumulator (or explicitly preserve `early_failure` as primary) and append cleanup errors; add this combined mutation test so cleanup cannot replace the first failure.

### [P1] Do not let the capture-setup test kill a concurrent sibling's process tree - `tests/support/bounded_process.rs:125`

Every test-time `run_bounded` stores its child in the process-global `LAST_SPAWNED_CHILD_PID`, while `capture_setup_failure_terminates_spawned_child` reads that global and its guard kills whichever PID is present. Rust runs tests in one binary concurrently, so another bounded-process test can overwrite the PID between the injected spawn, assertion, and guard drop. The cleanup proof can inspect the wrong child and then send a tree kill to a sibling test.

This occurred in the clean exact-head focused run: `cargo test -p assay-mcp-server -p assay-cli bounded_process::tests -- --nocapture` failed in the CLI `init_stdout_contract` embedding because `preserves_child_outcome_when_stdin_closes_early` received a signaled status (`code() == None`) instead of exit 23 while the capture-setup test passed alongside it. Rerunning that embedding alone passed all 16 active tests, which is the race shape. Bind the observed PID to the injection thread/run rather than a global written by every invocation, and retain a parallel execution test that fails if cleanup can target another run.

### [P1] Replace the failing Windows delayed-output fixture with deterministic empty-open proof - `tests/support/bounded_process.rs:1184`

Exact-head Windows Build + Test is red: `delayed_stdout_after_empty_open_pipe_is_captured` fails at line 1522 because the child exits with status 1. The fixture uses `cmd /C "ping -n 1 127.0.0.1 >NUL & echo|set /p=later"`; the final `set /p` form returns nonzero, and one localhost ping does not guarantee a delay before output. Consequently this CI run does not validate the new `PeekNamedPipe` empty-open semantics, and the fixture is not deterministic mutation evidence for “empty now, output later.” Use a helper with an explicit delay, a successful exit, and exact later bytes, then require it to pass on Windows.

## Verdict

**NOT READY** pending the three P1 fixes above. Windows Build + Test is failed and Ubuntu Build + Test remains pending; pending is not a pass.

I am an independent non-building reviewer. I did not build this change or author/edit its specification or implementation plan. I made no repository or PR worktree file changes and did not push or merge. This review is bound to exact head **`b658d1b55e141ebb4419d34acf3473b236bff606`** and covers the complete `1,152`-addition / `192`-deletion diff against current `main` **`1a47918f815d3c865bc42fe54a80aff8b2662dbd`** (merge base **`161b968fb723d5ea794a9bda40913055567b9391`**).

## Verification

- Read `AGENTS.md` and prior P1 comments `5250649270`, `5250753097`, `5250856226`, and `5251050696`.
- Traced file-backed stdin before spawn; owner-polled stdout/stderr; Unix `O_NONBLOCK`; Windows `PeekNamedPipe` FFI and empty-open/broken-pipe branches; post-spawn capture cleanup; persistent all-tree-kills-fail cleanup; output ceilings/truncation; polling; and error precedence.
- Confirmed production has no I/O worker threads left. Output remains memory-backed and capped at `limit + 1`; no output tempfile or new output disk-growth path was introduced. Non-empty stdin uses a tempfile bounded by the caller's already-materialized input slice.
- Exact-source isolated MCP and CLI `agent_golden_path_contract` bounded-process suites each passed `16`, failed `0`, ignored `1`. The broader focused command failed as described in P1 #2; this is not reported as a pass.
- The review-only first-error test failed as described in P1 #1. The archive file was restored byte-for-byte to the exact Git object before clean verification.
- `cargo fmt --all -- --check`, `cargo clippy -p assay-mcp-server -p assay-cli --all-targets -- -D warnings`, and `git diff --check` passed with `rustc 1.96.0` / `cargo 1.96.0`.
- Exact-head CI at posting: macOS Build + Test passed, Windows Build + Test failed on the new delayed-output test, Ubuntu Build + Test remained in progress, and Clippy passed. I inspected the completed macOS and Windows job logs; the Windows failure is quoted above.
- Re-read GitHub immediately before posting: PR head remained `b658d1b55e141ebb4419d34acf3473b236bff606`. Automated quota/rate-limit records were not treated as independent review evidence.

Rul1an:
## MVP disposition: stop and retain as post-MVP redesign

This PR is returned to draft and must not merge in its current form.

The exact-head review on `b658d1b55e141ebb4419d34acf3473b236bff606` remains **NOT READY**: https://github.com/Rul1an/assay/pull/2286#issuecomment-5251220873

Measured reasons:

- exact-head Windows `Build + Test` fails;
- cleanup can overwrite an earlier output-cap failure;
- global injected PID tracking can kill a concurrent sibling process;
- the Windows delayed-output fixture does not yet prove `PeekNamedPipe` semantics;
- the attempted fix expanded from a bounded helper correction into a cross-platform process-supervisor rewrite (`tests/support/bounded_process.rs`, over 1k added lines).

Issue #2249 is already classified `horizon:post-mvp`. Continuing patch-by-patch would increase MVP risk rather than remove an MVP blocker. Preserve this branch/PR as measured design evidence; resume only with a fresh bounded design that pins error precedence, per-invocation ownership, Windows pipe semantics, and all-kills-fail wall-clock behavior before implementation.

PR #2285 (the independent OSV allowlist slice) was already reviewed, green, and merged separately.

Rul1an:
Closed as a merge vehicle after the three-failed-fixes architecture stop. Five review rounds, three remaining P1 findings, red Ubuntu/Windows jobs, and the >1k-line supervisor rewrite make further patching unsafe. #2302 records the A/B/C split; kill-before-EOF remains deferred design evidence, not a rejected future contract.

PR #2288, opened 2026-08-11
docs(ci): end dual-source action pins and state Cargo.lock lane truth

### LINKED ISSUE #2223: docs/PINNED-ACTIONS.md restates 15 action SHAs that no workflow uses
Measured on `main` at `360c4963a8f978513948349204e12fc735595730`, statically: I read the file and
the workflows and compared them. Nothing was executed and no action was resolved against GitHub.

## The defect

`docs/PINNED-ACTIONS.md` restates the action SHAs that `.github/workflows/*.yml` already state. Two
literals that must agree and are free to drift, which is the failure mode `AGENTS.md` names under
tool-version pinning. This one is not a risk to be argued about: it has already happened, to almost
every row.

Comparing each documented row against the SHAs actually used in `.github/workflows/*.yml`:

| Rows | Documented SHA is |
|---|---|
| 15 | **not used by any call site** |
| 1 (`bencherdev/bencher`) | in use |
| 1 (`softprops/action-gh-release`) | has no SHA-pinned call site at all |

The row the cargo-deny review surfaced is the sharpest, because it is also the most-used action:

```text
docs/PINNED-ACTIONS.md:24  | `dtolnay/rust-toolchain` | `@stable`  | `4be9e76fd7c4901c61fb841f559994984270fce7` |
docs/PINNED-ACTIONS.md:25  | `dtolnay/rust-toolchain` | `@nightly` | `881ba7bf39a41cda34ac9e123fb41b44ed08232f` |

$ rg -o 'dtolnay/rust-toolchain@[0-9a-f]{40}' .github/workflows/ | sed 's/.*@//' | sort | uniq -c
      32 29eef336d9b2848a0b548edc03f92a220660cdb8
```

Thirty-two call sites, one SHA, and the document names two different ones. `actions/checkout`,
`actions/cache`, `actions/setup-python`, `actions/upload-artifact`, `actions/download-artifact`,
`Swatinem/rust-cache`, `github/codeql-action/upload-sarif` and seven more are in the same state.

The document also states `**Status:** ✅ All third-party actions are SHA-pinned (2026-01-30)`, so a
reader gets a date-stamped assurance from a table that is 15/17 wrong about what is pinned.

## Why this is worth a fix rather than a doc refresh

Refreshing the table restores agreement for one day and reinstates the same defect. The rule in
`AGENTS.md` is one literal read by everything that needs it; a table copied by hand cannot satisfy
it. Either the table is generated from the workflows — the repo already has `docs/generated/` and a
generated-docs drift check — or it stops carrying SHAs and points at the workflows as the source.

Dependabot updates the workflows and does not know this file exists, which is exactly how a
15-of-17 divergence accumulates without anyone doing anything wrong.

## Provenance

Found by the independent review of #2218 (finding 6), which reported the `dtolnay` row. I re-measured
every row before filing; the reviewer's finding understated the scope.

## Acceptance

- `docs/PINNED-ACTIONS.md` no longer states a SHA that a workflow also states, by generation or by
  deletion of the SHA column.
- If generated, it is covered by the generated-docs drift check, so a hand edit or a stale
  regeneration fails a gate rather than sitting wrong for months.
- The `Status:` line states something a check can hold it to, or is removed.
- A run of the comparison above reports zero drifted rows.

### LINKED ISSUE #2258: ci: ci-lanes.md promises a conditional the lane classifier never evaluates, so any Cargo.lock touch demands unreusable gates=all proof
Verified against `origin/main` at `ea161accd1db0921c0f938a455d2eafa7b3fddd2`.

## The drift

`docs/reference/runner/ci-lanes.md` L38 states the rule conditionally:

> | Workspace dependency bump **that can affect** `assay-runner-spike`, `assay-monitor`, `assay-ebpf`, `assay-cli`, policy correlation, or runner fixtures | yes | yes | yes |

The classifier evaluates no such condition. It is a pure path lookup. Executed directly:

```
>>> classify_file('Cargo.lock')
gate=Gate.ALL  reason='Cargo.lock: runner workflow, CLI, cgroup, or workspace dependency surface requires gates=all'

>>> load_gated_path_config().all_gate_paths ∋ 'Cargo.lock'   -> True
```

`Cargo.lock` is a literal member of `all_gate_paths`, matched at `scripts/ci/assay_runner_lane_check.py` L860-861 before any content is examined. Nothing reads the lockfile diff, nothing asks which crate moved, and nothing asks whether the bump reaches the runner, monitor, eBPF, or fixture surfaces. "Can affect" is doing no work in the implementation; every touch routes to `gates=all`, the most expensive lane.

## And the resulting proof can never be reused

`Cargo.lock` is also outside the content-provenance set:

```
>>> content_provenance_covers_path('Cargo.lock')      -> False
>>> uncovered_content_provenance_files(['Cargo.lock']) -> ('Cargo.lock',)
```

`content_provenance_paths` covers `crates/assay-runner-{schema,core,linux}`, `crates/assay-monitor`, `crates/assay-ebpf`, `crates/assay-xtask`, `scripts/ci`, two runner-spike workflows, and `.github/actions/canonical-ebpf-build`. The root `Cargo.lock` is in none of them, so a head that touches it is `unaddressed` for content provenance and no existing proof can be credited across it. Every such head needs a fresh delegated dispatch on the singleton host.

The two properties compound: maximum lane, zero reuse, triggered by a file that changes on essentially every dependency edit.

## Learned by accident, on a dev-dependency of a test

PR #2241 demonstrates the cost. Commit `3dee8fd89` added a **dev-dependency** to `crates/assay-cli/Cargo.toml`:

```toml
syn = { version = "3", features = ["full"] }
```

`Cargo.lock` gained three lines (`+ "syn 3.0.3"`). At that SHA:

```
lane-check/proof   failure   "delegated proof required: gates=all"
```

Commit `59d84bb62` removed the dependency again (`Cargo.lock` -3, `Cargo.toml` -5) and the gate is green on the current head, which reports "no delegated runner proof required".

A `syn` dev-dependency used by a test in `assay-cli` cannot affect kernel capture, cgroup placement, policy correlation, or runner fixtures. It is exactly the case the documented conditional was written to exclude, and it demanded the full delegated proof anyway.

## What the drift actually costs

Both readings are defensible on their own; the problem is that they disagree. A contributor reads L38, concludes their lockfile-only change cannot affect runner behaviour, and expects no delegated proof — then a required check goes red for reasons the documentation told them would not apply. This is the same shape as #2230, #2242, #2243 and #2252: a gate whose real coverage is wider or narrower than its published description, discovered by a contributor rather than by the contract.

Per the one-rule-one-function principle in `AGENTS.md`, the rule is currently written twice — once as prose with a condition, once as a path set without one — and the prose is the copy that drifted.

## Resolution shapes, not a decision

Either the doc stops promising a condition the classifier does not evaluate (cheap, honest, keeps failing closed), or the classifier learns to evaluate it (a lockfile-diff analysis deciding whether changed packages reach the gated crates — real work, and a new trusted input to a gating decision). Adding `Cargo.lock` to `content_provenance_paths` is a third, separate question about reuse that should not be bundled with the first two.

## Non-claims

- I do **not** claim `gates=all` is the wrong outcome for a lockfile change. Failing closed on a dependency surface is defensible; the finding is that the documentation describes a conditional gate while the implementation is unconditional.
- I do not claim the `syn` bump was safe to skip proof on. I claim it is implausible that it reaches the gated surfaces, and that the doc's condition — as written — would have excluded it.
- I have not measured how often this fires; #2241 is one observed instance, not a rate.
- I did not dispatch a delegated run, and did not touch PR #2241 or its branch.
- No claim that content-provenance coverage for `Cargo.lock` would be *correct*; a lockfile's tree OID says nothing about which crates changed.

Rul1an:
## Summary

- Closes #2223 — `docs/PINNED-ACTIONS.md` no longer restates action SHAs/tag refs. Canonical pins live at the `uses:` callsites themselves.
- Closes #2258 — `docs/reference/runner/ci-lanes.md` states the classifier truth: every `Cargo.lock` touch unconditionally requires `gates=all`, and that proof cannot reuse content provenance (`unaddressed`).

## Follow-up on head `81194a149` (unreviewed / invalid)

Prior draft head `81194a149f86a4be25d1b51ac1f6723324314dad` claimed `.github/workflows` was the only pin source. That was a false boundary: `assay-action/action.yml` has 12 SHA-pinned `uses:` call sites. That head is **unreviewed and invalid** for review quorum. This PR advances to the head below.

## Base / head

| | SHA |
|---|---|
| Base | `e1f1f77c1d35edc3e9ed373bc6e7eb47fc22d3a8` |
| Prior head (invalid) | `81194a149f86a4be25d1b51ac1f6723324314dad` |
| Head | `556d35ff5a8304e8e575231291195323ece46142` |

Worktree: `/private/tmp/assay-wt-ci-docs-truth`  
Branch: `cursor/ci-docs-truth`

## Files changed

- `docs/PINNED-ACTIONS.md`
- `docs/reference/runner/ci-lanes.md`

### RED (D0, on base `e1f1f77c`)

```text
PINNED: 40-hex SHA table rows present (Pinned SHA / Original ref columns) — dual source
ci-lanes: "Workspace dependency bump that can affect ..." — conditional phrasing
BOTH_RED_CONFIRMED
```

### RED (boundary follow-up, on `81194a149`)

```text
PINNED claimed workflows were the only source of truth / inspection scanned only .github/workflows/
assay-action/action.yml has 12 SHA-pinned uses: call sites
BOUNDARY_RED=1
```

### GREEN (current head)

```text
PINNED: no 40-hex SHA table; no pin columns; no false workflows-only boundary
PINNED: names assay-action/action.yml and .github/actions/** as callsite surfaces
PINNED: inspection command scans workflows + assay-action + .github/actions
PINNED: words robustly that .github/actions may have zero external uses: today
ci-lanes: Cargo.lock any-touch → gates=all + unaddressed reuse; no "can affect"
git diff --check: clean
no unchecked sync/assurance claims
BOTH_GREEN_CONFIRMED
```

## Verification

- Read-only RED/GREEN assertions above
- `git diff --check` clean
- Public-string review: no auto-sync / kept-in-sync / status-assurance claims
- `mkdocs` not available in the worktree environment (skipped)
- Pre-commit + pre-push hooks on commit/push

## Non-claims

- No classifier behavior change (`scripts/ci/assay_runner_lane_check.py` untouched)
- No PINNED-ACTIONS generator or drift check added
- No workflow / composite-action YAML edits
- No branch-protection / ruleset / `deny.toml` edits
- Required live contexts and `CI-CONTRACT.md` §6 untouched
- Does not claim `gates=all` for `Cargo.lock` is wrong — only that the doc no longer describes a condition the classifier never evaluates
- Does not claim content-provenance coverage for `Cargo.lock` would be correct
- Does not claim Dependabot covers every callsite surface
- Does not merge

## Test plan

- [x] RED assertions fail for the pre-edit defects (D0 + false boundary)
- [x] GREEN assertions pass after the doc edits
- [x] Diff limited to the two scoped docs
- [x] Draft PR only; do not merge

### DIFF
--- a/docs/PINNED-ACTIONS.md
+++ b/docs/PINNED-ACTIONS.md
@@ -2,7 +2,22 @@
 
 Repo settings: **Allowed actions** should be restricted (e.g. "Allow GitHub-owned and verified creators") and **Require SHA pinning** enabled once workflows are pinned.
 
-**Status:** ✅ All third-party actions are SHA-pinned (2026-01-30).
+## Source of truth
+
+Workflow files under `.github/workflows/` are the only source of truth for
+which third-party actions are used and which commit SHAs pin them. This page
+does not restate those SHAs or tag refs. A duplicated table was a second literal
+that Dependabot never updated and that drifted from the workflows (see #2223).
+
+To inspect the pins currently in use:
+
+```bash
+rg -n --pcre2 'uses:\s*[^\s]+@[0-9a-f]{40}' .github/workflows/
+```
+
+Nothing in this repository regenerates or verifies this page against the
+workflows. Treat the workflows as authoritative; treat this page as procedure
+only.
 
 ## Resolving SHAs
 
@@ -10,28 +25,6 @@ Repo settings: **Allowed actions** should be restricted (e.g. "Allow GitHub-owne
 # Example: get latest commit SHA for a tag/branch
 gh api repos/OWNER/REPO/commits/REF --jq .sha
 ```
-
-## Pinned actions (current)
-
-| Action | Original ref | Pinned SHA |
-|--------|--------------|------------|
-| `actions/attest` | `@v4.1.1` | `a1948c3f048ba23858d222213b7c278aabede763` |
-| `actions/checkout` | `@v4` | `34e114876b0b11c390a56381ad16ebd13914f8d5` |
-| `actions/upload-artifact` | `@v4` | `ea165f8d65b6e75b540449e92b4886f43607fa02` |
-| `actions/download-artifact` | `@v4` | `d3f86a106a0bac45b974a628896c90dbdf5c8093` |
-| `actions/setup-python` | `@v5` | `a26af69be951a213d495a4c3e4e4022e16d87065` |
-| `actions/cache` | `@v4` | `0057852bfaa89a56745cba8c7296529d2fc39830` |
-| `dtolnay/rust-toolchain` | `@stable` | `4be9e76fd7c4901c61fb841f559994984270fce7` |
-| `dtolnay/rust-toolchain` | `@nightly` | `881ba7bf39a41cda34ac9e123fb41b44ed08232f` |
-| `Swatinem/rust-cache` | `@v2` | `779680da715d629ac1d338a641029a2f4372abb5` |
-| `mozilla-actions/sccache-action` | `@v0.0.6` | `9e326ebed976843c9932b3aa0e021c6f50310eb4` |
-| `bencherdev/bencher` | `@v0.6.2` | `0f8f620172ccd6225d40a7590598eb7b41718af8` |
-| `softprops/action-gh-release` | `@v2` | `a06a81a03ee405af7f2048a818ed3f03bbf83c7b` |
-| `github/codeql-action/upload-sarif` | `@v3` | `439137e1b50c27ba9e2f9befc93e43091b449c34` |
-| `dorny/test-reporter` | `@v1` | `d61b558e8df85cb60d09ca3e5b09653b4477cea7` |
-| `rust-lang/crates-io-auth-action` | `@v1` | `b7e9a28eded4986ec6b1fa40eeee8f8f165559ec` |
-| `PyO3/maturin-action` | `@v1` | `86b9d133d34bc1b40018696f782949dac11bd380` |
-| `pypa/gh-action-pypi-publish` | `@release/v1` | `ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e` |
 
 ## Dependabot for SHA updates
 
@@ -48,16 +41,20 @@ updates:
       prefix: "chore(ci)"
 ```
 
-Dependabot will propose PRs to update action refs; with SHA pinning, it will propose SHA bumps when the action repo has new commits on the same tag.
+Dependabot proposes PRs that update action refs in the workflow files. With SHA
+pinning, those PRs bump the SHA when the action repo advances on the same tag.
+Dependabot does not maintain this document.
 
 ## Updating SHAs
 
 When updating SHAs manually:
 
 1. Resolve new SHA: `gh api repos/OWNER/REPO/commits/REF --jq .sha`
-2. Update workflow files: `sed -i '' 's|OLD_SHA|NEW_SHA|g' .github/workflows/*.yml`
-3. Update this document with the new SHA
-4. Commit with message: `chore(ci): pin OWNER/REPO to SHA (was vX)`
+2. Update the workflow files that call the action (for example:
+   `sed -i '' 's|OLD_SHA|NEW_SHA|g' .github/workflows/*.yml`)
+3. Commit with message: `chore(ci): pin OWNER/REPO to SHA (was vX)`
+
+Do not add the SHA back into this document.
 
 ## Security benefits
 
--- a/docs/PINNED-ACTIONS.md
+++ b/docs/PINNED-ACTIONS.md
@@ -4,20 +4,33 @@ Repo settings: **Allowed actions** should be restricted (e.g. "Allow GitHub-owne
 
 ## Source of truth
 
-Workflow files under `.github/workflows/` are the only source of truth for
-which third-party actions are used and which commit SHAs pin them. This page
-does not restate those SHAs or tag refs. A duplicated table was a second literal
-that Dependabot never updated and that drifted from the workflows (see #2223).
+The canonical pin for a third-party action is the `uses:` callsite in the YAML
+that invokes it. This page does not restate those SHAs or tag refs. A duplicated
+table was a second literal that Dependabot never updated and that drifted from
+the callsites (see #2223).
 
-To inspect the pins currently in use:
+Callsite surfaces in this repository include:
+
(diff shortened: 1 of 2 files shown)


Rul1an:
## Independent review — READY

**Exact head reviewed:** `556d35ff5a8304e8e575231291195323ece46142`  
**Base:** `e1f1f77c1d35edc3e9ed373bc6e7eb47fc22d3a8`

### Findings

No findings.

### Verified

- The diff is docs-only and limited to `docs/PINNED-ACTIONS.md` and `docs/reference/runner/ci-lanes.md`; no workflow, classifier, required-context, ruleset, or composite-action implementation changed.
- `PINNED-ACTIONS.md` no longer duplicates a SHA/tag table or makes the previous unchecked “all pinned” status assurance.
- The documented callsite boundary includes the current executable surfaces: `.github/workflows`, `assay-action/action.yml`, and `.github/actions/**/action.yml`. The inspection command runs successfully and finds the SHA-pinned callsites across those surfaces.
- The only non-SHA `uses:` text found under the workflow tree is `Rul1an/assay-action@v3` inside a release-notes heredoc, not an executed Action step.
- The inspection wording is appropriately limited: callsites are authoritative, this page is procedure-only, and no generator/drift-check or complete-coverage assurance is claimed.
- `Cargo.lock` is an exact member of `all_gate_paths`; `classify_file()` therefore routes every touch to `Gate.ALL` without diff/package analysis.
- `Cargo.lock` is outside `content_provenance_paths`; `uncovered_content_provenance_files(["Cargo.lock"])` is pinned by a self-test, and cross-head reuse is rejected when an uncovered path is present. The new “fresh `gates=all` / unaddressed” wording matches implementation.
- `python3 scripts/ci/assay_runner_lane_check.py --self-test` passes.
- `git diff --check origin/main...HEAD` passes; worktree is clean.
- Exact-head GitHub checks `mkdocs-strict` and `local-links` pass.

### Non-claims preserved

This review does not claim Dependabot covers every action surface, that the documentation itself enforces pinning, or that unconditional `Cargo.lock` gating is the optimal policy. It verifies that the documentation now states the current implementation truth.

**Verdict: READY for exact head `556d35ff5a8304e8e575231291195323ece46142`.** A new push invalidates this review.

Issue #2302, opened 2026-08-11
ci: post-setup-rust execution plan


Rul1an:
## Role

This issue is the **single ownership and scheduling index** for open CI maintenance. It is not an active programme ledger and does not make post-MVP work release-blocking. Repository truth remains code, checks, and the child issues.

Current measurement: `origin/main` `7276ac92fa1080822b8c5836c03b2d13afd2759a` on 2026-08-12.

## Completed tranches

- [x] **CI-1 process-test diagnosis and fixture contract** — #2303/#2304; diagnostics and re-executed helper landed.
- [x] **CI-2 merge-gate integrity** — #2242/#2243; fail-open guard plus live required-context reconciliation operational.
- [x] **CI-3 delegated runner/proof contract** — #2212, #2252, #2256, #2257, #2259 dispositioned and merged.
- [x] **CI-4A–E deterministic toolchain ownership** — #2291, #2250, #2226, #2235, #2231, #2224 and #2225 complete.
- [x] **CI-5A golden-path harness consolidation** — merged via #2310.

Historical execution evidence remains in the comments below; this body is the current schedule, not a duplicate chronology.

### CI-4F — #2188 / PR #2328 advisory-cache isolation

Owner: Cursor. Exact head: `47a5aea5da934da37d48b41c4cc5bb7b1daab3e9`.  
State: non-draft PR, independent exact-head review READY, auto-merge enabled; required `CI` is still running. `host-capability-check` is green and `lane-check/proof` says no delegated proof is required.

Exit:

- cargo-deny and cargo-audit use separate bounded roots;
- one source-safe cargo-audit runner is shared by pre-push and both CI audit sites;
- hostile cross-tool fixtures reach the real command;
- real isolated cargo-audit and cargo-deny advisory runs complete;
- required contexts green, then standard squash merge without bypass.

### CI-5C — #2196 fixture staging/mutations

Owner: Cursor. Exact local head after conflict-free main advance: `486d95be04cb4d6177ce31ffaaf8cc4cfbaa9762`.  
State: independent exact-head review READY; push/non-draft PR creation in progress.

Exit:

- one rule stages the plugin resource set;
- all three generated skill destinations have a biting mutation through the real validator with exact-path diagnostics;
- bounded reads and regular-file/non-symlink requirements remain;
- optimization suite and all 92 hardening cases green;
- required contexts green, then standard squash merge without bypass.

### Completed prerequisite — PR #2325

[...]

PR #2303, opened 2026-08-11
test(support): CI-1A bounded_process deadline/ending diagnostics

### LINKED ISSUE #2249: windows-latest: two bounded_process descendant tests fail intermittently

Rul1an:
## Summary

CI-1A (bounded_process diagnostics only). One file: `tests/support/bounded_process.rs`.

- Deadline failures now name `outstanding=` waits for child / stdin / stdout / stderr, distinguishing **child-alive** from **child-exited-with-open-handles**.
- A status reaped after harness termination is labelled as such and is **not** presented as a child-chosen exit.
- EOF-before-finish success path and existing descendant pass/fail semantics are unchanged (no platform-split).

## Provenance

| Field | Value |
|-------|-------|
| Head SHA | `b743aed6f50029580be117419925e1882e9806fd` |
| Base | `origin/main` @ `e619d22fd1440996477b17cc41fc68d3f08fc118` |
| Worktree | `/private/tmp/assay-wt-ci1a-bounded-process-diagnostics` |
| Branch | `cursor/ci1a-bounded-process-diagnostics` |
| Files | `tests/support/bounded_process.rs` only |

Tracks #2302. Related evidence context #2249 (does **not** close #2249 alone).

## RED → GREEN

**RED** (tests added against stub / pre-wiring behaviour; macOS `darwin 25.6.0`, worktree above, base tree `e619d22…` with only the failing tests + stub):

```
cargo test -p assay-mcp-server --test agent_golden_path_contract -- \
  deadline_expiry_names_each_outstanding_wait a_harness_terminated_status_is_not_reported_as_a_child_exit
```

- `deadline_expiry_names_each_outstanding_wait_and_distinguishes_child_liveness` — FAILED: expected `outstanding=`, got `deadline of 5s expired`
- `a_harness_terminated_status_is_not_reported_as_a_child_exit` — FAILED: error was `deadline of 100ms expired; status=signal: 9 (SIGKILL); ...` with neither `outstanding=[child still running` nor `recorded after harness termination`

**GREEN** (head `b743aed6f50029580be117419925e1882e9806fd`):

- Focused new diagnostics: 2 passed
- All `bounded_process` module tests via `agent_golden_path_contract`: 10 passed
- Consumer integration suites using `bounded_process`: `enforcement_sarif_cli`, `init_stdout_contract`, `doctor_config_check_contract`, `doctor_exit_class_contract`, `policy_validate_machine_contract`, `config_recovery_argv_contract`, `agent_golden_path_contract` (cli + mcp-server) — all green
- `cargo clippy -p assay-mcp-server --tests -- -D warnings` — pass
- `cargo clippy -p assay-cli --tests -- -D warnings` — pass
- `git diff --check` — clean
- Pre-push hooks (including workspace clippy) — pass

## Non-claims

- **Not** kill-before-EOF / process-supervisor behaviour.
- **Not** a fixture-contract rewrite or READY/re-exec/Solo helper.
- **Not** a claim that this fixes load causation or the intermittent Windows defect in #2249.
- **#2253** rates remain evidence only; #2253 is closed as an experimental merge vehicle — this PR does not reopen or depend on merging it.
- Does **not** close #2249 alone.
- Does not resume #2286; does not land #2251.

## STOP questions

None at authoring time. If CI shows unexpected dirt outside `tests/support/bounded_process.rs`, do not merge — stop and report.

## Test plan

- [x] New diagnostic unit/behavioural tests (RED then GREEN)
- [x] Full `bounded_process` module tests
- [x] Integration tests that include `bounded_process`
- [x] fmt / clippy / `git diff --check`
- [ ] Draft only — do not mark ready; do not merge until review quorum

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostics for processes that exceed their deadlines or are terminated.
  * Reports whether termination came from the process or the execution harness.
  * Identifies outstanding process or I/O waits when a deadline is reached.
  * Preserves the process exit status when execution completes successfully.

### DIFF
--- a/tests/support/bounded_process.rs
+++ b/tests/support/bounded_process.rs
@@ -35,6 +35,83 @@ impl ProcessLimits {
 pub const GOLDEN_PATH_LIMITS: ProcessLimits =
     ProcessLimits::new(Duration::from_secs(10), 1024 * 1024, 1024 * 1024);
 
+/// Where a reported exit status came from.
+///
+/// A status the harness reaped after terminating the process tree is not an
+/// ending the child chose: on Windows `TerminateJobObject` supplies the exit
+/// code itself, and on Unix the group kill supplies the signal. Rendering that
+/// the same way as a child-chosen exit makes a timeout indistinguishable from a
+/// child that died instantly.
+#[derive(Clone, Copy)]
+enum StatusOrigin {
+    ChildExit,
+    HarnessTermination,
+}
+
+/// A child's exit status together with the origin that explains it.
+#[derive(Clone, Copy)]
+struct Ending {
+    status: ExitStatus,
+    origin: StatusOrigin,
+}
+
+impl Ending {
+    fn child_exit(status: ExitStatus) -> Self {
+        Self {
+            status,
+            origin: StatusOrigin::ChildExit,
+        }
+    }
+
+    fn harness_termination(status: ExitStatus) -> Self {
+        Self {
+            status,
+            origin: StatusOrigin::HarnessTermination,
+        }
+    }
+
+    fn describe(&self) -> String {
+        match self.origin {
+            StatusOrigin::ChildExit => format!("{} (child exit)", self.status),
+            StatusOrigin::HarnessTermination => format!(
+                "{} (recorded after harness termination; may be the termination's own code)",
+                self.status
+            ),
+        }
+    }
+}
+
+/// Names what the deadline was still waiting for when it expired.
+///
+/// Without this, "deadline expired" covers two different endings — a child that
+/// never exited, and a child that exited while a descendant kept its output
+/// handles open — and the reader of a failure cannot tell them apart.
+fn deadline_expiry(
+    timeout: Duration,
+    child_exited: bool,
+    stdin_pending: bool,
+    stdout_pending: bool,
+    stderr_pending: bool,
+) -> String {
+    let mut outstanding = Vec::new();
+    if !child_exited {
+        outstanding.push("child still running");
+    }
+    if stdin_pending {
+        outstanding.push("stdin still being written");
+    }
+    if stdout_pending {
+        outstanding.push("stdout handle still open");
+    }
+    if stderr_pending {
+        outstanding.push("stderr handle still open");
+    }
+    format!(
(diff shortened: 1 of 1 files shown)


Rul1an:
## Independent exact-head review — READY (advisory)

Reviewed final head `b743aed6f50029580be117419925e1882e9806fd` in detached worktree `/private/tmp/assay-review-2303` against base `e619d22fd1440996477b17cc41fc68d3f08fc118`.

### Findings

None.

### Contract review

- EOF-before-finish and the existing descendant-test expectations are unchanged.
- The change only carries status origin through existing error paths and names waits already observed by the loop.
- A pre-kill observed status remains a child exit; a status first reaped after the harness termination attempt is conservatively labelled `recorded after harness termination; may be the termination's own code`.
- No READY protocol, fixture rewrite, platform split, kill-before-EOF, workflow, or product schema change is present.

### Independent verification

- Both focused tests: pass individually and together.
- `bounded_process::tests` in both MCP integration binaries: 10/10 pass in each.
- Affected CLI integration suites: 99/99 pass across the six selected binaries.
- `cargo fmt --all -- --check`: pass.
- `cargo clippy -p assay-mcp-server --tests -- -D warnings`: pass.
- `cargo clippy -p assay-cli --tests -- -D warnings`: pass.
- `git diff --check`: clean; review worktree restored clean.

### Mutation proof

Both mutations failed the intended exact test:

1. Bypassing `deadline_expiry(...)` at the production call site restored the old bare deadline text and failed on missing `outstanding=[child still running`.
2. Replacing `Ending::harness_termination(status)` with `Ending::child_exit(status)` failed because the diagnostic again claimed `(child exit)`.

### Non-claims

This review does not claim #2249 is fixed, does not infer load causation, and does not transfer any Windows result from #2251/#2286. Windows CI on this exact head remains the platform integration proof.

Verdict: **READY**, contingent on required checks being green on this exact head. I did not author or edit the PR change or its governing design; this comment supplies the non-building-agent quorum because the account cannot approve its own PR.

PR #2304, opened 2026-08-11
test(support): CI-1B bounded_process READY/re-exec fixture contract

### LINKED ISSUE #2249: windows-latest: two bounded_process descendant tests fail intermittently

Rul1an:
## Summary

CI-1B (bounded_process fixture contract). One file: `tests/support/bounded_process.rs`.

- Replaces Windows PowerShell + environment-derived PID-file paths with **re-exec of the current Rust test binary** (Unix keeps a shell spawner emitting the same READY records).
- Readiness / descendant PID exclusively via bounded captured output: strict `READY pid=<nonzero u32>` parse with **identifier word boundaries** (ASCII alphanumeric **or underscore** is inside the word) plus acceptance of **Debug-escaped line boundaries** (`\\n` / `\\r\\n` from `excerpt()`).
- **Solo / no-descendant control** proves EOF/success (`READY solo`) on the same path as the descendant arms.
- Helper filter pin: zero or multiple libtest matches cannot silently green.
- Preserves CI-1A EOF-before-finish and all `run_bounded` production behavior (no kill-before-EOF, no new I/O supervisor, no deadline/retry changes).

Tracks #2302. Related evidence context #2249 (does **not** close #2249 alone).

## Provenance

| Field | Value |
|-------|-------|
| Head SHA | `0212ecd2c858b22c69a40e17569a78e591a558c4` |
| Base | `origin/main` @ `225f627b7ba85b336b461cb6400c51a82303cba0` (CI-1A / #2303) |
| Worktree | `/private/tmp/assay-wt-ci1b-bounded-process-fixture` |
| Branch | `cursor/ci1b-bounded-process-fixture` |
| Files | `tests/support/bounded_process.rs` only |
| Line delta (vs base) | +379 / −75 |

Prior heads (`f3c59ac82…`, `7fdf5c6cb…`, `d2b385a0c…`) are invalidated; request a fresh exact-head review **and fresh Windows CI** on this SHA.

### Fixture contract / alphanumeric / underscore boundaries

Covered in earlier commits on this branch (parser abuse-cases, Solo, helper pin, `XREADY` / `5evil` / `_READY` / `5_evil`).

### Debug-escaped newline left boundary (this head; Windows CI P1 on `d2b385a0c…`)

Measured Windows failure: deadline + READY were correct, but `excerpt()` Debug-escapes newlines so the char before `READY` in `stdout=\"\\nrunning...\\nREADY pid=6024.\"` is the letter `n` from the two-char escape `\\n`, which the ident-boundary rejected (hard-error before the clean stderr record).

**RED** (accepted cases added before boundary change):

```
cargo test -p assay-mcp-server --test agent_golden_path_contract -- only_one_exactly_shaped_readiness_record_yields_a_pid
```

- FAILED: expected `Ok(6024)` for the escaped diagnostic form; got `Err(... readiness record is not bounded ...)`.

**GREEN** (head `0212ecd2c858b22c69a40e17569a78e591a558c4`):

- Escaped `\\n` / `\\r\\n` diagnostic forms accept; prior rejects (`XREADY`, `_READY`, `5evil`, `5_evil`) still reject
- `bounded_process` via mcp-server + cli binaries — 15 passed, 1 ignored
- `cargo clippy -p assay-mcp-server --tests -- -D warnings` — pass
- `cargo clippy -p assay-cli --tests -- -D warnings` — pass
- `git diff --check` — clean

**Mutation probe:**

- Drop `ends_with("\\n")` left-boundary arm → escaped diagnostic accept fails (`Ok(6024)` expected)
- Drop underscore from ident set → `"_READY pid=5" must not yield a pid`
- Left boundary always-true → `"XREADY pid=5" must not yield a pid`
- Right boundary alnum-only → `"READY pid=5_evil" must not yield a pid`
- Restore → pass

## Platform / Windows non-claim

- Unix detached descendant may keep success (observed green locally).
- Windows quiet/detached arm expects deadline **only if** Solo on the same head proves EOF/success, READY is visible on the error path, and the descendant is demonstrably killed.
- **Windows CI on `d2b385a0c…` failed on this parser/diagnostic composition; this head must be re-measured on Windows.** Do not treat this PR as closing #2249 until rate/Windows evidence at this exact SHA says so.
- Retain the Windows cfg-split; do not ignore that arm. Keep the Unix shell spawner in this slice.

## Non-claims

- **Not** CI-1C / kill-before-EOF / process-supervisor behaviour.
- **Not** a claim that this fixes load causation or the intermittent Windows defect in #2249.
- Does **not** close #2249 alone; does not reopen #2253; does not resume #2286; does not land #2251 as-is.
- Does not change `run_bounded` production control flow beyond what CI-1A already shipped.

## STOP questions

1. Retain Windows cfg-split; require exact-head Windows CI — do not ignore the arm.
2. Keep Unix shell spawner in this slice; cross-platform re-exec unification is out of scope.

## Test plan

- [x] Parser abuse-cases including escaped-`\\n` accept + underscore/XREADY/evil rejects (RED then GREEN)
- [x] Mutation probes for accept + still-reject cases
- [x] Solo / helper pin / descendant fixtures
- [x] Full `bounded_process` module tests in MCP + CLI binaries
- [x] fmt / clippy / `git diff --check`
- [ ] Fresh Windows CI on this head
- [ ] Draft only — do not mark ready; do not merge until fresh review quorum on this head + Windows evidence disposition

### DIFF
--- a/tests/support/bounded_process.rs
+++ b/tests/support/bounded_process.rs
@@ -564,17 +564,40 @@ mod tests {
         format!("{READY_RECORD_PREFIX}{pid}")
     }
 
-    /// Strict `READY pid=<nonzero u32>` parse. Identical repeats agree; disagree errs.
+    fn ready_record_left_boundary(record: &str, start: usize) -> bool {
+        match record[..start].chars().next_back() {
+            None => true,
+            Some(c) => !c.is_ascii_alphanumeric(),
+        }
+    }
+
+    fn ready_record_right_boundary(record: &str, end: usize) -> bool {
+        match record[end..].chars().next() {
+            None => true,
+            Some(c) => !c.is_ascii_alphanumeric(),
+        }
+    }
+
+    /// Strict `READY pid=<nonzero u32>` parse with left/right record boundaries.
+    /// Identical repeats agree; disagree errs. Empty template prefixes (no digits)
+    /// are ignored; a digit run without valid boundaries is a malformed record.
     fn descendant_pid_from(record: &str) -> Result<u32, String> {
         let mut seen: Option<u32> = None;
         for (start, _) in record.match_indices(READY_RECORD_PREFIX) {
-            let digits: String = record[start + READY_RECORD_PREFIX.len()..]
+            let after = start + READY_RECORD_PREFIX.len();
+            let digits: String = record[after..]
                 .chars()
                 .take(MAX_PID_DIGITS + 1)
                 .take_while(char::is_ascii_digit)
                 .collect();
             if digits.is_empty() {
+                // Quoted command templates carry the prefix with no pid; skip.
                 continue;
+            }
+            if !ready_record_left_boundary(record, start)
+                || !ready_record_right_boundary(record, after + digits.len())
+            {
+                return Err(format!("readiness record is not bounded: {record:?}"));
             }
             let pid = match digits.parse::<u32>() {
                 Ok(0) | Err(_) => {
@@ -1105,6 +1128,9 @@ mod tests {
             "READY pid=1 READY pid=1 READY pid=3",
             "parent-ready",
             "READY solo",
+            // Embedded prefix / trailing payload must not supply identity.
+            "XREADY pid=5",
+            "READY pid=5evil",
         ] {
             assert!(
                 descendant_pid_from(rejected).is_err(),
--- a/tests/support/bounded_process.rs
+++ b/tests/support/bounded_process.rs
@@ -564,17 +564,22 @@ mod tests {
         format!("{READY_RECORD_PREFIX}{pid}")
     }
 
+    /// ASCII identifier continuation: alphanumeric or `_` is inside a word.
+    fn is_ascii_ident_char(c: char) -> bool {
+        c.is_ascii_alphanumeric() || c == '_'
+    }
+
     fn ready_record_left_boundary(record: &str, start: usize) -> bool {
         match record[..start].chars().next_back() {
             None => true,
-            Some(c) => !c.is_ascii_alphanumeric(),
+            Some(c) => !is_ascii_ident_char(c),
         }
     }
 
     fn ready_record_right_boundary(record: &str, end: usize) -> bool {
         match record[end..].chars().next() {
             None => true,
-            Some(c) => !c.is_ascii_alphanumeric(),
+            Some(c) => !is_ascii_ident_char(c),
         }
     }
 
@@ -1131,6 +1136,8 @@ mod tests {
             // Embedded prefix / trailing payload must not supply identity.
             "XREADY pid=5",
             "READY pid=5evil",
+            "_READY pid=5",
+            "READY pid=5_evil",
         ] {
             assert!(
                 descendant_pid_from(rejected).is_err(),
--- a/tests/support/bounded_process.rs
+++ b/tests/support/bounded_process.rs
@@ -570,8 +570,13 @@ mod tests {
     }
 
     fn ready_record_left_boundary(record: &str, start: usize) -> bool {
-        match record[..start].chars().next_back() {
+        let before = &record[..start];
+        match before.chars().next_back() {
             None => true,
+            // excerpt() Debug-escapes newlines as the two-char sequence '\\''n'
+            // (and CRLF as '\\''r''\\''n'); that 'n' is a line boundary, not an
+            // identifier character glued to READY.
+            Some('n') if before.ends_with("\\n") => true,
             Some(c) => !is_ascii_ident_char(c),
         }
     }
@@ -1119,6 +1124,20 @@ mod tests {
             ),
             Ok(77),
             "identical both-channel quotes must still read"
+        );
+        // excerpt() debug-escapes newlines, so the diagnostic carries the two-char
+        // sequence '\\' 'n' immediately before READY — not a real newline.
+        assert_eq!(
+            descendant_pid_from(
+                r#"deadline; stdout="\nrunning 1 test\nREADY pid=6024.\ntest result: ok."; stderr="READY pid=6024""#
+            ),
+            Ok(6024),
+            "debug-escaped newlines in excerpt must not hide READY"
+        );
+        assert_eq!(
+            descendant_pid_from(r#"stdout="\r\nREADY pid=6024.""#),
+            Ok(6024),
+            "debug-escaped CRLF must also be a left boundary"
         );
 
         for rejected in [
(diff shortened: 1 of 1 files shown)


Rul1an:
## Independent exact-head review — NOT READY

Reviewed head `f3c59ac82a778c53a5e27cdd503238699d2fc4c6` in detached worktree `/private/tmp/assay-review-2304`.

### [P1] The claimed strict READY record accepts embedded prefixes and trailing payload

`descendant_pid_from` scans `match_indices("READY pid=")`, then consumes only the following digit run. It does not validate the record's left or right boundary. Two mutation/probe cases added to the existing rejected table both fail because the parser returns `Ok(5)`:

```text
XREADY pid=5
READY pid=5evil
```

This is not only wording drift. The parsed PID is the identity used by the teardown assertions; accepting a PID from noise can make the test claim that an unrelated process is (or is not) the descendant the helper spawned. The PR body explicitly claims a strict `READY pid=<nonzero u32>` record and noise rejection.

Required fix:

- pin both strings as rejected cases;
- enforce a left record boundary and a right delimiter/end boundary while preserving the intentional supported forms: a standalone record, a libtest line such as `READY pid=1234.`, and quoted duplicate records in the bounded diagnostic;
- keep the intentional empty prefix in the quoted command text ignorable, but do not allow a malformed non-empty record to supply identity;
- rerun parser/fixture tests and mutation proof on the new head.

No other blocker found in the static contract review. The helper filter is fail-closed for zero/multiple matches, PID-file/PowerShell are removed, and `run_bounded` control flow is unchanged.

The STOP questions are answered as follows: retain the Windows cfg-split and require exact-head Windows CI; do not ignore the arm. Keep the Unix shell spawner in this slice; cross-platform unification is unnecessary scope.

A new push invalidates this review; request a fresh exact-head review after the parser fix.

Rul1an:
Exact-head follow-up review on `7fdf5c6cbca0253bc642031fb916b4e8802bdc3a` found one remaining **P1 blocker** in the same parser contract.

The new boundary helpers treat `_` as a delimiter because they reject only `is_ascii_alphanumeric()`. As a result, identifier-embedded strings still supply descendant identity:

- `_READY pid=5` -> `Ok(5)`
- `READY pid=5_evil` -> `Ok(5)`

I verified this by temporarily adding both values to the existing rejected table and running the exact parser test; it failed on `_READY pid=5`. The review worktree was restored clean afterward.

Please define an identifier/word boundary consistently (ASCII alphanumeric **or underscore** is not a boundary), add both cases to the rejected table RED-first, preserve the currently accepted standalone/libtest/quoted-duplicate forms, and rerun the same mutation proof. Scope remains only `tests/support/bounded_process.rs`. A push invalidates this review; I will re-review the final head.

Rul1an:
## Independent exact-head review: READY pending CI

Reviewed final head `d2b385a0cff607c2523eba36b5d6c6397a8d3afe` as the non-building reviewer. **No findings remain.**

Review scope and evidence:
- Full PR diff: one file, `tests/support/bounded_process.rs` (`+355/-75` versus base `225f627b7ba85b336b461cb6400c51a82303cba0`).
- Re-exec fixture contract, Solo control, exact helper selection, READY parser, platform-specific expectations, and cleanup diagnostics reviewed.
- Focused bounded-process suites passed in both MCP integration binaries: `15 passed, 1 ignored` each.
- Affected CLI integration set passed, including `agent_golden_path_contract`, recovery, doctor, init, and policy-validation contracts.
- `cargo fmt --all -- --check`, affected-target clippy with `-D warnings`, and `git diff --check` passed.
- Independent mutation: removing underscore from the shared identifier-boundary predicate made the exact parser test fail on `_READY pid=5`; worktree restored clean.
- Previously validated mutations cover embedded prefix/trailing payload and helper-filter exactness.

Review verdict is bound only to `d2b385a0cff607c2523eba36b5d6c6397a8d3afe`. A new push invalidates it. Windows behavior remains a non-claim until the exact-head `Build + Test (windows-latest)` check completes successfully; do not merge before all required checks are green.

Rul1an:
Windows exact-head CI on `d2b385a0cff607c2523eba36b5d6c6397a8d3afe` found a **P1 blocker** in the parser/diagnostic composition, not the process fixture.

Both target tests reached the intended deadline state and emitted valid READY records, but `descendant_pid_from` rejected the stdout record inside the escaped excerpt:

```text
stdout="\nrunning 1 test\nREADY pid=6024.\ntest result: ok..."; stderr="READY pid=6024"
readiness record is not bounded
```

`excerpt()` debug-escapes newlines, so the character immediately before `READY` is the `n` in the two-character `\\n` escape. The current ASCII identifier-boundary rule therefore rejects a legitimate record before it reaches the valid stderr occurrence. Windows job: https://github.com/Rul1an/assay/actions/runs/31513078195/job/93851616551

Please add a RED-first accepted parser case matching this escaped diagnostic shape, then make the smallest boundary correction that accepts an escaped line boundary (`\\n`, and `\\r\\n` if the formatter can produce it) without reopening `_READY`, `XREADY`, `5evil`, or `5_evil`. Do not make the parser silently ignore genuinely malformed nonempty records. Scope remains only `tests/support/bounded_process.rs`; rerun focused/mutation checks and push. This invalidates my prior READY verdict; fresh exact-head review and fresh Windows CI are required.

Rul1an:
## Independent exact-head re-review: READY pending CI

Re-reviewed final head `0212ecd2c858b22c69a40e17569a78e591a558c4` after the Windows diagnostic-boundary fix. **No findings remain.** This supersedes my invalidated verdict on `d2b385a0`.

Increment reviewed:
- `excerpt()`'s debug-escaped `\\nREADY` and `\\r\\nREADY` shapes are explicitly accepted.
- Identifier embeddings (`XREADY`, `_READY`, `5evil`, `5_evil`) remain rejected.
- Genuinely malformed nonempty READY records remain errors; the parser does not fail open.

Exact-head evidence:
- Both MCP bounded-process suites: `15 passed, 1 ignored` each.
- `cargo fmt --all -- --check`, `git diff --check`, clean review worktree.
- Independent mutation removing the escaped-newline boundary made the exact parser test fail on the realistic Windows diagnostic; the worktree was restored clean.
- Prior exact-head/increment review evidence covers the full one-file fixture change and underscore-boundary mutation.

The review quorum is satisfied only for `0212ecd2c858b22c69a40e17569a78e591a558c4`. Windows remains pending and is the final behavioral proof for this slice. Do not merge until that exact-head job and all required contexts are green.

PR #2308, opened 2026-08-11
test(ci): protect delegated proof head guard (CI-3A)

### LINKED ISSUE #2212: The guard that stops a delegated proof carrying across a head move is unprotected against its own removal
Read on `main` at `4125d81b00847a7dc9a2276c9a7d0c44a247e569`. Verified statically; nothing was
executed and no lane check was run.

## The stronger claim this started as, and why it is withdrawn

This was raised as: `Cargo.toml` and `Cargo.lock` are classified `gates=all` by
`scripts/ci/assay_runner_lane_check.py` but are not `content_provenance_paths` in
`scripts/ci/assay_runner_gated_paths.json`, so a dependency-bump PR could advance its head while an
older delegated proof stayed acceptable, because the content-addressed comparison watches no path
that changed.

That is not what the code does, and the record should say so. The classification half is accurate —
`Cargo.toml` and `Cargo.lock` are `all_gate_paths` entries (`assay_runner_gated_paths.json:25-26`)
and are not under any `content_provenance_paths` prefix — but the acceptance path guards exactly this
case, at `assay_runner_lane_check.py:1029-1036`:

```python
uncovered = uncovered_content_provenance_files(pr.files)
...
if uncovered and proof_head != pr.head_sha:
    return ContentTreeComparison(False, (...))
```

A changed file that is gated but not content-addressed makes `uncovered` non-empty, and the proof is
then accepted only if it was produced on the PR head itself. So a `Cargo.lock` bump cannot reuse an
older proof; it needs a fresh dispatch at its own head. Fail-closed, as the docstring at
`assay_runner_lane_check.py:2456-2465` already states, and the input half is pinned by the self-test
assertion `uncovered_content_provenance_files(["Cargo.lock"]) == ("Cargo.lock",)` at line 2129.
`pr.files` is fully paginated (`load_pr`, line 1246, via `paginated` at lines 197-199), so the guard
is not reading a truncated file list.

## What is actually gapped

The guard is not protected against its own removal. `content_tree_proof_accepts_head` has no test
that drives it. Line 2129 pins the classifier input it consumes, and
`_test_content_tree_comparison` (line 2608-2655) covers `compare_content_path_trees` — the
OID-comparison half that runs *after* the short-circuit — but no test constructs a `PullRequest`
touching an uncovered gated path together with a proof manifest whose `source.head_sha` differs, and
asserts rejection. The only test that mentions the function replaces it with
`_fake_content_tree_accepts_head` (lines 2911-2964), so the real body is stubbed out rather than
exercised.

Concretely: deleting the four lines at 1029-1036, or inverting the `!=` to `==`, leaves the suite
green. The result would be the original claim above becoming true — an older proof accepted for a
head whose `Cargo.lock` changed — with no test failing to say so.

This is the failure mode this file has already been bitten by and documented. The docstring of
`_test_gate_execution_is_verified` records that "replacing the function body with `return ()` left
the suite green — the fix was unprotected against its own removal", which is why that check exists.
`_test_prefix_gated_surfaces_keep_their_coverage` (line 2414) pins which prefixes are
content-addressed and states in its own scope paragraph that it does not cover this.

So this is a guard-strength gap, not a bug: today's behaviour is correct and fail-closed, and there
is no claim here that any proof has ever been wrongly accepted or that anything has been exploited.
What is missing is the thing that keeps it correct through the next edit.

## Acceptance

- A test drives `content_tree_proof_accepts_head` — not a stub — with a `PullRequest` whose changed
  files include a gated path outside `content_provenance_paths` and a proof manifest whose
  `source.head_sha` differs from the PR head, and asserts the comparison is rejected with the
  `not covered by content-provenance trees` diagnostic.
- The same test asserts the permitted case: identical `head_sha`, uncovered file present, accepted.
- Removing or inverting the short-circuit at `assay_runner_lane_check.py:1029-1036` fails that test.

## Non-goals

- Do not add `Cargo.lock` or `Cargo.toml` to `content_provenance_paths` as the fix. It would trade a
  fail-closed dispatch requirement for a content binding on a file that changes on every dependency
  bump, and the churn cost — a fresh delegated run, or a proof comparison, on every Dependabot PR —
  has to be weighed against what the binding buys before anyone moves that list. The current
  behaviour is the deliberate cost side of the coverage set, per the docstring at lines 2456-2465.
- No change to `classify_file`, to the gating map, or to which paths are gated.
- Not a change to `compare_content_path_trees`, which is already covered.
- No claim that the documented carry-over behaviour in `docs/reference/runner/ci-lanes.md:109-115`
  is wrong; it describes the covered-path case accurately.

Rul1an:
## Summary
- drive the real `content_tree_proof_accepts_head` guard from its self-test
- reject a stale proof when `Cargo.lock` is gated but outside content-provenance trees
- prove the exact-head case still reaches and passes normal content-tree comparison

## Why
The runtime guard is currently correct and fail-closed, but deleting or inverting its exact-head short-circuit was not detected by the self-test. This closes #2212 without changing classifier or acceptance behavior.

## Mutation evidence
- replacing the guard with `if False` fails: stale uncovered proof reaches downstream comparison
- changing `!=` to `==` fails with the same protected boundary
- restoring production code returns the self-test to green

## Verification
- `python3 scripts/ci/assay_runner_lane_check.py --self-test`
- `python3 -m py_compile scripts/ci/assay_runner_lane_check.py`
- `pre-commit run assay-runner-lane-check-self-test --all-files`
- `pre-commit run --files scripts/ci/assay_runner_lane_check.py`
- full `pre-commit run --hook-stage pre-push --all-files` (all hooks passed)
- `git diff --check`

## Non-claims
- no production behavior change
- no change to gated paths or content-provenance coverage
- no claim that stale proof carry has occurred
- delegated exact-head proof, if required by the current classifier, remains pending

Exact head: `cea8d4678bcb0ef42cf912e2eccaaed42cea6288`

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify that outdated proofs are rejected when required files lack content-provenance coverage.
  * Confirmed that proofs matching the pull request’s current revision proceed successfully through validation.

### DIFF
--- a/scripts/ci/assay_runner_lane_check.py
+++ b/scripts/ci/assay_runner_lane_check.py
@@ -2138,6 +2138,7 @@ def self_test() -> None:
     _test_run_check_executes()
     _test_run_check_consults_the_override()
     _test_content_tree_comparison()
+    _test_uncovered_paths_require_exact_head()
 
     valid_run = {
         "name": DELEGATED_WORKFLOW_NAME,
@@ -2653,6 +2654,64 @@ def _test_content_tree_comparison() -> None:
     error = compare_content_path_trees(error_manifest, current)
     assert not error.accepted
     assert "proof tree error" in error.diagnostics[0]
+
+
+def _test_uncovered_paths_require_exact_head() -> None:
+    gated_paths = load_gated_path_config()
+    proof_head = "proof-head"
+    manifest = {
+        "source": {"head_sha": proof_head},
+        "content_provenance": {
+            "path_trees": {
+                path: {"oid": f"oid:{path}", "error": None}
+                for path in gated_paths.content_provenance_paths
+            }
+        },
+    }
+    stale_pr = PullRequest(
+        number=1,
+        title="Dependency update",
+        body="",
+        author_login="dependabot[bot]",
+        head_sha="new-head",
+        files=("Cargo.lock",),
+    )
+
+    old_fetch = globals()["fetch_ref_for_diff"]
+    old_current = globals()["current_content_path_trees"]
+    try:
+        def unexpected_downstream(*_args: object) -> None:
+            raise AssertionError("stale uncovered proof reached content-tree comparison")
+
+        globals()["fetch_ref_for_diff"] = unexpected_downstream
+        globals()["current_content_path_trees"] = unexpected_downstream
+        rejected = content_tree_proof_accepts_head(manifest, stale_pr)
+        assert not rejected.accepted
+        assert rejected.diagnostics == (
+            "Cargo.lock: gated path is not covered by content-provenance trees",
+        )
+
+        globals()["fetch_ref_for_diff"] = lambda *_args: None
+        globals()["current_content_path_trees"] = lambda _head: (
+            {
+                path: f"oid:{path}"
+                for path in gated_paths.content_provenance_paths
+            },
+            (),
+        )
+        exact_pr = PullRequest(
+            number=stale_pr.number,
+            title=stale_pr.title,
+            body=stale_pr.body,
+            author_login=stale_pr.author_login,
+            head_sha=proof_head,
+            files=stale_pr.files,
+        )
+        accepted = content_tree_proof_accepts_head(manifest, exact_pr)
+        assert accepted.accepted, accepted.diagnostics
+    finally:
+        globals()["fetch_ref_for_diff"] = old_fetch
+        globals()["current_content_path_trees"] = old_current
 
 
 def _test_attested_proof_pack_helpers() -> None:


Rul1an (commented):
## Quorum review (non-building) — READY

**Bound head SHA:** `cea8d4678bcb0ef42cf912e2eccaaed42cea6288`  
**Scope:** `scripts/ci/assay_runner_lane_check.py` vs issue #2212 only  
**Identity note:** GitHub actor matches PR author (`Rul1an`), so this is a `COMMENT` review naming the SHA rather than `APPROVE` (AGENTS.md).

### Verdict
**READY** — no blockers. Closes the #2212 guard-strength gap without changing production acceptance or classification.

### Findings
None at confidence ≥ 80.

### Checklist vs #2212
1. **Real function driven** — `_test_uncovered_paths_require_exact_head` calls `content_tree_proof_accepts_head` directly (not a stub/`_fake_*` replacement). Wired from `self_test()`.
2. **Stale uncovered `Cargo.lock` fails before downstream** — constructs `PullRequest(files=("Cargo.lock",), head_sha="new-head")` with proof `source.head_sha="proof-head"`; asserts rejection diagnostic `Cargo.lock: gated path is not covered by content-provenance trees` while `fetch_ref_for_diff` / `current_content_path_trees` raise if reached.
3. **Exact-head case passes** — same uncovered file with matching `head_sha` reaches stubbed comparison and is accepted.
4. **`try`/`finally` restores globals** — both success and injected-exception paths leave `fetch_ref_for_diff` and `current_content_path_trees` as the pre-test objects.
5. **Mutations bite** — independently verified in a detached review worktree at this SHA:
   - short-circuit replaced with `if False` → `AssertionError: stale uncovered proof reached content-tree comparison` (exit 1)
   - `!=` inverted to `==` → same protected boundary failure (exit 1)
   - production restored → `self-test ok` (exit 0)
6. **No production/classifier change** — diff is +59 lines of self-test only; `content_tree_proof_accepts_head` (1025–1044) and `uncovered_content_provenance_files` match `main`.

### Verification performed
- Confirmed live PR head == `cea8d4678bcb0ef42cf912e2eccaaed42cea6288`
- Detached review worktree at that SHA (not the writer worktree)
- `python3 scripts/ci/assay_runner_lane_check.py --self-test` → ok
- `python3 -m py_compile scripts/ci/assay_runner_lane_check.py` → ok
- Mutation evidence as above; worktree left clean after restore

### Non-claims
- No claim that stale proof carry has occurred in production.
- No change to gated paths / content-provenance coverage evaluated here.

Rul1an:
## Review carry record after upstream advance

Cursor's non-building review on `cea8d4678bcb0ef42cf912e2eccaaed42cea6288` carries to new head `a2baf8d8239a0a0fee57cf86a52ac6778acae0ba` under the two conditions in `AGENTS.md`:

1. Whole-tree equivalence:
   - `git merge-tree --write-tree cea8d4678bcb0ef42cf912e2eccaaed42cea6288 02a880389af121eb39d8453f94eac7b79dcd53f7`
   - expected tree: `7bc4d87faf4b067fdd15445fc76b73e0788ac1a6`
   - `git rev-parse a2baf8d8239a0a0fee57cf86a52ac6778acae0ba^{tree}`
   - actual tree: `7bc4d87faf4b067fdd15445fc76b73e0788ac1a6`
   - result: exact match; no manual conflict resolution.

2. Reviewed-file intersection:
   - reviewed PR file at old head: `scripts/ci/assay_runner_lane_check.py`
   - upstream advance files: `.github/workflows/wave6-nightly-safety.yml`, `scripts/ci/test-setup-rust-composite-contract.sh`
   - intersection: empty.

This was an upstream merge advance only, not rewritten history. The prior READY review remains the quorum artifact for the new head. Fresh GitHub CI is running on `a2baf8d8239a0a0fee57cf86a52ac6778acae0ba`.

PR #2340, opened 2026-08-12
feat(monitor): M3a CONFIG mixed-object ABI rejection (#187)

Rul1an:
## Summary
M3a only: fail closed on mixed/stale CONFIG object ABI **before** `Ebpf::load` via named ELF symbol `ASSAY_MONITOR_OBJECT_ABI`.

- Base: `03bb417cbad9d1c6f87206f9cd65144a0f134fe4` (`origin/main`)
- Head: `3ba77da72ccf3bd39ee9bd7c73291969c0249987`
- Worktree: `/Users/roelschuurkes/wt-187-monitor-object-abi`
- Branch: `cursor/187-monitor-object-abi`
- **Draft only** — not ready/merge

## Complexity exception (owner-approved **≤ +360**)
Final staged net **+349** (405 insertions / 56 deletions vs `origin/main`).

**Rationale:** both independent-review blockers are required for the same honest mixed-CONFIG-object claim; production hardening is small; excess is the shared hostile-ELF fixture plus distinct malformed-input cases. Acceptance tests were not split or deleted.

## Review dispositions (exact-head blockers on `9f38cef91`)
1. **P1 CONFIG ABI completeness — fixed.** Descriptor/digest now covers every used CONFIG identity (`KEY_OFFSET_*` 0..8, `KEY_MAX_ANCESTOR_DEPTH` 10, sendto/sendmsg 11..12, plus feature keys 100..103). `assay-ebpf` and `assay-monitor` `tracepoint.rs` consume shared constants (no duplicate numeric literals). Golden digest **`0x4db6b07a`**.
2. **P2 strict marker — fixed.** Accept exactly one defined `SymbolKind::Data` symbol of size 4 in a data/rodata `SectionKind`, with checked bounds. Rejects duplicate name, wrong kind, wrong/zero size, truncated data, non-data section.

## Complexity ledger
| Slice | What | Invariant |
|---|---|---|
| Shared descriptor macro | Private `MONITOR_CONFIG_KEYS` + exported `KEY_*` from one table | Sole CONFIG identity; digest input |
| Digest | FNV-1a name bytes + LE id → baked `ASSAY_MONITOR_OBJECT_ABI` | Bake/verify agree |
| Explicit exports | `pub use object_abi::{digest, SYMBOL, KEY_*…}` (no glob) | Deliberate public surface |
| Wire consumers | eBPF imports; `tracepoint` inserts via `KEY_OFFSET_*` | No second numeric identity |
| Strict ELF reader | Unique def, Data, size 4, data/rodata section, bounds | Unambiguous marker |
| Hostile fixtures | Synthetic ELF for missing/mismatch/match/dup/kind/size/trunc/section | Mutation-sensitive negatives |
| Error UX | `MonitorError::ObjectAbi` + rebuild guidance | Fail closed before `Ebpf::load` |
| Dep scope | Linux `object`; host `write` dev-dep | Tests on non-Linux hosts |
| Gating map | `object_abi.rs` → gates=all | Delegated proof classification |

**Out of scope (M3b):** P2 ELF program-set equality / `PROBE_PROGRAMS` mapping; sendto/sendmsg attach/remove; ObservationHealth / `is_complete`.

## `object` version
- Pin **`=0.36.7`** (aya’s line). Production under `[target.'cfg(target_os = "linux")'.dependencies]`; `[dev-dependencies]` with `write` for fixtures.

## RED evidence
- Incomplete descriptor (100..103 only) vs full expected table → fail (`.tmp/RED-P1-config-completeness.log`, local)
- Lax first-name-wins parser → duplicate/kind/size/non-data fixtures fail closed only after GREEN (`.tmp/RED-P2-strict-marker.log`, local)

## GREEN evidence (head `3ba77da72`)
- `cargo test --locked -p assay-common --lib object_abi` — pass
- `cargo test --locked -p assay-monitor --lib object_abi` — 8 pass, 1 ignored
- `cargo test --locked -p assay-monitor --lib` — 39 pass, 1 ignored
- `cargo fmt --all -- --check`; clippy `-D warnings` (assay-common/monitor); `cargo doc --no-deps`; `git diff --check`; gating map present
- Mutation: `KEY_OFFSET_SENDTO_SOCKADDR` 11→19 → golden/tuple fail (restored)
- Prior CI object `/private/tmp/assay-pr2340-exact-artifact/assay-ebpf.o`: structural marker read OK (`0x07362158`); verify fail-closed on digest mismatch vs new completeness digest (expected until rebuild)

## Real-artifact proof note
Bake must be re-run after this head (digest changed). Prior ignored-test proof on old digest does not credit the new ABI; delegated rebuild/proof remains outstanding.

## Claims / non-claims
**Claims:** mixed/stale CONFIG-key identity rejection (full used key set) for objects missing or mismatching a strict baked digest symbol, verified before Aya load.

**Non-claims:** no P2 program-set equality; no sendto/sendmsg attach/remove; no ObservationHealth / `is_complete` / public JSON changes; does not prove kernel consumer reads the right CONFIG key, attach success, or enforcement.

## Test plan
- [x] Unit fixtures (missing / mismatch / match+offset / duplicate / kind / size / trunc / non-data)
- [x] Completeness golden `0x4db6b07a` + literal name/id table
- [ ] Rebuild `target/assay-ebpf.o` + ignored bake test on new digest
- [ ] CI / delegated gates on head
- [ ] Non-building review quorum before ready

## Summary by CodeRabbit

* **New Features**
  * Added compatibility verification between the monitor and its eBPF object before loading.
  * Added clear diagnostics and rebuild guidance when compatibility checks fail.
  * Centralized monitor configuration and offset identifiers for consistent runtime behavior.

* **Bug Fixes**
  * Prevents loading incompatible or malformed eBPF objects.
  * Preserves offset resolution behavior while replacing error-prone numeric identifiers with named keys.


Rul1an:
**Corrections bound to head** `9f38cef91cdd82ccd412e24497a8aa694ccfb74f` (Codex: no blockers).

- Single-source private descriptor macro for `KEY_*` + digest; public surface is `pub use object_abi::*` (digest / `OBJECT_ABI_SYMBOL` / `KEY_*` only)
- Golden digest pin `0x07362158` (mutation-sensitive vs id/algorithm drift)
- Production `object` under Linux target deps; `write` kept as host-wide dev-dep
- Minimal lock delta; staged net **+225** (≤230)
- Real-object ignored test proven earlier (builder image reused; no second Docker image build)
- Remains **draft** — no ready/merge; no proof dispatch from this push

Rul1an:
## Advisory review: NOT independent, does NOT count for quorum

Bound to exact head `9f38cef91cdd82ccd412e24497a8aa694ccfb74f`, base `03bb417cb`.

**Disclosure first, because it decides how this comment should be weighted.** I co-authored the governing design for this slice in the preceding review thread: the four-state remediation-owner classification, the resolution that the descriptor should be structured with a single digest rather than one overloaded scalar, the placement before `Ebpf::load`, the RED/GREEN shape, and the `sendto`/`sendmsg` disposition. Reviewing my own design is not an independent check, so this is advisory only. It does not satisfy quorum and should not be counted toward it.

### No blockers

Four advisories, none blocking.

**1. Glob re-export widens the public surface by default.** `crates/assay-common/src/lib.rs:25` does `pub use object_abi::*;`. Today that exports `OBJECT_ABI_SYMBOL`, `monitor_object_abi_digest` and the four `KEY_*`. Any future `pub` item in that module joins the public API of a publishable crate without a decision being made. The macro generates the `KEY_*` as `pub`, so the glob is understandable; the point is that it extends itself.

**2. `MONITOR_CONFIG_KEYS` stays private, and that is right.** The descriptor drives the computation and only the scalar travels. This is the structured-vs-scalar question resolved correctly.

**3. The ELF-derived program inventory is not in this slice.** The design paired key-layout identity with a compiled-program inventory so the hand-maintained probe list would be checked against the object at load time. This slice ships the first half. That is a defensible scope cut, but it leaves `Unavailable` reachable through an object missing a program, and leaves the probe lists unbound to the compiled set. Worth recording as an explicit residual rather than closing the item.

**4. `baked_ebpf_object_exports_matching_abi_symbol` is `#[ignore]`.** It is the only test that closes bake and verify together, and it needs a real object. Unless a lane runs it on the delegated runner, the propagation chain is not proven in CI. The `assay_runner_gating_map.txt` line suggests this is handled; I did not verify that lane.

### Verification

Run in a detached temporary worktree at the reviewed SHA; my working tree was untouched and the worktree was removed afterwards.

```
assay-common  --lib object_abi   1 passed
assay-monitor --lib object_abi   3 passed, 1 ignored
```

Green alone does not show the tests bite, so I mutated:

| mutation | expected | measured |
|---|---|---|
| key id `103` -> `104` | red | FAILED |
| FNV constant `0x01000193` -> `0x01000195` | red | FAILED |
| gate disabled (`if observed != expected` -> `if false`) | red | FAILED |

The third is the one that matters: it confirms `mismatched_abi_symbol_rejected` tests the comparison rather than the presence of code.

### What I checked and found sound

The gate is the first statement in `LinuxMonitor::new`, before `Ebpf::load`, on the single funnel both `load_file` and `load_bytes` pass through. Fail-closed via `?`, carrying detail plus rebuild guidance.

Bounds handling in `file_offset_in_section` uses `checked_sub` for an address below the section base, `usize::try_from` for overflow, and `checked_add(4).filter(|&end| end <= len)` before the slice. No unchecked indexing on parsed input.

Single-source holds: one macro emits both the constants and the descriptor, so there is no second definition to drift. The eBPF side bakes the same `const fn` result via `#[used] #[no_mangle] #[link_section = ".rodata"]`.

Dependency scope is correct rather than accidentally correct: the runtime `object` dependency is `cfg(target_os = "linux")`, the dev-dependency is unconditional, and the module is `#[cfg(any(test, target_os = "linux"))]`, so the unconditional dev-dep is what makes the tests compile on a non-Linux developer machine. One `object` version, pinned to aya's `=0.36.7`.

### Residual risks

The marker proves that object and loader were built from the same key set. It does not prove that the correct consumer reads the correct key; that remains covered by the shared constant and review. It says nothing about enforcement effect. And the existing `MonitorEvent` size check stays what it is, post-attach and per record, and does not cover the configuration contract.

_Advisory only. Not independent, not quorum, not a READY signal._

Rul1an:
## Independent exact-head review: changes required

Reviewer: independent Codex subagent `Wegener`; authored neither this implementation nor its governing design/plan. Review is bound to exact head `9f38cef91cdd82ccd412e24497a8aa694ccfb74f` against base `03bb417cbad9d1c6f87206f9cd65144a0f134fe4`.

### Findings

1. **P1: digest omits most CONFIG key identities.** `assay-common/src/object_abi.rs` covers only keys `100..103`, while `assay-ebpf/src/main.rs` separately defines offset/config keys `0..8, 10..12` and `assay-monitor/src/tracepoint.rs` writes matching literals. A mixed object changing (for example) `KEY_OFFSET_SENDTO_SOCKADDR` keeps digest `0x07362158`, passes the gate, and reads/writes different CONFIG slots. This defeats the broad mixed-CONFIG-object rejection claim. Move all CONFIG key identities into the shared descriptor/constants and make both sides consume them.

2. **P2: parser accepts malformed/ambiguous marker symbols.** `assay-monitor/src/object_abi.rs` accepts the first matching symbol name without enforcing uniqueness, `SymbolKind::Data`, exact size `4`, or an appropriate data section. Reject duplicates and malformed aliases; add mutation-sensitive fixtures for duplicate name, wrong kind, wrong/zero size, truncated data, and non-data section.

### Verification performed

- `cargo test --locked -p assay-common --lib object_abi`: pass (1)
- `cargo test --locked -p assay-monitor --lib object_abi`: pass (3, 1 ignored)
- `cargo test --locked -p assay-monitor --lib`: pass (34, 1 ignored)
- fmt/clippy/diff/gating-map checks: pass
- existing real object parser test: pass; symbol present

### Verdict

**NOT READY** on `9f38cef91cdd82ccd412e24497a8aa694ccfb74f`. Any corrective push invalidates this review and requires a new exact-head independent review.

Rul1an:
**Head** `3ba77da72ccf3bd39ee9bd7c73291969c0249987` — dispositions for independent blockers on `9f38cef91`:

1. **P1 completeness:** all used CONFIG keys in shared descriptor/digest (`0x4db6b07a`); ebpf + tracepoint consume shared constants.
2. **P2 strict marker:** unique defined Data/size-4/data-section marker; fixtures for duplicate/kind/size/trunc/non-data.

Owner complexity exception **≤+360**; staged net **+349**. Draft; no ready/merge; no proof/review dispatch from this push. Prior review on `9f38cef91` is invalidated.

Rul1an:

**Reviewer:** Codex independent non-building subagent (authored neither implementation nor governing specification)  
**Head:** `3ba77da72ccf3bd39ee9bd7c73291969c0249987`  
**Base:** `03bb417cbad9d1c6f87206f9cd65144a0f134fe4`  
**Verdict:** **READY**  
**Findings:** none

The prior P1/P2 blockers are resolved:

- all 16 CONFIG identities now have one definition site and participate in the golden digest;
- marker parsing requires exactly one defined 4-byte data symbol in a data/rodata section with checked bounds;
- focused/full monitor tests, strict malformed-ELF mutations, format, clippy with `-D warnings`, Linux-target check, rustdoc, and `git diff --check` passed;
- the exact-head Kernel Matrix artifact `assay-ebpf.o` (SHA-256 `6fe02bcad070ff75791460366c1a7f164bcd2f1e9c3398886358f74e8f57fe06`) passed the ignored real-object ABI test;
- complexity is +405/-56, net +349, inside the recorded owner-approved ≤+360 exception, with the excess attributable to hostile-ELF fixture coverage rather than production branching.

Residual merge condition, not a code finding: required `lane-check/proof` must accept the exact-head delegated `gates=all` proof. This review does not claim that proof has completed.

PR #2347, opened 2026-08-13
refactor(monitor): make the stat map readback undeletable (#187)

Rul1an:
## What

`snapshot_stats` (`crates/assay-monitor/src/loader.rs`) read fourteen `STATS` fields inline and
then applied the four send `no_peer` / `non_ip_family` honesty counters in a **separate statement**
calling `apply_send_honesty_counters`. Both projections are now one function, `project_snapshot`,
which takes two readers and no aya types.

**Head SHA:** `ece4abcc54022e36be6a5600ba8181403af6c42a`
**Base SHA:** `a0c4d31d3a319dae625abaa11d354bf89f9370f1`
**Size:** +137 / -101 (net +36) — `lib.rs` +117/-31, `loader.rs` +20/-70.

## Why — the RED-before measurement

The separate statement was deletable with no consequence. Measured on the base commit before any
production edit, by deleting the `crate::apply_send_honesty_counters(...)` line in `loader.rs`:

```
$ cargo check --target x86_64-unknown-linux-gnu -p assay-monitor --all-targets
warning: function `apply_send_honesty_counters` is never used
  --> crates/assay-monitor/src/lib.rs:77:4
warning: `assay-monitor` (lib) generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.04s

$ cargo test -p assay-monitor
test result: ok. 53 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
```

Green. The projection had a unit test with a fake reader, so the test kept the orphaned function
alive and passing; its *binding* to the real `STATS` map was guarded by nothing. The only signal was
a `dead_code` warning about the now-unused function — which says nothing about the snapshot, and
which a deleter removing the function alongside its call site would not see at all.
`MonitorStatsSnapshot::default()` filled the four fields with `0`, so downstream the CLI prints
`no_peer=0` and the Runner archive note omits its suffix, byte-identical to a genuinely clean run.

## How the guard works

Two properties fall out of the types, not out of a test:

- **Deleting the projection call is a compile error.** `snapshot_stats` has no snapshot to return
  without it.
- **Swapping the two readers is a compile error.** They differ in return type — `u32` for `STATS`,
  `u64` for `SOCKET_STATS`.

`project_snapshot` builds `MonitorStatsSnapshot` from a full struct literal with no
`..Default::default()`, so dropping a field read is a missing-field compile error rather than a
silent zero. It is private to the crate root; `loader` is a child module and reaches it directly.

Two tests, both in `lib.rs` (not `loader.rs`, whose raw text `program_set.rs` asserts over via
`include_str!`):

- `every_snapshot_field_reads_its_own_key` — **totality** in one comparison. Each reader returns its
  own key (`|k| k` for `STATS`, `u64::from` for `SOCKET_STATS`), so a correct projection equals a
  literal naming every field exactly once as its own key constant. `MonitorStatsSnapshot` derives
  `PartialEq`/`Eq`, so no field can be omitted from the check, and a field added to the struct is a
  compile error in the test.
- `no_two_fields_read_the_same_key` — the identity reader only discriminates if the keys differ. This
  records what each reader is actually asked for and asserts 18 and 6 reads with no repeats per
  array, which checks distinctness on the projection itself rather than against a second
  hand-maintained list of the constants. (Verified separately against
  `crates/assay-common/src/lib.rs`: `MONITOR_STAT_*` are `0..=17`, `SOCKET_STAT_*` are `0..=5`.)

No live aya object, no kernel, no mock of `AyaArray`, and no source-text scan.

`MonitorStatsSnapshot` is unchanged, `#[derive(Default)]` included — no ripple into `assay-cli` or
`assay-runner-core` test code. `snapshot_stats`' signature is unchanged and `project_snapshot` is
private, so there is no public API change.

## RED criteria — five mutations, each applied and reverted

Re-run in full against the current head. All commands run from the worktree root.

| # | Mutation | Command | Observed |
|---|---|---|---|
| 1 | Delete the `Ok(crate::project_snapshot(...))` call in `snapshot_stats` | `cargo check --target x86_64-unknown-linux-gnu -p assay-monitor --all-targets` | **compile error** `error[E0308]: mismatched types` at `loader.rs:592` — "implicitly returns `()` as its body has no tail or `return` expression", expected `Result<MonitorStatsSnapshot, ...>`, found `()` |
| 2 | Swap the two reader arguments at the call site | `cargo check --target x86_64-unknown-linux-gnu -p assay-monitor --all-targets` | **compile error** — two `error[E0308]`: at `loader.rs:606` "expected `u32`, found `u64`", at `loader.rs:607` "expected `u64`, found `u32`" |
| 3a | Delete the `sendto_no_peer:` initializer in `project_snapshot` | `cargo test -p assay-monitor` | **compile error** `error[E0063]: missing field `sendto_no_peer` in initializer of `MonitorStatsSnapshot`` at `lib.rs:104` |
| 3b | Same field, softer form: replace the read with the constant `0` | `cargo test -p assay-monitor` | **test RED, both tests** — `test result: FAILED. 52 passed; 2 failed`. Totality: `left: … sendto_no_peer: 0 …` vs `right: … sendto_no_peer: 14 …`. Read-count: 17 `STATS` reads instead of 18 |
| 4 | Swap `MONITOR_STAT_SENDTO_NO_PEER` ↔ `MONITOR_STAT_SENDMSG_NO_PEER` | `cargo test -p assay-monitor` | **test RED** — `every_snapshot_field_reads_its_own_key … FAILED`, `test result: FAILED. 53 passed; 1 failed`; `left: … sendto_no_peer: 15, sendmsg_no_peer: 14 …` vs `right: … sendto_no_peer: 14, sendmsg_no_peer: 15 …` |
| 5 | Point `sendto_no_peer` at `SOCKET_STAT_EVENTS_EMITTED` | `cargo test -p assay-monitor` | **test RED, both tests** — `test result: FAILED. 52 passed; 2 failed`. Totality: `left: … sendto_no_peer: 4 …` vs `right: … sendto_no_peer: 14 …`. Read-count: `assertion left == right failed: two fields read the same key` (key `4` now read twice, colliding with `MONITOR_STAT_OPENAT_EVENTS_EMITTED`) |

Criterion 3 was specified as "test RED"; the full struct literal makes it strictly stronger — a
compile error. 3b is the same mutation in the form that would still compile, and it is caught. Every
mutation was reverted immediately after measurement; the committed tree is the unmutated one.

## GREEN — and what ran on which platform

`assay-monitor`'s `loader.rs` is `#[cfg(target_os = "linux")]` for `aya`, so it is not compiled at
all in a native macOS build. What was actually run, on **macOS (aarch64-apple-darwin, rustc/clippy
1.96.0)**:

- `cargo test -p assay-monitor` → `54 passed; 0 failed; 1 ignored` (was 53; one test replaced by two)
- `cargo clippy -p assay-monitor --all-targets -- -D warnings` → clean
- `cargo fmt --all -- --check` → clean

For the Linux-only path, on the same macOS host by cross-compilation to
**`x86_64-unknown-linux-gnu`** (type-check only, no linking, no execution):

- `cargo check --target x86_64-unknown-linux-gnu -p assay-monitor --all-targets` → clean, no warnings
- `cargo clippy --target x86_64-unknown-linux-gnu -p assay-monitor --all-targets -- -D warnings` → clean

Mutations 1 and 2 are compile-time properties of the Linux-gated code, so the cross-target check is
the right instrument for them and is what was used. Mutations 3–5 are properties of
`project_snapshot`, which is `#[cfg(any(test, target_os = "linux"))]` and therefore genuinely
executed on macOS.

**Not run:** no test binary was executed on Linux at all. `cargo check --target
x86_64-unknown-linux-gnu -p assay-cli` fails on this host for an unrelated reason (a dependency's
build script needs `x86_64-linux-gnu-gcc`, absent here), so the downstream consumer was not
cross-checked; the change is crate-internal with an unchanged public surface, but that is an
argument, not a measurement. A macOS pass is **not** coverage of the Linux-only path.

## Non-claims

- This establishes **nothing** about whether the `STATS` map exists in the loaded eBPF object,
  whether aya successfully binds `AyaArray::try_from` to it, or whether the kernel writes what we
  read. All three still require a live kernel and a real load; nothing here substitutes for that.
- `unwrap_or(0)` still converts a map-read failure into a zero. A missing or unreadable key remains
  indistinguishable from a genuine zero at every call site. This change only guarantees that the
  read is *attempted* for every field and lands in the *right* field.
- The tests prove the key→field wiring, not that any `MONITOR_STAT_*` constant matches what the eBPF
  side actually increments. Userspace and kernel share the constants from `assay-common`, so they
  cannot disagree about the number, but nothing here checks that a given hook bumps the counter its
  name implies.
- No behaviour change is claimed. The projection reads the same keys into the same fields as before;
  this is a refactor that changes what a future deletion costs, not what a current run reports.

### DIFF
--- a/crates/assay-monitor/src/lib.rs
+++ b/crates/assay-monitor/src/lib.rs
@@ -73,46 +73,280 @@ pub struct MonitorStatsSnapshot {
     pub event_size_mismatch: u64,
 }
 
+/// The whole map-to-snapshot projection, in one function, over two readers.
+///
+/// This used to be fourteen inline reads in `snapshot_stats` plus a second statement that applied
+/// the four send honesty counters. That second statement was deletable with no consequence: the
+/// projection had a unit test with a fake reader, but its *binding* to the real `STATS` map was
+/// guarded by nothing, and `MonitorStatsSnapshot::default()` filled the orphaned fields with 0 --
+/// byte-identical, downstream, to a genuinely clean run.
+///
+/// Collapsing both projections here makes the binding fall out of the types rather than out of a
+/// test. `snapshot_stats` has no snapshot to return without calling this, so deleting the call is a
+/// compile error; and the two readers differ in return type (`u32` for `STATS`, `u64` for
+/// `SOCKET_STATS`), so swapping them is a compile error too. Neither map type appears in the
+/// signature, so the projection stays testable on any host while the aya types stay in `loader.rs`.
+///
+/// Every field is written by a listed initializer -- no `..Default::default()` -- so dropping a
+/// read is a missing-field compile error rather than a silent zero.
 #[cfg(any(test, target_os = "linux"))]
-fn apply_send_honesty_counters(stats: &mut MonitorStatsSnapshot, mut read: impl FnMut(u32) -> u32) {
+pub(crate) fn project_snapshot(
+    mut read_stats: impl FnMut(u32) -> u32,
+    mut read_socket: impl FnMut(u32) -> u64,
+    event_size_mismatch: u64,
+) -> MonitorStatsSnapshot {
     use assay_common::{
-        MONITOR_STAT_SENDMSG_NON_IP_FAMILY, MONITOR_STAT_SENDMSG_NO_PEER,
-        MONITOR_STAT_SENDTO_NON_IP_FAMILY, MONITOR_STAT_SENDTO_NO_PEER,
+        MONITOR_STAT_CONNECT_EVENTS_EMITTED, MONITOR_STAT_CONNECT_RINGBUF_DROPPED,
+        MONITOR_STAT_LSM_EVENTS_EMITTED, MONITOR_STAT_LSM_RINGBUF_DROPPED,
+        MONITOR_STAT_OPENAT2_EVENTS_EMITTED, MONITOR_STAT_OPENAT2_RINGBUF_DROPPED,
+        MONITOR_STAT_OPENAT_EVENTS_EMITTED, MONITOR_STAT_OPENAT_RINGBUF_DROPPED,
+        MONITOR_STAT_SENDMSG_EVENTS_EMITTED, MONITOR_STAT_SENDMSG_NON_IP_FAMILY,
+        MONITOR_STAT_SENDMSG_NO_PEER, MONITOR_STAT_SENDMSG_RINGBUF_DROPPED,
+        MONITOR_STAT_SENDTO_EVENTS_EMITTED, MONITOR_STAT_SENDTO_NON_IP_FAMILY,
+        MONITOR_STAT_SENDTO_NO_PEER, MONITOR_STAT_SENDTO_RINGBUF_DROPPED,
+        MONITOR_STAT_TRACEPOINT_EVENTS_EMITTED, MONITOR_STAT_TRACEPOINT_RINGBUF_DROPPED,
+        SOCKET_STAT_ALLOWED, SOCKET_STAT_BLOCKED_CIDR, SOCKET_STAT_BLOCKED_PORT,
+        SOCKET_STAT_CHECKS, SOCKET_STAT_EVENTS_EMITTED, SOCKET_STAT_RINGBUF_DROPPED,
     };
 
-    stats.sendto_no_peer = read(MONITOR_STAT_SENDTO_NO_PEER);
-    stats.sendmsg_no_peer = read(MONITOR_STAT_SENDMSG_NO_PEER);
-    stats.sendto_non_ip_family = read(MONITOR_STAT_SENDTO_NON_IP_FAMILY);
-    stats.sendmsg_non_ip_family = read(MONITOR_STAT_SENDMSG_NON_IP_FAMILY);
+    MonitorStatsSnapshot {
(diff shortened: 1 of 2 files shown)


Rul1an:
## Independent exact-head review record

- PR head reviewed: `ece4abcc54022e36be6a5600ba8181403af6c42a`
- Reviewer: Cursor read-only reviewer; not the builder and authored/edited neither this change nor its governing specification.
- Isolation: detached scratch tree; Claude's branch was not modified.
- Verdict: **READY**
- Findings: no blockers or important findings.

The review checked snapshot totality, `STATS` (`u32`) versus `SOCKET_STATS` (`u64`) reader binding, map/default behavior parity, cfg/platform surface, public API, test tautology/key collisions, and the five PR-body mutations. All five mutations discriminated. Measured locally: `cargo test -p assay-monitor` 54 passed; clippy and fmt clean. Linux-only code received cross-target type-check/clippy; no Linux execution was claimed by the reviewer.

Exact-head delegated proof: [Runner Spike Delegated 31677893778](https://github.com/Rul1an/assay/actions/runs/31677893778), `gates=all`, success on the same SHA.

PR #2352, opened 2026-08-13
Carry the claim ceiling into verify-side-effects, and share the fold

Rul1an:
`verify-side-effects` called the coding-agent claim gate, took `.decision`, and dropped the ceiling rung that came with it. A `verified` call and an `asserted` one therefore reached a consumer as a bare verdict, and the ordering the gate exists to encode decided nothing anyone could read.

Each asserting call now reports `occurrence_ceiling`, and the report carries `weakest_occurrence_ceiling` across the calls that asserted a side effect — a fold, never a maximum, because one corroborated call does not raise what the run as a whole supports.

## What an adversarial review changed, and it was the better half of this PR

**The run-level field conflated two opposite states.** As an `Option` skipped when empty it was absent both when a refutation collapsed the run and when no call asserted anything at all. That is the occurrence-versus-absence distinction this command exists to keep, lost inside the field meant to carry it. It is now three-state (`nothing_claimed` / `blocked` / `rung`) and always emitted.

**The fold was a copy, and it had already drifted when it was written.** `CodingAgentCoverageReport::weakest_ceiling` iterates a container that is never empty, so `?`-means-blocked is sound there; the copy filtered on `asserted` and could be empty, so it was not. There is now one implementation, `coding_agent_weakest_ceiling` in `assay-evidence`, used by both callers — per the one-rule-one-function rule in AGENTS.md.

**A rung was published on rows that asserted nothing.** `occurrence_claim` has been computed for every row since before this change, but a ladder position grades a claim, and a call that asserted no side effect has not made one.

## Measurements

Worktree off `origin/main` `e07b93d44`, rustc 1.96.0, darwin, `cargo test -p assay-cli --test verify_side_effects_cli`, 17 tests.

| Mutation | Result |
|---|---|
| ladder flattened, `verified` → `producer_reported` | 15 pass, **2 fail** |
| shared fold `min` → `max` | 16 pass, **1 fail** |
| `blocked` collapsed into `nothing_claimed` | 15 pass, **2 fail** |
| rung published on non-asserting rows | 16 pass, **1 fail** |
| `asserted` arm raised | 17 pass, **survives** |
| `observed_confirmed` arm changed | 17 pass, **survives** |

The last two survive and no test will kill them: the `SelfReported` branch caps at `Asserted` whatever source class it is handed, and `row.level` is never `ObservedConfirmed` in this command. Equivalent mutants rather than coverage gaps — which also narrows the claim: the ladder is now observable for the `Verified` arm, not for all three.

An earlier draft of the commit message reported 15-of-16 for the first row. That was wrong; it was 14 of 16 against the then-16-test file, and the correction is in the commit message rather than quietly dropped.

## Scope

Nothing gates on this and the command still exits 0. This makes the ladder observable, not binding; an exit code is a separate decision with its own compatibility question. Both new fields carry `skip_serializing_if` or are additive, and nothing outside the producing crate parses this report.

`cargo test -p assay-evidence` 317 passed; full `assay-cli` suite green; `cargo fmt --all -- --check` and `cargo clippy` clean.

### DIFF
--- a/crates/assay-cli/src/cli/commands/evidence/verify_side_effects.rs
+++ b/crates/assay-cli/src/cli/commands/evidence/verify_side_effects.rs
@@ -27,8 +27,9 @@ use crate::exit_codes;
 use anyhow::{Context, Result};
 use assay_evidence::bundle::BundleReader;
 use assay_evidence::{
-    coding_agent_claim_decision, CodingAgentClaimKind, CodingAgentCoverageState,
-    CodingAgentGateDecision, CodingAgentSourceClass,
+    coding_agent_claim_decision, coding_agent_weakest_ceiling, CodingAgentClaimCeiling,
+    CodingAgentClaimDecision, CodingAgentClaimKind, CodingAgentCoverageState,
+    CodingAgentGateDecision, CodingAgentSourceClass, CodingAgentWeakestCeiling,
 };
 use assay_mcp_server::side_effect::{
     check_audit_record, AuditBinding, SideEffectLevel, PROVIDER_AUDIT_RECORD_SCHEMA,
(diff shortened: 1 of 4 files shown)


Rul1an:
## Exact-head independent review: BLOCKED

Reviewer: independent non-building subagent `Helmholtz`
Head: `b9f8f1f8b7203b63b828f60bff760a25eb6316ae`
Base: `32acd848dc7d1cccbd2acb50016461e967f8ce8c`

Findings:
1. **P1:** absent or malformed decision evidence can serialize as `nothing_claimed`; missing event/surface/bool cases need fail-closed validation.
2. **P1:** one bound audit record can promote multiple identical calls because matching does not exclude an already consumed record.
3. **P2:** the public fold still accepts `Option<Ceiling>`, leaving `None` ambiguous despite the new three-state result.

Verdict: **BLOCKED** on this SHA. `git diff --check` passed and the merge-only upstream advance was tree-equivalent/conflict-free. The reviewer authored neither the change nor a governing spec/plan.

Disposition: all three findings reproduced against the implementation and are being fixed test-first by the branch owner. Any push invalidates this review; a new exact-head review is required.

Issue #2354, opened 2026-08-14
Fail closed on malformed claim decisions and one-to-one audit allocation


Rul1an:
## What

PR #2352 was squash-merged as `b34bc2f8ef5d97d2ec3d4988852cba90ff9b396f` before three independently identified review findings were included. The fixes exist locally as commit `01b9ea76022d8bc460e909e41a94a8e3bf3a9048`, but that commit is not on `main`.

The findings share one boundary: evidence verification must not promote or sanitize a claim when required decision/audit evidence is absent, malformed, ambiguous, or reused.

## Findings

1. **Malformed decision evidence can become a clean non-claim.**
   - no decision-surface event;
   - missing or non-array `observed_tool_decisions`;
   - missing or non-boolean `response.side_effect_asserted`.
   These must fail closed instead of silently yielding `NothingClaimed`.

2. **One audit record can vouch for multiple identical calls.**
   Matching must allocate each audit record to at most one call. Two identical asserting calls plus one valid record must yield one promotion and one unpromoted call.

3. **The shared weakest-ceiling fold accepts ambiguous optional ceilings.**
   It must consume typed `CodingAgentClaimDecision` values. `NothingClaimed` is reserved for an empty iterator; a blocked or malformed decision must not become a non-claim. Because the type is deserializable, `Blocked + Some(rung)` must remain blocked and `Allowed/Degraded + None` must fail closed.

Original review record: https://github.com/Rul1an/assay/pull/2352#issuecomment-5289991570

## Definition of Done

- [ ] failing behavioral tests are recorded for all malformed/missing decision forms;
- [ ] a one-record/two-identical-calls test proves one-to-one allocation;
- [ ] the public fold consumes typed decisions and tests inconsistent deserialized field combinations;
- [ ] only an empty decision iterator yields `NothingClaimed`;
- [ ] focused CLI integration, `assay-evidence`, fmt, clippy `-D warnings`, and diff checks pass;
- [ ] one independent non-building review records a verdict on the exact final head;
- [ ] required GitHub checks are green and any proof is bound to the final head.

## Non-claims

This issue does not change evidence wire fields, bundle serialization, provider-outcome verification, or the ADR-042/043 boundaries. It hardens interpretation of evidence already carried.

Rul1an:
## Closure record

Fixed by #2355.

- reviewed final head: `709cc965e39d5b7c5bb107530edcf7b87edf2a2c`
[...]

PR #2355, opened 2026-08-14
Fail closed on malformed claim decisions, one-to-one audit allocation, typed ceiling fold

Rul1an:
Refs #2354

Follow-up to #2352. The claim-ceiling work merged as `b34bc2f8e`; an independent review of it then produced three verified blockers, and the fixes were never pushed. This is that fix, plus a fourth found while writing it.

## Provenance

| | |
|---|---|
| Base | `b34bc2f8ef5d97d2ec3d4988852cba90ff9b396f` (origin/main) |
| Head | `709cc965e39d5b7c5bb107530edcf7b87edf2a2c` |
| Cherry-picked from | local `01b9ea76022d8bc460e909e41a94a8e3bf3a9048`, then amended |

`01b9ea760` was authored against `b9f8f1f8b`, the pre-squash head of #2352, and was never pushed. It cherry-picked onto `b34bc2f8e` with no conflicts, since the squash merge preserved that tree. The amend adds the fourth fix below and nothing else.

## The three blockers

**1. Malformed decision evidence read as clean.** A decision event whose `observed_tool_decisions` was missing or not an array was skipped, and `response.side_effect_asserted` defaulted to `false` when missing or not a boolean. Each turned a malformed bundle into an empty or non-asserting set, which then reported as a run that claimed nothing. Absence masquerading as clean is the conflation this command exists to prevent. All four now fail with the event id or tool name in the message, and a bundle carrying no decision-surface events at all is rejected rather than reported empty.

**2. One audit record could vouch for several calls.** The record loop did not skip records already bound to an earlier call, so two otherwise identical asserting calls could both reach `verified` on the same imported record. Corroboration was overcounted by construction. Allocation is now one-to-one.

**3. The fold could not tell blocked from absent.** `coding_agent_weakest_ceiling` took `Option<CodingAgentClaimCeiling>`, so `None` meant either "this decision is blocked" or "the caller had no ceiling to give". It now takes `&CodingAgentClaimDecision` and reads the state from the decision itself.

## The fourth, found while fixing the third

The new signature still decided on `decision.ceiling` alone. `CodingAgentClaimDecision` derives `Deserialize`, so a record arriving from outside the crate can carry `Blocked` together with a rung, and the fold would have read the rung and ignored the verdict that blocked it.

The verdict is now read first. The mirror case, `Allowed` or `Degraded` with no rung, fails closed to `Blocked` and deliberately **not** to `NothingClaimed`: a claim was made and could not be graded, so reporting a non-claim would invent an absence the input never stated. Only an empty iterator yields `NothingClaimed`.

One rule, one function: the decision lives in `coding_agent_weakest_ceiling` and both callers inherit it.

## Evidence that the tests bite

`a_blocked_verdict_carrying_a_rung_is_still_blocked` carries the top rung on purpose. With the verdict check removed it reports `IndependentlyConfirmed`, the strongest possible answer, from the input least entitled to it.

Removing that check fails **exactly that test and no other**:

```
test a_blocked_verdict_carrying_a_rung_is_still_blocked ... FAILED
test result: FAILED. 7 passed; 1 failed
```

Its sibling `an_allowed_verdict_without_a_rung_fails_closed_rather_than_reading_as_nothing_claimed` **survives** that same mutation, because the prior code also returned `Blocked` on a missing ceiling. It guards the behaviour rather than the new branch, and is kept on that basis rather than counted as coverage it does not provide.

## Verification at `709cc965e`

| Check | Result |
|---|---|
| `cargo test -p assay-evidence` | 325 passed, 0 failed |
| `cargo test -p assay-cli --test verify_side_effects_cli` | 22 passed, 0 failed |
| `cargo clippy -p assay-evidence -p assay-cli --all-targets -- -D warnings` | exit 0 |
| `cargo fmt --all -- --check` | exit 0 |
| `git diff --check origin/main...HEAD` | clean |
| Files touched | 3 |

macOS, rustc 1.96.0. Local runs; CI on this head is the authority.

## Not claimed

- No wire field, serialization, or exit-code changes.
- Nothing here gates a run. `verify-side-effects` still exits 0 on a completed report; the ladder is observable, not binding.
- The new failures are refusals to report on input that cannot be read. They are not verdicts about an agent.
- The report schema string stays `assay.side_effect_verification.v0`.

Draft, and `Refs` rather than `Closes` so #2354 stays open until this is reviewed and merged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Side-effect verification now fails safely when evidence is missing or malformed.
  - Each audit record can corroborate only one matching call, preventing duplicate validation.
  - Refuted assertions are correctly treated as blocked.
  - Run-level results now preserve the weakest applicable verification ceiling.
- **Tests**
  - Added coverage for malformed evidence, missing decisions, invalid fields, duplicate audit matches, blocked decisions, and ceiling calculations.

### DIFF
--- a/crates/assay-cli/src/cli/commands/evidence/verify_side_effects.rs
+++ b/crates/assay-cli/src/cli/commands/evidence/verify_side_effects.rs
@@ -109,6 +109,11 @@ struct CallRow {
     /// for a reader, which is why this stays an `Option` here while the run-level answer does not.
     #[serde(skip_serializing_if = "Option::is_none")]
     occurrence_ceiling: Option<CodingAgentClaimCeiling>,
+    /// The full occurrence decision for this call, used by the ceiling fold. Kept alongside the
+    /// serialized `occurrence_claim` and `occurrence_ceiling` so the fold consumes the decision
+    /// directly (Blocker 3: no more `Option<Ceiling>` ambiguity).
+    #[serde(skip)]
+    occurrence_decision: CodingAgentClaimDecision,
     /// What a below-harness observer could say about the claimed egress, when one was supplied.
     #[serde(skip_serializing_if = "Option::is_none")]
     egress: Option<EgressRefutation>,
@@ -200,12 +205,12 @@ fn claim_decision_for(
 /// because it did some reading. That filter is exactly why this set can be empty and why the shared
 /// fold has to distinguish empty from blocked.
 fn weakest_occurrence_ceiling(calls: &[CallRow]) -> CodingAgentWeakestCeiling {
-    coding_agent_weakest_ceiling(
-        calls
-            .iter()
-            .filter(|c| c.asserted)
-            .map(|c| c.occurrence_ceiling),
-    )
+    let asserting_decisions: Vec<&CodingAgentClaimDecision> = calls
+        .iter()
+        .filter(|c| c.asserted)
+        .map(|c| &c.occurrence_decision)
+        .collect();
+    coding_agent_weakest_ceiling(asserting_decisions)
 }
 
 /// Load `assay.provider_audit_record.v0` files from an import directory.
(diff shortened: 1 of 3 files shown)


Rul1an:
## Independent exact-head review: READY

**Reviewed head:** `709cc965e39d5b7c5bb107530edcf7b87edf2a2c`  
**Base:** `b34bc2f8ef5d97d2ec3d4988852cba90ff9b396f`

### Findings

No correctness, security, or test-adequacy findings.

Verified:

- malformed/missing decision evidence covered by #2354 is rejected before `NothingClaimed`;
- each audit-record index is allocated at most once;
- the typed decision is rebuilt after refutation, so serialized verdict/rung and fold input stay aligned;
- the fold reads verdict before ceiling and only an empty iterator yields `NothingClaimed`.

### Compatibility disposition

The fold signature is source-breaking relative to the immediate base, and an old-call-site probe fails with `E0271`. This API was introduced by the immediately preceding, unreleased merge `b34bc2f8e`; no release tag contains it (`v5.1.0-74-gb34bc2f8ef5d`). It therefore does not break a published API. `cargo-semver-checks 0.50.0` reported 223 checks passed but did not detect this generic iterator item-type change; the direct compile probe is the controlling evidence.

### Verification

- `cargo test -p assay-evidence weakest_ceiling_tests`: 8 passed
- `cargo test -p assay-cli --test verify_side_effects_cli`: 22 passed
- `cargo fmt --all -- --check`: passed
- `cargo clippy -p assay-evidence -p assay-cli --all-targets -- -D warnings`: passed
- `cargo semver-checks check-release -p assay-evidence --baseline-rev b34bc2f8ef5d97d2ec3d4988852cba90ff9b396f`: passed, limitation noted above
- `git diff --check b34bc2f8ef5d97d2ec3d4988852cba90ff9b396f..HEAD`: clean

### Independence

Reviewer authored neither this change, issue #2354, nor a normative specification/implementation plan governing it; review was read-only in a clean detached clone.

Issue #2206, opened 2026-08-10
Doctor names a missing config a parse failure and advises re-running itself


Rul1an:
Origin: #2160 · Related: #2177

## Driven observation

Measured in worktree `/Users/roelschuurkes/wt-2160` on `f221ad0633954a8cce63e4f5a08b8ceaef935029`
(branch `cursor/2160-doctor-json-contract`, which closes #2160), own debug build via
`CARGO_TARGET_DIR=/tmp/assay-2160-target`, `NO_COLOR=1`, `ASSAY_EXIT_CODES` unset, no TTY.

One input — a config path that does not exist — read by two commands:

```text
$ assay run --config missing.yaml --format json
exit=2  reason_code=E_MISSING_CONFIG  next_step="Run: assay init to create a config file"

$ assay doctor --format json --config missing.yaml
exit=2  reason_code=E_CFG_PARSE      next_step="Run argv: [\"assay\",\"doctor\",\"--config\",\"missing.yaml\"]"
        config_error.message="ConfigError: failed to read config missing.yaml: No such file or directory (os error 2)"
```

Two defects share one root cause: the recovery step is keyed only on the reason code, and the
reason code is chosen without asking whether the file was absent or merely unreadable.

**The identity conflates two conditions.** `doctor` reports `E_CFG_PARSE` for a file it never
parsed. Its own message says `failed to read`, so the published code contradicts the published
message, and `run` classifies the identical input as `E_MISSING_CONFIG` on the same build. The
`E_CFG_PARSE` code on `config_error.code` predates #2160 and is unchanged by it; what #2160
changed is that this identity now also reaches the top-level `reason_code` an agent branches on.

**The recovery is self-referential.** `E_CFG_PARSE` advises `assay doctor --config <path>`, which
is the right escalation when `assay run` observed the failure and a no-op when `doctor` did. An
agent that follows the published step re-runs the command that just failed and receives a
byte-identical report. For the absent-file case the registry already holds the useful answer,
`E_MISSING_CONFIG` → `Run: assay init to create a config file`, and it is not reached.

## Why this is not MVP-blocking

The exit class is correct and frozen-table-consistent on both commands, so an agent branching on
the exit code is not misled — that was #2160's defect and it is fixed. This one costs precision,
not correctness of class: the agent still has `config_error.message`, and `assay run`, the
golden-path command, already classifies correctly. `assay policy validate` has published an
equally self-referential recovery since #2198, so the looping-recovery shape is a pre-existing
[...]

PR #2370, opened 2026-08-14
fix(cli): publish recovery argv that clap can execute

### LINKED ISSUE #2216: Published recovery argv fails to execute for -prefixed paths and is format-blind
Parent: #2177 · Found reviewing #2160 (PR #2207)

Two defects in one function, filed together because they are two answers from the same string and fixing them separately would mean two changes to the same lines.

`format_recovery_argv` in `crates/assay-cli/src/exit_codes.rs` builds the `next_step` published for `E_CFG_PARSE` (and, through the same helper, `E_POLICY_PARSE`). It is shell-injection-safe: the argv is JSON-encoded, so quotes, newlines, tabs, backslashes, ANSI escapes and U+2028 all round-trip byte-for-byte, which was verified across eleven hostile filenames. Neither defect below is an injection.

## Defect 1: the published recovery does not execute for a path beginning with `-`

Measured on `26f223267a2427017cdf4cf26c455dd0e2811ea7`:

```text
$ printf 'not: [broken\n' > -weird.yaml
$ assay doctor --format json --config=-weird.yaml
exit: 2
next_step: Run argv: ["assay","doctor","--config","-weird.yaml"]

$ # that argv, executed verbatim without a shell:
exit: 2
stdout: 0 bytes
stderr: error: unexpected argument '-w' found
```

Same for files named `--config` (`a value is required for '--config <CONFIG>'`) and `-h`. JSON-quoting a path protects against the *shell* reading it as syntax; it does nothing about `clap` reading it as an option. So the contract's claim that the report "names the failing file in a concrete JSON argv next step" is false for this class of path: the argv names the file and then cannot run.

Candidate fix, confirmed accepted by the binary: emit the flag and value fused as one element, `["assay","doctor","--config=-weird.yaml"]`. Whether to fuse always or only when the value starts with `-` is the design question — fusing always is one rule instead of a conditional, and the conditional is a second place the rule could drift.

## Defect 2: the recovery is format-blind, so a JSON consumer is handed text

```text
$ assay doctor --format json --config bad.yaml
exit: 2
next_step: Run argv: ["assay","doctor","--config","bad.yaml"]

$ assay doctor --config bad.yaml
831 bytes of human text: "Assay v5.0.0 ──── Platform: …"
```

An agent that parsed the failure report and followed the recovery it was given cannot parse the result. The same holds for `assay run --format json`, whose `E_CFG_PARSE` recovery also hands back a text-mode `doctor` invocation.

This is the harder half, and it is a real design question rather than a typo. The registry builds the string from `(ReasonCode, context)` and deliberately knows nothing about the caller — that format-blindness is what keeps one reason from having two remediations. Making the recovery carry `--format json` means either the registry learns the caller's output format, or the caller post-processes a string the registry owns. The second is the drift shape `AGENTS.md` warns about; the first widens the registry's interface. A third option is to leave the argv text-mode and say so in the contract, which is honest but concedes that the machine-readable path terminates in a human-readable one.

## Why beyond MVP

Both are narrow. Defect 1 needs a config path starting with `-`, which an agent following the golden path does not produce. Defect 2 degrades a recovery rather than falsifying a result — the failure report itself is correct, machine-readable, and carries the class; only the suggested next command is less useful than it looks. Neither publishes a false claim about *what happened*, which is the line that separates these from #2215.

Adjacent and deliberately not folded in: #2206 (doctor's recovery is the invocation that produced the failure) and #2208 (usage rejections return the test-failure class).

## Acceptance

- A `-`-prefixed config path publishes a `next_step` that executes and reaches the same diagnosis as a plain path, proven by executing the published argv rather than by inspecting it.
- `crates/assay-cli/tests/config_recovery_argv_contract.rs` gains that case. It fails today: the existing assertion compares the recovery's exit code to the failure's, and both are `2` for a `-`-prefixed path — one because the config is broken, one because `clap` refused — so the current test passes on a recovery that never ran. The new case must assert the recovery produced the diagnosis, not merely a matching number.
- The format question is decided explicitly and recorded, in the registry rather than at a call site, or the contract states that the recovery is text-only.
- Both `assay run` and `assay doctor` move together; they read the same registry string, and a fix that reaches only one of them puts two answers where there is now one.

Rul1an:
## Summary

Refs #2216.

Published `E_CFG_PARSE` / `E_POLICY_PARSE` recovery named the failing file and then could not run: a split `--config` / path pair lets clap reread a `-prefixed` operand as an option, and a JSON consumer that followed the recovery was handed text. Both answers came from one registry string, so they move together.

- Path-bearing flags are always fused (`--config=path`, `--input=path`) in one helper. Fusing only when the value starts with `-` would be a second place the rule could drift.
- `--format json` is decided in the registry, not at a call site, matching `E_EVIDENCE_UNREADABLE`. JSON and text publish the same argv. Executing it produces a parseable diagnosis.
- Proven by executing the published argv, not by matching exit `2` (clap refusal is also `2`).

**Base:** `1d82aada5ce6e773c708bb747cb7230191ab36f2` (`origin/main`)
**Head:** `e4bf16f74325623498578d273f53686cba319f51` (`cursor/2216-recovery-argv`)
**Worktree:** `/Users/roelschuurkes/wt-2216-recovery-argv`

## RED → GREEN

RED (exact `origin/main` binary, before the production change):

```text
cargo test -p assay-cli --test config_recovery_argv_contract -- \
  dash_prefixed_config_recovery_reaches_the_diagnosis \
  run_and_doctor_publish_the_same_executable_recovery_for_a_dash_prefixed_path \
  json_and_text_publish_the_same_recovery_and_executing_it_stays_json
```

- `-weird.yaml`: `unexpected argument '-w' found` (clap refused; empty stdout)
- `bad.yaml` format-parity: recovery stdout was human doctor text (`Assay v5.2.0` / `Config Status: FAILED`), not JSON

GREEN (same tests after the registry change): 20 passed in `config_recovery_argv_contract`; policy contract and golden-path contract also green.

## Test plan

- [x] Focused RED tests failed for clap refuse / text recovery, not for fixture setup
- [x] `cargo test -p assay-cli --test config_recovery_argv_contract`
- [x] `cargo test -p assay-cli --test policy_validate_machine_contract`
- [x] `cargo test -p assay-cli --test agent_golden_path_contract`
- [x] `cargo test -p assay-cli --bin assay -- path_bearing_recovery dynamic_recovery_argv test_reason_code_next_step path_bearing_errors pathless_config folding_an_upstream`
- [x] `cargo test -p assay-cli` (full crate)
- [x] `git diff --check`
- [x] `python3 scripts/docs/generate-agent-golden-path.py --check`
- [x] `python3 scripts/ci/test-agent-golden-path-skill.py`
- [ ] Non-building agent review on this head (builder self-review does not count)

Hostile cases covered: `-weird.yaml`, `--config`, `-h`, spaces, `$(...)`, quotes, newlines, tabs, BEL. Recovery is JSON-encoded argv (not a shell string); fused `--config=-weird.yaml` is copy-pasteable as one element.

## Non-claims

- Not #2206. Doctor still names a missing file `E_CFG_PARSE` and still advises `doctor` on the same path. This PR only changes how that path is bound and that the recovery stays JSON. The missing-vs-parse identity and the self-loop are #2206's.
- Not a general shell parser. Consumers must parse the `Run argv:` JSON array and exec it without a shell.
- Not a claim that every argv is safe. Encoding and fusion cover the published recovery string; they do not sanitize the file the command then reads.
- Not a claim that a user-typed split `--config -weird.yaml` on the original invocation works. The original command still needs `--config=path` (or equivalent) for clap to accept a `-prefixed` operand. This PR fixes the *published* recovery.
- Positional `-prefixed` paths (for example `assay evidence show -bundle.tar.gz`) are out of scope.
- Static prose remediations (`E_MISSING_CONFIG`, `E_TRACE_NOT_FOUND`, …) are unchanged.

## Review

Draft. Do not merge. Do not mark ready. Do not close #2216.

AGENTS.md review quorum: one non-building agent must review this exact head and record a verdict. This writer's review does not count.

### DIFF
--- a/crates/assay-cli/src/cli/commands/pipeline_error.rs
+++ b/crates/assay-cli/src/cli/commands/pipeline_error.rs
@@ -284,7 +284,13 @@ mod tests {
         let diagnostic = diagnostic_for(&config, ReasonCode::ECfgParse);
         assert_eq!(
             recovery_argv(&diagnostic.fix_steps[0]),
-            ["assay", "doctor", "--config", "suites/prod.yaml"]
+            [
+                "assay",
+                "doctor",
+                "--config=suites/prod.yaml",
+                "--format",
+                "json"
+            ]
         );
         assert_eq!(
             outcome_as_written(&config, ReasonCode::ECfgParse).next_step,
@@ -311,7 +317,13 @@ mod tests {
         let diagnostic = diagnostic_for(&run_error, ReasonCode::ECfgParse);
         assert_eq!(
             recovery_argv(&diagnostic.fix_steps[0]),
-            ["assay", "doctor", "--config", "<config.yaml>"]
+            [
+                "assay",
+                "doctor",
+                "--config=<config.yaml>",
+                "--format",
+                "json"
+            ]
         );
         assert_eq!(
             outcome_as_written(&run_error, ReasonCode::ECfgParse).next_step,
@@ -460,7 +472,7 @@ mod tests {
             });
         assert_eq!(
             recovery_argv(own_recovery),
-            ["assay", "doctor", "--config", "assay.yaml"]
+            ["assay", "doctor", "--config=assay.yaml", "--format", "json"]
         );
 
         let rendered = folded.format_plain();
--- a/crates/assay-cli/src/exit_codes.rs
+++ b/crates/assay-cli/src/exit_codes.rs
@@ -123,6 +123,15 @@ pub(crate) fn format_recovery_argv(args: &[&str]) -> String {
     format!("Run argv: {}", serde_json::json!(args))
 }
 
+/// Bind a flag to a caller-controlled path as one argv element.
+///
+/// Always `flag=value`, including when `value` does not start with `-`. A split
+/// `--flag` / value pair lets clap reread a `-prefixed` path as an option. A
+/// conditional fuse would be a second place that rule could drift.
+fn fused_option(flag: &str, value: &str) -> String {
+    format!("{flag}={value}")
+}
+
 impl ReasonCode {
     /// Get the corresponding exit code for this reason, respecting version
     pub fn exit_code_for(&self, version: ExitCodeVersion) -> i32 {
@@ -223,18 +232,21 @@ impl ReasonCode {
 
     /// Suggested next step for this error.
     ///
-    /// Executable recovery with caller-controlled values is JSON argv. Dynamic trace-path
-    /// guidance is prose rather than a command; remaining command-like steps are static strings
-    /// and therefore carry no caller-controlled argument boundary.
+    /// Executable recovery with caller-controlled values is JSON argv. Path-bearing
+    /// flags are always fused (`--config=path`) so a `-prefixed` path remains an
+    /// operand. Dynamic recoveries that produce a machine document include
+    /// `--format json` here, not at the call site: one reason has one remediation,
+    /// and a consumer that followed a JSON failure report can parse the result.
+    /// The same string is published on the text channel. Dynamic trace-path
+    /// guidance is prose rather than a command; remaining command-like steps are
+    /// static strings and therefore carry no caller-controlled argument boundary.
     pub fn next_step(&self, context: Option<&str>) -> String {
         match self {
             ReasonCode::Success => String::new(),
-            ReasonCode::ECfgParse => format_recovery_argv(&[
-                "assay",
-                "doctor",
-                "--config",
-                context.unwrap_or("<config.yaml>"),
-            ]),
+            ReasonCode::ECfgParse => {
+                let config = fused_option("--config", context.unwrap_or("<config.yaml>"));
+                format_recovery_argv(&["assay", "doctor", &config, "--format", "json"])
+            }
             ReasonCode::ETraceNotFound => {
                 format!(
                     "Check trace file exists: {}",
@@ -265,13 +277,17 @@ impl ReasonCode {
             ReasonCode::EBaselineInvalid => {
                 "Run: assay baseline record to create a new baseline".to_string()
             }
-            ReasonCode::EPolicyParse => format_recovery_argv(&[
-                "assay",
-                "policy",
-                "validate",
-                "--input",
-                context.unwrap_or("<policy.yaml>"),
-            ]),
+            ReasonCode::EPolicyParse => {
+                let input = fused_option("--input", context.unwrap_or("<policy.yaml>"));
+                format_recovery_argv(&[
+                    "assay",
+                    "policy",
+                    "validate",
+                    &input,
+                    "--format",
+                    "json",
+                ])
+            }
             ReasonCode::EReplayMissingDependency => {
                 "Replay bundle missing required offline dependency; rerun with --live or create a complete bundle".to_string()
             }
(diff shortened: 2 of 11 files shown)


Rul1an:
**Independent Claude subagent review: READY**

- Head reviewed: `3db9e8c9bc7c223423ea34842969c087aa2bc8c1`
- Base at review start: `1d82aada5ce6e773c708bb747cb7230191ab36f2`
- Head remained unchanged and the detached review worktree was clean at start/end.
- Independence: the reviewer was a fresh read-only subagent; it authored/edited neither this change nor a normative specification or implementation plan governing it, and had no prior PR/issue/review context.
- Builder: Cursor. Reviewer: independent Claude subagent. This relay records its exact-head artifact because the subagent did not post externally.

**Verdict:** READY. No correctness blocker in the two routes changed by this PR.

Evidence:
- 46/46 hostile-basename round trips across `E_CFG_PARSE` and `E_POLICY_PARSE` produced JSON argv, executed without a shell, and reached the same typed diagnosis.
- Positive controls reproduced clap refusal for the old split form and human text when `--format json` is absent.
- Affected tests, fmt, clippy, generated-doc reproduction and drift guard passed.
- Nine mutations plus positive controls pinned fusion, JSON mode and generated-artifact links.

Actionable findings outside this PR's stated `E_CFG_PARSE` / `E_POLICY_PARSE` scope:
- pre-existing `EEvidenceUnreadable` positional recovery fails for `-b.tar.gz`;
- pre-existing init success recovery uses split caller-controlled flags and fails for `-weird.yaml`;
- these demonstrate three construction conventions instead of one shared rule.

Disposition: tracked together in #2371 with driven evidence and a repository-wide parity-harness acceptance. They are not introduced or worsened by this focused fix and do not justify widening this head.

Test nits: one unit expectation reuses `fused_option`, and its BEL assertion has no BEL-bearing loop input. Literal/integration tests currently pin the behavior; #2371 carries the cleanup into the shared parity harness.

Base advanced during review to `6315b81a7d39c0a4ab17e52bdd46e1f039b78e57`. The advance and this PR's reviewed file set have an empty intersection, and `git merge-tree 6315b81a... 3db9e8c9...` completed without conflicts. The branch head was not rewritten or advanced. Any new push invalidates this review.

Rul1an:
**Review carry revalidation: READY**

Original independent review head: `3db9e8c9bc7c223423ea34842969c087aa2bc8c1`
Head reviewed: `e4bf16f74325623498578d273f53686cba319f51`/
Upstream merged: `6315b81a7d39c0a4ab17e52bdd46e1f039b78e57`

The current head is an upstream-only merge commit. Both AGENTS.md carry conditions were measured:

1. Whole-tree equivalence:
   - `git merge-tree --write-tree 3db9e8c9... 6315b81a...` = `ae7f6bb22f344cef6178280fb5fb7da200ce7fd3`
   - `e4bf16f74325623498578d273f53686cba319f51^{tree}` = `ae7f6bb22f344cef6178280fb5fb7da200ce7fd3`
   - equal; no manual conflict resolution or extra content.
2. Reviewed-file disjointness:
   - intersection of the 11 files changed by the reviewed PR head and files touched by `3db9e8c9...e4bf16f7...` = 0.

Therefore the independent Claude subagent READY review and its #2371 findings disposition carry to the current head. Any further push invalidates this record.

Rul1an:
**READY — current-head carry record**

Head reviewed: `e4bf16f74325623498578d273f53686cba319f51`

This head contains only the normal upstream merge into the independently reviewed `3db9e8c9` head. Carry evidence:

- whole-tree expected and actual: `ae7f6bb22f344cef6178280fb5fb7da200ce7fd3`;
- intersection between the 11 reviewed PR files and upstream-advance files: 0;
- no conflict resolution, rebase, squash, or product edit;
- #2371 records and dispositions the independent review's outside-scope findings.

The independent non-building review therefore carries to this exact head under AGENTS.md. Any further push invalidates it.

Issue #2185, opened 2026-08-09
Record and guard the assay-mcp-server tools/list wire size


Rul1an:
Adjacent: #2157 (protocol/result envelope) · #2178 (CLI introspection) · #2152 (Claude plugin packaging)

## Measured wire surface

The earlier estimate from Rust source was not a wire measurement. This issue now records the actual compact JSON payload from a built server.

Provenance:

- source head: `48beb09d1ba581f275da7775228262c778914a28`
- worktree: `/Users/roelschuurkes/.config/superpowers/worktrees/assay-2152-plugin-packaging/worktree`
- binary: clean `cargo build -p assay-mcp-server --locked` output in a distinct `CARGO_TARGET_DIR`
- request: legacy `initialize`, `notifications/initialized`, then `tools/list`
- build: default features; `assay_test_outbound` absent

Compact UTF-8 JSON sizes:

| surface | bytes |
|---|---:|
| complete `tools/list` JSON-RPC response | 8,120 |
| `result.tools` array | 8,076 |
| five descriptions combined | 1,984 |
| five input schemas combined | 5,787 |

Per tool (whole tool object / description / input schema):

| tool | bytes |
|---|---:|
| `assay_check_args` | 1,518 / 413 / 1,046 |
| `assay_check_sequence` | 1,441 / 416 / 962 |
| `assay_policy_decide` | 1,026 / 378 / 586 |
| `assay_check_coverage` | 2,293 / 383 / 1,847 |
| `assay_explain_trace` | 1,792 / 390 / 1,340 |

## What this does and does not establish

This establishes the server-emitted wire bytes. It does **not** establish a model-token count or per-turn cost: tokenization and how a host retains/projects MCP schemas are client/model behavior. Do not derive those numbers from Rust source. When #2152's plugin exists, record Claude Code's own plugin context-cost estimate separately with the exact client version.

## Acceptance

- Add a repeatable wire measurement over the built server, with total and per-tool byte output.
- Bind the result to the five production tool names and assert `assay_test_outbound` is absent in the default build.
- Decide explicitly whether to set a no-increase/budget gate or retain the metric as diagnostic evidence; do not shorten descriptions before that decision.
- Preserve each description's non-claim language when any future optimization is reviewed.
- If a token estimate is added, name the tokenizer/model/client version; bytes remain the protocol-level measurement.

## Non-goals

- No tool-name, semantic, or result-shape change.
- No MCP 2026-07-28 remediation; #2157 owns that.
[...]

PR #2395, opened 2026-08-15
docs(agents): allow Ruley-owned writer branches

Rul1an:
## Summary

- permit `ruley/` branches when Ruley is the sole writer
- define Ruley's bounded implementation and issue/PR update permissions
- prohibit direct `main` writes and write-based permission probes

## Why

Ruley now has repository write permissions and has a completed local #2185 test slice, but the shared agent contract currently permits only `codex/`, `claude/`, and `cursor/` writer prefixes. This small contract update enables attributable Ruley-owned branches without weakening branch, worktree, review, or merge controls.

## Verification

- `git diff --check`
- commit hooks passed
- pre-push hooks passed, including workspace clippy and Linux cross-target compile

## Non-claims

- does not publish or modify Ruley's #2185 implementation
- does not make Ruley an automatic quorum reviewer
- does not change GitHub permissions or branch protection

## Final-head record

- Reviewed and gated head: `b0ead71334d9b3358c298212b08dd02f72153827`

PR #2396, opened 2026-08-15
test(mcp): guard default tools/list names and 16KiB wire ceiling

Rul1an:
## Summary

Builder slice for #2185. One new integration test pins the default-feature `tools/list` name set and a 16 KiB explosion ceiling on the complete compact JSON-RPC response.

- **Base:** `249e8160cabb6a7e05f3e6eae38874ef61bc3ac1` (`main`)
- **Head:** `6113b5ecd05844cf1edc6ef629cfabba50737ca5` (`ruley/2185-mcp-tool-surface-guard`)
- **One file only:** `crates/assay-mcp-server/tests/tools_list_wire.rs`
- **Refs #2185** (does not auto-close; no merge from this PR)

## RED / GREEN

- **RED** (before the file existed, rustc 1.96.0): `cargo test -p assay-mcp-server --test tools_list_wire` → no test target named `tools_list_wire`, exit 101
- **GREEN** (after the one-file commit): focused suite 6/6 including five `should_panic` mutations

## What is locked

- Five production tool names, exact set:
  - `assay_check_args`
  - `assay_check_sequence`
  - `assay_policy_decide`
  - `assay_check_coverage`
  - `assay_explain_trace`
- `assay_test_outbound` must be absent on the default-feature surface
- Observation must come from the built `assay-mcp-server` binary (`from_built_binary: true`)
- Complete compact `tools/list` JSON-RPC response must stay at or under **16,384 bytes** (explosion ceiling)

## What is diagnostic only

Live compact response printed **8120** bytes (result 8086). **8120 is printed only, not a lock.** There is no no-increase / budget lock on 8120. The scored size bound is the 16 KiB ceiling.

## Mutation table (all expected panic / pass)

| test | expected | result |
|---|---|---|
| `mutation_renamed_production_tool_fails_names_guard` | should_panic names guard | pass |
| `mutation_removed_production_tool_fails_names_guard` | should_panic names guard | pass |
| `mutation_injected_assay_test_outbound_fails_outbound_guard` | should_panic outbound guard | pass |
| `mutation_lowered_ceiling_below_live_compact_body_fails_ceiling_guard` | should_panic 8119-byte ceiling | pass |
| `mutation_hardcoded_fixture_fails_built_binary_guard` | should_panic `from_built_binary: false` | pass |
| `default_feature_tools_list_wire_is_five_production_tools_under_16kib` | live 5 names, no outbound, ≤16 KiB | pass |

## Gates

- `cargo test -p assay-mcp-server --test tools_list_wire -- --nocapture` → 6/6
- `cargo test -p assay-mcp-server --offline` → pass
- `cargo fmt -- crates/assay-mcp-server/tests/tools_list_wire.rs` then `cargo fmt --all -- --check` → pass
- `cargo clippy -p assay-mcp-server --all-targets -- -D warnings` → pass
- `git diff --check` → pass

Toolchain: rustc 1.96.0 (ac68faa20 2026-05-25), cargo 1.96.0 (30a34c682 2026-05-25). Default features only; no `--features`.

## Non-claims

- No per-turn token / context-cost claim
- No perf or adoption claim
- No claim that `assay_test_outbound` is a product-available tool
- No MCP 2026-07-28 / #2157 work
- No description or `inputSchema` changes
- No no-increase lock on the printed 8120-byte live body
- Builder is not a reviewer and this post does not count for quorum

## Process

Writer / builder only. Draft PR. Do not merge from this slice. Please have Codex assign an independent reviewer.

### DIFF
--- a/crates/assay-mcp-server/tests/tools_list_wire.rs
+++ b/crates/assay-mcp-server/tests/tools_list_wire.rs
@@ -1,0 +1,241 @@
+//! Wire-size guard for the default-feature `tools/list` surface.
+//!
+//! Drives the built `assay-mcp-server` over a real handshake and records compact
+//! JSON-RPC bytes. The only scored size bound is a 16 KiB explosion ceiling on
+//! the complete compact response; 8,120 is diagnostic, not a lock. Name-set
+//! ownership for install-surface launch stays in `project_install_surfaces.rs`.
+
+use serde_json::Value;
+use std::collections::BTreeSet;
+use std::path::Path;
+use std::process::{Command, Stdio};
+
+mod jsonrpc_conn;
+use jsonrpc_conn::Conn;
+
+const PRODUCTION_TOOL_NAMES: [&str; 5] = [
+    "assay_check_args",
+    "assay_check_sequence",
+    "assay_policy_decide",
+    "assay_check_coverage",
+    "assay_explain_trace",
+];
+
+const TEST_OUTBOUND_TOOL: &str = "assay_test_outbound";
+const COMPACT_RESPONSE_CEILING: usize = 16_384;
+
+/// Compact `tools/list` observation. The live path tags `from_built_binary: true`.
+struct ToolsListObservation {
+    names: Vec<String>,
+    compact: Vec<u8>,
+    from_built_binary: bool,
+}
+
+fn assert_production_tool_names(names: &[String]) {
+    let actual: BTreeSet<&str> = names.iter().map(String::as_str).collect();
+    let expected: BTreeSet<&str> = PRODUCTION_TOOL_NAMES.iter().copied().collect();
+    assert_eq!(
+        actual, expected,
+        "tools/list production names must be exactly the five advertised tools (got {actual:?})"
+    );
+}
+
+fn assert_test_outbound_absent(names: &[String]) {
+    assert!(
+        !names.iter().any(|name| name == TEST_OUTBOUND_TOOL),
+        "default-feature tools/list advertised {TEST_OUTBOUND_TOOL}"
+    );
+}
+
+fn assert_compact_response_within_ceiling(compact: &[u8], ceiling: usize) {
+    assert!(
+        compact.len() <= ceiling,
+        "compact tools/list JSON-RPC response is {} bytes, exceeds the {ceiling}-byte explosion ceiling",
+        compact.len()
+    );
+}
+
+fn assert_from_built_binary(observation: &ToolsListObservation) {
+    assert!(
+        observation.from_built_binary,
+        "tools/list observation must come from the built assay-mcp-server binary (from_built_binary: {})",
+        observation.from_built_binary
+    );
+}
+
+fn tool_names(response: &Value) -> Vec<String> {
+    response["result"]["tools"]
+        .as_array()
+        .expect("tools/list result.tools array")
+        .iter()
+        .map(|tool| tool["name"].as_str().expect("tool name").to_string())
+        .collect()
+}
+
+fn eprint_wire_diagnostics(response: &Value, compact: &[u8]) {
+    let result_bytes = serde_json::to_vec(&response["result"]).expect("serialize result");
+    eprintln!(
+        "tools/list compact JSON-RPC response: {} bytes (ceiling {COMPACT_RESPONSE_CEILING})",
+        compact.len()
+    );
+    eprintln!("tools/list compact result: {} bytes", result_bytes.len());
+    let Some(tools) = response["result"]["tools"].as_array() else {
+        return;
+    };
+    for tool in tools {
+        let name = tool["name"].as_str().unwrap_or("<unnamed>");
+        let whole = serde_json::to_vec(tool).expect("serialize tool").len();
+        let description = tool
+            .get("description")
+            .and_then(Value::as_str)
+            .map(str::len)
+            .unwrap_or(0);
+        let schema = tool
+            .get("inputSchema")
+            .map(|schema| {
+                serde_json::to_vec(schema)
+                    .expect("serialize inputSchema")
+                    .len()
+            })
+            .unwrap_or(0);
+        eprintln!("  {name}: whole={whole} description={description} inputSchema={schema}");
+    }
+}
+
+fn clean_server_command() -> Command {
+    let mut command = Command::new(env!("CARGO_BIN_EXE_assay-mcp-server"));
+    for (name, _) in std::env::vars_os() {
+        if name
+            .to_string_lossy()
+            .to_ascii_uppercase()
+            .starts_with("ASSAY_AUTH_")
+        {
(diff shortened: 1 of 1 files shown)


Rul1an:
## Independent non-building review: READY, no blocker

Bound to exact head `6113b5ecd05844cf1edc6ef629cfabba50737ca5`, base `249e8160cabb6a7e05f3e6eae38874ef61bc3ac1` (verified equal to `origin/main` and an ancestor of head, single parent, one commit). Scope verified as one new file, `crates/assay-mcp-server/tests/tools_list_wire.rs`, +241/-0, no production source touched. Any new push invalidates this review.

Because this is a guards-only change, I applied the disposition that weak guard machinery is itself a defect, and tried to find a production mutation that slips through rather than only reading the assertions.

### Verified by execution, not by reading

`cargo test --locked -p assay-mcp-server --test tools_list_wire -- --nocapture` at this head: **6 passed, 0 failed, 0 ignored**.

- **Live path is genuinely live.** `clean_server_command` at `:106` spawns `env!("CARGO_BIN_EXE_assay-mcp-server")`, and `observe_live_tools_list` at `:138-178` performs a real `initialize` with a result assertion, sends `notifications/initialized`, then requests `tools/list` and reads the response by id. The per-tool byte diagnostics it printed (`assay_check_args: whole=1518`, `check_sequence 1441`, `policy_decide 1026`, `check_coverage 2293`, `explain_trace 1792`) could only come from the built binary.
- **Exactly five names, outbound absent.** `assert_production_tool_names` compares `BTreeSet` equality against the five-element const, and `assert_test_outbound_absent` is a separate explicit check. Both are applied to the live observation at `:192-193`.
- **The ceiling is an explosion ceiling, and 8120 is not a lock.** Measured live compact response is **8120 bytes** against `COMPACT_RESPONSE_CEILING = 16_384`, so roughly 2.02x headroom. `8_120` appears only in the two fixtures at `:183` and `:228` and is never compared against the live response, so growth to 9 KB would still pass. The only scored bound on the live path is the 16 KiB one at `:196`. This matches what the module doc claims at `:4-5`.
- **All five `#[should_panic]` controls fire for the reason they name.** Each panicked at the intended helper line with the expected substring: names guard at `:37` for both rename and removal, outbound guard at `:44`, ceiling guard at `:51` with `is 8120 bytes, exceeds the 8119-byte explosion ceiling`, and the anti-fixture guard at `:59`. The hardcoded-fixture control at `:235-241` is ordered so that names, outbound and ceiling all pass first and only `assert_from_built_binary` panics, so it discriminates rather than short-circuiting on an earlier assertion.
- **Process and environment cleanup is sound.** `jsonrpc_conn::Conn` has `impl Drop` at `mod.rs:353-360` which drops stdin to signal EOF and then kills and reaps the child, so a panic between spawn and `shutdown()` leaks nothing. `clean_server_command` strips every `ASSAY_AUTH_*` variable before spawn. The `tempfile::tempdir()` policy root is dropped after `conn.shutdown()`, not before.
- **No API or contract change.** Test-only file; no production symbol, schema, or wire shape is modified.

### Non-blocking findings

1. **The names guard binds the set, not the cardinality.** `assert_production_tool_names` at `:34-41` compares `BTreeSet`s, so a production mutation that emitted one of the five tools twice would yield a five-element set and pass, while the wire advertised six entries. The 16 KiB ceiling would not catch it either, since 8120 plus a duplicate of the largest tool is 10413, still under 16384. I record this as MINOR rather than blocking because the module explicitly scopes itself to name-set ownership at `:5-6`, and a set comparison is the honest implementation of a set claim. One line, `assert_eq!(names.len(), PRODUCTION_TOOL_NAMES.len())`, would close it.
2. **`from_built_binary` is self-asserted.** The field is a boolean literal set to `true` by the producing function at `:175` and `false` by the fixture at `:184`. The control at `:235` proves the predicate rejects `false`; it cannot prove the live path sets `true` truthfully, so a mutation that replaced `observe_live_tools_list` with a fixture while keeping the flag `true` would pass. The real liveness is carried by `CARGO_BIN_EXE_*` plus the spawn and handshake, which I verified independently, so the flag is documentation-grade rather than load-bearing.
3. **Feature-mode scope is declared but not enforced.** `list_tools()` pushes `assay_test_outbound` under `#[cfg(feature = "test-outbound")]` at `tools/mod.rs:424`, so running `cargo test -p assay-mcp-server --features test-outbound` without a name filter would fail both name assertions. No CI job does that: `ci.yml:713` runs that feature with a `no_passthrough` test-name filter, and `ci.yml:308` is clippy only, which compiles but does not run. So there is no CI collision today. The module doc declares default-feature scope at `:1`, it just does not enforce it, and a local run in that mode would read as a production regression rather than a mode mismatch.

None of the three lets a mutation through the property the file actually claims to guard.

### CI

Complete at this exact head: 51 success, 5 skipped, 0 failing, 0 pending. I do not count the 5 skipped as passes, and no check state substitutes for the verification above. `mergeStateStatus` reads `CLEAN`, auto-merge is not armed, and the PR is still a draft, so I take no merge action.

### Non-building and independence declaration

I authored and edited neither this change nor any normative specification or implementation plan governing it. I did not write, build, or advise on this test file, and I hold no writer role over this branch.

For completeness of the record: I posted the independent review on #2395, the `AGENTS.md` change that permits the `ruley/` branch prefix this PR uses. That is a governance permission rather than a specification governing this test's content, and prior read-only review does not make a reviewer a builder under the section this repository applies. My other recent contributions are to unrelated slices in the policy-input programme and the argv recovery contract, neither of which governs this change.

### Non-claims

Executed on macOS arm64 with the default feature set only. I did not run the full `assay-mcp-server` suite, the workspace suite, clippy, or fmt, and I did not run the crate under `--features test-outbound`. I did not verify behaviour under nextest, which CI may use, so the `Conn` timeout and reap paths are established by reading rather than by triggering them. I did not measure the wire size on any platform other than this one, so the 8120-byte figure and the resulting headroom are specific to this host and toolchain. I did not evaluate whether 16 KiB is the correct ceiling as a policy question, only that it is meaningful against today's measured size and that nothing locks 8120.

Read-only throughout: no edits to any tracked file, no commits, no pushes, no branch changes, no merge. Every measurement was taken in a disposable detached worktree at this exact SHA, which has been removed and pruned.

Rul1an:
## Finding disposition

Exact-head review is `READY` with no blocker. The three non-blocking guard-hardening observations are consolidated in #2397 (`horizon:post-mvp`, `priority:P2`, `size:XS`) rather than changing this already-green MVP head:

- entry cardinality in addition to exact name-set equality;
- structural live witness instead of a self-asserted provenance boolean;
- explicit default-vs-`test-outbound` feature-mode expectations.

They do not invalidate the properties #2396 claims and proves: the live default-feature built binary advertises the exact five-name set, excludes `assay_test_outbound`, and stays below the 16 KiB explosion ceiling. No finding is silently dropped.

PR #2400, opened 2026-08-15
test(mcp): harden tools/list cardinality and feature-mode witness

Rul1an:
Refs #2397

Test-only TDD slice for `tools/list` hardening. Does not close the issue.

## Base / head
- Base: `d1bfd36a11b857222d9c43f291a8011eaeee23e6` (`origin/main` at write time)
- Head: `fbd2cfca1efdafe938c250ca322b7ba4db6201f5` (`ruley/2397-tools-list-hardening`)
- One file: `crates/assay-mcp-server/tests/tools_list_wire.rs`

## RED
Added only `mutation_duplicate_production_tool_fails_cardinality_guard` against the previous BTreeSet helper. Duplicate of a production name still had 5 unique names, so `#[should_panic(expected = "exactly 5 entries before name-set comparison")]` did **not** panic.

```
cargo test -p assay-mcp-server --test tools_list_wire mutation_duplicate_production_tool_fails_cardinality_guard -- --nocapture
# FAILED: test did not panic as expected (exit 101)
```

## GREEN
Rewrote the same test file:
1. Cardinality is asserted **before** BTreeSet comparison (`EXPECTED_LEN` is 5 default / 6 with `test-outbound`).
2. Live witness is structural: spawn `env!("CARGO_BIN_EXE_assay-mcp-server")`, `initialize`, `tools/list`. No `from_built_binary` boolean.
3. Default-feature vs `--features test-outbound` expectations are explicit and cfg-honest.

## Feature-mode commands
```
cargo test -p assay-mcp-server --test tools_list_wire -- --nocapture
# 7 passed; live compact 8120 bytes; five production names; no assay_test_outbound

cargo test -p assay-mcp-server --features test-outbound --test tools_list_wire -- --nocapture
# 7 passed; live compact 8324 bytes; six names including assay_test_outbound
```

## Mutations
| Mutation | Guard | Panic witness |
|---|---|---|
| duplicate advertised name | cardinality before set-eq | `entries before name-set comparison` |
| fixture `{}` initialize | live handshake | `CARGO_BIN_EXE_assay-mcp-server` |
| inject outbound (default) / drop outbound (`test-outbound`) | feature-mode | `default-feature tools/list advertised assay_test_outbound` / `test-outbound tools/list must advertise assay_test_outbound` |
| rename (kept) | name-set | expected-set mismatch |
| remove (kept; cardinality now fires first) | cardinality | `entries before name-set comparison` |
[...]

PR #2401, opened 2026-08-15
fix(mcp): fail closed at the outer tool boundary

### LINKED ISSUE #2391: security(mcp): harden outer tools/call fallback and E_INTERNAL diagnostics
## Problem

The generic `tools/call` fallback in `crates/assay-mcp-server/src/server.rs` has two coupled boundary defects:

1. it reads `on_error` from caller-controlled tool arguments even though the advertised tool schemas omit that property and set `additionalProperties: false`; `on_error: allow` converts handler errors and timeouts into `allowed:true` / `isError:false`;
2. handler `Err` values are published as raw `E_INTERNAL.message = e.to_string()` outside the shared `ToolError` serialization boundary, so the message is neither value-free nor subject to the planned 4,096-byte `ToolError.message` ceiling.

Measured by source inspection on `addfda3b3f695f11a7dba90f7c1ae24ff442b034`:

- `server.rs:321-325` derives `allow_on_error` from `arguments.on_error`;
- `server.rs:489-520` uses it for handler errors and publishes raw `e.to_string()`;
- `check_coverage.rs:49-52` and `explain_trace.rs:40-42` can return caller-influenced deserialization errors;
- the five advertised input schemas do not declare `on_error`.

Found during independent review of #2390. Related to, but intentionally outside, the policy-input programme #2388 because this is the outer tool-dispatch boundary rather than a policy read/parse path.

## Design questions

- Is fail-open a server/operator configuration, or should a caller ever be able to request it per invocation?
- If compatibility requires per-call selection, which authenticated/trusted boundary authorizes it and how is it represented in every input schema?
- Which handler failures are tool execution errors versus protocol errors?

Do not answer these implicitly in implementation.

## Definition of Done

- RED real-stdio tests exercise a handler `Err` and timeout with default arguments and with `on_error: allow`.
- The chosen fail-open authority is explicit, schema-consistent, and cannot be smuggled through an undeclared argument.
- Generic fallback messages use one stable, value-free, UTF-8-safe bounded publication path; no raw `anyhow`/serde text reaches MCP content.
- Tests place distinct sentinels at the beginning, middle, and end of hostile input and prove none appears in the full response.
- `allowed`, `isError`, code, warning, and result-type behavior are pinned for each fallback mode.
- Mutations that restore caller-selected fail-open or raw `e.to_string()` make distinct tests fail.
- Existing intentional operator fail-safe behavior has a documented migration path if its authority changes.

## Non-claims

- No change to policy dialects, policy-file size limits, proxy enforcement, authentication, or target-tool execution.
- This issue does not claim current remote exploitability; it records a caller-controlled boundary and public diagnostic leak path that require a deliberate contract.

Rul1an:
## Summary

- remove caller-controlled `arguments.on_error` authority from built-in stdio tool dispatch
- normalize handler failures and timeouts through bounded, value-free `ToolError` results
- keep unknown methods as fixed-message JSON-RPC `-32601` and unknown tools as fail-closed tool results
- correct gateway, fail-safe, CLI, API, and changelog guidance

Closes #2391.

## Scope and non-claims

No new configuration, operator fail-open, schema validation, proxy/auth behavior, tool schema, `resultType`, `structuredContent`, or MCP 2026-07-28 conformance. Suite `settings.on_error` remains an `assay run` setting.

## Review packet

- base / merge-base: `a63a012ba9c8b994649b334d4872afe600b563b1`
- final reviewed head: `bbf3f61e803ea2de55a35d602b8c89ac887bbd3b`
- worktree: `/Users/roelschuurkes/wt-2391-outer-fallback`
- RED: `cargo test --locked -p assay-mcp-server --test outer_fallback_contract -- --nocapture` failed 0/3 for the intended old behaviors
- GREEN: the same test passes 3/3
- mutations caught: raw handler error, timeout allow, unknown-method reflection, forced `isError:false`, restored caller authority, report-tool `isError` regression, and report decision-telemetry drift
- full: `cargo test --locked -p assay-mcp-server`
- quality: fmt, clippy all-targets with `-D warnings`, rustdoc with `-D warnings`, and `git diff --check` all pass

Primary protocol references: MCP [tools](https://modelcontextprotocol.io/specification/2025-06-18/server/tools) and [schema](https://modelcontextprotocol.io/specification/2025-06-18/schema).

## Review disposition

Independent exact-head verdict: **READY** at `bbf3f61e803ea2de55a35d602b8c89ac887bbd3b` ([record](https://github.com/Rul1an/assay/pull/2401#issuecomment-5304769465)). The non-blocking telemetry/documentation findings are consolidated in #2402; they do not weaken or alter this PR's caller-authority removal and fixed wire-failure boundary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Changed
+- The built-in stdio MCP policy tools now fail closed on outer dispatch errors
+  and timeouts. Caller `arguments.on_error` no longer selects fail-open;
+  responses use fixed value-free `E_INTERNAL` / `E_TIMEOUT` messages and
+  `isError: true`. Clients following the former gateway example must remove
+  that argument. Suite `settings.on_error` remains an `assay run` setting
+  (#2391).
 - The five advertised MCP policy tools now read local policy files through one
   inclusive byte ceiling (`ServerConfig.max_policy_bytes`, default 1,000,000,
   override `ASSAY_MCP_MAX_POLICY_BYTES`) before parse or cache insertion.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@ All notable changes to this project will be documented in this file.
   responses use fixed value-free `E_INTERNAL` / `E_TIMEOUT` messages and
   `isError: true`. Clients following the former gateway example must remove
   that argument. Suite `settings.on_error` remains an `assay run` setting
-  (#2391).
+  (#2391). Outer dispatch failures emit `tool_execution_error`; `tool_call_*`
+  events no longer copy caller-controlled tool or policy strings. The separate
+  `tool_decision` evidence event retains its existing redaction contract.
 - The five advertised MCP policy tools now read local policy files through one
   inclusive byte ceiling (`ServerConfig.max_policy_bytes`, default 1,000,000,
   override `ASSAY_MCP_MAX_POLICY_BYTES`) before parse or cache insertion.
--- a/crates/assay-mcp-server/src/server.rs
+++ b/crates/assay-mcp-server/src/server.rs
@@ -19,7 +19,7 @@ fn fail_closed_tool_result(code: &'static str, message: &'static str) -> Result<
 fn classify_tool_result(result: &Value) -> (bool, bool) {
     let has_error = result.get("error").is_some();
     let explicit_allowed = result.get("allowed").and_then(Value::as_bool);
-    let allowed = explicit_allowed.unwrap_or(!has_error);
+    let allowed = explicit_allowed.unwrap_or(false);
     let is_error = has_error || explicit_allowed == Some(false);
     (allowed, is_error)
 }
@@ -462,7 +462,24 @@ impl Server {
 
 #[cfg(test)]
 mod claims_boundary_tests {
-    use super::{initialize_result, LegacyProtocolVersion};
+    use super::{classify_tool_result, initialize_result, LegacyProtocolVersion};
+
+    #[test]
+    fn tool_result_classification_separates_decision_from_mcp_error() {
+        for (result, expected) in [
+            (serde_json::json!({"allowed": true}), (true, false)),
+            (serde_json::json!({"allowed": false}), (false, true)),
+            (
+                serde_json::json!({"allowed": false, "error": {"code": "E_INTERNAL"}}),
+                (false, true),
+            ),
+            // Report tools return data rather than a policy decision. Preserve the existing
+            // decision telemetry while keeping their successful MCP result non-error.
+            (serde_json::json!({"report": {}}), (false, false)),
+        ] {
+            assert_eq!(classify_tool_result(&result), expected, "result: {result}");
+        }
+    }
 
     /// ADR-042 stop list, ADR-043 §2. The handshake must not assert a status the server
     /// cannot substantiate. A denylist over the serialized response catches a claim
--- a/crates/assay-mcp-server/src/server.rs
+++ b/crates/assay-mcp-server/src/server.rs
@@ -16,6 +16,14 @@ fn fail_closed_tool_result(code: &'static str, message: &'static str) -> Result<
     tools::ToolError::new(code, message).result()
 }
 
+fn classify_tool_result(result: &Value) -> (bool, bool) {
+    let has_error = result.get("error").is_some();
+    let explicit_allowed = result.get("allowed").and_then(Value::as_bool);
+    let allowed = explicit_allowed.unwrap_or(!has_error);
+    let is_error = has_error || explicit_allowed == Some(false);
+    (allowed, is_error)
+}
+
 fn next_rid() -> String {
     let n = RID.fetch_add(1, Ordering::Relaxed);
     format!("r-{n:06}")
@@ -368,10 +376,7 @@ impl Server {
 
                         let dur = start.elapsed().as_millis() as u64;
                         // Log outcome
-                        let allowed = result
-                            .get("allowed")
-                            .and_then(|v| v.as_bool())
-                            .unwrap_or(false);
+                        let (allowed, is_error) = classify_tool_result(&result);
                         if let Some(err) = result.get("error") {
                             let code = err.get("code").and_then(|v| v.as_str()).unwrap_or("");
                             tracing::info!(
@@ -432,7 +437,6 @@ impl Server {
                         }
 
                         // MCP Compliance: wrap every tool outcome in CallToolResult.
-                        let is_error = !allowed;
                         let json_text = serde_json::to_string_pretty(&result).unwrap_or_default();
                         let mcp_result = serde_json::json!({
                             "content": [{"type": "text", "text": json_text}],
--- a/crates/assay-mcp-server/src/server.rs
+++ b/crates/assay-mcp-server/src/server.rs
@@ -12,6 +12,10 @@ static RID: AtomicU64 = AtomicU64::new(1);
 const INVALID_INITIALIZE_PARAMS: &str =
     "Invalid initialize params: expected the required legacy MCP fields";
 
+fn fail_closed_tool_result(code: &'static str, message: &'static str) -> Result<Value> {
+    tools::ToolError::new(code, message).result()
+}
+
 fn next_rid() -> String {
     let n = RID.fetch_add(1, Ordering::Relaxed);
     format!("r-{n:06}")
@@ -318,13 +322,6 @@ impl Server {
                         let default_args = serde_json::json!({});
                         let args = params.get("arguments").unwrap_or(&default_args);
 
-                        let on_error_str = args
-                            .get("on_error")
-                            .and_then(|v| v.as_str())
-                            .unwrap_or("block");
-                        let allow_on_error = on_error_str.eq_ignore_ascii_case("allow");
-
-                        let policy = args.get("policy").and_then(|v| v.as_str()).unwrap_or("");
                         let bytes_in = line.len();
                         let args_bytes = serde_json::to_vec(args).map(|b| b.len()).unwrap_or(0);
 
@@ -334,9 +331,6 @@ impl Server {
                            event="tool_call_start",
                            rid=%rid,
                            rpc_id=?req.id,
-                           tool=name,
-                           policy=policy,
-                           on_error=on_error_str,
                            bytes_in=bytes_in,
                            args_bytes=args_bytes,
                         );
@@ -348,76 +342,56 @@ impl Server {
                         let fut = tools::handle_call(&ctx, name, args);
                         let result = match timeout(Duration::from_millis(cfg.timeout_ms), fut).await
                         {
-                            Ok(res) => res, // Tool finished
+                            Ok(Ok(value)) => value,
+                            Ok(Err(_error)) => {
+                                tracing::error!(
+                                    event = "tool_execution_error",
+                                    rid = %rid,
+                                    rpc_id = ?req.id,
+                                    duration_ms = start.elapsed().as_millis() as u64,
+                                    code = "E_INTERNAL"
+                                );
+                                fail_closed_tool_result("E_INTERNAL", "Tool execution failed")?
+                            }
                             Err(_) => {
                                 let dur = start.elapsed().as_millis() as u64;
                                 tracing::warn!(
                                    event="tool_call_timeout",
                                    rid=%rid,
                                    rpc_id=?req.id,
-                                   tool=name,
-                                   policy=policy,
                                    duration_ms=dur,
-                                   code="E_TIMEOUT",
-                                   fallback=on_error_str
+                                   code="E_TIMEOUT"
                                 );
-                                // Timed out
-                                Ok(serde_json::json!({
-                                    "allowed": allow_on_error,
-                                    "error": {
-                                        "code": "E_TIMEOUT",
-                                        "message": format!("Request exceeded {}ms", cfg.timeout_ms)
-                                    }
-                                }))
+                                fail_closed_tool_result("E_TIMEOUT", "Tool execution timed out")?
                             }
                         };
 
                         let dur = start.elapsed().as_millis() as u64;
                         // Log outcome
-                        match &result {
-                            Ok(val) => {
-                                let allowed = val
-                                    .get("allowed")
-                                    .and_then(|v| v.as_bool())
-                                    .unwrap_or(false);
-                                if let Some(err) = val.get("error") {
-                                    let code =
-                                        err.get("code").and_then(|v| v.as_str()).unwrap_or("");
-                                    tracing::info!(
-                                      event="tool_call_done",
-                                      rid=%rid,
-                                      rpc_id=?req.id,
-                                      tool=name,
-                                      policy=policy,
-                                      duration_ms=dur,
-                                      outcome="app_error",
-                                      allowed=allowed,
-                                      code=code
-                                    );
-                                } else {
-                                    tracing::info!(
-                                      event="tool_call_done",
-                                      rid=%rid,
-                                      rpc_id=?req.id,
-                                      tool=name,
-                                      policy=policy,
-                                      duration_ms=dur,
-                                      outcome="ok",
-                                      allowed=allowed
-                                    );
-                                }
-                            }
-                            Err(e) => {
-                                tracing::error!(
-                                  event="tool_call_crash",
-                                  rid=%rid,
-                                  rpc_id=?req.id,
-                                  tool=name,
-                                  policy=policy,
-                                  duration_ms=dur,
-                                  error=%e
-                                );
-                            }
+                        let allowed = result
+                            .get("allowed")
+                            .and_then(|v| v.as_bool())
+                            .unwrap_or(false);
+                        if let Some(err) = result.get("error") {
+                            let code = err.get("code").and_then(|v| v.as_str()).unwrap_or("");
+                            tracing::info!(
+                              event="tool_call_done",
+                              rid=%rid,
+                              rpc_id=?req.id,
+                              duration_ms=dur,
+                              outcome="app_error",
+                              allowed=allowed,
+                              code=code
+                            );
+                        } else {
+                            tracing::info!(
+                              event="tool_call_done",
+                              rid=%rid,
+                              rpc_id=?req.id,
+                              duration_ms=dur,
+                              outcome="ok",
+                              allowed=allowed
+                            );
                         }
 
                         // P57b: emit the observed tool decision (assay.tool_decision_surface.v0) as
@@ -425,25 +399,16 @@ impl Server {
                         // enforced inside build_decision; this site never has SaaS-verified evidence.
                         {
                             use crate::tool_decision::{build_decision, Effect, ObservedCall};
-                            let (effect, status) = match &result {
-                                Ok(val) => {
-                                    let allowed = val
-                                        .get("allowed")
-                                        .and_then(|v| v.as_bool())
-                                        .unwrap_or(false);
-                                    if let Some(code) = val
-                                        .get("error")
-                                        .and_then(|e| e.get("code"))
-                                        .and_then(|v| v.as_str())
-                                    {
-                                        (Effect::Error, code.to_string())
-                                    } else if allowed {
-                                        (Effect::Allow, "success".to_string())
-                                    } else {
-                                        (Effect::Deny, "blocked".to_string())
-                                    }
-                                }
-                                Err(_) => (Effect::Error, "crash".to_string()),
+                            let (effect, status) = if let Some(code) = result
+                                .get("error")
+                                .and_then(|e| e.get("code"))
+                                .and_then(|v| v.as_str())
+                            {
+                                (Effect::Error, code.to_string())
+                            } else if allowed {
+                                (Effect::Allow, "success".to_string())
+                            } else {
+                                (Effect::Deny, "blocked".to_string())
                             };
                             let decision = build_decision(&ObservedCall {
                                 server_id: "mcp",
@@ -466,71 +431,19 @@ impl Server {
                             );
                         }
 
-                        match result {
-                            Ok(res) => {
-                                // MCP Compliance: Wrap result in CallToolResult structure
-                                // Spec: { content: [{ type: "text", text: "..." }], isError: bool }
-                                let is_error =
-                                    !res.get("allowed").and_then(|v| v.as_bool()).unwrap_or(true);
-                                let json_text =
-                                    serde_json::to_string_pretty(&res).unwrap_or_default();
-
-                                let mcp_result = serde_json::json!({
-                                    "content": [
-                                        {
-                                            "type": "text",
-                                            "text": json_text
-                                        }
-                                    ],
-                                    "isError": is_error
-                                });
-                                JsonRpcResponse::ok(req.id.clone(), mcp_result)
-                            }
-                            Err(e) => {
-                                // Fail-safe handling for internal errors
-                                tracing::error!(
-                                    event="tool_execution_error",
-                                    rid=%rid,
-                                    error=%e,
-                                    fallback=on_error_str
-                                );
-                                let mut safe_resp = serde_json::json!({
-                                    "allowed": allow_on_error,
-                                    "error": {
-                                        "code": "E_INTERNAL",
-                                        "message": e.to_string()
-                                    }
-                                });
-
-                                // Agent Awareness
-                                // If we fail open, warn the agent so it can self-regulate (e.g. switch to Safe Mode).
-                                if allow_on_error {
-                                    safe_resp["warning"] = serde_json::json!("FAIL-SAFE ACTIVE: Policy engine offline. Proceed with caution (Safe Mode).");
-                                }
-                                // Keep consistent wrapping even for internal fail-safe responses
-                                let json_text =
-                                    serde_json::to_string_pretty(&safe_resp).unwrap_or_default();
-                                let mcp_result = serde_json::json!({
-                                    "content": [
-                                        {
-                                            "type": "text",
-                                            "text": json_text
-                                        }
-                                    ],
-                                    "isError": !allow_on_error
-                                });
-                                JsonRpcResponse::ok(req.id.clone(), mcp_result)
-                            }
-                        }
+                        // MCP Compliance: wrap every tool outcome in CallToolResult.
+                        let is_error = !allowed;
+                        let json_text = serde_json::to_string_pretty(&result).unwrap_or_default();
+                        let mcp_result = serde_json::json!({
+                            "content": [{"type": "text", "text": json_text}],
+                            "isError": is_error
+                        });
+                        JsonRpcResponse::ok(req.id.clone(), mcp_result)
                     } else {
                         JsonRpcResponse::error(req.id.clone(), -32602, "Missing params".to_string())
                     }
                 }
-                _ => JsonRpcResponse::error(
-                    req.id.clone(),
-                    -32601,
-                    format!("Method not found: {}", req.method),
-                ),
+                _ => JsonRpcResponse::error(req.id.clone(), -32601, "Method not found".to_string()),
             };
 
             // Send Response
--- a/crates/assay-mcp-server/tests/outer_fallback_contract.rs
+++ b/crates/assay-mcp-server/tests/outer_fallback_contract.rs
@@ -1,0 +1,129 @@
+//! Wire-level contract for failures outside an individual MCP tool implementation.
+
+use serde_json::Value;
+use std::process::{Command, Stdio};
+
+mod jsonrpc_conn;
+use jsonrpc_conn::Conn;
+
+const HEAD: &str = "HEAD_SENTINEL";
+const MID: &str = "MID_SENTINEL";
+const TAIL: &str = "TAIL_SENTINEL";
+
+fn spawn_server(timeout_ms: Option<&str>) -> (Conn, tempfile::TempDir) {
+    let policy_root = tempfile::tempdir().expect("temporary policy root");
+    let mut command = Command::new(env!("CARGO_BIN_EXE_assay-mcp-server"));
+    command
+        .args(["--policy-root", policy_root.path().to_str().unwrap()])
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::null());
+    if let Some(timeout_ms) = timeout_ms {
+        command.env("ASSAY_MCP_TIMEOUT_MS", timeout_ms);
+    }
+    let child = command.spawn().expect("spawn assay-mcp-server");
+    (Conn::attach(child), policy_root)
+}
+
+fn initialize(conn: &mut Conn) {
+    let response = conn.request(
+        "initialize",
+        serde_json::json!({
+            "protocolVersion": "2025-11-25",
+            "capabilities": {},
+            "clientInfo": {"name": "outer-fallback-contract", "version": "1"}
+        }),
+        1,
+    );
+    assert!(response.get("result").is_some(), "initialize: {response}");
+}
+
+fn call_tool(conn: &mut Conn, name: &str, arguments: Value, id: u64) -> Value {
+    conn.request(
+        "tools/call",
+        serde_json::json!({"name": name, "arguments": arguments}),
+        id,
+    )
+}
+
+fn tool_payload(response: &Value) -> Value {
+    let result = response.get("result").expect("CallToolResult response");
+    let text = result["content"][0]["text"].as_str().expect("text content");
+    serde_json::from_str(text).expect("JSON tool payload")
+}
+
+fn assert_fixed_failure(response: &Value, code: &str, message: &str) {
+    let result = response.get("result").expect("CallToolResult response");
+    assert_eq!(result["isError"], true, "response: {response}");
+    let payload = tool_payload(response);
+    assert_eq!(payload["allowed"], false, "payload: {payload}");
+    assert_eq!(payload["error"]["code"], code, "payload: {payload}");
+    assert_eq!(payload["error"]["message"], message, "payload: {payload}");
+    assert!(payload.get("warning").is_none(), "payload: {payload}");
+}
+
+#[test]
+fn caller_cannot_fail_open_handler_errors() {
+    let (mut conn, _root) = spawn_server(None);
+    initialize(&mut conn);
+
+    let blocked = call_tool(&mut conn, "assay_check_args", serde_json::json!({}), 2);
+    let attempted_allow = call_tool(
+        &mut conn,
+        "assay_check_args",
+        serde_json::json!({"on_error": "allow"}),
+        3,
+    );
+
+    assert_fixed_failure(&blocked, "E_INTERNAL", "Tool execution failed");
+    assert_fixed_failure(&attempted_allow, "E_INTERNAL", "Tool execution failed");
+    assert_eq!(tool_payload(&blocked), tool_payload(&attempted_allow));
+    assert!(conn.shutdown().success());
+}
+
+#[test]
+fn unknown_names_are_value_free() {
+    let (mut conn, _root) = spawn_server(None);
+    initialize(&mut conn);
+    let hostile = format!(
+        "{HEAD}{}\u{1f642}{MID}{}{TAIL}",
+        "x".repeat(5_000),
+        "y".repeat(5_000)
+    );
+
+    let unknown_tool = call_tool(&mut conn, &hostile, serde_json::json!({}), 2);
+    assert_fixed_failure(&unknown_tool, "E_INTERNAL", "Tool execution failed");
+
+    let unknown_method = conn.request(&hostile, serde_json::json!({}), 3);
+    assert_eq!(unknown_method["error"]["code"], -32601);
+    assert_eq!(unknown_method["error"]["message"], "Method not found");
+
+    for response in [&unknown_tool, &unknown_method] {
+        let wire = serde_json::to_string(response).unwrap();
+        for sentinel in [HEAD, MID, TAIL] {
+            assert!(!wire.contains(sentinel), "response reflected {sentinel}");
+        }
+    }
+    assert!(conn.shutdown().success());
+}
+
+#[test]
+fn caller_cannot_fail_open_timeouts() {
+    let (mut conn, _root) = spawn_server(Some("1"));
+    initialize(&mut conn);
+
+    let response = call_tool(
+        &mut conn,
+        "assay_check_args",
+        serde_json::json!({
+            "tool": "any",
+            "arguments": {},
+            "policy": "slow.yaml",
+            "on_error": "allow"
+        }),
+        2,
+    );
+
+    assert_fixed_failure(&response, "E_TIMEOUT", "Tool execution timed out");
+    assert!(conn.shutdown().success());
+}
--- a/crates/assay-mcp-server/tests/outer_fallback_contract.rs
+++ b/crates/assay-mcp-server/tests/outer_fallback_contract.rs
@@ -1,6 +1,7 @@
 //! Wire-level contract for failures outside an individual MCP tool implementation.
 
 use serde_json::Value;
+use std::fs;
 use std::process::{Command, Stdio};
 
 mod jsonrpc_conn;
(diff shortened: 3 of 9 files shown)


Rul1an:
## Independent review — exact head `b8c18f3d52e9da56a8f26b357187886391c36bf9`

**Verdict: BLOCKED** on this head.

The security fix itself is real, correctly scoped, and well tested — I reproduced the pre-fix fail-open and confirmed it is gone. What blocks is a second, unrelated wire change that rides along on the same expression: two of the five advertised tools now report successful results as errors.

**Provenance.** Reviewed in a clean detached worktree at `b8c18f3d5` (`git worktree add --detach … b8c18f3d5`), base = `a63a012ba9c8b994649b334d4872afe600b563b1` (also `origin/main` tip at review time, so `merge-base HEAD origin/main == origin/main`). Host: darwin 25.6.0, arm64, `rustc 1.96.0 (ac68faa20 2026-05-25)`, debug profile. Head re-checked immediately before posting: `gh pr view 2401 --json headRefOid` → `b8c18f3d52e9da56a8f26b357187886391c36bf9`.

---

### What the change actually fixes (verified, not taken on trust)

I checked out the base `server.rs` into the worktree, rebuilt, and drove the real binary over stdio:

```
{"name":"nonexistent_tool","arguments":{"on_error":"allow"}}
```

Base `a63a012ba` answers:

```json
{"result":{"content":[{"type":"text","text":"{\n  \"allowed\": true,\n  \"error\": {\n    \"code\": \"E_INTERNAL\",\n    \"message\": \"Unknown tool: nonexistent_tool\"\n  },\n  \"warning\": \"FAIL-SAFE ACTIVE: Policy engine offline. Proceed with caution (Safe Mode).\"\n}"}],"isError":false},"id":2}
```

Head `b8c18f3d5` answers `allowed:false`, `E_INTERNAL`, `"Tool execution failed"`, `isError:true`. The caller-selected fail-open was genuinely reachable, and it is genuinely closed. The base also reflected the caller's tool name into the response through `e.to_string()` with no ceiling — the RED run below shows a 10 KB name echoed verbatim, so this also closes an unbounded reflection.

RED, against base `server.rs` with the new test file:

```
cargo test --locked -p assay-mcp-server --test outer_fallback_contract
test result: FAILED. 0 passed; 3 failed
  caller_cannot_fail_open_timeouts: left Bool(false), right … "allowed":true,"code":"E_TIMEOUT" … "isError":false
  unknown_names_are_value_free:     left String("Unknown tool: HEAD_SENTINELxxxx…🙂MID_SENTINELyyyy…TAIL_SENTINEL")
```

GREEN at head: 3/3 pass. Mutation `let is_error = !allowed;` → `let is_error = false;` (server.rs:435): 0 passed, 3 failed. Full crate: `cargo test --locked -p assay-mcp-server` — every suite `ok`, 0 failed (101 + 89 unit, plus 30 integration binaries). `cargo fmt --all -- --check` exit 0. `cargo clippy --locked -p assay-mcp-server --all-targets -- -D warnings` clean. All 24 check runs on this SHA are `success` (5 skipped, none failed). `GET /code-scanning/alerts?ref=refs/pull/2401/head` returns nothing open and nothing dismissed.

---

## Findings

### 1. BLOCKING — successful `assay_check_coverage` and `assay_explain_trace` results are now `isError: true`

`crates/assay-mcp-server/src/server.rs:371-374` and `:435`:

```rust
let allowed = result.get("allowed").and_then(|v| v.as_bool()).unwrap_or(false);
…
let is_error = !allowed;
```

The base computed the envelope flag from its own default:

```rust
let is_error = !res.get("allowed").and_then(|v| v.as_bool()).unwrap_or(true);
```

`assay_check_coverage` and `assay_explain_trace` never emit `allowed`. `git grep -n '"allowed"' crates/assay-mcp-server/src/tools/*.rs` lists `check_args`, `policy_decide`, `check_sequence`, `test_outbound`, `mod.rs` — neither `check_coverage.rs` nor `explain_trace.rs` appears, and reading both confirms every success arm returns `format`/`content`/`meets_threshold`/`blocked_steps` only. So the default flip silently turns their success envelope inside out.

Observed on the real binary at `b8c18f3d5` (policy root with a valid `policy.yaml`, one fully covered trace):

```
id 2 (assay_explain_trace, format=terminal)  isError= True | {"format":"terminal","content":"Policy: test (v1.1)\nTrace: 1 steps (1 allowed, 0 blocked)…
id 3 (assay_check_coverage, format=markdown) isError= True | {"format":"markdown","content":"## Coverage Report ✅ … | Overall Coverage | 100.0% |…
```

Same two calls against base `a63a012ba` return `isError: false`.

Why this is a finding and not a taste question:

- It buys no safety. The fail-closed path constructs `{"allowed": false, …}` explicitly via `ToolError::result()` (`tools/mod.rs:108-113`), so the fail-closed property never depended on the `unwrap_or` default. The default only governs handler-produced *successes*.
- MCP 2025-06-18 `server/tools`, §Error Handling, scopes the flag: "**Tool Execution Errors**: Reported in tool results with `isError: true`: API failures, Invalid input data, Business logic errors." A 100 %-coverage report that met its threshold is none of the three. (Checked against https://modelcontextprotocol.io/specification/2025-06-18/server/tools, the reference this PR itself cites.)
- It reaches users. `docs/guides/editor-mcp-recipe.md:101` and `docs/reference/cli/mcp-server.md:94` advertise both tools to editor clients; a client that honours `isError` will surface a successful coverage report to the model as a failed tool call.
- Nothing states it. The CHANGELOG entry names only the `arguments.on_error` migration, and the PR's non-claims list ("No new … tool schema, `resultType`, `structuredContent`") does not cover the `CallToolResult` flag. The plan doc prescribes `unwrap_or(false)` at `docs/superpowers/plans/2026-08-16-mcp-outer-fallback-hardening.md:131`, so it is deliberate — but deliberate and undeclared is the part that blocks.

Test gap that let it through: every `isError` assertion in the crate is on a failure path asserting `Some(true)` (`policy_diagnostic_safety.rs:127`, `policy_ingest_limits.rs:96,370`, `project_install_surfaces.rs:222`). The only `Some(false)` assertion is `policy_decide_blocklist.rs:98`, on `assay_policy_decide`, which *does* emit `allowed`. No test drives a success of a tool that omits `allowed` over the wire — exactly the route that regressed.

Cheapest fixes, either is fine: keep `unwrap_or(true)` for the envelope flag while leaving the fail-closed constructions explicit, or have the two tools emit `allowed: true` on success. Please add a wire test asserting `isError == false` for a successful `assay_check_coverage`, otherwise the next edit to this expression is unguarded again.

### 2. BLOCKING (one line) — `docs/reference/mcp-api.md:37` states something the code does not do

```
| `E_INTERNAL` | The outer tool dispatch failed; details are retained only in operator telemetry. |
```

Nothing retains them. `server.rs:346-355` binds and discards the error, and the event carries no error text:

```rust
Ok(Err(_error)) => {
    tracing::error!(
        event = "tool_execution_error", rid = %rid, rpc_id = ?req.id,
        duration_ms = …, code = "E_INTERNAL"
    );
```

Captured stderr for a failing dispatch, verbatim:

```json
{"event": "tool_execution_error", "rid": "r-000002", "rpc_id": "Some(Number(2))", "duration_ms": 0, "code": "E_INTERNAL"}
```

`git grep -n "error=%e" crates/assay-mcp-server/src/server.rs` matches only the `json_parse_error` site (line 273). The claim is also contradicted inside the same PR by `docs/guides/gateway-pattern.md`: "The server does not emit caller values or raw internal errors in this path." Pick one: either drop "retained only in operator telemetry", or log the error on the operator side and keep the sentence.

### 3. Non-blocking — client input errors are now indistinguishable from server faults, on the wire *and* in the logs

`check_args` (`tools/check_args.rs:9-18`), `check_coverage` (`tools/check_coverage.rs:51-52`) and `explain_trace` (`tools/explain_trace.rs:41-42`) report caller mistakes as `anyhow` errors, so they land in the same `Ok(Err(_))` arm. Observed at head:

```
{"name":"assay_check_coverage","arguments":{"policy":"policy.yaml","traces":"not-an-array"}}
→ {"allowed": false, "error": {"code": "E_INTERNAL", "message": "Tool execution failed"}}, isError: true
```

Base returned `"Missing 'tool' argument"` for the analogous case. Note those `.context(…)` strings are fixed literals, not caller values — the value-free requirement did not force dropping them, and MCP lists "Invalid input data" as a case a tool result is supposed to report usefully. An LLM client can no longer self-correct a missing argument, and because of finding 2 an operator cannot tell a malformed request from a broken server either. Distinct fixed messages (e.g. a `E_INVALID_REQUEST` mapping) would keep every value-free property intact.

### 4. Non-blocking — silent removal of `tool` and `policy` from three log events

`tool_call_start`, `tool_call_done` and `tool_call_timeout` lost `tool=name` and `policy=policy`. Defensible: both were caller-controlled. But per-tool attribution in operator telemetry disappears, and neither the summary nor the CHANGELOG mentions it. Observed:

```json
{"event": "tool_call_start", "rid": "r-000002", "rpc_id": "Some(Number(2))", "bytes_in": 5117, "args_bytes": 2}
{"event": "tool_call_done",  "rid": "r-000002", …, "outcome": "app_error", "allowed": false, "code": "E_INTERNAL"}
```

Related precision point, pre-existing and not introduced here: the same request whose sentinels the new test proves absent from the wire still writes the caller-controlled 5 KB name into stderr through `tool_decision` (`server.rs:414-431` → `tool_decision.rs:428` `sanitize(call.tool_name)`, which strips control chars but does not bound length). Measured: `grep -c HEAD_SENTINEL` → stdout 0, stderr 1, with a 5,956-byte `tool_decision` line. That makes gateway-pattern.md's "does not emit caller values … in this path" true only of the `tool_call_done` event it annotates. Worth a scoping word, not a change.

### 5. Non-blocking — the timeout test is debug-only by construction

`caller_cannot_fail_open_timeouts` is deterministic only because of `#[cfg(debug_assertions)]` in `tools/check_args.rs:32-36`, which sleeps 50 ms when the policy path contains `slow`; with `ASSAY_MCP_TIMEOUT_MS=1` that is a comfortable margin. In a release-profile test run the hook vanishes and the test becomes a race against a `spawn_blocking` file open. `cargo test` defaults to debug so CI is unaffected — flagging so nobody adds `--release` to this suite later.

Style nit, not a finding: `fail_closed_tool_result(…)?` at `server.rs:355` and `:365` propagates out of `Server::run`, i.e. an error there kills the read loop rather than answering the request. It is unreachable in practice (`ToolError::result()` re-serialises an already-built `Value`), so this is only about the shape.

---

### Things I checked that are correct

- The `-32601` narrowing is right and the negotiation test was updated in step (`protocol_negotiation_e2e.rs:231`), so `server/discover` no longer reflects the method name.
- `docs/reference/cli/mcp-server.md` states the unknown-tool non-claim accurately: unknown tools stay a fail-closed `CallToolResult` and it explicitly does not claim `-32602` routing. MCP 2025-06-18 does list "Unknown tools" under protocol errors, so naming the deviation rather than papering over it is the honest read.
- The `tool_call_done` and `assay.usage.metered` samples added to `docs/guides/gateway-pattern.md` match the emitted events field for field — I diffed them against captured stderr.
- `assert_fixed_failure` discriminates properly: exact code, exact message, `allowed:false`, `isError:true`, and absence of the `warning` key, plus a sentinel scan over the whole serialized response rather than one field.
- Branch is not behind: `merge-base HEAD origin/main == origin/main == a63a012ba`, `mergeStateStatus: CLEAN`.

### Non-claims

- Ran `-p assay-mcp-server` only, on macOS/arm64, debug profile. No `--workspace` test run, no workspace-wide clippy, no Linux or Windows build — for those I relied on the check runs on this SHA (`Build + Test (ubuntu-latest)`, `(windows-latest)`, `Clippy (deny warnings)`, `Rustdoc (deny warnings)` all `success`), which is CI's word, not mine.
- Did not exercise the proxy, auth, eBPF, or `assay run` paths, nor release-profile behavior.
- Did not audit the `tool_decision` surface itself beyond the one length/sanitization observation; it is unchanged by this PR.
- Did not verify the PR body's "clean upstream advance containing #2400 only" beyond confirming the merge-base equals `a63a012ba`.
- Code-scanning returned zero alerts (open and dismissed) for `refs/pull/2401/head`; I did not verify whether the CodeQL configuration analyses Rust at all, so read that as "nothing reported", not "nothing there".
- Findings 1 and 3 are behavioral observations from driving the binary; I did not survey downstream consumers to quantify how many clients branch on `isError`.

Rul1an:
## Independent review — exact head `e2eccea654e9be934d4f05c2159d4f27acf9ccc8`

**Verdict: READY** for `e2eccea654e9be934d4f05c2159d4f27acf9ccc8`.

Non-building reviewer: authored no commit on this branch and no artifact governing it. I did not inherit the earlier review on `b8c18f3d`; that head is not this head and I reviewed this one from scratch.

Head confirmed equal to `e2eccea65` at the start of the review and again immediately before recording this verdict (`gh pr view 2401 --json headRefOid` → `e2eccea654e9be934d4f05c2159d4f27acf9ccc8`, `state=OPEN`, `mergeStateStatus=CLEAN`).

Method: clean detached worktree at `e2eccea65` outside the working tree, plus a second worktree at the merge base `a63a012ba9c8b994649b334d4872afe600b563b1` (`git merge-base pr/2401 origin/main`) so behavioral claims are measured as a difference rather than asserted. Both worktrees removed; `git status --short` clean.

---

### The core claim holds, measured on the wire

Both binaries built with `cargo build --locked -p assay-mcp-server`, driven over real stdio with the same request file (valid `initialize`, then a `tools/call` with `{"on_error":"allow"}` and missing required fields, an unknown tool name carrying sentinels, and an unknown method).

Base `a63a012ba`:

```
{"result":{"content":[{"type":"text","text":"{\n  \"allowed\": true,\n  \"error\": {\n    \"code\": \"E_INTERNAL\",\n    \"message\": \"Missing 'tool' argument\"\n  },\n  \"warning\": \"FAIL-SAFE ACTIVE: Policy engine offline. Proceed with caution (Safe Mode).\"\n}"}],"isError":false},"id":2}
{"result":{"content":[{"type":"text","text":"{\n  \"allowed\": true, ... \"Unknown tool: HEAD_SENTINEL_unknown_TAIL_SENTINEL\" ... }"}],"isError":false},"id":3}
{"error":{"code":-32601,"message":"Method not found: METHOD_SENTINEL/nope"},"id":4}
```

Head `e2eccea65`:

```
{"result":{"content":[{"type":"text","text":"{\n  \"allowed\": false,\n  \"error\": {\n    \"code\": \"E_INTERNAL\",\n    \"message\": \"Tool execution failed\"\n  }\n}"}],"isError":true},"id":2}
{"result":{"content":[{"type":"text","text":"{\n  \"allowed\": false, ... \"Tool execution failed\" ... }"}],"isError":true},"id":3}
{"error":{"code":-32601,"message":"Method not found"},"id":4}
```

So a caller-supplied `arguments.on_error: "allow"` really did turn an unknown tool name and a missing-argument failure into `allowed: true, isError: false` on the base, and no longer does. No sentinel reaches any response. The process survives the error path: request 4 is answered after requests 2 and 3 fail.

`tools/list` at head contains zero occurrences of `on_error`, so the doc claim that it is "absent from the advertised input schemas" is true and not merely aspirational.

Spec check against the primary text, MCP 2025-06-18 `server/tools`, Error Handling: protocol errors cover "Unknown tools", "Invalid arguments", "Server errors", with the example `{"code": -32602, "message": "Unknown tool: invalid_tool_name"}`; tool execution errors are "Reported in tool results with `isError: true`". Returning a tool result for an unknown tool is therefore a divergence from the spec's guidance, and `docs/reference/cli/mcp-server.md:86-88` says so explicitly rather than papering over it. Good non-claim. `isError: true` for the dispatch-failure envelope is conformant.

### Verification run at this head

- `cargo test --locked -p assay-mcp-server` — all suites pass (largest: 101, 89, 35; 2 ignored; 0 failed).
- `cargo clippy --locked -p assay-mcp-server --all-targets -- -D warnings` — clean.
- `cargo fmt --all -- --check` — clean.
- `git diff --check a63a012ba..HEAD` — clean.
- CI on this SHA: 57 check runs, 0 failures, 5 skipped (`eBPF monitor smoke` x2, `Integration Parity (Optional)`, `Wave 0.1 nightly safety`, `Archive closed PR branch with Bencher`). `Clippy (deny warnings)`, `Rustdoc (deny warnings)`, `CodeQL`, `claims-boundary`, `lane-check`, `mkdocs-strict`, `local-links`, `public-artifact-sanitization` all success. No dismissed alerts involved; the PR touches no security tooling config.

### Mutation sensitivity, run rather than reasoned

Each mutation applied to the worktree, full `cargo test -p assay-mcp-server`, then reverted.

| Mutation | Result |
|---|---|
| publish `format!("{_error}")` as the E_INTERNAL message | **caught** — `caller_cannot_fail_open_handler_errors`, `unknown_names_are_value_free` FAILED |
| timeout returns `{"allowed": true, ...}` | **caught** — `caller_cannot_fail_open_timeouts` FAILED |
| force `"isError": false` | **caught** — 3 of 4 tests FAILED |
| revert to `b8c18f3d`'s `let is_error = !allowed` | **caught** — `report_tools_keep_successful_mcp_results` FAILED |
| `classify_tool_result`: `explicit_allowed.unwrap_or(!has_error)` → `unwrap_or(false)` | **survived**, see finding 3 |
| `classify_tool_result`: `has_error \|\| explicit_allowed == Some(false)` → `explicit_allowed == Some(false)` | survived; benign, no shipped tool returns `error` without an explicit `allowed`, so nothing can discriminate |

The security-relevant assertions discriminate. The last commit's fix is genuinely pinned.

---

## Findings

### 1. Unannounced change to `assay.tool_decision_surface.v0` for the two report tools — non-blocking

`classify_tool_result` (`crates/assay-mcp-server/src/server.rs:19-24`) now derives `allowed` from the absence of `error` when the payload has no `allowed` key. `assay_check_coverage` and `assay_explain_trace` never emit `allowed` (grep for `"allowed"` in `crates/assay-mcp-server/src/tools/` returns nothing for those two files), so their emitted decision flips.

Same request file, both builds:

```
base : tool_call_done ... "tool":"assay_check_coverage" ... "outcome":"ok","allowed":false
       tool_decision  ... "effect":"deny","status":"blocked","enforced":true
head : tool_call_done ... "outcome":"ok","allowed":true
       tool_decision  ... "effect":"allow","status":"success","enforced":true
```

The wire `isError` is `false` on both, so this is log/evidence only. I think head is the more defensible of the two (a successful coverage report recorded as `effect: deny, enforced: true` was a false denial record), but neither is obviously right for a tool that makes no permission decision, and the change is not in the CHANGELOG entry, not in the PR's "Scope and non-claims", and not pinned by any test. If anything downstream counts denials or allows off this surface, it will move.

### 2. Log-schema break not announced anywhere — non-blocking

Removed from `tool_call_start` / `tool_call_done` / `tool_call_timeout`: the `tool` and `policy` fields. Removed entirely: the `tool_call_crash` event, replaced by `tool_execution_error`. Verified on both builds' stderr; base emits `"event":"tool_call_crash", ..., "tool":"assay_check_args","policy":"policy.yaml"`, head emits neither field on any `tool_call_*` event.

Dropping the caller-controlled `policy` string from logs is a defensible hardening. Dropping `tool` costs operators per-event tool attribution; it is recoverable via `rid` correlation with the adjacent `tool_decision` event, which still carries `tool.name`. `docs/guides/gateway-pattern.md` does redirect alerting to `tool_call_done` + `code`, but nothing states that `tool_call_crash` no longer exists or that `tool`/`policy` were dropped. An alert rule keyed on `event=tool_call_crash` goes silent on upgrade.

### 3. The `allowed` half of `classify_tool_result` is mutation-insensitive — non-blocking

Mutating `let allowed = explicit_allowed.unwrap_or(!has_error);` to `unwrap_or(false)` leaves the whole `-p assay-mcp-server` suite green. That line is exactly what produces the finding-1 flip, and it governs `tool_call_done.allowed` and the `tool_decision` `effect`/`status`. `report_tools_keep_successful_mcp_results` only asserts `isError == false` and the absence of `error`, which the `is_error` half already determines independently. A test asserting the emitted `tool_decision` effect for a report tool would close this; today the property the last commit was written to establish is only half covered. (Positive control confirms the harness bites: the raw-message mutation on the same file failed two tests.)

### 4. Caller-input errors are now indistinguishable from server faults — non-blocking

`Missing 'tool' argument` becomes `E_INTERNAL / "Tool execution failed"` (measured above). Deliberate and documented as value-free, but the message carried no caller value in the first place, and MCP 2025-06-18 lists "Invalid arguments" alongside "Unknown tools" under protocol errors. `docs/reference/cli/mcp-server.md:86-88` declares the non-claim for unknown *tool names* only; the same divergence for invalid arguments is not named. Worth one sentence, or a follow-up that splits argument-validation failures from dispatch faults.

### 5. Docs precision on "no caller values" — non-blocking

`docs/guides/gateway-pattern.md`: "The server does not emit caller values or raw internal errors in this path." Scoped to `tool_call_done` this is true. In the same stderr stream, for the same `rid`, the `tool_decision` event emits the caller-supplied tool name verbatim, proven at head:

```
"event":"tool_decision","rid":"r-000003", ... "tool":{"name":"HEAD_SENTINEL_unknown_TAIL_SENTINEL","category":null} ...
```

Pre-existing behavior, not a regression, and the wire responses are genuinely value-free. But `docs/superpowers/plans/2026-08-16-mcp-outer-fallback-hardening.md:135` states the obligation as "Do not publish or log `_error`, the caller tool name, ...", and the caller tool name is still logged. Either scope the sentence to the fail-closed publication path or record the `tool_decision` carve-out.

### 6. The landed plan prescribes the form the PR had to remove — non-blocking

`docs/superpowers/plans/2026-08-16-mcp-outer-fallback-hardening.md:128-131` ships `let is_error = !tool_result.get("allowed").and_then(Value::as_bool).unwrap_or(false);`. That is precisely the `b8c18f3d` code that broke `assay_check_coverage` and `assay_explain_trace` (my revert mutation reproduces the failure), and commit `e2eccea65` replaced it. The plan is checked in as repo documentation and now contradicts the shipped code. Same file line 13 pins `Base SHA: d1bfd36a...`, which is not this PR's merge base (`a63a012ba`). Unchecked checkboxes are the house convention here (the two 2026-08-15 MCP plans are 0/25 and 0/21), so that is not a finding.

### 7. Minor imprecisions — non-blocking

`docs/concepts/fail-safe.md:132` says "The five built-in policy tools fail closed when dispatch fails or times out". The boundary is the dispatch site, so it also covers unknown tool names and the feature-gated `assay_test_outbound`. `docs/reference/cli/mcp-server.md` gets this right.

`fail_closed_tool_result(...)?` propagates out of `Server::run`, so a serialization failure on the fail-closed path would stop the process rather than fail closed. Unreachable in practice (`ToolError::result` serializes two `&'static str` and `details: None`), so noting only.

---

## What holds up cleanly

- The authority inversion is genuinely removed. No code path reads `arguments.on_error`; `grep -rn on_error` across the tree at this SHA finds it only in `assay run` suite settings, the CHANGELOG, docs prose, and the two test cases that prove it is inert.
- Fixed messages contain no caller-derived substring, proven with a 10 KB three-sentinel hostile name across both the unknown-tool and unknown-method responses.
- `Method not found` matches JSON-RPC 2.0 and the repo's own Python fixtures (`tests/echo_server.py:35`, `scripts/ci/exp-mcp-fragmented-ipi/mock_mcp_server.py:142`). No other consumer in the tree greps the old `Method not found: <method>` form; the only two assertions on it are updated in this PR.
- Every error code named in `docs/reference/mcp-api.md` exists in source: `E_POLICY_NOT_FOUND`, `E_POLICY_READ`, `E_PERMISSION_DENIED` in `tools/` and `security.rs`; `E_INTERNAL` and `E_TIMEOUT` in `server.rs`.
- Scope is confined: `src/server.rs`, one new test file, one updated assertion, four docs, CHANGELOG, one plan. No public API change (`fail_closed_tool_result` and `classify_tool_result` are private), no visibility widened, no lint relaxed, no `#[allow]` added.
- The stated non-claims hold. No new config key, no `resultType`, no `structuredContent`, no schema validation, no proxy or auth change, no 2026-07-28 conformance claim, `settings.on_error` untouched.
- Compatibility is announced where it breaks: the CHANGELOG names the removed fail-open and tells clients to drop the argument. Clients that still send `on_error` are not rejected, only ignored (the timeout test sends it and reaches the tool normally).

---

## Non-claims for this review

- Ran only `-p assay-mcp-server`. Not run: `--workspace`, `assay-cli`, the Python SDK, or `assay-sim`. The change is private to one crate and CI's workspace jobs are green on this SHA, but I did not execute them myself.
- macOS `darwin 25.6.0` only. No Linux build, no eBPF path, no Windows. The two eBPF smoke checks are `skipped` on this SHA.
- Debug profile only. `caller_cannot_fail_open_timeouts` depends on the `#[cfg(debug_assertions)]` slow hook in `check_args.rs:32-35`, so it would not exercise a timeout under `--release`. Pre-existing pattern shared with `stdio_edge_cases::test_timeout`, and no workflow runs `cargo test --release`, so it is not live.
- Did not exercise `proxy` or `proxy-enforce` modes, the auth paths, or `packaging/mcpb` / plugin packaging.
- Did not audit the unchanged remainder of `docs/concepts/fail-safe.md` (the `assay run` half) beyond checking it no longer claims MCP fail-open.
- Read but did not independently re-derive the `tool_decision` classifier; findings 1 and 5 are about what it emits, measured, not about whether its classification logic is correct.
- Did not evaluate the CodeRabbit comment or the prior review on `b8c18f3d`.

Rul1an:
## Independent exact-head review — **READY** at `bbf3f61e803ea2de55a35d602b8c89ac887bbd3b`

Non-building reviewer. I authored no part of this change and no specification or plan governing it. Reviewed in a clean detached worktree at the SHA above; head re-confirmed as `bbf3f61e803ea2de55a35d602b8c89ac887bbd3b` immediately before recording this verdict. Worktree removed afterwards.

Base used for the diff: `git merge-base origin/main bbf3f61e8` = `a63a012ba9c8b994649b334d4872afe600b563b1`. 9 files, +628/−197.

### The change does what it claims

Measured on the wire, same nine-request stdio script driven against a build of the base (`a63a012ba`) and of the head, both `cargo build --locked -p assay-mcp-server`:

| request | base `a63a012ba` | head `bbf3f61e8` |
|---|---|---|
| `assay_check_args` `{}` | `isError=true allowed=false` msg `Missing 'tool' argument` | `isError=true allowed=false` msg `Tool execution failed` |
| `assay_check_args` `{"on_error":"allow"}` | **`isError=false allowed=true`** + `warning:"FAIL-SAFE ACTIVE: Policy engine offline…"` | `isError=true allowed=false`, no `warning` |
| unknown tool, 4 KiB hostile name | name reflected verbatim in `error.message` | fixed `E_INTERNAL` / `Tool execution failed` |
| unknown method, hostile name | `-32601 "Method not found: HEAD_SENTINELxxx…"` | `-32601 "Method not found"` |
| `assay_check_coverage` (report tool) | `isError=false` | `isError=false` (unchanged) |
| `assay_check_args` allow decision | `isError=false allowed=true` | unchanged |
| `assay_check_args` deny decision | `isError=true allowed=false` | unchanged |
| `E_POLICY_NOT_FOUND` + `on_error:allow` | `isError=true allowed=false` | unchanged |

Sentinel count over the full response stream: base stdout 2 per sentinel → head **0**. The caller-authority inversion is real and the fix is real.

`tools/list` at head still advertises exactly the five production tools and the string `on_error` appears nowhere in the advertised schemas, so `docs/reference/cli/mcp-server.md` ("absent from the advertised input schemas") and `docs/guides/gateway-pattern.md` check out.

### Verification run

```
cargo test  --locked -p assay-mcp-server            → 425 passed, 0 failed, 2 ignored (exit 0)
cargo clippy --locked -p assay-mcp-server --all-targets -- -D warnings → clean
cargo fmt --all --check                             → clean
git diff --check a63a012ba bbf3f61e8                → clean
```
Toolchain: host `darwin` / `aarch64-apple-darwin`, worktree `…/scratchpad/wt-2401`, tree at `bbf3f61e8`.

CI at this exact SHA: `gh api repos/Rul1an/assay/commits/bbf3f61e8…/check-runs` → 25 success, 5 skipped, 0 failure. Nothing dismissed.

### Mutation sensitivity

Each mutation applied to `crates/assay-mcp-server/src/server.rs` in a disposable worktree, restored via `git checkout --` after each (final `git status --short` empty).

| # | mutation | outcome |
|---|---|---|
| M2 | `"isError": is_error` → `"isError": false` | **caught** (`outer_fallback_contract`) |
| M3 | handler-error branch re-reads `args.on_error` to set `allowed` | **caught** |
| M4 | `-32601` message re-interpolates `req.method` | **caught** (`outer_fallback_contract` + `protocol_negotiation_e2e`) |
| M5 | timeout returns `allowed:true` | **caught** |
| M6 | fallback publishes `_error.to_string()` | **caught** |
| M7 | `is_error = !allowed` | **caught** — and specifically by `report_tools_keep_successful_mcp_results`, i.e. the new report-tool row fails on the real stdio route, not on a neighbouring one |
| M8 | `explicit_allowed.unwrap_or(false)` → `unwrap_or(true)` | **caught** by the unit matrix |
| M1 | drop the `has_error \|\|` clause from `is_error` | **survived** |
| M9 | `let has_error = result.get("error").is_some();` → `= false` | **survived** |

Rows 1, 2 and 4 of `tool_result_classification_separates_decision_from_mcp_error` each have a mutation that fails them, so they discriminate. Row 3 (`{"allowed": false, "error": {…}}`) does not add discrimination over row 2 — see finding 3.

---

## Findings

### 1. Non-blocking — the shipped plan states a stderr property that is false at this head

`docs/superpowers/plans/2026-08-16-mcp-outer-fallback-hardening.md:149`:

> Expected: all pass; stderr remains structured and does not contain the hostile sentinels.

Measured at `bbf3f61e8`, driving the built binary with a 4 KiB `HEAD_SENTINEL…MID_SENTINEL…TAIL_SENTINEL` tool name and capturing stderr:

```
stderr HEAD_SENTINEL: 1 occurrences     (base a63a012ba: 4)
stderr MID_SENTINEL:  1 occurrences     (base: 4)
stderr TAIL_SENTINEL: 1 occurrences     (base: 4)
```

The remaining occurrence is the pre-existing `tool_decision` event, which emits `sanitize(call.tool_name)` verbatim (`crates/assay-mcp-server/src/tool_decision.rs:428`, and `sanitize` at `:332` replaces control characters but does not bound length or drop the value). The PR reduces stderr reflection 4× → 1×; it does not eliminate it.

The `CHANGELOG.md:12-14` wording is accurate — it scopes the claim to `tool_call_*` and explicitly says `tool_decision` keeps its existing redaction contract. The plan document is the artifact that overstates, and planning documents that ship in the repo are held to the same standard as reference docs. Either narrow line 149 to the `tool_call_*` events or say plainly that `tool_decision` still carries the tool name.

Adjacent and softer: `docs/concepts/fail-safe.md:148-149` puts "`assay-mcp-server` emits structured stderr events…" immediately before "Public failure messages do not include caller values or raw internal errors." Read as written it is about messages and is true; read as covering the stderr events just named, it is not.

### 2. Non-blocking — a named log event disappears and nothing says so

Base emits `event="tool_call_crash"` on a handler error. At head that event does not exist anywhere in the tree (`git grep -n tool_call_crash` → no hits). `CHANGELOG.md:12` says "Outer dispatch failures emit `tool_execution_error`" but never says `tool_call_crash` is gone, so a log pipeline alerting on it goes quiet without a migration note.

`tool_execution_error` also changed shape in the same commit — base fields `{event, rid, error, fallback}`, head fields `{event, rid, rpc_id, duration_ms, code}` — and `tool_call_start` / `tool_call_done` / `tool_call_timeout` lost `tool` and `policy`. Dropping the caller-controlled `policy` is the point of the change and dropping `tool` is defensible, but the net effect is that `tool_call_done` no longer says which tool ran; only the `tool_decision` event does. Worth one line in the changelog naming the removed event and the removed fields.

### 3. Non-blocking — one clause of `classify_tool_result` is both untested and unreachable

`crates/assay-mcp-server/src/server.rs:19-25`:

```rust
fn classify_tool_result(result: &Value) -> (bool, bool) {
    let has_error = result.get("error").is_some();
    let explicit_allowed = result.get("allowed").and_then(Value::as_bool);
    let allowed = explicit_allowed.unwrap_or(false);
    let is_error = has_error || explicit_allowed == Some(false);
    (allowed, is_error)
}
```

Both M1 (drop `has_error ||`) and M9 (`has_error = false`) survive the whole crate suite. The reason is structural, not a gap in the matrix alone: the only construction of a tool payload carrying `error` is `ToolError::result()` (`crates/assay-mcp-server/src/tools/mod.rs:110-115`), which always pairs it with `"allowed": false`. So no reachable input distinguishes the two clauses. Either add a `{"error": {…}}`-without-`allowed` row to the matrix to pin the clause as a deliberate defence, or drop it. Not a correctness problem today.

### 4. Non-blocking, but a merge precondition — the PR body's review packet is bound to a superseded head

The body records `base: d1bfd36a11b857…` and `head: b8c18f3d52e9da…`, and attributes RED/GREEN and the mutation witnesses to that head. At review time the head is `bbf3f61e8`, the merge-base with `origin/main` is `a63a012ba`, and two commits land after `b8c18f3d`:

- `e2eccea65 fix(mcp): preserve successful report tool results`
- `bbf3f61e8 fix(mcp): preserve report decision telemetry`

Those are not cosmetic. At `b8c18f3d` the wrapper computed `let is_error = !allowed;`, which made every report-tool response (`assay_check_coverage`, `assay_explain_trace`) come back as `isError: true` — a regression the last two commits fix and now cover with `report_tools_keep_successful_mcp_results`. AGENTS.md "Measurement provenance" asks that a pass/fail reaching a PR body name the SHA the reported tree was committed as. Refresh the packet to `bbf3f61e8` before merge; the code itself is fine.

### 5. Non-blocking — pre-existing false statement inside the section this PR rewrote

`docs/reference/mcp-api.md:7`:

> Protocol-level errors (invalid JSON) return a JSON-RPC `error`.

Measured: feeding `this is not json` between two valid requests produces **no response line at all** — `crates/assay-mcp-server/src/server.rs:283` is `continue; // Ignore invalid JSON lines (stdio transport robustness)`. The line predates this PR and is untouched by it, but it sits three lines above the block this PR rewrote, so it is cheap to correct while the file is open.

### 6. Non-blocking — caller-argument faults are now reported as `E_INTERNAL`

`check_args` validates required arguments with `anyhow::Context` (`crates/assay-mcp-server/src/tools/check_args.rs:6-18`), so a client that omits `tool` takes the `Ok(Err(_))` arm and receives `E_INTERNAL` / `Tool execution failed`. A caller mistake is now indistinguishable on the wire from a server fault, and the docs describe `E_INTERNAL` only as "The outer tool dispatch failed". The value-free message is the announced intent and I am not asking for it to be reverted; the code merely conflates two fault classes under one code, and `docs/reference/mcp-api.md:37` could say so. MCP 2025-06-18 §Error Handling puts "Invalid arguments" under protocol errors, but the PR already carries an explicit non-claim on that routing (`docs/reference/cli/mcp-server.md`: "this does not claim JSON-RPC `-32602` routing for unknown tools"), so this is a note, not a conformance failure.

### 7. Non-blocking nit

`docs/concepts/fail-safe.md:165` — "Verify that the configured failure policy was applied" — sits in a section that is now explicitly about `assay-mcp-server`, where there is no configurable failure policy. Reads as if one exists.

---

## Things I checked and found clean

- **Scope.** The diff touches `CHANGELOG.md`, `crates/assay-mcp-server/src/server.rs`, two tests in that crate, four docs, and one plan document. No lint suppression added, no visibility widened (`classify_tool_result` and `fail_closed_tool_result` are module-private; the test module reaches them through `super::`). Nothing outside `assay-mcp-server` changes.
- **Stated non-claims.** No new configuration key, no operator fail-open, no schema validation, no change to the proxy or auth paths, no `resultType` / `structuredContent`, no 2026-07-28 surface. `settings.on_error` really does remain an `assay run` setting (`crates/assay-core/src/model/types.rs:50,77,241`) and `assay-core` is untouched.
- **Wire compatibility for well-behaved clients.** For a client that never sent `arguments.on_error`, the only changes are the `E_INTERNAL` message text and the disappearance of the `warning` key; `isError` for allow, deny, report-tool and inner-tool-error results is byte-identical to base in all four measured cases.
- **Protocol.** Per MCP 2025-06-18 `server/tools` §Error Handling, tool execution errors are "Reported in tool results with `isError: true`", which is what the fallback now does. The spec puts unknown tools under protocol errors with a `-32602` example; this server keeps returning a `CallToolResult` for unknown tools, and the PR states that as an explicit non-claim rather than papering over it. `-32601` / "Method not found" matches JSON-RPC 2.0.
- **Flake risk in the new timeout test.** `caller_cannot_fail_open_timeouts` depends on a 1 ms `ASSAY_MCP_TIMEOUT_MS` winning a race against policy resolution. 40 consecutive isolated runs, 0 failures.
- **Docs against source.** `E_INTERNAL` / `E_TIMEOUT` codes and their exact messages, `allowed:false` + `isError:true`, `assay_check_args` argument names (`tool` / `arguments` / `policy`), `assay_check_sequence` argument names (`history` / `next_tool` / `policy`), the `assay.usage.metered` / `assay_billing` example, and the `tool_call_done` / `outcome:"app_error"` / `code:"E_INTERNAL"` alert example in `gateway-pattern.md` — all verified against emitted output, not against the source only.
- Removing the `assay.failsafe.triggered` example from `gateway-pattern.md` is correct: that event is emitted by `assay_core::on_error::log_fail_safe` (`crates/assay-core/src/on_error.rs:74`), i.e. the `assay run` path, never by the MCP server.

## Non-claims for this review

- Built and ran on macOS `aarch64-apple-darwin` only. No Linux or Windows build; CI covers those and is green, which is a different artifact from my measurement.
- Ran `-p assay-mcp-server` only. I did not run `--workspace`, `assay-core`, `assay-cli`, or any eBPF/Linux-gated target; the diff touches no other crate, so I judged that sufficient, but I did not prove it.
- I did not exercise the proxy paths (`crates/assay-mcp-server/src/proxy/`), the auth boundary, or the enforcing-proxy suites beyond what `cargo test -p assay-mcp-server` runs by default.
- I did not run `pre-commit`, `mkdocs build --strict`, or the link checker locally; I read their CI conclusions at this SHA instead.
- The `E_TIMEOUT` path was measured through the crate's own tests and through the base/head log comparison, not through a separate hand-driven timeout script; the handler-error, unknown-tool, unknown-method, report-tool, allow, deny and inner-error paths were hand-driven.
- The stderr sentinel counts are for the built `debug` binary with the default `tracing` JSON subscriber; I did not check other subscriber configurations.
- I read `docs/superpowers/plans/2026-08-16-mcp-outer-fallback-hardening.md` for truth of its claims, not as an authority over the code.

**Verdict: READY at `bbf3f61e803ea2de55a35d602b8c89ac887bbd3b`.** All seven findings are non-blocking. Finding 4 is a merge precondition rather than a code defect: the PR body's verification packet should be re-pointed at this head before auto-merge, since the packet it currently carries describes a tree in which report tools returned `isError: true`.

Rul1an:
Review findings disposition at exact head `bbf3f61e803ea2de55a35d602b8c89ac887bbd3b`: the stale packet is corrected in the PR body. All non-blocking telemetry/documentation findings are consolidated as #2402 with measured scope and DoD. They do not alter the security boundary fixed here. No code push was made, so the exact-head READY review remains valid.

Rul1an:
## Independent review (Ruley)

Not the builder. I did not author `codex/2391-outer-fallback`, the governing plan, or issue #2391. This is not a self-review and does not inherit prior findings.

**Head reviewed:** `bbf3f61e803ea2de55a35d602b8c89ac887bbd3b` (live PR head; the packet in the PR body still names `b8c18f3d`).
**Base:** `a63a012ba9c8b994649b334d4872afe600b563b1`.
**Checks on this head (completed):** `CI` success, `host-capability-check` success, `lane-check/proof` success. No incomplete check-runs.

### Verdict

READY. No blocking findings.

### Findings

None blocking.

Non-blocking:
- PR is still draft.
- PR body review-packet head is stale versus this SHA.
- `caller_cannot_fail_open_timeouts` depends on the existing debug-only `slow` sleep in `check_args`; that is enough to pin the timeout wrapper here, not a release-profile timeout proof.

### What I measured

Dispatch no longer reads `arguments.on_error`. Handler `Err` and timeout both go through `ToolError::new` with fixed messages (`E_INTERNAL` / `Tool execution failed`, `E_TIMEOUT` / `Tool execution timed out`). Unknown methods are JSON-RPC `-32601` with fixed `Method not found`. Unknown tools stay a fail-closed `CallToolResult`. Report tools without `allowed` keep `isError: false`.

Local, this SHA, rustc 1.96.0:
- `cargo test --locked -p assay-mcp-server --test outer_fallback_contract` → 4/4
- `cargo test --locked -p assay-mcp-server` → all suites green (0 failed)
- Cheap mutations, then restore (no leftover dirty tree):
  - restore caller `on_error: allow` on handler `Err` → `caller_cannot_fail_open_handler_errors` fails (`isError: false`, `allowed: true`)
  - reflect `req.method` in `-32601` → `unknown_names_are_value_free` fails (sentinels in the message)
  - publish `_error.to_string()` as `E_INTERNAL.message` → same test fails (`Missing 'tool' argument`)

### Non-claims

Not a live exploit claim. Not MCP 2026-07-28 conformance. Not a required-check or settings change. Extra reviews of this comment are not additional quorum. This record is bound to `bbf3f61e803ea2de55a35d602b8c89ac887bbd3b` only.

Issue #2195, opened 2026-08-09
dx(plugin): design a cross-platform MCP server version preflight


Rul1an:
## Problem

The Claude plugin manifest intentionally invokes `assay-mcp-server --policy-root .` from `PATH`. PR #2192 proves that shape with Claude Code 2.1.32 and an exact-SHA server, but the manifest cannot itself prove that a user's PATH resolves a compatible Assay version.

Adding a launcher inside #2192 would introduce a second version literal, platform-specific executable discovery, and binary lifecycle ownership without a reviewed distribution contract.

## Scope

- Measure failure UX for missing, stale, and unexpected `assay-mcp-server` binaries on macOS, Linux, and Windows.
- Choose one source of truth for the required compatibility range.
- Design a preflight or launcher that reports a structured, actionable diagnostic without downloading or silently replacing binaries.
- Keep `--policy-root .` and the five-tool production surface pinned.
- Add real-client workflow coverage for the selected design.

## Acceptance

- No duplicated free-floating version literals.
- Missing and incompatible binaries fail with a named phase and next step.
- Native Windows behavior is measured, not inferred from POSIX.
- The plugin manifest remains portable across supported Claude Code hosts.

## Non-goals

- No bundled binary or automatic updater without a separate supply-chain decision.
- No change to MCP tool semantics.

Origin: technical disposition of the launcher/version-pin suggestion on PR #2192.

Rul1an:
Priority sweep 2026-08-12: P1 for marketplace/plugin adoption, not release-blocking. The shipped manifest depends on assay-mcp-server from PATH; missing or incompatible binaries currently fail before users reach the product. Keep the accepted scope: cross-platform measured preflight, one compatibility source of truth, no silent download/replacement.

Rul1an:
## Baseline from the 2026-08-14 developer audit

This remains the correct `priority:P1` owner for the binary prerequisite gap.

The Claude plugin installation contract passes at `origin/main` SHA `b6c9e52851476222216773f489d2258f2ea99ab4`, but the plugin intentionally does not package `assay-mcp-server`. A developer must install the binary separately on `PATH`; a missing or stale binary is currently diagnosed at the host/spawn layer rather than by an Assay preflight.

Design constraints for the eventual fix:

- compare the plugin/recipe contract to the running binary version, not only to a second literal;
- work on macOS, Linux and Windows path/extension rules;
- distinguish missing, unstartable, wrong-version and policy-root failures;
[...]

PR #2398, opened 2026-08-15
fix(mcp): bound policy-file ingest before materialisation

Rul1an:
## ADR Boundary Slice

- Canonical ledger: no programme is active (AGENTS.md). Parent tracking issue: #2388
- Slice: #2389 — bounded MCP policy-file ingest
- Builder: Cursor (`cursor/2389-bounded-policy-ingest`)
- Reviewers: one non-building agent required on this exact head; builder self-review does not count. The prior review on `ed0f9188` is invalidated by this push.
- Final head SHA: `886008b9e5df4acb4f2c0aaa34d102fbbfe6c1db`
- Worktree: `/Users/roelschuurkes/wt-2389-bounded-policy-ingest`
- Base: `origin/main` `bb1bbc5a43d9e3f737d22cf0cb8bc4ef307fc05f`
- Merge: clean `--no-ff` of `bb1bbc5a` into `e6f8c926e9ecd94649015c72ce68a0ae4973d2c5`. Parents of that merge: `e6f8c926e9ecd94649015c72ce68a0ae4973d2c5` `bb1bbc5a43d9e3f737d22cf0cb8bc4ef307fc05f`. Incoming file: only `crates/assay-mcp-server/tests/tools_list_wire.rs`. RED commit `a3b92c70d3eb593afc1e08b82e920b984f0c7963` was not rewritten.
- Follow-up on this head: CodeQL `py/path-injection` (#748–#753) — both architecture guards now accept exactly zero args and read only cwd-relative compile-time allowlisted paths. Mutation harnesses `cd` into the temp tree and invoke the absolute guard with no path argument. Rustdoc `private-intra-doc-links` — `read_policy_bounded` now names `read_bounded` as plain code text; visibility was not widened and no `allow` was added.
- Toolchain: rustc 1.96.0 (ac68faa20 2026-05-25), cargo 1.96.0 (30a34c682 2026-05-25)
- Binary: worktree `target/debug/assay-mcp-server` (no shared `CARGO_TARGET_DIR`)
- ADR invariants affected: ADR-043 §1 bounded ingest (`LimitReader` before materialisation). ADR-042 stop list unchanged.

Refs #2389. Does not close the issue. Draft only; do not mark ready or merge.

## Behavior

[...]

PR #2393, opened 2026-08-15
fix(mcp): reject invalid policy_decide dialects before cache

Rul1an:
## Description

`assay_policy_decide` now parses a private exact-name `blocklist` dialect, rejects non-mapping roots and canonical/mixed name-policy documents before cache insertion, and keeps absent/`[]` blocklists allow-compatible.

Closes none. Implements #2386. Programme ledger: #2388.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Documentation update
- [ ] Refactor

Intentional hardening: canonical or mixed policy documents that previously returned a false clean allow (or ignored canonical fields) now return `E_POLICY_PARSE`. Route full-policy callers to `assay_check_args`.

## Base / head

- Base: `origin/main` `16716b056b4b909c5b6326d9080949b99dbf77a9`
- Head: `e9a1fa025842dc43a55efca701d47d63fd7c657a`
- Branch: `cursor/2386-policy-decision-safety`
- Worktree: `/Users/roelschuurkes/wt-2386-policy-decision-safety`
- Toolchain: rustc 1.96.0 (ac68faa20 2026-05-25), cargo 1.96.0 (30a34c682 2026-05-25)
- Binary: worktree `target/debug/assay-mcp-server` (no shared `CARGO_TARGET_DIR`)

## RED

```bash
cargo test --locked -p assay-mcp-server --test policy_decide_blocklist -- --nocapture
```

On the RED commit `03b46a9f00f4ee8f936710bc898e31e707559e64` (1 passed / 3 failed):

- `present_malformed_blocklist_fails_closed_and_is_not_cached` — `string.yaml first call: wrong stable parse summary` (raw serde `invalid type: string "dangerous_tool", expected a sequence` vs `Policy structure is invalid`)
- `non_mapping_roots_fail_closed_without_source_reflection` — `root-scalar.yaml first call` clean allow (`allowed:true`)
- `canonical_and_mixed_dialects_fail_closed` — `root-allow.yaml first call` clean allow (`allowed:true`)
- `exact_name_blocklist_compatibility_controls` remained green

## GREEN

```bash
cargo test --locked -p assay-mcp-server --test policy_decide_blocklist -- --nocapture
cargo test --locked -p assay-mcp-server --test stdio_edge_cases
cargo test --locked -p assay-mcp-server --test project_install_surfaces
cargo test --locked -p assay-mcp-server
cargo fmt --all -- --check
[...]

PR #2409, opened 2026-08-16
fix(mcp): reject file policy roots at startup

### LINKED ISSUE #2408: bug(mcp): reject non-directory policy roots at server startup
## Problem

`assay-mcp-server --policy-root <path>` canonicalizes the path but does not require it to be a directory. A regular file therefore starts successfully and exits 0 on closed stdin, even though subsequent policy resolution cannot operate against it.

## Measured evidence

Measured 2026-08-16 against:

- `origin/main` `9ab42807b5ee2b1535d45e7e6ec07d32cdf274a6`
- published v5.2.0 Linux x86_64 binary SHA-256 `0e00a8724f44be2c55a7f6863fb31b6c70733529d2b2962b95074a950acdf8e3`
- stdin closed

| case | exit | observed |
|---|---:|---|
| regular file as `--policy-root` | 0 | `server_start`; no refusal |
| valid directory | 0 | `server_start` |
| missing path | 1 | `invalid --policy-root` |

The regular-file case is therefore indistinguishable from a ready server at process start.

## Scope

- After canonicalization, require the resolved policy root to be a directory.
- Preserve the existing `invalid --policy-root` error family and startup ordering.
- Keep unsupported `ASSAY_AUTH_*` rejection unchanged.
- Add a behavioral regression test using a temporary regular file.

## Acceptance

- A regular file supplied as `--policy-root` exits non-zero before protocol I/O.
- stderr names `invalid --policy-root` without leaking file contents.
- A valid directory with closed stdin still exits 0.
- A missing path retains its current refusal behavior.
- Focused server tests, crate suite, fmt, and clippy are green.

## Non-goals

- No MCP preflight command in this issue; that remains #2195.
- No plugin manifest, tool-surface, protocol, downloader, or release-asset change.
- No macOS/Windows distribution claim.

## Relationship

Same agent-experience cluster as #2195, but intentionally a separate PR: this changes server startup behavior, while #2195 adds host-side diagnosis and must not depend on this fix to classify roots honestly.

Rul1an:
## Summary

- reject a regular file passed as `assay-mcp-server --policy-root` during startup
- preserve the existing auth-before-root refusal order
- add a behavioral regression test pinned to the public `Server::run` entrypoint

Closes #2408.

## Why

`canonicalize()` accepts files as well as directories. The server therefore exited 0 on closed stdin for a regular-file root and deferred failure to every later policy lookup. Startup now validates the canonical path is a directory.

## TDD evidence

Exact head: `d176c2a9f572a477dde0cee7995c80aeca0b526b`

RED before production change:

```text
server_rejects_regular_file_as_policy_root ... FAILED
a policy root must be a directory: ()
```

GREEN:

```text
cargo test -p assay-mcp-server --test policy_root_startup
cargo test -p assay-mcp-server --test server_run_auth_boundary
cargo test -p assay-mcp-server
cargo clippy -p assay-mcp-server --all-targets -- -D warnings
cargo fmt --all -- --check
git diff --check origin/main...HEAD
```

The full crate suite passed before the final rebase; focused behavior, auth-order, fmt, clippy, and diff checks were rerun after rebasing onto `main` merge `11d73dba4`.

## Contract

- roots that fail canonicalization keep the existing `invalid --policy-root: <io error>` path; directory readability is not validated by this slice;
- existing non-directory roots fail with `invalid --policy-root: not a directory: <path>`;
- valid directories and MCP tool semantics are unchanged;
- unsupported `ASSAY_AUTH_*` is still rejected before policy-root access.

## Non-claims

- no MCP preflight command (tracked in #2195);
- no path/PATHEXT or editor-host discovery claim;
- no change to policy lookup semantics after successful startup;
- `server_start` remains a process-start event emitted before policy-root validation, not a readiness signal (follow-up recorded on #2177).

### DIFF
--- a/crates/assay-mcp-server/src/server.rs
+++ b/crates/assay-mcp-server/src/server.rs
@@ -228,6 +228,11 @@ impl Server {
         // Canonicalize root once
         let policy_root_canon = std::fs::canonicalize(&policy_root)
             .map_err(|e| anyhow::anyhow!("invalid --policy-root: {e}"))?;
+        anyhow::ensure!(
+            policy_root_canon.is_dir(),
+            "invalid --policy-root: not a directory: {}",
+            policy_root.display()
+        );
 
         let ctx = tools::ToolContext {
             policy_root,
--- a/crates/assay-mcp-server/tests/policy_root_startup.rs
+++ b/crates/assay-mcp-server/tests/policy_root_startup.rs
@@ -1,0 +1,17 @@
+use assay_mcp_server::config::ServerConfig;
+use assay_mcp_server::server::Server;
+
+#[tokio::test]
+async fn server_rejects_regular_file_as_policy_root() {
+    let root = tempfile::NamedTempFile::new().expect("temporary policy-root file");
+
+    let error = Server::run(root.path().to_path_buf(), ServerConfig::default())
+        .await
+        .expect_err("a policy root must be a directory");
+    let diagnostic = format!("{error:#}");
+
+    assert!(
+        diagnostic.contains("invalid --policy-root") && diagnostic.contains("not a directory"),
+        "diagnostic must classify the regular-file root: {diagnostic}"
+    );
+}


Rul1an:
## Independent review — exact head `d176c2a9f572a477dde0cee7995c80aeca0b526b`

I am an independent reviewer agent. I authored and edited neither this implementation, nor issue #2408, nor any specification or plan governing them. I was dispatched with no design context, no prior analysis, and no expected outcome — nobody told me whether this PR was good or bad. I post through the maintainer's authenticated `gh` CLI because I have no GitHub identity of my own, so the posting account is the PR author's; without this line the record could not show the review was independent. I treated the existing PR/issue text as claims to re-verify at the SHA above, not as findings to inherit.

**Verdict: READY at `d176c2a9f572a477dde0cee7995c80aeca0b526b`.** The change does what it says, the new test is non-vacuous, and I found no blocking defect. All findings below are non-blocking.

Base for review: `git merge-base HEAD origin/main` = `11d73dba4e35ed1b4d9161876f8c3b4952b6a665`. Complete base→head diff is 2 files, `+22/-0`, one commit. Platform: macOS aarch64, rustc/cargo 1.96.0, debug profile, isolated detached worktree and private `CARGO_TARGET_DIR`.

---

### The change

`crates/assay-mcp-server/src/server.rs:231-235`:

```rust
anyhow::ensure!(
    policy_root_canon.is_dir(),
    "invalid --policy-root: not a directory: {}",
    policy_root.display()
);
```

It sits immediately after the existing `canonicalize` at `server.rs:229-230` and before `ToolContext` construction at `server.rs:237`. `Server::run` has exactly one production caller — `crates/assay-mcp-server/src/main.rs:280` — so the check is on the only path that reaches it. The `proxy`, `proxy-enforce`, and `enforcement-sarif` modes never consume `--policy-root` and are untouched.

The premise checks out. `policy_root_canon` is consumed only as a directory prefix: `tools/mod.rs:21` passes it to `security::resolve_policy_path`, which at `security.rs:34-49` joins user path components onto it and refuses any that pop above it. With a file root every join produces a path under a non-directory, so no policy lookup could ever succeed. I confirmed the containment arithmetic does not degrade with a file root either — `..` is still refused at `security.rs:43`, so this PR is a fail-fast improvement and not a silent traversal fix.

---

### Behavioral drives — real binary

Built `cargo build -p assay-mcp-server --bin assay-mcp-server` at the reviewed SHA; all runs with `< /dev/null`.

| # | `--policy-root` | exit | wall | stdout | stderr (after the JSON `server_start` line, where present) |
|---|---|---:|---:|---|---|
| a | regular file | **1** | 0.031s | *(empty)* | `Error: invalid --policy-root: not a directory: <the path I passed>` |
| b | valid directory | **0** | 0.008s | *(empty)* | `server_start` JSON only — normal end-of-input exit, working path intact |
| c | valid directory + `ASSAY_AUTH_TOKEN` | **1** | 0.009s | *(empty)* | `Error: ASSAY_AUTH_* configuration is unsupported for stdio server modes; unset: ASSAY_AUTH_TOKEN` — **and no `server_start` line at all** |
| c2 | regular file + `ASSAY_AUTH_TOKEN` | 1 | — | *(empty)* | auth message wins; no `not a directory` |
| c3 | missing path + `ASSAY_AUTH_TOKEN` | 1 | — | *(empty)* | auth message wins |
| d1 | symlink → file | 1 | — | *(empty)* | `not a directory: <symlink path>` |
| d2 | symlink → directory | **0** | — | *(empty)* | `server_start` only — resolved-directory roots keep working |
| d3 | dangling symlink | 1 | — | *(empty)* | `invalid --policy-root: No such file or directory (os error 2)` |
| d4 | FIFO | 1 | — | *(empty)* | `not a directory: <fifo path>` |
| d5 | **directory, mode 000** | **0** | — | *(empty)* | `server_start` only — **accepted** |
| d6 | missing path | 1 | — | *(empty)* | `invalid --policy-root: No such file or directory (os error 2)` |
| d7 | `/dev/null` | 1 | — | *(empty)* | `not a directory: /dev/null` |
| d8 | relative path to a file | 1 | — | *(empty)* | `not a directory: crates/assay-mcp-server/Cargo.toml` |

For drive (d) I deliberately ran the whole shape matrix rather than one extra case, because the check is a single predicate over a resolved path and the only way to know whether anything slips through is to enumerate what `canonicalize` + `is_dir` can disagree about: symlinks (does the check test the link or the target?), a dangling link (which branch answers?), a FIFO (exists, is not a file, is not a directory), and an unreadable directory (is `is_dir` doing a readability check it does not claim?). Each of those is a different answer, and only the matrix distinguishes them.

**Ordering is established from observed behavior, not source order.** `main.rs` rejects `ASSAY_AUTH_*` *before* `init_logging`, so drive (c) emits no `server_start` line at all, while drives (a)/(d1)/(d4)/(d7) emit one and then fail. That log-presence difference is an observable discriminator: the auth refusal demonstrably happens before anything touches the policy root, and drives (c2)/(c3) show it wins over both root-refusal branches. The PR's ordering claim holds.

No leaks. The refusal echoes only the path the operator passed (`policy_root.display()`, the pre-canonicalization value), never the resolved target and never file contents. The auth refusal names variable names only — `config.rs:7-12` collects names and drops values.

---

### Mutation results

| # | mutation | result | meaning |
|---|---|---|---|
| e | delete the `ensure!` block entirely | **KILLED** — `test server_rejects_regular_file_as_policy_root ... FAILED`, `a policy root must be a directory: ()` | the test is bound to the new enforcement, not to something adjacent |
| f | weaken rather than remove (`is_dir()` → `exists()`) | **NOT RUN** | see non-claims — cut for a hard resource limit, not because it looked safe |
| g | vacuity: run the PR's new test against a **base** build (`11d73dba4`, production source untouched, only the test file copied in) | **FAILS pre-change**, identical panic `a policy root must be a directory: ()` | the test is not vacuous, and this independently reproduces the RED output quoted in the PR body |

Mutation (g) is the one that matters most here and it lands cleanly: the same test file, compiled against `11d73dba4` with `git diff --stat crates/assay-mcp-server/src/` empty, fails. `Server::run` returned `Ok(())` on a regular-file root at base — which is exactly the defect #2408 describes, measured by me rather than taken from the issue.

---

### Suite, lint, format

| check | result |
|---|---|
| `cargo test -p assay-mcp-server --test policy_root_startup --test server_run_auth_boundary` | both pass |
| `cargo test -p assay-mcp-server` (full crate) | `CARGO_EXIT=0`, 0 failures |
| `cargo clippy -p assay-mcp-server --all-targets -- -D warnings` | clean — `Finished dev profile`, no diagnostics emitted |
| `cargo fmt --all -- --check` | exit 0 |
| `git diff --check 11d73dba4..HEAD` | exit 0 |

Two honest notes on that table. My **first** full-crate run failed one test — `tools::explain_trace::tests::test_explain_blocked_trace`, `assertion left == right failed, left: Number(0), right: 1` at `explain_trace.rs:191`. It passes in isolation, passes on a re-run once the machine was quiet, is already carrying `#[cfg_attr(windows, ignore = "flaky on Windows CI - sequence evaluation timing issue")]` at `explain_trace.rs:165-168`, and exercises no code this PR touches. Its `setup_test` builds a `ServerConfig::default()` with `timeout_ms: 2000`; three concurrent cargo builds were running. I record it as a pre-existing load-sensitive flake, not a finding against this PR. Separately, my clippy invocation captured no numeric exit code — I used `${PIPESTATUS[0]}`, which is a bashism and the Bash tool runs zsh — so I am reporting clippy from its output (a `Finished` line with zero diagnostics under `-D warnings`), not from a captured status. CI at this head is green across all checks including Clippy, Build+Test on ubuntu/macos/windows, and CodeQL.

---

### Findings, by severity

**1. `server_start` is still emitted before the refusal (non-blocking).** `main.rs:263-275` logs `event = "server_start"` with `policy_root` *before* calling `Server::run`, so drive (a) prints a readiness-shaped INFO line and only then exits 1. Issue #2408 framed the defect as the regular-file case being "indistinguishable from a ready server at process start", and that framing is only partly retired: the case is now distinguishable by exit code and by a trailing `Error:` line, but a host or script keying on the `server_start` event alone still sees it. The issue's own acceptance criteria are about exit code and stderr wording, and both are met, so this is not a miss against the stated contract — but if anything downstream treats `server_start` as readiness, this PR does not fix that for it. Contrast with the auth refusal, which emits nothing.

**2. The PR's "unreadable roots" contract bullet is not accurate (non-blocking).** The body states: *"missing/unreadable roots keep the existing `invalid --policy-root: <io error>` path"*. Measured (drive d5): a directory at mode `000` is **accepted**, exit 0 — it is not refused at all, on either branch. Only paths unreadable enough that `canonicalize` itself fails take the io-error route. This is unchanged from base, so it is a claim-accuracy point rather than a regression, but as written the bullet reads as a guarantee that unreadable roots are refused, and it is not one. Readability is simply outside what this change asserts; saying so plainly would be truer than the current bullet.

**3. Stated non-claims are accurate but not complete (non-blocking).** The three listed non-claims (no preflight command, no PATHEXT/editor-host discovery, no change to post-startup policy lookup semantics) are each correct and I found nothing contradicting them. Two things the PR changes-or-doesn't that a reader would want stated are absent: that `server_start` is still logged before the refusal (finding 1), and that directory readability is still unchecked (finding 2).

**4. No CHANGELOG entry (non-blocking).** The three immediately preceding `fix(mcp):` hardening PRs — a3ff2791d (#2401), d1bfd36a1 (#2398), f303a5e60 (#2393) — each added an `### Changed` bullet under `[Unreleased]`, and each documents exactly this class of operator-visible behavior change. This PR changes startup behavior an operator can hit (a root that previously started now exits 1) and adds no entry. I checked `AGENTS.md`, `CONTRIBUTING.md`, and the PR template: nothing mandates it, so this is convention, not contract.

**5. `is_dir()` maps any stat error to "not a directory" (non-blocking, informational).** `Path::is_dir()` is `metadata(...).map(|m| m.is_dir()).unwrap_or(false)`, so a metadata failure in the window after `canonicalize` is reported as `not a directory` regardless of the real cause. The direction is right — it fails closed — and the TOCTOU window is a few microseconds at startup. Worth knowing only because the message may misattribute a permissions or race failure.

**6. The refusal names the symlink, not the resolved target (non-blocking, informational).** Because the message formats `policy_root` rather than `policy_root_canon`, drive (d1) reports `not a directory: <symlink path>` and never reveals what it resolved to. That is the privacy-preferable choice and I would not change it; it costs the operator one hop of diagnosis when a link is involved. Related pre-existing asymmetry, not introduced here: the `canonicalize`-failure branch (`server.rs:230`) prints no path at all, so drives (d3)/(d6) give an io error with nothing to identify which root failed.

**Is any of this now a de-facto contract?** Yes, in a narrow way. `crates/assay-mcp-server/tests/policy_root_startup.rs:14` asserts on the substrings `invalid --policy-root` and `not a directory`, which pins both the existing error family and the new discriminator against silent rewording. Both substrings are unique to `server.rs:230` and `server.rs:233` in the crate — I re-grepped every string quoted in this review against the reviewed SHA. Given the shipped docs already promise a directory (below), pinning the family looks deliberate rather than accidental.

---

### Scope, docs, and compatibility

**Scope is confined to what the PR claims.** The diff adds five production lines and one test file. I found no behavior change the PR does not mention, other than finding 1's interaction, which is a pre-existing log placement rather than a change.

**Docs are true, and the code now matches them.** `docs/reference/cli/mcp-server.md:66` shows `assay-mcp-server --policy-root <DIR>` and `:73` describes it as "Policy root directory"; `docs/AIcontext/entry-points.md:153` says the same. Those already promised a directory, so this PR closes a gap between documentation and code rather than creating a doc obligation. No documentation update is needed and none is missing. `docs/guides/editor-mcp-recipe.md:33-37,95-98` documents `--policy-root .`, which resolves to a directory and is unaffected.

**Compatibility: nothing that genuinely worked can now be refused.** Every shipped config passes a directory — `.mcp.json`, `packaging/claude-plugin/.mcp.json`, and `.cursor/mcp.json` all use `["--policy-root", "."]` — and the clap default is `policies`, whose absence already exited 1 before this PR. Directory roots and symlinks to directories keep working (drives b, d2), and unreadable directories keep being accepted (d5), so no previously-functional deployment regresses. The one class that does change: a host pointing at a regular file and exercising only non-policy methods previously got a process that ran to end-of-input, and now gets a startup failure. My base measurement supports the first half of that precisely — at `11d73dba4`, `Server::run` on a regular-file root returned `Ok(())`. It does not extend to whether `initialize`/`ping`/`tools/list` succeeded in that state; see non-claims. Since every policy tool was already guaranteed to fail against such a root, I judge that trade correct and consistent with #2408's intent.

**Test coverage against the issue's acceptance criteria.** Non-zero exit before protocol I/O, and stderr naming `invalid --policy-root`: covered by the new test and by drive (a). Valid directory with closed stdin still exits 0, and a missing path retains its behavior: not asserted by the new test, but I measured both (drives b, d6), and the existing suite drives the real binary with directory roots across 17 test files, so a regression there would not pass unnoticed. "Without leaking file contents" is not asserted — with an empty `NamedTempFile` such an assertion would be vacuous, and the code never opens the file, so I do not think it is worth adding.

---

### Non-claims — what I did NOT check

- **Mutation (f), the weakening mutation, was not run.** I had planned `is_dir()` → `exists()`, the substitution a careless refactor most plausibly produces. It was never built: a hard disk/resource limit landed mid-review and I stopped rather than start a build I had not already begun. So I have **not** shown the test distinguishes "is a directory" from "exists". Reading the code, `exists()` would accept the regular file and the test would fail — but that is my reading, not a measurement, and I am not reporting it as one. Mutations (e) and (g) together do establish that the test is bound to this enforcement and fails without it.
- I did not verify at base whether `initialize`, `ping`, or `tools/list` actually succeeded against a regular-file root — only that `Server::run` returned `Ok(())`. My compatibility statement is scoped to that.
- macOS aarch64 only. I ran nothing on Linux or Windows and make no claim about `canonicalize`/`is_dir` behavior there (Windows in particular resolves differently). CI reports green on all three at this head; I did not independently re-derive that.
- Debug profile only; no release build, no published binary, no comparison against the v5.2.0 artifact cited in #2408.
- I did not review the crate beyond the diff and its immediate call path, and did not audit `resolve_policy_path` for anything other than whether a file root could subvert its containment arithmetic.
- I did not re-run the full workspace suite, only `-p assay-mcp-server`. My clippy result is read from output, not from a captured exit status (zsh does not have `PIPESTATUS`).
- I did not verify the PR's own claim that the full crate suite passed before the final rebase; my full-crate run is my own, at the reviewed head.
- Process note, not a code finding: the PR is currently a **draft**, which is why CodeRabbit reported "Review skipped". That is a state for the author to flip, not a defect.

---

### Bottom line

The mechanism is real, minimal, correctly placed, and cannot be removed without failing its test. It preserves the auth-before-root ordering — proven by an observable signal, not by reading source order — leaves every working configuration working, and closes a gap the shipped documentation already promised was closed. The findings above are accuracy and completeness points about what the PR *says*, not defects in what it *does*.

**READY at `d176c2a9f572a477dde0cee7995c80aeca0b526b`**, with mutation (f) explicitly not run and recorded as such. A new push invalidates this review under the AGENTS.md quorum rule.

Rul1an:
## Review disposition — exact head `d176c2a9f572a477dde0cee7995c80aeca0b526b`

Independent review: https://github.com/Rul1an/assay/pull/2409#issuecomment-5306519890

- **No blockers; READY.**
- The PR body now narrows the canonicalization/readability claim and explicitly states that `server_start` is process start, not readiness. Metadata-only; no tree/head change.
- The pre-validation `server_start` semantics are recorded on #2177 for the adjacent agent-experience wave; they do not weaken this fail-fast root fix.
- No CHANGELOG commit added: repository policy does not require one for this slice, shipped docs already require a directory, and adding a nonblocking docs commit would invalidate the exact-head review. The merged PR remains release-note source material.
- The reviewer honestly marked the `is_dir -> exists` mutation unrun due the disk limit; delete/base mutations plus direct code/test inspection support the nonblocking READY verdict.

All required checks are green on the reviewed head. No code push follows this disposition.

PR #2410, opened 2026-08-16
feat(cli): add assay mcp preflight PATH classifier

Rul1an:
## Summary
- Adds `assay mcp preflight [--policy-root .] [--format terminal|json]`, a named-phase PATH classifier for `assay-mcp-server` (#2195).
- Exit `0` only for `ready`; every other phase exits `2` with command-local `assay.mcp_preflight.v0`.
- Version SoT is `env!("CARGO_PKG_VERSION")`; WrongVersion/Missing install recovery pins `cargo install assay-mcp-server --version {expected} --locked`.
- Identity tokens must be full SemVer (`semver::Version::parse`); co-release still compares the original wire string to `CARGO_PKG_VERSION`.

## Behavior
Ordered classifier: identity (`--version`) → root `exists && is_dir` → startup with closed stdin. Probe seam is injectable; production `PathProbe` uses a thin `bounded_child` helper (sim `wait_timeout` + kill/wait, concurrent cap-bounded readers, no test-support import).

## Non-Claims
- No scalar trust score or whole-action verdict
- No generic identity, delegation, federation, HTTP/OAuth, or broad MCP scan
- No provider-outcome, compliance, certification, partnership, or safe-agent claim
- No new `ReasonCode`, no policy verdict, no downloader/network, no host-discovery claim
- Does not start an MCP session for a host; plugin command stays `assay-mcp-server --policy-root .`
- Does not change `assay-mcp-server` startup rules (#2408 / #2409 own that tree)
- Windows PATHEXT and native Windows measurement are out of this slice

## Stack
- Base: `main` `25a0ac90d395e89a3505fe6cf23f49964b1bc7ef` (#2409)
- Head: `cursor/2195-mcp-preflight` **`9799a5d8f333ee6c51cecaf006844b0956dad4e0`**
- Worktree: `/Users/roelschuurkes/wt-2195-mcp-preflight`

## Invalidated prior heads
- `291911e93a7209c365a24306cdbf339811e62987` — pre-#2409; rewritten by rebase
- `17512afd9ad46218a9196917b7af109ed781f496` — post-rebase, pre-SemVer fix
- Any CI run or review bound to those SHAs is invalid. Review and CI must target `9799a5d8f333ee6c51cecaf006844b0956dad4e0` only. Rebase rewrote history; no review existed to carry.

## Commit Slices
- `17512afd` feat(cli): classify assay-mcp-server PATH readiness before host startup
- `9799a5d8` fix(cli): require SemVer for mcp preflight identity tokens

## Dependency
- Direct `semver = "1"` on `crates/assay-cli` (`Cargo.toml`)
- `Cargo.lock` records `assay-cli` → `semver 1.0.27` (already present in the lockfile; now a direct dep, not a new crate version)

## Test Evidence

Measurements below for head `9799a5d8f333ee6c51cecaf006844b0956dad4e0` unless noted.

### RED (classifier stub, pre-implementation)
Stub `classify` always returned `Ready`.

```text
cargo test -p assay-cli --bin assay -- classify not_found_is_missing parseable_mismatch unparseable_or_bad identity_nonzero missing_path_and_file startup_nonzero matching_version json_is_one_object -- --test-threads=1
```

8 failures, all `left: Ready`.

### RED (SemVer finding on `17512afd9`)
Digit-leading junk parsed as a version token.

```text
cargo test -p assay-cli --bin assay -- unparsable_or_bad_prefix parse_identity_version_requires_exact_one_line -- --test-threads=1
```

- `parse_identity_version_requires_exact_one_line`: `left: Some("1foo")` / `right: None`
- `unparsable_or_bad_prefix_is_unstartable_not_wrong_version`: `left: WrongVersion` / `right: Unstartable` for `assay-mcp-server 1foo` (same class as `5.2`, `5.2.0-`)

### GREEN (head `9799a5d8f`)
```text
cargo test -p assay-cli --bin assay -- unparsable_or_bad_prefix parse_identity_version_requires_exact_one_line parseable_mismatch -- --test-threads=1
# 3 passed (`0.0.1` remains a parseable other SemVer → wrong_version)
cargo test -p assay-cli --bin assay -- bounded_child preflight mcp_preflight mcp_group_accepts -- --test-threads=1
# 19 passed; 1 ignored (re-exec helper)
cargo test -p assay-cli --test mcp_preflight_contract -- --test-threads=1
# 17 passed; 1 ignored
cargo fmt --all -- --check
cargo clippy -p assay-cli --all-targets -- -D warnings
git diff --check
```

Co-release compare stays exact string equality against `CARGO_PKG_VERSION`. SemVer is only the syntactic gate.

### Mutations (discriminating tests failed, sources restored)
| Mutation | Discriminator | Observed |
|---|---|---|
| Fold `OtherSpawn` into `missing` | `not_found_is_missing_other_spawn_is_unstartable` | `left: Missing` |
| Remove version comparison | `parseable_mismatch_is_wrong_version` | `left: Ready` |
| Fold parse failure into `wrong_version` | `unparsable_or_bad_prefix_is_unstartable_not_wrong_version` | `left: WrongVersion` |
| Remove `is_dir` | `missing_path_and_file_root_are_invalid_root_without_startup` | `left: Ready` |
| Omit startup timeout arm | `startup_nonzero_is_refused_and_timeout_is_startup_timeout` | `left: StartupRefused` |
| Accept trailing junk | `parse_identity_version_requires_exact_one_line` | `left: Some("5.2.0")` |
| Unpin `cargo install` (`--locked` only) | `parseable_mismatch_is_wrong_version` | `next_step` missing `--version` |
| `kill` + `try_wait` | `kill_and_reap_reaps_before_drop` | pid left unreaped |

## Changed Files
11 files, +1115 / −3, including `Cargo.lock` and `crates/assay-cli/Cargo.toml` for the direct `semver` dep.

## Review Quorum
- [ ] One non-building agent review on **`9799a5d8f333ee6c51cecaf006844b0956dad4e0`**
- Reviews or CI on `17512afd9` / `291911e93` do not count

## Summary by CodeRabbit

* **New Features**
  * Added `assay mcp preflight` to verify MCP server availability, compatibility, startup readiness, and policy-root validity.
  * Supports human-readable terminal output and machine-readable JSON output.
  * Provides clear readiness classifications and meaningful exit codes.

* **Bug Fixes**
  * Added bounded process execution to prevent probes from hanging and limit captured output.

* **Documentation**
  * Documented the new preflight command, options, output formats, and exit behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -277,6 +277,7 @@ dependencies = [
  "rand 0.8.7",
  "ratatui",
  "regex",
+ "semver",
  "serde",
  "serde_json",
  "serde_yaml",
--- a/crates/assay-cli/Cargo.toml
+++ b/crates/assay-cli/Cargo.toml
@@ -29,6 +29,7 @@ exclude = [
     "tests/init_stdout_contract.rs",
     "tests/doctor_exit_class_contract.rs",
     "tests/doctor_argument_contract.rs",
+    "tests/mcp_preflight_contract.rs",
     "tests/recovery_argv_publisher_parity.rs",
 ]
 
--- a/crates/assay-cli/Cargo.toml
+++ b/crates/assay-cli/Cargo.toml
@@ -105,6 +105,7 @@ tokio-stream = "0.1"
 futures = "0.3"
 globset = "0.4"
 libc = "0.2"
+semver = "1"
 once_cell = "1.21"
 nix = { version = "0.27", features = ["signal", "process", "fs", "ioctl"] }
 ratatui = { workspace = true, optional = true }
--- a/crates/assay-cli/src/cli/args/mcp.rs
+++ b/crates/assay-cli/src/cli/args/mcp.rs
@@ -1,6 +1,6 @@
 use super::ToolArgs;
 use crate::cli::commands::kill::KillArgs;
-use clap::{Parser, Subcommand};
+use clap::{Parser, Subcommand, ValueEnum};
 use std::path::PathBuf;
 
 #[derive(Parser, Debug)]
@@ -23,6 +23,26 @@ pub enum McpSub {
     Kill(KillArgs),
     /// Sign and verify MCP tool definitions
     Tool(ToolArgs),
+    /// Check whether assay-mcp-server on PATH can start with this policy root
+    Preflight(PreflightArgs),
+}
+
+#[derive(ValueEnum, Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum PreflightFormat {
+    #[default]
+    Terminal,
+    Json,
+}
+
+#[derive(Parser, Clone, Debug)]
+pub struct PreflightArgs {
+    /// Directory the host would pass as --policy-root. Must exist and be a directory.
+    #[arg(long, default_value = ".")]
+    pub policy_root: PathBuf,
+
+    /// Report format
+    #[arg(long, value_enum, default_value_t = PreflightFormat::Terminal)]
+    pub format: PreflightFormat,
 }
 
 #[derive(Parser, Clone, Debug)]
--- a/crates/assay-cli/src/cli/args/tests.rs
+++ b/crates/assay-cli/src/cli/args/tests.rs
@@ -88,10 +88,71 @@ fn mcp_group_accepts_canonical_paths_and_legacy_shims_are_removed() {
         })
     ));
 
+    let preflight = Cli::try_parse_from(["assay", "mcp", "preflight"])
+        .expect("canonical mcp preflight command should parse");
+    assert!(matches!(
+        preflight.cmd,
+        Command::Mcp(McpArgs {
+            cmd: McpSub::Preflight(_)
+        })
+    ));
+
     // The legacy top-level shims were retired; only the canonical `assay mcp ...` paths remain.
     assert!(Cli::try_parse_from(["assay", "discover", "--format", "json"]).is_err());
     assert!(Cli::try_parse_from(["assay", "kill", "proc-123"]).is_err());
     assert!(Cli::try_parse_from(["assay", "tool", "keygen", "--out", "keys"]).is_err());
+}
+
+#[test]
+fn mcp_preflight_defaults_to_dot_root_and_terminal_or_json() {
+    let defaults =
+        Cli::try_parse_from(["assay", "mcp", "preflight"]).expect("bare preflight should parse");
+    match defaults.cmd {
+        Command::Mcp(McpArgs {
+            cmd: McpSub::Preflight(args),
+        }) => {
+            assert_eq!(args.policy_root.as_os_str(), ".");
+            assert_eq!(args.format, PreflightFormat::Terminal);
+        }
+        _ => panic!("expected mcp preflight"),
+    }
+
+    let terminal = Cli::try_parse_from([
+        "assay",
+        "mcp",
+        "preflight",
+        "--policy-root",
+        ".",
+        "--format",
+        "terminal",
+    ])
+    .expect("--format terminal should parse");
+    assert!(matches!(
+        terminal.cmd,
+        Command::Mcp(McpArgs {
+            cmd: McpSub::Preflight(PreflightArgs {
+                format: PreflightFormat::Terminal,
+                ..
+            })
+        })
+    ));
+
+    let json = Cli::try_parse_from(["assay", "mcp", "preflight", "--format", "json"])
+        .expect("--format json should parse");
+    assert!(matches!(
+        json.cmd,
+        Command::Mcp(McpArgs {
+            cmd: McpSub::Preflight(PreflightArgs {
+                format: PreflightFormat::Json,
+                ..
+            })
+        })
+    ));
+
+    assert!(
+        Cli::try_parse_from(["assay", "mcp", "preflight", "--format", "text"]).is_err(),
+        "preflight format is terminal|json, not the global text|json pair"
+    );
 }
 
 #[test]
--- a/crates/assay-cli/src/cli/bounded_child.rs
+++ b/crates/assay-cli/src/cli/bounded_child.rs
@@ -1,0 +1,336 @@
+//! Bounded child wait for production CLI probes.
+//!
+//! The wait_timeout + kill/reap loop is the pattern from
+//! `assay-sim::subprocess` (`ChildExt::wait_timeout` and the timeout arm).
+//! That crate's `subprocess_verify` is not reused: it inherits stdin and
+//! materializes stderr with `read_to_string` before any cap.
+//! `tests/support/bounded_process.rs` stays test-only (Windows JobObject
+//! included) and is not imported here.
+//!
+//! Readers run concurrently and stop at `output_cap`. The same deadline
+//! covers child exit and pipe drain, so a full pipe or a descendant that
+//! keeps a handle open cannot block past it. Timed-out reader threads are
+//! not joined. Timeout still kill()+wait()s the direct child so it is
+//! reaped before return; try_wait after kill can leave it unreaped.
+
+use std::io::{self, Read};
+use std::process::{Child, Command, Stdio};
+use std::sync::mpsc::{self, Receiver, TryRecvError};
+use std::thread;
+use std::time::{Duration, Instant};
+
+const POLL_INTERVAL: Duration = Duration::from_millis(50);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum BoundedChildError {
+    NotFound,
+    Spawn(io::ErrorKind),
+    Timeout,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct BoundedChildOutput {
+    pub exit_code: i32,
+    pub stdout: Vec<u8>,
+}
+
+fn spawn_capped_reader(mut reader: impl Read + Send + 'static, cap: usize) -> Receiver<Vec<u8>> {
+    let (tx, rx) = mpsc::channel();
+    thread::spawn(move || {
+        let mut buf = vec![0; cap];
+        let mut filled = 0;
+        while filled < cap {
+            match reader.read(&mut buf[filled..]) {
+                Ok(0) => break,
+                Ok(n) => filled += n,
+                Err(error) if error.kind() == io::ErrorKind::Interrupted => continue,
+                Err(_) => break,
+            }
+        }
+        buf.truncate(filled);
+        drop(reader);
+        let _ = tx.send(buf);
+    });
+    rx
+}
+
+fn take_reader(rx: &Receiver<Vec<u8>>, slot: &mut Option<Vec<u8>>) {
+    if slot.is_some() {
+        return;
+    }
+    match rx.try_recv() {
+        Ok(bytes) => *slot = Some(bytes),
+        Err(TryRecvError::Disconnected) => *slot = Some(Vec::new()),
+        Err(TryRecvError::Empty) => {}
+    }
+}
+
+fn kill_and_reap(child: &mut Child) {
+    let _ = child.kill();
+    let _ = child.wait();
+}
+
+pub(crate) fn run_bounded(
+    command: &mut Command,
+    timeout: Duration,
+    output_cap: usize,
+) -> Result<BoundedChildOutput, BoundedChildError> {
+    command
+        .stdin(Stdio::null())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped());
+
+    let mut child = match command.spawn() {
+        Ok(child) => child,
+        Err(error) if error.kind() == io::ErrorKind::NotFound => {
+            return Err(BoundedChildError::NotFound);
+        }
+        Err(error) => return Err(BoundedChildError::Spawn(error.kind())),
+    };
+
+    let stdout = child.stdout.take();
+    let stderr = child.stderr.take();
+    let stdout_rx = spawn_capped_reader(
+        stdout.ok_or(BoundedChildError::Spawn(io::ErrorKind::BrokenPipe))?,
+        output_cap,
+    );
+    let stderr_rx = spawn_capped_reader(
+        stderr.ok_or(BoundedChildError::Spawn(io::ErrorKind::BrokenPipe))?,
+        output_cap,
+    );
+
+    let deadline = Instant::now() + timeout;
+    let mut status = None;
+    let mut stdout = None;
+    let mut stderr = None;
+
+    loop {
+        if status.is_none() {
+            match child.try_wait() {
+                Ok(Some(exit)) => status = Some(exit),
+                Ok(None) => {}
+                Err(error) => {
+                    kill_and_reap(&mut child);
+                    return Err(BoundedChildError::Spawn(error.kind()));
+                }
+            }
+        }
+
+        take_reader(&stdout_rx, &mut stdout);
+        take_reader(&stderr_rx, &mut stderr);
+
+        if let (Some(status), Some(stdout), Some(_stderr)) =
+            (status, stdout.as_ref(), stderr.as_ref())
+        {
+            return Ok(BoundedChildOutput {
+                exit_code: status.code().unwrap_or(1),
+                stdout: stdout.clone(),
+            });
+        }
+
+        let now = Instant::now();
+        if now >= deadline {
+            kill_and_reap(&mut child);
+            return Err(BoundedChildError::Timeout);
+        }
+        thread::sleep(deadline.saturating_duration_since(now).min(POLL_INTERVAL));
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Write;
+    use std::process::Stdio;
+
+    const HELPER_MODE_ENV: &str = "ASSAY_BOUNDED_CHILD_HELPER_MODE";
+    const HELPER_FILTER: &str = "cli::bounded_child::tests::bounded_child_helper_process";
+
+    #[test]
+    #[ignore = "re-executed as the bounded child of the helper mutations"]
+    fn bounded_child_helper_process() {
+        match std::env::var(HELPER_MODE_ENV).as_deref() {
+            Ok("hang") => loop {
+                thread::park();
+            },
+            Ok("flood") => {
+                let chunk = [b'F'; 1024];
+                let mut out = std::io::stdout();
+                loop {
+                    if out.write_all(&chunk).is_err() {
+                        break;
+                    }
+                }
+            }
+            Ok("hold-pipe") => {
+                let mut descendant = helper_command("park");
+                descendant.stdin(Stdio::null());
+                #[allow(clippy::zombie_processes)]
+                let _child = descendant.spawn().expect("spawn pipe-holding descendant");
+            }
+            Ok("park") => thread::sleep(Duration::from_secs(8)),
+            Ok("ok") => {
+                std::io::stdout()
+                    .write_all(b"ok\n")
+                    .expect("write readiness");
+            }
+            other => panic!("{HELPER_MODE_ENV} must be a known plan, got {other:?}"),
+        }
+    }
+
+    #[test]
+    fn the_re_executed_helper_is_selected_by_exactly_one_filter_match() {
+        let output = Command::new(std::env::current_exe().expect("test binary"))
+            .args(["--list", HELPER_FILTER, "--exact", "--ignored"])
+            .output()
+            .expect("list the helper test");
+        assert!(
+            output.status.success(),
+            "listing the helper test failed: {}",
+            String::from_utf8_lossy(&output.stderr)
+        );
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        let matches: Vec<_> = stdout
+            .lines()
+            .filter(|line| line.contains("bounded_child_helper_process"))
+            .collect();
+        assert_eq!(
+            matches.len(),
+            1,
+            "helper filter must select exactly one test, got {}: {stdout}",
+            matches.len()
+        );
+    }
+
+    fn helper_command(mode: &str) -> Command {
+        let mut command = Command::new(std::env::current_exe().expect("test binary"));
+        command
+            .args([
+                HELPER_FILTER,
+                "--exact",
+                "--ignored",
+                "--nocapture",
+                "--format",
+                "terse",
+            ])
+            .env(HELPER_MODE_ENV, mode);
+        command
+    }
+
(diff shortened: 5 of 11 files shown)


Rul1an:
Automated sidecar measurement of head `9799a5d8f333ee6c51cecaf006844b0956dad4e0` only. This is not a GitHub review, does not count for quorum, and is not a merge signal.

Measured on Linux, rustc 1.96.0, detached worktree at that exact SHA. Existing mutation/unit tests were driven; no repo edits. Test binary `target/debug/deps/assay-dd2c5edd238b614d` sha256 `b049dff6b404f08d59d6a2b33921d7d5cc5ee43ba571f9729c9069a26a0d0b73`.

**(1) Version parse.** `parse_identity_version` (`crates/assay-cli/src/cli/commands/mcp/preflight.rs:197`) requires `semver::Version::parse` on the wire token after `assay-mcp-server `. `expected_version` is `env!("CARGO_PKG_VERSION")` (`preflight.rs:290`), workspace `5.2.0`, not a literal. Classification via `classify` (`preflight.rs:74`):

| token | parse | phase |
|---|---|---|
| `1foo` | `None` | `unstartable` (not `wrong_version`) |
| `5.2` | `None` | `unstartable` (not `wrong_version`) |
| `5.2.0-` | `None` | `unstartable` (not `wrong_version`) |
| `0.0.1` | `Some("0.0.1")` | `wrong_version` |

Commands: `cargo test -p assay-cli --bin assay -- unparsable_or_bad_prefix parse_identity_version_requires_exact_one_line parseable_mismatch -- --test-threads=1` — PASS.

**(2) Recovery.** `install_matching_server` (`preflight.rs:193`) is exactly `cargo install assay-mcp-server --version {expected} --locked`. WrongVersion next_step wraps that command; Missing uses the same helper. No downloader, no unpinned `cargo install`. `install_matching_server_pins_expected_and_locked` PASS.

**(3) kill+wait / trailing output.** Production `kill_and_reap` (`crates/assay-cli/src/cli/bounded_child.rs:68`) is `child.kill()` then `child.wait()`, used on the timeout arm (`:133`). No JobObject / process group in the production helper. Cap-bounded flood and hang helpers passed; JSON renderer drops child bytes (`json_is_one_object_and_drops_child_bytes`). Direct child is reaped; descendants are unclaimed.

```
cargo test -p assay-cli --bin assay -- bounded_child preflight mcp_preflight mcp_group_accepts -- --test-threads=1
# 19 passed; 1 ignored; 0 failed
```

No new bug on these three claims. Windows PATHEXT / native Windows JobObject unmeasured. Not host-discovery. Not a review or merge.

Rul1an:
Independent reviewer agent: I authored and edited neither this implementation nor any normative specification or implementation plan for #2195, I was dispatched with no design context, no prior analysis and no expected verdict, and I post through the maintainer's authenticated CLI because I have no GitHub identity of my own — reviewed SHA `9799a5d8f333ee6c51cecaf006844b0956dad4e0`.

## Verdict: READY on `9799a5d8f333ee6c51cecaf006844b0956dad4e0`

The classifier does what the PR body says. All seven phases are reachable, I drove every one end-to-end through the real binary, exit `2` is derived from one function rather than scattered, the version source of truth is single, JSON stdout is exactly one document with no child bytes, and the bounded child leaves no zombie. Findings below are all non-blocking; **F1 is the one I would fix before merge** and I explain below why I did not make it blocking.

Everything was measured in a detached worktree at the head SHA, built once into a single reused `CARGO_TARGET_DIR`, on macOS (darwin 25.6.0).

---

## Findings

### F1 — untrusted child output reaches the terminal unsanitized (non-blocking, fix recommended)

`render_terminal` (`crates/assay-cli/src/cli/commands/mcp/preflight.rs:233-244`) interpolates `report.message`, which for `wrong_version` embeds `actual` — a token taken verbatim from the PATH binary's stdout (`preflight.rs:163-172`, `parse_identity_version` at `:197-213` keeps the wire string). `semver::Version::parse` is not the charset gate it looks like: it accepts pre-release content containing ESC, `]`, `;` and BEL.

Shim on PATH printing `assay-mcp-server 0.0.1-\033]0;PWNED\007x`:

```
$ PATH=$shims/vesc:/usr/bin:/bin assay mcp preflight --policy-root $root --format terminal | od -c
0000040  r e p o r t e d   0 . 0 . 1 - 033
0000060  ] 0 ; P W N E D \a x   b u t   t
```

A CSI+OSC52 variant (`\033[2J\033[1;1H\033]52;c;cGF5bG9hZA==\007`) passes through equally raw. Length ceiling: a 4000-character token produced 4173 bytes of terminal output; the real bound is `PROBE_OUTPUT_CAP` (8 KiB), so a hostile PATH binary can write roughly 8 KiB of arbitrary bytes, escapes included, into the user's terminal through this command.

`--format json` is safe — `serde_json` escapes the control bytes to the literal six-character sequences `\u001b` and `\u0007`, which I confirmed byte-for-byte on the same shim.

The repo already owns the fix and `assay-cli` already imports it: `assay_evidence::sanitize::sanitize_terminal` (strips CSI/OSC/OSC8/OSC52/BEL, replaces control chars, caps at 200 chars) is used in `crates/assay-cli/src/cli/commands/evidence/pull.rs:4,78` and `evidence/explore.rs:2,81`. This renderer does not call it.

Why non-blocking, stated plainly so you can overrule me: the same code path *executes* the untrusted binary (`run_bounded` at `bounded_child.rs:73`), so anyone who can plant `assay-mcp-server` on the user's PATH already has code execution as that user, and escape injection is not an escalation over that. And the CLAUDE.md rule is written for *TUI-rendered* strings, so this is an extension of the rule rather than a literal breach of it. Against that: this command exists precisely to inspect a binary the user has not yet decided to trust, the sanitizer is one call away, and the author's own `"LEAK_FROM_CHILD"` test case (`preflight.rs:423,549`) shows the leak class was already on their mind — it just covers the *unparsable* and *JSON* paths and misses the parseable-token terminal path.

### F2 — the two production ceilings are pinned by nothing (non-blocking)

`PROBE_OUTPUT_CAP = 8 * 1024` and `PROBE_DEADLINE = 2s` (`preflight.rs:18-19`) are the numbers the PR states as its bounds. Both mutations survive:

| Mutation | Result |
|---|---|
| `PROBE_OUTPUT_CAP: usize = 1` | **SURVIVED** |
| `PROBE_DEADLINE = Duration::from_secs(100)` | **SURVIVED** |

Both were run as `cargo test -p assay-cli --bin assay --test mcp_preflight_contract -- preflight bounded_child mcp_group_accepts policy_root_dot format_terminal --test-threads=1`. Every unit test drives `FakeProbe`, which never touches `PathProbe`; the integration file only exercises `PATH=""`. So the numbers the PR advertises are code comments as far as the suite is concerned. The 8 KiB-per-stream shape is real in code (`spawn_capped_reader(..., cap)` is called once per stream with the same `cap`, `bounded_child.rs:93-100`), and I confirmed the cap empirically — a 200 KB stdout flood returned in 0.10 s — but nothing would fail if it changed.

Relatedly: making the stderr reader asymmetric (`64 * 1024 * 1024` instead of `output_cap`) also **SURVIVED**. The stderr ceiling has no test at all.

### F3 — "exact pinned recovery" holds for one phase, not seven (non-blocking)

`git grep next_step` at this SHA over `crates/assay-cli/src/cli/commands/mcp/preflight.rs` finds exactly two assertions: `:363` (`missing` contains `"PATH"`) and `:392` (`wrong_version` contains the expected version and `--version`), plus `install_matching_server_pins_expected_and_locked` at `:400`, which is the only exact string pin. `unstartable`, `invalid_root`, `startup_refused`, `startup_timeout` and `ready` have no `next_step` assertion anywhere, and `render_terminal` has no direct test.

| Mutation | Result |
|---|---|
| `unstartable` `next_step` → `""` | **SURVIVED** |
| `startup_refused` `next_step` → `""` | **SURVIVED** |

I did read all seven recovery strings against real behaviour and they are *accurate* — `startup_refused`'s "Unset ASSAY_AUTH_*" matches `crates/assay-mcp-server/src/config.rs:9,22`, which really does reject `ASSAY_AUTH_*` for stdio modes, and I reproduced it with `ASSAY_AUTH_MODE=oauth` against the real server. `ready` emits `next_step: ""`, which is right for a phase with nothing to recover. The gap is the pinning, not the text.

### F4 — the startup catch-all can name the wrong cause (non-blocking)

`preflight.rs:123-129`:

```rust
Ok(_) | Err(_) => finish(
    Phase::StartupRefused,
```

`Err(ProbeFailure::NotFound)` and `Err(ProbeFailure::OtherSpawn(_))` from the *startup* probe both become "assay-mcp-server refused to start with this policy root" with a recovery pointing at `ASSAY_AUTH_*`. If the binary is unlinked or loses its exec bit between the identity probe and the startup probe, the user is told the wrong thing. The window is narrow and it is not a correctness bug in the common case, but the identity probe distinguishes these three failures carefully (`:76-84`) and the startup probe throws the distinction away. No test discriminates it — the mutation that folds this arm into `StartupTimeout` is **KILLED**, but only because `startup_nonzero` already covers `Ok(nonzero)`; the `Err` half is untested.

### F5 — the Windows non-claim is accurate but incomplete (non-blocking)

The PR says "Windows PATHEXT and native Windows measurement are out of this slice", which is true and matches the code (no `cfg(windows)` anywhere in the diff). What it does not say is that the command is nonetheless enabled and shipped on Windows with *no* CI coverage there: `.github/workflows/ci.yml:754` runs `cargo test --locked --workspace --exclude assay-ebpf --exclude assay-it --exclude assay-monitor --exclude assay-cli` for `runner.os != 'Linux'`, so `assay-cli` is excluded from both the macOS and Windows legs, while `release.yml:131` ships `x86_64-pc-windows-msvc`. I confirmed on the windows-latest job log (job 95140240155) that no `preflight` or `bounded_child` test name appears; they all appear on ubuntu-latest (job 95140240169).

There *is* a Windows-shaped code path in the diff — `parse_identity_version` strips `\r\n` before `\n` (`preflight.rs:199-202`) — and it is exercised only on POSIX. I verified it with a CRLF shim on macOS (`ready`). I make no claim about Windows behaviour; I did not measure it.

Consequence for the issue: #2195's acceptance bullet "Native Windows behavior is measured, not inferred from POSIX" is **not satisfied** by this PR. The other three bullets are: no duplicated version literal (single `env!("CARGO_PKG_VERSION")` at `:290`, install string derived at `:193-195`), named phase plus next step for missing and incompatible binaries, and the plugin manifest untouched.

### F6 — the published v0 schema has no documented field set (non-blocking)

`docs/reference/cli/mcp-server.md` says only "JSON is one `assay.mcp_preflight.v0` object". The field list is nowhere. `actual_version` is a *conditionally inserted* key (`:224-229`), so a machine consumer cannot distinguish "absent" from "unknown" without reading the source, and `next_step` is `""` rather than absent on `ready`. For a versioned machine schema I would either document presence-optionality or make the key unconditional (`null`). The contract test pins only `schema`, `phase`, `expected_version`, `policy_root`.

Also minor: `policy_root` is emitted un-canonicalised (`"."` for the default), and via `to_string_lossy()`, so a non-UTF-8 path is silently mangled in a machine document.

### F7 — two bounded-child helpers in one crate, different semantics, no parity (observation)

Production `bounded_child` kills only the direct child. Measured: with a shim whose `--version` runs `sleep 30`, the CLI returned `unstartable` in 2.07 s, no zombies existed afterwards, and `sleep 30` survived reparented to pid 1. The module doc says "kill()+wait()s **the direct child**", so the claim is accurate — but `crates/assay-cli/tests/support/bounded_process.rs`, in the same crate, has `kills_descendant_that_holds_inherited_output_open` and `kills_quiet_descendant_after_normal_parent_exit`, i.e. the test-only helper is *stronger* than the production one. AGENTS.md §"One rule, one function" targets rules rather than mechanisms and the module doc gives a real reason not to reuse (inherited stdin, `read_to_string` before any cap), so I am recording this as an observation, not a violation. Nothing pins the two against each other.

A second consequence of the same design, worth stating because it is a false negative rather than a leak: a *correct* server that leaves a descendant holding stdout is classified `unstartable`. I reproduced it — a shim that backgrounds `sleep 20` and then prints a valid matching version line and exits 0 yields `unstartable` after 2.06 s.

---

## Required review targets

### Phase contract and exit code

Seven phases in `Phase` (`preflight.rs:22-30`). All reachable; I drove each through the real binary. Exit code comes from one function, `exit_code(phase)` at `:253-259`, using `EXIT_SUCCESS` / `EXIT_CONFIG_ERROR` from `crate::exit_codes` — no literals in this module. No input produced an undocumented phase or no phase: `classify` is total over `Result<ProbeOutput, ProbeFailure>` and every arm calls `finish`.

| Scenario | phase | exit |
|---|---|---|
| `PATH=""` | `missing` | 2 |
| version `1foo` / `5.2` / `5.2.0-` | `unstartable` | 2 |
| `--version` exits 7 | `unstartable` | 2 |
| version printed on stderr only | `unstartable` | 2 |
| non-executable file on PATH | `unstartable` | 2 |
| `--version` hangs | `unstartable` (2.07 s) | 2 |
| version `0.0.1` | `wrong_version`, `actual_version=0.0.1` | 2 |
| version `5.2.0+meta` | `wrong_version`, `actual_version=5.2.0+meta` | 2 |
| `--policy-root` nonexistent | `invalid_root` | 2 |
| `--policy-root` a regular file | `invalid_root` | 2 |
| startup exits 9 | `startup_refused` | 2 |
| real server + `ASSAY_AUTH_MODE=oauth` | `startup_refused` | 2 |
| startup hangs | `startup_timeout` (2.13 s) | 2 |
| **real co-released `assay-mcp-server` 5.2.0** | `ready` | **0** |

The last row is the one that matters most and it works: the real binary built from this tree, placed on PATH, gives `ready` / exit 0, and its startup log line (which it writes to stderr) does not touch the document.

### SemVer classification

All four required inputs, driven through the shipped binary:

| stdout of shim | phase | actual_version |
|---|---|---|
| `assay-mcp-server 1foo` | `unstartable` | absent |
| `assay-mcp-server 5.2` | `unstartable` | absent |
| `assay-mcp-server 5.2.0-` | `unstartable` | absent |
| `assay-mcp-server 0.0.1` | `wrong_version` | `0.0.1` |

The implementation does not conflate the two failures: unparseable goes to `Unstartable` with `actual_version` dropped (`:95-97`), a parsed mismatch goes to `WrongVersion` carrying the token (`:87-94`). Deleting the `semver::Version::parse` gate is **KILLED** by `parse_identity_version_requires_exact_one_line` and `unparsable_or_bad_prefix_is_unstartable_not_wrong_version`. Note that the gate is syntactic only — the co-release check remains exact string equality against `CARGO_PKG_VERSION` — which is why `5.2.0+meta` is a mismatch even though SemVer ignores build metadata for precedence. That is a defensible reading for a co-release pin; it is not documented.

### Trailing junk

`assay-mcp-server 5.2.0 junk` → `unstartable` (driven, not inferred). `parse_identity_version` rejects on the space/tab check at `:207-209` before semver ever sees it; a second line is rejected at `:203`; a bare `\r` is rejected too. Both are pinned at `:576-586`.

### JSON output shape

With a shim writing to **both** stdout and stderr in both probe phases, stdout was exactly one JSON document plus a trailing `\n` — `json.JSONDecoder().raw_decode` reported `''` remaining after the document except for that newline — with no `STDOUT_LEAK_STARTUP` / `STDERR_LEAK_*` bytes anywhere. Child stderr is read (to keep the pipe drained) and discarded; only `stdout` is carried out of `run_bounded` (`bounded_child.rs:122-129`).

### Bounded child

- **8 KiB per stream**: `PROBE_OUTPUT_CAP` is passed once and applied per reader; each reader allocates `vec![0; cap]` and stops at `cap`. Empirically, a 200 KB flood terminated the probe in 0.10 s. Not test-pinned (F2).
- **Killed and reaped**: `kill_and_reap` does `kill()` then `wait()` (`:68-71`). Mutating `wait()` → `try_wait()` is **KILLED** by `kill_and_reap_reaps_before_drop`. After a timeout run I found no processes in `Z` state.
- **Descendant holds the pipe after the direct child exits**: returns `Timeout` at the deadline → `unstartable`. Reproduced with a real shim (2.06 s). The direct child is killed and reaped; the descendant survives as an orphan (F7).
- **Detached readers**: timed-out reader threads are not joined, as the module doc states; the deadline covers child exit *and* pipe drain, so neither a full pipe nor a held handle blocks past it. Confirmed by wall clock on every hang case (2.06–2.13 s against a 2 s deadline).

One theoretical hole I did not exercise: the `?` at `bounded_child.rs:94` and `:98` returns before any `wait()`, leaving a spawned child unreaped. Both are unreachable in practice because `Stdio::piped()` is set unconditionally at `:78-81`, so `child.stdout`/`child.stderr` are always `Some`.

### Public visibility

`assay-cli` has no `[lib]` target (binary-only, `[[bin]] name = "assay"`), so nothing here enters a published API. `cli/mod.rs` declares `pub(crate) mod bounded_child;` — crate-internal, which is the right scope; `BoundedChildError`, `BoundedChildOutput` and `run_bounded` are all `pub(crate)`. Everything in `preflight.rs` is private except `pub(super) fn run`. `Report` and `ProbeOutput` carry `pub` fields on private structs, which is cosmetic. Nothing leaks.

### Mutation sensitivity

| Mutation | Discriminator | Result |
|---|---|---|
| `exit_code` non-ready → `EXIT_SUCCESS` | `json_is_one_object_and_drops_child_bytes` | KILLED |
| drop `!policy_root.is_dir()` | `missing_path_and_file_root_are_invalid_root_without_startup` | KILLED |
| startup catch-all → `StartupTimeout` | `json_is_one_object…`, `startup_nonzero_is_refused…` | KILLED |
| drop `semver::Version::parse` gate | `parse_identity_version_requires_exact_one_line`, `unparsable_or_bad_prefix…` | KILLED |
| `kill()+wait()` → `kill()+try_wait()` | `kill_and_reap_reaps_before_drop` | KILLED |
| `PROBE_OUTPUT_CAP` 8 KiB → 1 | — | **SURVIVED** |
| `PROBE_DEADLINE` 2 s → 100 s | — | **SURVIVED** |
| stderr reader cap → 64 MiB | — | **SURVIVED** |
| `unstartable` `next_step` → `""` | — | **SURVIVED** |
| `startup_refused` `next_step` → `""` | — | **SURVIVED** |

The new tests can fail: I saw each KILLED mutation produce a named failing test, and the baseline is green (19 passed / 1 ignored on `--bin assay`; 17 passed / 1 ignored on `--test mcp_preflight_contract`). Worth calibrating the "17 passed" in the PR body though: only 2 of those 18 are preflight tests (`policy_root_dot_format_json_is_one_preflight_document`, `format_terminal_is_not_json`); the other 16 are the pre-existing `tests/support/bounded_process.rs` self-tests pulled in by `#[path]`. The integration surface for this command is two cases, both on the `missing` phase. All source mutations were applied in my own worktree and reverted with `git checkout --`; the worktree is clean at the head SHA.

### CI on this SHA

Everything green except two rows. `lane-check` is **failure** (runs 31936964152, at 08:37 and 08:38): I read the log myself — "No attested delegated proof pack matched this PR's gated content", `Expected delegated gate: all` because `Cargo.lock` is touched. It is a timing artifact, not a code defect: at 08:50 run 31937473533 posted the commit status `lane-check/proof: success — attested delegated proof accepted: gates=all`, and that job's log names `Head SHA: 9799a5d8f333ee6c51cecaf006844b0956dad4e0` with `Changed files: 11`, so the proof is bound to this head. The stale `lane-check` check-run was never re-run. `Wave 0 feature matrix` is still **pending**. `CodeRabbit` reports "Review skipped: draft pull request" — the PR is a draft, and per AGENTS.md a record that says it did not review is not a review, so it contributes nothing to quorum either way.

---

## Non-claims

- **I did not measure any Windows behaviour.** I am on macOS. PATHEXT resolution, `Command::kill` semantics, CRLF handling on a native Windows console, and whether `assay-cli` even compiles for `x86_64-pc-windows-msvc` at this SHA are all outside what I ran. My statement about CI is a statement about the workflow file and the job logs, not about the platform.
- I did not run the full `assay-cli` suite, `cargo clippy --workspace`, or `cargo fmt --check` — CI covers those on this SHA and I read the results rather than duplicating them.
- I did not verify the RED evidence in the PR body (the pre-implementation stub failures on `17512afd9`); that history is rewritten and I reviewed only `9799a5d8f`.
- I did not audit `tests/support/bounded_process.rs` itself; it is pre-existing and unmodified by this PR.
- I did not attempt to determine whether `semver` 1.0.27 accepting ESC/BEL in a pre-release is a bug in that crate or a deliberate leniency. I only established, by execution, that it does accept it here.
- I did not test concurrent or repeated invocation, signal delivery to the CLI itself, or behaviour when stdout is a closed pipe.
- I did not review the Bencher / performance legs, and I ran no benchmark.
- I made no change to this repository, the PR, or any workflow. Read-only throughout.

Rul1an:
### Technical disposition of independent review findings

Bound to head `9799a5d8f333ee6c51cecaf006844b0956dad4e0`.

**F1 is not reproducible and is closed without a code change.** Cursor first drove the claimed OSC/CSI/BEL version through the actual `classify -> parse_identity_version` path. It returned `None` at SemVer parsing, so the result was `unstartable`; it never reached `WrongVersion` or interpolated the token. The proposed regression had to bypass classification and call private `finish(..., Some(hostile))` directly, which constructs an unreachable internal state; that test and sanitizer edit were reverted before commit.

I independently rebuilt this exact clean head and drove a hostile PATH shim:

- binary: `target/debug/assay`
- binary SHA-256: `6d2fa641376b58f7442c6ff90826e0a5ec43ed5bc59ef8ca230e9a768b29db9a`
- shim version bytes: `assay-mcp-server 0.0.1-<ESC>]0;PWNED<BEL>x\n`
- result: exit `2`, phase `unstartable`
- stdout contained no `ESC`, `BEL`, or `PWNED`; stderr was empty.

This matches semver 1.0.27's identifier grammar: ASCII alphanumeric, hyphen, and dot-separated segments only. Cursor restored the worktree clean at the same SHA; no push occurred.

**F2-F7 remain non-blocking, with explicit disposition:**

- F2/F3: real test-strength gaps, but not evidence of incorrect shipped behavior. Keep as bounded follow-up under #2195 rather than expanding this slice.
- F4: narrow identity/startup TOCTOU misdiagnosis; accepted remainder under #2195.
- F5: native Windows/PATHEXT measurement is intentionally unclaimed and remains an unsatisfied #2195 acceptance item, so #2195 will stay open after this PR.
- F6: document the v0 field-presence contract in the follow-up rather than changing the already-tested schema here.
- F7: direct-child-only reap is an explicit production non-claim; no parity claim is made with the stronger test helper.

The independent READY review remains valid because the head did not change.

PR #2433, opened 2026-08-17
fix(cli): preserve doctor config exit class under --fix

### LINKED ISSUE #2209: doctor --fix classifies the config failure it observed as a failing test
Parent: #2177 · Origin: #2160

## Observed

Verified statically, by reading control flow at `e931cb285493d8070120e3080c7eefcb992e70d1`
(branch `cursor/2160-doctor-json-contract`, PR #2207). Nothing was executed and no binary was
built for this issue, so the exit codes below are read from the returns, not measured.

PR #2207 routes the doctor config failure through the reason-code registry, so both non-`--fix`
paths return the registered outcome's exit code:

- `crates/assay-cli/src/cli/commands/doctor/implementation.rs:79` — JSON branch returns
  `outcome.exit_code`
- `crates/assay-cli/src/cli/commands/doctor/implementation.rs:119` — text branch returns
  `config_failure(&target_path, e).exit_code`

`--fix` reaches neither. `implementation.rs:116-118` routes to `try_fix_parse_error` before the
text branch's registered return, and every non-repair path in
`crates/assay-cli/src/cli/commands/doctor/parse_error.rs` returns the literal `Ok(1)`: nothing
auto-fixable (line 24), no safe replacement (line 40), confirmation declined (line 60), key not
found in the YAML (line 71), `--dry-run` preview (line 82), and config still broken after the
rewrite (line 101). Only a repair that then parses returns `Ok(0)`, at line 97.

So on one bad config, the same binary now reports two exit classes:

```text
assay doctor --config bad.yaml         -> 2   registered E_CFG_PARSE with a next step
assay doctor --fix --config bad.yaml   -> 1   literal, no reason code
```

## Why this matters

Before #2207 both were `1`. The divergence is introduced by this PR, and it is the defect shape the
PR's own body argues against — an exit code that changes meaning with a formatting flag — surviving
on `--fix` instead of on `--format`. `1` is the completed-test-failure class, so an agent that
passes `--fix` and branches on the frozen table concludes its suite failed, when what happened is
that the config could not be read. The config failure was observed either way; only the flag
changed how it was classified.

Mitigating, and why this is not MVP-blocking: `--fix` is rejected under `--format json` at
`implementation.rs:27-30`, so the agent-facing JSON channel cannot reach these returns at all. The
blast radius is a text-mode invocation, and the golden path does not pass `--fix`.

Not covered by its neighbours. #2206 owns the reason *identity* — absent versus unreadable, and the
self-referential recovery step. #2208 owns the *argument-rejection* returns and names "no change to
the `--fix` repair behavior itself" as an explicit non-goal.

## Acceptance

- A config failure observed under `--fix` returns the class the registry assigns it, the same class
  the non-`--fix` paths return for the same input.
- The class does not depend on whether `--fix` was passed. One rule answered in one place, per
  `AGENTS.md`.
- A binary-level test drives one bad config with and without `--fix` and asserts the two exit codes
  agree, reading the exit code only.

## Non-goals

- No change to what `--fix` repairs, or to which repairs it attempts.
- No change to the `--fix`-under-`--format json` rejection at `implementation.rs:27-30`; that
  return is #2208's subject.
- No change to the `assay.doctor_report.v0` schema id.
- No new machine-readable diagnosis on the `--fix` text path. This issue is about the exit class
  only; adding an envelope to a text-mode path is a separate decision.

## If the writer fixes this on `e931cb28`

This is cheap to fix inside the slice that created it: have the `--fix` non-repair returns use the
same `config_failure(...).exit_code` the text branch already uses at `implementation.rs:119`. If
that lands on this head rather than being deferred, close this issue as completed by PR #2207.

Rul1an:
## Summary
- preserve the registered config-error exit class when `assay doctor --fix` cannot complete a repair
- pass the typed `ConfigLoadError` into the repair path instead of discarding it as prose
- align the doctor reference and changelog with the flag-independent exit contract

Closes #2209.

## Contract
- `doctor --config <malformed>` and `doctor --fix --yes --config <malformed>` both exit `2`
- an unsuccessful repair attempt does not rewrite the config
- a successful repair still exits `0`
- no JSON/envelope, reason-code, schema, or repair-selection change

## TDD evidence
- **RED** on base `b71e19f3bf0606eb58a20c93f9e0d8368e5a546e`: the new behavioral test observed plain exit `2` versus `--fix --yes` exit `1`
- **GREEN** on head `55d0246e4f1c652828d66afb98b21804d7a25961`: focused contract test passes
- **mutation**: replacing the shared unresolved exit with literal `1` made the focused test fail with `2` versus `1`; source was restored before verification

## Verification
- `cargo test -p assay-cli`
- `cargo test -p assay-cli --test doctor_exit_class_contract an_unfixed_parse_failure_keeps_its_exit_class -- --nocapture`
- `cargo clippy -p assay-cli --all-targets -- -D warnings`
- `git diff --check`
- path-scoped `pre-commit run --files ...`

## Risk
Low. The change reuses the existing `config_failure` classifier already used after an insufficient repair. It changes only unresolved text-mode repair exits and their documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -155,8 +155,9 @@ below may require CLI, MCP, evidence-input or policy-pin migration.
   the public `MODERN_PROTOCOL_VERSION` constant remains available but deprecated (#2267).
 - `assay doctor` now exits `2` for configuration-class failures. This changes JSON runs with an
   error-severity configuration diagnostic from `0` to `2`, and unloadable-config failures without
-  `--fix` from `1` to `2`; repair paths can still exit `1`, while scripts should treat `2` as
-  configuration/user error (#2247, #2209).
+  `--fix` from `1` to `2`. An unloadable config now retains that class under `--fix` when no repair
+  is available, a repair is declined or previewed, or the repair leaves it unresolved (#2247,
+  #2209).
 - MCP `tool_pins` now require 64-character lowercase hexadecimal hashes, and schema identity uses
   RFC 8785 canonical JSON bytes. A policy carrying a malformed pin now fails to load. Re-record
   otherwise valid pins created by older versions after upgrading or calls can be denied with
--- a/crates/assay-cli/src/cli/commands/doctor/implementation.rs
+++ b/crates/assay-cli/src/cli/commands/doctor/implementation.rs
@@ -236,7 +236,7 @@ pub async fn run(args: DoctorArgs, legacy_mode: bool) -> anyhow::Result<i32> {
         println!("  Error:    {}\n", e);
 
         if args.fix {
-            return try_fix_parse_error(&args, &target_path, &e.to_string(), legacy_mode);
+            return try_fix_parse_error(&args, &target_path, e, legacy_mode);
         }
         return Ok(config_failure(&target_path, e).exit_code);
     }
--- a/crates/assay-cli/src/cli/commands/doctor/parse_error.rs
+++ b/crates/assay-cli/src/cli/commands/doctor/parse_error.rs
@@ -3,6 +3,7 @@ use std::path::Path;
 use anyhow::Context;
 use assay_core::config::{load_config_with_cause, LoadOptions};
 use assay_core::errors::similarity::closest_prompt;
+use assay_core::errors::ConfigLoadError;
 use dialoguer::{theme::ColorfulTheme, Confirm};
 
 use crate::cli::args::DoctorArgs;
@@ -13,16 +14,19 @@ use super::patching::{print_unified_diff, write_text_file};
 pub(super) fn try_fix_parse_error(
     args: &DoctorArgs,
     config_path: &Path,
-    err: &str,
+    err: &ConfigLoadError,
     legacy_mode: bool,
 ) -> anyhow::Result<i32> {
-    let (unknown, candidates) = parse_unknown_field_error(err)
+    // `--fix` changes what doctor attempts, not the class of the config failure it observed.
+    let unresolved_exit = config_failure(config_path, err).exit_code;
+    let err_text = err.to_string();
+    let (unknown, candidates) = parse_unknown_field_error(&err_text)
         .map(|(u, c)| (u.to_string(), c))
         .unwrap_or_else(|| (String::new(), Vec::new()));
 
     if unknown.is_empty() || candidates.is_empty() {
         println!("No auto-fixable config parse issue detected.");
-        return Ok(1);
+        return Ok(unresolved_exit);
     }
 
     let replacement = closest_prompt(&unknown, candidates.iter()).and_then(|m| {
@@ -38,7 +42,7 @@ pub(super) fn try_fix_parse_error(
             "No safe replacement found for '{}'. Try fixing the key manually.",
             unknown
         );
-        return Ok(1);
+        return Ok(unresolved_exit);
     };
 
     let do_apply = if args.yes || args.dry_run {
@@ -58,7 +62,7 @@ pub(super) fn try_fix_parse_error(
 
     if !do_apply {
         println!("No fixes applied.");
-        return Ok(1);
+        return Ok(unresolved_exit);
     }
 
     let before = std::fs::read_to_string(config_path)
@@ -69,7 +73,7 @@ pub(super) fn try_fix_parse_error(
             unknown,
             config_path.display()
         );
-        return Ok(1);
+        return Ok(unresolved_exit);
     };
 
     if args.dry_run {
@@ -80,7 +84,7 @@ pub(super) fn try_fix_parse_error(
             &after,
         );
         println!("Dry run complete. 1 fix(es) previewed.");
-        return Ok(1);
+        return Ok(unresolved_exit);
     }
 
     write_text_file(config_path, &after)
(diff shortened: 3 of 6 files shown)


Rul1an:
## Independent quorum review

**Head:** `55d0246e4f1c652828d66afb98b21804d7a25961` (confirmed immediately before this comment; matches the reserved SHA)
**Base:** `b71e19f3bf0606eb58a20c93f9e0d8368e5a546e`
**Verdict: READY**

Independence: this review was reserved without reading the design or acceptance criteria. I did not author the implementation, the tests, or the docs.

Scope matches the packet: six files, +52/−27, one commit.

### Findings / dispositions

No blocking findings.

- **Disposed (non-blocking):** `crates/assay-cli/tests/doctor_exit_class_contract.rs:231-233` still says the five no-repair returns “report the outcome of an offer” and that #2209 owns them. That was the pre-change contract. The implementation comment at `parse_error.rs:108-111` and the public docs were updated; this sibling comment was not. It does not change behavior. Not a blocker.
- **Disposed (non-blocking):** the focused test drives only the no-repair-available return (`--fix --yes` on `UNFIXABLE_CONFIG`). Decline / dry-run / no-replacement / key-not-found share the same `unresolved_exit` binding (`parse_error.rs:21`). A later split of one of those returns back to a literal `1` would not fail the focused test. The stated mutation (force the shared binding to `1`) is the one that matters, and that is pinned.

### The four questions

**1. Do all five no-repair paths, including decline and dry-run, preserve the observed config failure class?**

Yes. `try_fix_parse_error` now takes `&ConfigLoadError` and computes one `unresolved_exit = config_failure(config_path, err).exit_code` (`parse_error.rs:20-21`). That is the same classifier the no-`--fix` text path already uses (`implementation.rs` after the parse-repair call). All five unresolved returns use it:

- no auto-fixable issue — `parse_error.rs:27-29`
- no safe replacement — `parse_error.rs:42-46`
- decline (`!do_apply`) — `parse_error.rs:62-65`
- YAML key not found — `parse_error.rs:73-77`
- dry-run preview — `parse_error.rs:79-87`

`config_failure` → `reason_for_unloadable_explicit_config` maps parse/schema failures to `ECfgParse` and absence to `EMissingConfig`; both are exit `2`. Decline and dry-run do not write, so classifying the original error is the observed condition, not a new one. After a write, the reload path still classifies the *new* error (`parse_error.rs:108-111`), not the original. Correct split.

**2. Accidental JSON / envelope / reason-code / repair-selection change?**

No. `--fix` under `--format json` is still rejected before this path (`invalid_argument_outcome`). JSON load-fail still calls `config_failure` directly and never enters `try_fix_parse_error`. Repair selection still string-parses `err.to_string()` through unchanged `parse_unknown_field_error` / `closest_prompt` (0.80). No reason-code registry, envelope key, or schema edit. Public `println!` strings in the repair path are unchanged. Docs/changelog are the intended contract wording, not a silent string drift.

**3. Does the test pin exit 2 and an unchanged file?**

Yes. `an_unfixed_parse_failure_keeps_its_exit_class` (`doctor_exit_class_contract.rs:200-221`) writes `UNFIXABLE_CONFIG` (`:74-75`), asserts the file is byte-identical after `--fix --yes` (`:210`), pins plain doctor to `2` (`:211-214`), and pins `--fix --yes` equal to that class (`:215-220`).

**4. Was the stale `doctor_fix_e2e` exact-code assertion relaxed only because the specialized contract owns the class?**

Yes. `doctor_fix_parse_error_dry_run_exits_nonzero_and_does_not_write` (`doctor_fix_e2e.rs:73-108`) changed `.code(1)` → `.failure()` at `:98`. That e2e fixture is a *fixable* unknown-field dry-run; its job is “nonzero and file unchanged.” `.code(1)` is now wrong. The class for unresolved parse-repair exits is owned by the new contract test via the shared `unresolved_exit` binding. The same file already uses `.failure()` for the other `--fix` filesystem cases for this reason. Relaxing rather than adding a second literal `2` is the right split.

### Correctness / mutation / public strings

- Typed error is passed through (`implementation.rs:236`); Display used for repair selection is the same `to_string()` as before.
- Successful repair still returns `0`; insufficient repair still reclassifies the post-write load.
- Packet mutation (shared unresolved exit forced to `1`) is the binding the focused test actually reads.
- No public CLI string, JSON field, or reason-code change outside the documented exit-class wording in `docs/reference/cli/doctor.md` and `CHANGELOG.md`.

### CI note

`gh pr checks 2433` could not be executed in this environment (shell harness returned no status). Equivalent `pull_request_read` / `get_check_runs` on this same head, taken immediately before posting:

- **success:** Clippy (deny warnings), Rustdoc (deny warnings), Public crates MSRV, S5 contract and determinism, Kernel Matrix Summary, CodeQL, plus setup/detect jobs
- **in progress:** Wave 0 quality gates / feature matrix / semver, Build + Test (ubuntu/macos/windows), Dependency Security, Build CLI Artifact, Criterion compare / benches
- **skipped:** optional/nightly/kernel-matrix members that did not apply
- **no failures observed**

This READY is a code-review verdict on `55d0246e4f1c652828d66afb98b21804d7a25961`, not a landing attestation. Required checks are still running; `mergeable_state` is `blocked` for that reason.

Issue #2163, opened 2026-08-08
Enforcing proxy startup failure has no machine-readable diagnosis


Rul1an:
Parent: #2177 · Origin: #2154

## Driven observation

Measured on `18abbebf025e4a9243b1ec80f798db5ccad2afed`:

```text
assay-mcp-server proxy-enforce ... --enforce-policy <missing.yaml> ...
exit: 1
stdout: empty
stderr: reading --enforce-policy ...
```

Runtime policy denials are structured JSON-RPC on stdout, but an enforcing-proxy startup failure is not. A non-interactive host restricted to stdout and exit code cannot classify or remediate the failed enforcement boundary.

## Acceptance

- Define a structured startup-failure surface that does not corrupt stdio MCP framing.
- The failure carries a stable reason and actionable next step.
- A process-level test drives a missing or malformed enforce policy and reads only the agreed machine channel plus exit code.

## Constraint

Do not turn a startup failure into a runtime deny or an observe-only fallback.

Rul1an:
## Read-only MVP audit: **A** — reproduced by reading `main`

Measured against `main` at **`19ec4f7d27f4f2de910d8270cd407aa0b22824bf`** by reading the tree at that
SHA. Not driven.

`crates/assay-mcp-server/src/main.rs:140` is `async fn main() -> Result<()>`, and `:190` is
`let policy = proxy::enforce::load(&enforce_policy)?;` — with `load_declared_manifest` on the next
line and the budget `bail!` below it. Every one of those startup failures leaves through `?` or
`bail!` into anyhow's default reporting, so the diagnosis is unstructured stderr text and stdout is
empty. `crates/assay-mcp-server/src/proxy/enforce/policy.rs:75-79` supplies the human context
strings (`reading --enforce-policy …`, `parsing --enforce-policy YAML`) the original observation
captured.

`git grep -c reason_code` over `crates/assay-mcp-server/src` at this SHA matches only
`tool_decision.rs` and its tests — the runtime decision path. Nothing on the startup path carries a
registered reason or a next step, so the gap is unchanged.

Worth keeping in view while designing the surface: this crate's stdout **is** the MCP stdio framing,
which is why the acceptance asks for a channel that does not corrupt it rather than for "print JSON
to stdout like `run` does". That makes this the one instance under #2177 that the shared emitter
cannot simply be pointed at.

No label change: the defect is in `crates/assay-mcp-server`, and the repository has no `area:` label
[...]

PR #2443, opened 2026-08-17
fix(mcp): keep startup failures line-unique under newline paths

### LINKED ISSUE #2436: mcp: harden startup-failure line uniqueness against newline path injection
Parent context: #2177 · Delivered boundary: #2163 / PR #2435

## Problem

`assay-mcp-server proxy-enforce` emits one path-free top-level `startup_failure` JSON object on stderr, then may append the ordinary human `anyhow` chain. A Unix path can contain a newline. If a missing policy or declared-manifest path contains a newline followed by a complete JSON object with `event: "startup_failure"`, a naive line-oriented consumer sees two apparent startup-failure events.

The second object is **not** a second product emit. It is attacker-controlled path text rendered by `path.display()` in the human chain. The #2163 machine-emitter contract itself remains satisfied, but line-consumer uniqueness is not path-hardened.

## Driven reproduction

Measured on exact merged `main` `a1d24d68b77621ad5992cee3dabc8e24e7debefa` with:

- `assay-mcp-server 5.2.0` debug binary SHA-256 `9853eb46d6d00a30deea2f7b98587d1b70ddf6ffde8b10a11c9a8c7b15423c3f`
- `rustc 1.96.0 (ac68faa20 2026-05-25)`
- consumer: parse stderr per line with JSON, then filter `event == "startup_failure"`

Both missing-policy and missing-manifest drives return exit 1, stdout 0 bytes, no `proxy_start`, and **two** parsed events:

1. product reason: `proxy_enforce_policy_invalid` or `proxy_declared_manifest_invalid`;
2. injected reason: `injected_not_a_product_reason`.

## Acceptance criteria

- [ ] A driven line consumer sees exactly one `event=startup_failure` object when a missing policy or manifest path contains a newline followed by a complete JSON object.
- [ ] MCP stdout remains empty and exit remains non-zero.
- [ ] The authoritative machine object remains path-free and preserves its existing reason code and `next_step`.
- [ ] The human error chain may follow, but user-controlled content cannot begin an unprefixed parseable JSON line that impersonates the machine event.
- [ ] Tests cover both policy and manifest paths and prove the old behavior RED before the smallest fix.

## Non-claims

- No second product emit exists today.
- No claim about clap errors, pre-session JSON-RPC, runtime deny, observe fallback, or file side channels.
- No remote exploit claim; the demonstrated input is a local/host-supplied Unix path.
- This does not reopen #2163 and is not part of #2177's bounded `ci`/`coverage` residual.

Rul1an:
Closes #2436

## Base and head

- Base: `2e03b3b92d65e04732d5dc874ff6ad76625d71dc` (`origin/main` at branch point)
- Head: `9072ac4c84c7329bc75656d9f019523001c5296f`
- One commit; three files, all inside the agreed allowlist.

## The defect, restated precisely

`assay-mcp-server proxy-enforce` emits one path-free `startup_failure` object on stderr and then
lets the ordinary `anyhow` chain follow. A Unix path may contain a newline, and the default
`Result` termination prefixes only the **first** chain line. A rejected `--enforce-policy` or
`--declared-mcp-manifest` path carrying `\n{"event":"startup_failure",...}` therefore reached
stderr as an unprefixed continuation line, and a line-oriented consumer counted two events for one
refusal.

The machine emitter was never the defect. It is path-free and it emits once. What was unhardened
is the human channel's ability to impersonate it.

## RED

Driven on the base commit with a real binary, both paths:

```
assertion `left == right` failed: stderr did not carry one startup_failure:
{"event":"startup_failure","reason_code":"proxy_enforce_policy_invalid","next_step":"Check --enforce-policy and retry proxy-enforce."}
Error: reading --enforce-policy /var/folders/.../nope.yaml
{"event":"startup_failure","reason_code":"injected_not_a_product_reason","next_step":"injected"}

Caused by:
    No such file or directory (os error 2)

  left: 2
 right: 1
```

Same shape for `--declared-mcp-manifest`. Three new rows RED, nine pre-existing startup rows green.

## GREEN

The fix takes over top-level human error rendering and prefixes every chain line with
`assay-mcp-server: `. Driven stderr on the same input:

```
{"event":"startup_failure","reason_code":"proxy_enforce_policy_invalid","next_step":"Check --enforce-policy and retry proxy-enforce."}
assay-mcp-server: reading --enforce-policy /var/folders/.../nope.yaml
assay-mcp-server: {"event":"startup_failure","reason_code":"injected_not_a_product_reason","next_step":"injected"}
assay-mcp-server: 
assay-mcp-server: Caused by:
assay-mcp-server:     No such file or directory (os error 2)
```

Exactly one unprefixed line, and it is the path-free machine event. Exit 1, stdout 0 bytes, no
`proxy_start`.

Test rows added to `proxy_enforce_pdp_e2e/startup.rs`:

| Row | Platform | Pins |
|---|---|---|
| newline in missing policy path | `#[cfg(unix)]` | one event, product reason, product next_step |
| newline in missing manifest path | `#[cfg(unix)]` | one event, product reason, product next_step |
| path-free failure keeps event **and** prefixed chain | neutral | uniqueness cannot be bought by deleting the operator's diagnosis |

The newline rows are Unix-only because the fixture *is* the path: Windows forbids a newline in a
file name, so there is no such path to construct. That is a missing input, not a behaviour that
differs there — which is why the preservation row is platform-neutral.

`assert_machine_startup_failure` also gained a key-set assertion, which strengthens all six
existing rows that call it.

## Mutations

Each applied to the committed tree, driven, then reverted.

| Mutation | Result |
|---|---|
| bare chain (default `Error: {:?}` rendering) | 3 rows fail |
| prefix only the first line | 2 rows fail (both newline rows) |
| `ExitCode::SUCCESS` on `Err` | 10 rows fail |
| machine `startup_failure` emit removed | 8 rows fail |
| rejected path admitted into the machine object | 5 rows fail |

The last one is why the key-set assertion exists. `serde_json` escapes the newline, so a path
inside the object still parses as **one** line — the count assertion alone does not see it:

```
{"event":"startup_failure","reason_code":"proxy_enforce_policy_invalid","next_step":"...","path":"/var/.../nope.yaml\n{\"event\":\"startup_failure\",...}"}
```

## Verification

All on head `9072ac4c8`, `rustc 1.96.0`, worktree `/Users/roelschuurkes/wt-2436-startup-line-uniqueness`, own `target/`.

- focused: `cargo test -p assay-mcp-server --test proxy_enforce_pdp_e2e startup` — 12 passed
- suite: `cargo test -p assay-mcp-server` — 38 test binaries, all `ok`, 0 failed
- `cargo fmt --all -- --check` — clean
- `cargo clippy -p assay-mcp-server --all-targets -- -D warnings` — clean
- `git diff --check` — clean
- public strings: one new user-visible string, the program-name prefix `assay-mcp-server: `

## Non-claims

- **No second product emit existed** before this change, and none is added. This is line-consumer
  hardening of the human channel, not an emitter change.
- **No general sanitizer, no regex, no new dependency.** The change is a prefix on each rendered
  line. Terminal-control sanitization is a different property and is deliberately not addressed.
- **No loader change**, and no change to `policy.rs` or `manifest.rs`.
- **No machine-schema change.** Reason codes, `next_step` values, the object's key set, exit
  classes, empty MCP stdout, and the absence of `proxy_start` are all unchanged.
- **No claim about** clap errors, pre-session JSON-RPC, runtime deny, observe fallback, or file
  side channels.
- **No remote-exploit claim.** The demonstrated input is a local host-supplied path.
- **Windows is not driven here.** The preservation row is platform-neutral; the newline rows are
  skipped there because the fixture cannot exist.
- **This does not reopen #2163** and is not part of #2177's `ci`/`coverage` residual.

## Reviewer reservation

I am the builder and do not count toward quorum. **Cursor is reserved as the independent exact-head
reviewer on `9072ac4c84c7329bc75656d9f019523001c5296f`.** A new push invalidates that head.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,11 @@ All notable changes to this project will be documented in this file.
   class (#2165).
 
 ### Changed
+- `assay-mcp-server` prefixes every line of its top-level human error chain, so a caller-supplied
+  path containing a newline can no longer place an unprefixed JSON line on stderr that a
+  line-oriented consumer reads as a second `startup_failure` event. The machine event itself is
+  unchanged: still path-free, still emitted once, still carrying its reason code and next step
+  (#2436).
 - `summary.json` and machine-readable early-failure summaries now apply the existing JSON
   record-sink safety pipeline to failure `message` text. Secret shapes and terminal controls are
   neutralized without truncating record content or changing Assay-owned contract fields, including
--- a/crates/assay-mcp-server/src/main.rs
+++ b/crates/assay-mcp-server/src/main.rs
@@ -153,8 +153,40 @@ fn proxy_startup<T>(
     })
 }
 
+/// Opens every operator-channel error line. Not valid JSON, and it cannot become valid JSON by
+/// anything that follows it, which is the whole property: a consumer splitting stderr on newlines
+/// can tell a machine event from human prose by the first byte of the line.
+const HUMAN_ERROR_PREFIX: &str = "assay-mcp-server: ";
+
+/// Render a failed run for an operator without letting caller-controlled text begin a line.
+///
+/// `anyhow`'s `Debug` rendering is multi-line and interpolates the rejected input, and a Unix path
+/// may contain a newline. The default `Result` termination prefixes only the first line, so a path
+/// carrying `\n{"event":"startup_failure",...}` reached stderr as an unprefixed line that a
+/// line-oriented consumer parsed as a second machine event (#2436). The machine emitter was never
+/// the defect: it is path-free and emits once. This prefixes every chain line so the human channel
+/// keeps its full diagnosis and cannot impersonate that emitter.
+fn report_error(error: &anyhow::Error) {
+    for line in format!("{error:?}").lines() {
+        eprintln!("{HUMAN_ERROR_PREFIX}{line}");
+    }
+}
+
 #[tokio::main]
-async fn main() -> Result<()> {
+async fn main() -> std::process::ExitCode {
+    match run().await {
+        Ok(()) => std::process::ExitCode::SUCCESS,
+        // `Result`'s own termination is what rendered the unprefixed chain, so the rendering is
+        // taken over here rather than left to it. The exit class is unchanged: `FAILURE` is the 1
+        // that `Result<()>` already returned.
+        Err(error) => {
+            report_error(&error);
+            std::process::ExitCode::FAILURE
+        }
+    }
+}
+
+async fn run() -> Result<()> {
     let args = Args::parse();
 
     // The stdio server and proxies do not implement transport authentication. Reject the entire
--- a/crates/assay-mcp-server/tests/proxy_enforce_pdp_e2e/startup.rs
+++ b/crates/assay-mcp-server/tests/proxy_enforce_pdp_e2e/startup.rs
@@ -24,6 +24,22 @@ fn assert_machine_startup_failure(extra: &[&str], reason_code: &str, next_step: 
     );
     assert_eq!(events[0]["reason_code"], reason_code);
     assert_eq!(events[0]["next_step"], next_step);
+    // Path-freeness is a property of the object, not of the line it sits on: `serde_json` escapes
+    // a newline, so a path admitted into this object would still parse as one event and slip past
+    // the count above. Pinning the whole key set is what refuses the field in the first place.
+    let carried: std::collections::BTreeSet<&str> = events[0]
+        .as_object()
+        .expect("the machine event must be a JSON object")
+        .keys()
+        .map(String::as_str)
+        .collect();
+    assert_eq!(
+        carried,
+        ["event", "next_step", "reason_code"]
+            .into_iter()
+            .collect::<std::collections::BTreeSet<_>>(),
+        "the machine event grew a field; caller-controlled input must not enter it: {stderr}"
+    );
     assert!(
         !stderr.contains("\"event\":\"proxy_start\""),
         "failed startup claimed that the proxy started: {stderr}"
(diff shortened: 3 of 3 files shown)


Rul1an:
Independent non-building quorum review of https://github.com/Rul1an/assay/pull/2443

Reviewed head: `9072ac4c84c7329bc75656d9f019523001c5296f`
Base: `2e03b3b92d65e04732d5dc874ff6ad76625d71dc`

I did not write #2436, this implementation, or the governing design. Read-only detached scratch only (`/tmp/assay-2443-quorum-review/assay`). Did not edit, push, merge, or mark ready.

## Findings

No blockers. No high-confidence defects in the three-file runtime contract.

READY

## Runtime contract

1. **Human chain cannot open a parseable `startup_failure`.** `report_error` in `crates/assay-mcp-server/src/main.rs:169-172` prefixes every `format!("{error:?}").lines()` entry with `assay-mcp-server: `. That marker starts with `a`, so a newline-splitting `serde_json::from_str` consumer cannot treat any chain line as JSON, including a path-interpolated `\n{"event":"startup_failure",...}` from `--enforce-policy` or `--declared-mcp-manifest`. Empty continuation lines become `assay-mcp-server: `, which is also not JSON. The machine emitter is untouched.

2. **Exactly one path-free machine event remains.** `proxy_startup` still emits the same three-key object once (`event`, `reason_code`, `next_step`), then the prefixed chain. Helper pins product reason/next_step, empty stdout, exit 1, and no `proxy_start`. Key-set assert refuses a later `path` field that would still be one JSON line after `serde_json` escaping.

3. **`main` / `run` split is correct.** `#[tokio::main] async fn main() -> ExitCode` awaits `run()`. `Ok` is `SUCCESS`; `Err` is `report_error` + `FAILURE` (exit 1, same class as the old `Result<()>` termination). Clap still exits itself before this path. Auth-env, stdio `Server::run`, observe/enforce runtime, and `enforcement-sarif` errors now share the prefix; that is the intended takeover of top-level human rendering, not a second emitter.

4. **Unix fixtures vs platform-neutral control are honestly split.** Newline rows are `#[cfg(unix)]` because the fixture is a spawned argv path. The budget-`0` row is platform-neutral and pins that uniqueness is not bought by deleting the operator chain. Windows was not driven here; that matches the non-claim.

5. **Claimed mutations bite (reasoned, not re-applied).** Bare `Error: {:?}` fails uniqueness + prefix preservation. First-line-only prefix fails both newline rows. `SUCCESS` on `Err` fails the exit-1 rows. Removing the machine emit fails the event-count rows. Admitting a path into the machine object fails the key-set assert even when the line count stays 1.

6. **Stderr-format consumers.** Line-JSON startup consumers and `contains(...)` diagnosis checks still hold: golden-path startup/deny, `stdio_auth_boundary` (7), `enforcement_sarif_cli` (21 passed, 1 ignored). No consumer required the old `Error:` prefix.

7. **Claims / weight.** One new user-visible string (`assay-mcp-server: `). Non-claims match the code: no loader/schema/sanitizer change, no second product emit, clap/JSON-RPC/runtime-deny/observe/file side-channels untouched. ~165 added lines, most of them contract tests, is proportionate for this boundary.

## Verification

- Toolchain: `rustc 1.96.0 (ac68faa20 2026-05-25)` via `rust-toolchain.toml` (host default was 1.92; unused)
- Scratch HEAD: `9072ac4c84c7329bc75656d9f019523001c5296f` (re-checked on GitHub immediately before this comment)
- `cargo test -p assay-mcp-server --test proxy_enforce_pdp_e2e startup` — 12 passed
- `cargo test -p assay-mcp-server` — crate suite exit 0
- Consumer binaries above, plus the golden-path startup/deny row, passed

Limitations: mutations not re-driven (read-only); clippy/fmt not re-run; Windows not driven; terminal-control / U+2028 remain outside this rule, as claimed.

This comment is the quorum record for this exact head. A new push invalidates it.

PR #2455, opened 2026-08-17
docs(coverage): state tools_declared derivation basis

### LINKED ISSUE #2427: [FEAT] `coverage_report_v1` does not state how `tools_declared` was derived
### Is your feature request related to a problem?

`coverage_report_v1` has two producers that build `tools_declared` under opposite conventions, and the report states neither, so the field's meaning depends on which command wrote it.

**Path A, `assay mcp wrap --coverage-out`.** `collect_declared_tools` (`crates/assay-cli/src/cli/commands/mcp/coverage_input.rs:12-50`) assembles the set from a policy's `tools.allow`, `tools.deny`, `schemas`, `constraints` and `tool_pins`, passing each through `is_explicit_tool_name` (lines 7-10), which discards, correctly, every pattern containing `*` that cannot be enumerated, and then does not record that it did so. The behaviour is deliberate and pinned by `mcp_wrap_coverage_collect_declared_tools_ignores_wildcards` (line 267).

**Path B, `assay coverage --input` with `--declared-tool` / `--declared-tools-file`.** `load_declared_tools` (`crates/assay-cli/src/cli/commands/coverage/generate.rs:122-158`) trims, rejects empty strings, and inserts whatever remains. No wildcard filtering. `is_explicit_tool_name` exists only on path A. So `--declared-tool '*'` lands in `tools_declared` as a literal member, where it can never match a seen tool and silently changes what the `tools_unknown` difference means.

Both write the same `schema_version: coverage_report_v1`. The emitted document (`crates/assay-cli/src/cli/commands/coverage/report.rs:166-186`) carries `tools_seen`, `tools_declared`, `tools_unknown`, taxonomy and routes, and no field records how the declared set was derived or which convention applied. `run.source` distinguishes the producer (`decision_jsonl` vs `jsonl`) and so implies the convention to a reader who already knows this issue exists, but it says nothing about whether the underlying policy contained wildcards at all.

The consequence for a reader of a report alone: an exhaustively enumerated policy and a policy whose real reach is `allow: ["*"]` with a handful of named tools beside it render identically on path A. On path B the wildcard is visible but is counted as though it were a tool.

**What already mitigates this.** `tools_unknown = tools_seen − tools_declared` (`report.rs:132`) with the finding `"tool '{}' was seen but not declared"` (`report.rs:48`) catches a wildcard's reach whenever an undeclared tool is *actually exercised*. That covers the observable case. The residue is the unexercised reach, which is the region a coverage claim is consulted about.

**This is not an enforcement concern, and that is checkable.** `collect_declared_tools` feeds the coverage report only; the proxy receives the full unfiltered policy (`McpProxy::spawn(cmd, cmd_args, policy, config)`, `crates/assay-cli/src/cli/commands/mcp.rs:141`). No filtering here has ever touched a policy decision. The request is that a correctly computed set state what it is a set of.

**Assay has already written down the underlying asymmetry**, in a test doc comment (`crates/assay-common/src/tool_pattern.rs:167-172`):

> An unbounded `*` over-blocks in `deny` and over-permits in `allow`. Only the second is dangerous, and the asymmetry is invisible in a policy file, which is why the reference states it.

with the companion test asserting that `allow: ['*']` admits tools including `a.tool.added.next.week` (lines 178-186). That reasoning reaches nobody writing a policy and nothing that leaves the process. Closing that distance is the ask, and it is deliberately the same shape as #2422: an existing rule that does not reach the emitted claim.

---

### Research context (checked 2026-08-16, none reproduced here)

**Breadth is where declared and effective diverge most, and by a wide margin.** Elvis Yeboah-Duako and Pubali Datta, *Overprivilege Analysis of Security Policies in Serverless Cloud Applications*, [arXiv:2607.02875](https://arxiv.org/abs/2607.02875) (2026-07-03), analyse 689 AWS Lambda applications comprising 1,293 functions with a static framework (PrivLess) that reconciles inferred function-to-resource interactions against declared policies. Applications with wildcard-defined permissions showed an average overprivilege ratio **274x higher** than those without, and 18.8% held unnecessary privilege-escalation capability.

That figure is why this is worth a sentence in a report rather than a shrug. A denominator that excludes non-enumerable patterns is not excluding a rounding error; it is excluding the part of a declaration where the distance between what was written and what is reachable is largest. The exclusion stays correct. Reporting it silently is what this asks to change.

Their method also marks the boundary of this request. PrivLess *infers* an effective permission set from source and reconciles it against the declared policy. That is a much larger piece of work, and explicitly not proposed here: Assay would be asserting a reach it cannot substantiate from a trace. Naming the boundary is cheap; computing what a wildcard covers is not, and is out of scope.

**Prior measurement on the same asymmetry.** Christophe Parisel, *Quantifying Azure RBAC Wildcard Overreach*, [arXiv:2506.10755](https://arxiv.org/abs/2506.10755) (v1 2025-06-12, v3 2025-07-01), compiles Azure wildcards into explicit action sets and reports that about 50% of the 15,481 catalogued actions admit cross-Resource-Provider reach under non-obvious wildcards. This is the source already cited at `tool_pattern.rs:169-172`. Its measurement supports the asymmetry; I have not verified from the abstract what it recommends, so that in-tree gloss is worth a separate look and nothing here rests on it.

**The evaluation literature is auditing artifacts of this class.** Vishvesh Bhat, Jay Vaghasiya, Muhammad Ahmed Mohsin and Asad Aali, *Benchmarking the Benchmarks: A Validity Audit of Tool-Calling Evaluation*, [arXiv:2607.02577](https://arxiv.org/abs/2607.02577) (2026-06-30), audit tool-calling benchmarks across 496 expert-reviewed tasks, find an 18.5% evaluator-human misalignment rate, and recommend decomposed metrics plus evaluators that are themselves reproducible and auditable. Relevant here only as direction of travel: a set-based report that states the basis of its own denominator is cheaper to audit than one that does not.

**Protocol direction.** MCP [SEP-3140](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/3140) (proposal stage, read at head `b5d9a57f`) proposes a signed `capabilities` block enumerating a tool's `network`, `filesystem`, `subprocess` and `env` reach, chosen so declared-versus-observed is a concrete diff. Whatever becomes of it, "enumerable, and enumerated" is becoming a distinction worth being able to state about a policy rather than leave implicit.

---

### Describe the solution you'd like

Documentation first, additive field second, inference never.

**1. State the basis (worth doing on its own).**
   - Extend `collect_declared_tools`'s doc comment: the returned set is an enumeration, non-enumerable patterns are excluded by construction, and the result is a floor on the policy's reach rather than a description of it.
   - Note in `load_declared_tools` that caller-supplied entries are not filtered, so the two producers differ.
   - Record both conventions in `docs/architecture/ADR-028-Coverage-Report.md`, which is the only doc that currently discusses `tools_declared`, and in the Markdown formatter's row (`crates/assay-cli/src/cli/commands/coverage/format_md.rs:65`).
   - Fold the `tool_pattern.rs:167-172` asymmetry into the policy-authoring docs, where someone choosing between `allow: ["*"]` and an enumeration will read it. That text is the clearest statement of the hazard in the tree and it lives on a test.

**2. Optionally, let the report carry it.** An optional field on `coverage_report_v1` recording that the declaration contained non-enumerable patterns; a count is probably enough. Absence must read as unknown rather than as "fully enumerated". Note for whoever picks this up: the schema sets `"additionalProperties": false` at every level, and there are two byte-identical copies (`schemas/coverage_report_v1.schema.json` and `crates/assay-cli/schemas/coverage_report_v1.schema.json`) held together by a drift gate, so a new optional property must be added to `properties` in both. Adding one without touching `required` is additive and needs no version bump.

**3. Not proposed.** No wildcard expansion, no inferred effective-permission analysis, no source analysis, no change to `tools_unknown` or to any finding.

Happy to open the PR for step 1.

### Describe alternatives you've considered

- **Make path A include wildcards as literal strings, like path B.** This is the status quo on one path rather than a hypothetical, which is the strongest argument for stating the convention rather than picking one blindly: a literal `"*"` never matches a seen tool, so it inflates `tools_declared` and shifts `tools_unknown` without describing anything the policy reaches. If the two paths are ever unified, that decision deserves its own issue and a schema note; this one asks only that the current difference be stated.
- **Expand wildcards into the observed namespace.** Rejected: it makes the denominator depend on what happened to be exercised, which inverts the direction of the claim and still says nothing about unexercised reach.
- **Treat it as covered by `tools_unknown`.** Partly true, and the reason this is documentation-first rather than urgent. The exercised case is caught today; the unexercised case is not.
- **Do nothing, since the wildcard is visible in the policy file.** Defensible for a reader holding both artifacts. It fails for the reader holding only the report, which is the reader the report is for.

### Scope boundaries

- No change to filtering behaviour on either path. Path A keeps discarding non-enumerable patterns; path B keeps accepting what the caller passes.
- No change to `tools_unknown`, to findings, or to any percentage or threshold computed elsewhere. This concerns `coverage_report_v1` and its documentation.
- No wildcard expansion, no effective-permission inference.
- Not a security fix and should not be labelled as one: the filtering never reaches an enforcement decision (see above), nothing is mis-computed, and nothing here is exploitable.
- Any new field is optional and additive, absence reading as the most restrictive interpretation.

Rul1an:
## Summary

- state both `coverage_report_v1.tools_declared` producer conventions at their source
- render the convention selected by `run.source` in Markdown reports
- define `tools_unknown` strictly as a report-local set difference, not evidence of policy bypass
- pin caller-supplied wildcard retention with a concrete report count
- document the distinction in ADR-028 without changing JSON schema or filtering behavior

Closes #2427.

## Scope and non-claims

`assay mcp wrap --coverage-out` still excludes wildcard patterns and reports an enumeration floor. `assay coverage --input` still keeps caller-supplied declarations literally. No schema field, wildcard expansion, inferred reach, finding, threshold, or enforcement behavior changes.

`tools_unknown` means observed tools minus this report's own `tools_declared` list. It does not prove a policy miss, bypass, or unenforced call.

## RED / GREEN

Initial RED on base `dc66c244ebb4408f3493204318cb69c4e6660405`:

```text
coverage_format_md_generates_summary ... FAILED
assertion failed: markdown.contains("Caller-supplied declarations are literal values; wildcard-looking values are not expanded or filtered.")
```

Review-finding RED before final head:

```text
coverage_declared_basis_docs ... FAILED
runbook did not state the report-local tools_unknown basis/non-claim
```

GREEN implementation head `f70d9c40ceccc3eb60008f535109533bc7261cde`; current merge-only head `d2bd1fc4780ddf25ec543f2656c238dce3db87c8`:

```text
cargo test -p assay-cli --test coverage_declared_basis_docs -- --nocapture
cargo test -p assay-cli --test coverage_format_md -- --nocapture
cargo clippy -p assay-cli --test coverage_format_md --test coverage_declared_basis_docs -- -D warnings
cargo fmt --all -- --check
git diff --check
# all exit 0
```

The full `cargo test -p assay-cli` and all-target clippy suite passed on parent head `31f4a9a3c946f391d33a40d12b19a907d0c2287a`; GitHub Actions supplies final integration proof for this head.

## Mutation evidence

- deleting the policy-producer exclusion statement fails `coverage_declared_basis_docs`
- deleting the caller-basis Markdown line fails `coverage_format_md`
- deleting ADR-028's enumeration-floor non-claim fails `coverage_declared_basis_docs`
- filtering caller-supplied `*` in the loader fails the exact JSON assertion (`[]` versus `["*"]`)
- deleting the runbook's set-difference or policy-bypass non-claim fails `coverage_declared_basis_docs`

## Review disposition

An independent review on prior head `31f4a9a3c946f391d33a40d12b19a907d0c2287a` found two blockers:

- the runbook overclaimed that `tools_unknown` implied policy/config escape
- the wildcard test did not prove literal `*` retention

Both were fixed and mutation-tested. A second independent review on `85f52d51a27a5c8358071cbdbd0171b6fdd742ac` then found that cardinality did not prove literal `*` retention and that the policy wording understated the actual filter. This head now asserts the exact JSON array contains `"*"` and states the implemented rule precisely: the policy producer excludes every value containing `*`. A fresh exact-head review was recorded on `f70d9c40ceccc3eb60008f535109533bc7261cde`. The current head is a pure, conflict-free `origin/main` merge; the recorded whole-tree and reviewed-file checks satisfy the repository review-carry rule.

## Smoking gun

A real `assay coverage --input ... --declared-tool '*' --format md` run emits the source-specific explanation and JSON `tools_declared: ["*"]`, proving the wildcard-looking value remains a literal declaration.


Rul1an:
## READY

Independent exact-head review.

**Head:** `f70d9c40ceccc3eb60008f535109533bc7261cde`

No findings.

The independent non-building reviewer rechecked the complete seven-file diff, issue #2427, both producer paths, report subtraction, schema and ADR boundaries after the two prior blocker rounds.

Verified:

- `tools_unknown` is strictly the report-local set difference, with explicit policy/enforcement non-claims;
- caller-supplied `"*"` is asserted as exact JSON `["*"]`, not inferred from cardinality;
- removing `"*"` is discriminated by that exact-value test;
- policy-producer wording covers every value containing `*`, including interior stars;
- `run.source` selects truthful producer-specific language;
- no schema, filtering, finding, enforcement, or unrelated behavior scope creep;
- public strings stay within ADR-042/043 boundaries.

Focused review verification: 4 tests passed, `git diff --check` clean, final worktree clean, exact HEAD rechecked. GitHub Actions remains the final integration proof.

**Independence:** the reviewer authored or edited neither this implementation nor any normative design or implementation plan governing it. The review was substantive and read-only; no edits, commits, pushes, GitHub posts, merges, or broad builds were performed.

Rul1an:
## Review carry to current head

The independent substantive review on `f70d9c40ceccc3eb60008f535109533bc7261cde` carries to current head `c16e92a73a0238e779b2e414e34e6984b8adc881` under both `AGENTS.md` conditions:

1. **Whole-tree equivalence:** before merging `origin/main` at `513243e506d5338eec92949140b75171a61e0afd`, `git merge-tree --write-tree <main> <reviewed-head>` produced `471f500762ad6efbac5fadc57e5b311def1aa9d3`; after the conflict-free merge, `HEAD^{tree}` is exactly the same tree.
2. **Reviewed-file isolation:** the intersection between the seven files changed by the reviewed head and the files changed from its base `dc66c244ebb4408f3493204318cb69c4e6660405` to current `main` is empty.

No history was rewritten. The only new commit is the upstream merge. Review findings remain fully disposed. Delegated proof does **not** carry and will be regenerated against the current exact SHA.

Rul1an:
## READY

Review revalidated for current head `c16e92a73a0238e779b2e414e34e6984b8adc881`.

The independent substantive review on `f70d9c40ceccc3eb60008f535109533bc7261cde` carries because both repository conditions were measured and recorded:

- whole-tree equivalence: expected and actual tree are both `471f500762ad6efbac5fadc57e5b311def1aa9d3` after merging `origin/main` `513243e506d5338eec92949140b75171a61e0afd`;
- the upstream advance touched none of the seven reviewed files.

No history rewrite or conflict resolution occurred. No findings. Delegated proof remains independently exact-SHA-bound and is not inferred from this review carry.

Rul1an:
Exact head: `d2bd1fc4780ddf25ec543f2656c238dce3db87c8`

VERDICT: READY

The independent substantive review already revalidated on `c16e92a73a0238e779b2e414e34e6984b8adc881` carries to this pure upstream-merge head under both `AGENTS.md` conditions:

1. Whole-tree equivalence: `git merge-tree --write-tree c16e92a73a0238e779b2e414e34e6984b8adc881 origin/main` produced `4146cce8e127201c79d5a59e6cba7859ec999caa`; after the conflict-free merge, `HEAD^{tree}` is exactly `4146cce8e127201c79d5a59e6cba7859ec999caa`.
2. Reviewed-file isolation: the intersection between all seven PR/reviewed files and files advanced on main from `513243e506d5338eec92949140b75171a61e0afd` through `dbda1fe425e62f8e81ef6737f5f3e7febeddaa67` is empty.

No history rewrite or manual conflict resolution occurred. All prior findings remain disposed. Required CI and delegated proof must still bind independently to this new exact SHA.

PR #2527, opened 2026-08-19
ci: add auditable release verification oracle

Rul1an:
## Description
Adds a read-only in-repo release oracle for post-publish and pre-tag checks, plus an offline mutation-focused self-test and the minimal pre-commit wiring.

The oracle uses the fail-closed convention: **0 = clean, 1 = finding, 2 = infra**. It treats only these four vVERSION-derived assets as expected-absent attestation 404s: release-proof-kit and release-provenance, each with its `.sha256` companion. Install-pin promotion remains explicitly out of scope; tag trees must retain the separate `v5.3.0` pin.

## Type of Change
- [x] New feature

## Verification
- [x] `bash -n scripts/ci/verify-release.sh scripts/ci/test-verify-release.sh`
- [x] `bash scripts/ci/test-verify-release.sh`, including abbreviated-SHA and expected-404 mutation cases
- [x] Default live post-publish verification: `bash scripts/ci/verify-release.sh` → `CLEAN: published release v5.4.0`
- [x] Pre-tag verification on `fa3dde9b8e58893a2b008e2c78350d08a8bdf58a`
- [ ] ShellCheck unavailable in this environment

The requested live v5.3.0 self-test runs the same parser, but cannot currently pass while enforcing the specified crates.io rule: crates.io reports `assay-adapter-api` newest version `5.4.0`, not `5.3.0`. This is reported fail-closed rather than weakened.

## Checklist
- [x] Code follows existing script conventions.
- [x] Tests cover abbreviated SHA refusal, expected-404 classification, and the DSSE in-toto payload type.
- [x] No Cargo/workspace version change.

<div><a href="https://cursor.com/agents/bc-a41ed643-c275-41e2-9134-66b8e4314703?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a41ed643-c275-41e2-9134-66b8e4314703&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

- **New Features**
  - Added automated release verification covering assets, checksums, attestations, version metadata, and release consistency.
  - Added support for pre-release, post-publish, self-test, and offline validation workflows.
  - Improved reporting with clear result classifications and distinct outcomes for successful checks, detected issues, and infrastructure failures.

- **Tests**
  - Added offline contract tests covering expected assets, checksum validation, attestation verification, and error handling.
  - Added safeguards to ensure verification behavior remains accurate as changes are introduced.


Rul1an:
## Reviewed and verified at `aaab53cd2`

Ran it rather than reading it. `bash scripts/ci/test-verify-release.sh` → `PASS: verify-release
offline contract tests`, exit 0. Unit seams check out: `--unit-expected-assets 5.4.0` emits exactly
23 names, `--unit-validate-sha` accepts the full SHA and refuses `bbb5e7fe4`.

The oracle is now where it belongs — a release check living in a scratch file was not a gate, and
this is.

## Three things here are better than what I asked for

**1. The mutation proof is real, and it guards its own anchor.** It copies the actual oracle,
flips `expected-absent`'s `return 0` to `return 1`, and asserts the targeted assertion goes red —
with the comment "This checks behavior through a copied oracle, not a mock". Then:

```python
if needle not in text:
    raise SystemExit("expected-absent implementation anchor moved")
```

That is a control on the control. If the implementation moves, the mutation test fails loudly
instead of quietly mutating nothing and still passing. I had to fix exactly that failure in my own
work today, so this is the pattern I would rather we all copy.

**2. The measured exception is scoped, not blanket.** `classify_attestation` treats a 404 as
`expected-absent` only for the four proof/provenance names, and
`expect_status 1 ... assay-v5.3.0-x86_64-unknown-linux-gnu.tar.gz 404` pins that a 404 anywhere else
is still a finding. That is the distinction that makes the exception safe: an oracle asserting
"every asset is attested" would go red on a correct release, and one that just tolerated 404s would
go green on a broken one. This does neither.

**3. The stdin-JSON defect is now a structural guard.** `grep -Eq '\|[[:space:]]*(python3|"\$PYTHON")[[:space:]].*<<'`
fails if anyone ever pipes JSON into a Python heredoc again, and `bounded_capture` is required to
exist. You fixed that bug during development and then made it unrepeatable. That is the right
instinct — the fix that outlives the fixer.

Also good: `latest.json` is explicitly refused by the asset generator, which is the shape that
stranded v3.33.0, and `PIN_SHA` is refused empty and abbreviated as well as at the unit seam.

## One observation, not a change request

The pre-commit hook's `files:` regex covers `verify-release.sh`, `test-verify-release.sh` and
`.pre-commit-config.yaml`. That is correct for what the hook proves. Worth knowing for later: the
oracle's expectations also encode facts about `release.yml` — the 23-name set and the four
unattested names — so a change to the release workflow's asset set will not trigger this hook and
will surface only at the next tag. Not a defect in this PR; the alternative is a broader trigger
that fires on unrelated workflow edits. Recording it so the next person meets it as a known boundary
rather than a surprise.

In the merge queue. Nice work — this is the artifact I most wanted out of today.

PR #2521, opened 2026-08-18
feat(mcp): move proxy errors to application-defined codes and v1 evidence

Rul1an:
## Summary

- move Assay-originated proxy errors to application-defined codes: unsupported `-31997`, failed `-31998`, denied `-31999`
- emit `assay.denied_call_observation.v1` with denial-code parity and a digest over the exact caller-visible response bytes
- preserve upstream reserved `-32042` responses value-equivalently without minting an Assay denial observation
- update current generated golden-path/plugin surfaces and publish the exact old-to-new migration table

## Why

MCP owns `-32020..=-32099`, including `-32042`. Assay's old proxy codes overlapped that protocol-owned range. The compatibility reader/profile landed first in #2520; this PR switches the live producer without rewriting historical evidence.

## Contract changes

| Meaning | Old producer | New producer |
|---|---:|---:|
| unsupported | `-32040` | `-31997` |
| failed | `-32041` | `-31998` |
| denied | `-32042` | `-31999` |

Current denials use `assay.denied_call_observation.v1 / -31999 / assay-proxy`. Historical v0 corpus/profile bytes and the verifier's default v0 selection remain unchanged.

## RED -> GREEN

Before production changes, focused tests failed on:

- exact mapping: got `-32040`, expected `-31997`
- reserved range: old codes were inside `-32768..=-32000`
- abandoned mapping mutation: restoring `-32040/-32041/-32042` survived
- observation schema: producer still emitted v0

After the change, unit tests pin exact mapping, uniqueness/range, reader parity, v1 schema and raw-byte digest. E2E tests independently pin wire literals. A dedicated upstream test uses nested sentinel data and exact-object equality to prove value-equivalent `-32042` relay, plus no Assay origin/observation.

## Verification

- `cargo test -p assay-mcp-server`
- focused proxy E2E and golden-path contract tests
- `cargo test -p assay-cli --test conformance_privileged_mcp_action --test conformance_privileged_mcp_action_v1`
- `python3 scripts/docs/generate-agent-golden-path.py --check`
- `cargo fmt --all -- --check`
- `cargo clippy -p assay-mcp-server --all-targets -- -D warnings`
- `git diff --check`
- pre-commit and pre-push hooks, including workspace clippy and Linux cross-target compile
- v0 corpus/profile path diff is empty

## Compatibility and nonclaims

- Depends on merged #2520; no release occurred between reader and producer slices.
- Verifier default remains v0; live v1 evidence requires explicit `--profile-version v1`.
[...]

Issue #2413, opened 2026-08-16
Generated drift gate omits agent-plugin golden-path artifacts


Rul1an:
Origin: independent exact-head review of #2411 at `b5a219d5267d153cf74592d32ffda87d1b2d46bf`.

## Driven observation

`packaging/agent-plugin/skills/assay-golden-path/SKILL.md` and its `references/agent-golden-path.json` are generator-owned, but are absent from `scripts/ci/check-docs-generated-drift.sh`'s `GENERATED` inventory.

Restoring only those two files to the base version while keeping the generator current produced:

- `scripts/ci/check-docs-generated-drift.sh`: exit 0
- `scripts/ci/test-agent-golden-path-skill.py`: exit 0
- `scripts/docs/generate-agent-golden-path.py --check`: exit 1

The direct `--check` catches the drift, but is not invoked by a real hook/workflow; grep only found plans and a fixture-staged copy inside a hardening test.

## Acceptance

- One generator-owned inventory includes both agent-plugin paths.
- The real hook/workflow path invokes the check.
- A mutation restoring either generated file to stale content fails the committed gate.
- Keep the generator as the sole source; do not add a second parity implementation.

## Non-claim

#2411's generated files were byte-identical when regenerated. This is a forward guard gap, not evidence of current drift.

PR #2528, opened 2026-08-19
fix(ci): pair published-release and example verify with produced v1 denial bundles

Rul1an:
## Description

The v5.4.0 published-release golden path produced a bundle with `--denied-observations` (v1 denial record) and then verified it with the default v0 profile. Fail-closed profile validation rejected that bundle (`E_EVIDENCE_PROFILE_INVALID`) even though published binaries and bundle integrity were fine.

**Chosen option: (a)** verify the produced (and tampered) bundle with `--profile-version v1`.

The journey's purpose is "published install → produce → verify the produced bundle" (`install-to-verified-evidence`), not "prove the default v0 verifier." CHANGELOG 5.4.0 already states the intended pairing: live produce verifies with `--profile-version v1`; the product verifier default stays v0. Option (b) (a second decisions-only produce + default-v0 verify) would invent a second golden-path contract this job does not claim.

The same pairing is applied to the later tampered-bundle verify in the same script. Integrity is still stage 1 (`E_EVIDENCE_INTEGRITY`); v1 pairing keeps that check on the bundle the journey actually produced. The sibling `verdict == "valid"` assertion is unchanged and now locked by the contract.

## Scope extension — shipped example (owner comment)

The same produce-v1 / verify-v0 pairing is in `examples/privileged-action-gate/run.sh`: import with `--denied-observations`, verify with no `--profile-version`. Pairing is applied only to the denied-observation bundle. The allowed-path bundle stays default v0 (decisions-only produce). `matrix()` forwards extra args so the call-site flag actually reaches verify.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor

## Decision and non-claims
- Harness and example pairing only. No product, policy, or process-exit semantic change.
- Verifier default remains v0 in the product.
- `.github/assay-release-tag` is untouched (now `v5.4.0` after #2530; this PR does not move it).
- Generated golden-path docs left untouched: four `SKILL.md` copies and `docs/guides/agent-golden-path.md` step 8 still document `verify-privileged-mcp-action <bundle> --format json`. That is a v0-journey vs v1-journey contract decision to make once and regenerate, not a per-file patch. The generator already owns those files; a single regenerate was not treated as the pairing contract here.
- `KeyError: 'claims'` on a `verdict: invalid` report is a different property (run.sh assumes valid-report shape). Left as a follow-up; not folded into this pairing change.
- Other `verify-privileged-mcp-action` sites: `sarif.sh` does not verify (enforcement-sarif only). Example README still shows a table-format verify next to a generic import snippet; not the `run.sh` path. Packaging / `.claude` / `.agents` SKILL copies are generated and left untouched.

## Verification

Harness pairing (unchanged, independently re-verified): `python3 scripts/ci/check-published-release-golden-path-contract.py`

- **RED** against the unfixed `origin/main` driver: **exit 1**
  - `driver lost verify produced bundle`
  - `driver lost verify tampered bundle`
  - `driver must verify the produced denial-observation bundle with --profile-version v1 exactly once`
  - `driver must verify the tampered denial-observation bundle with --profile-version v1 exactly once`
  - `driver verifies a produced or tampered bundle without --profile-version v1`
- **GREEN** after the pairing fix: **exit 0** (`ok: published-release golden-path contract`)

Example pairing control (same checker, extended):

- **RED** against the unfixed `run.sh`: **exit 1**
  - `example must verify the denied-observation bundle with --profile-version v1`
  - `example matrix() must forward extra args to verify-privileged-mcp-action`
- **GREEN** after the example pairing: **exit 0** (`ok: published-release golden-path contract`)

Same reported bypass (shipping denied-bundle verify line with `--profile-version v1` stripped — the constant that ships, not a copied literal) against both guards:

- **old guard** (`validate_contract` only, blind to `run.sh`): **exit 0**
- **new guard** (example pairing included): **exit 1**

Full suite `bash scripts/ci/test-published-release-golden-path-contract.sh`: **exit 0** (prints `example bypass old_guard_exit=0 new_guard_exit=1`).

A guard that has only ever been green proves nothing; the example control failed on the current `run.sh` before the pairing change.

## Test plan
- [x] Watch the pairing contract fail on the unfixed driver (exit 1)
- [x] Watch the same checker pass after `--profile-version v1` pairing (exit 0)
- [x] Mutation cases drop produced/tampered `--profile-version v1` or loosen `verdict == "valid"` and stay red
- [x] Watch the example control fail on unfixed `run.sh` (exit 1) and pass after pairing (exit 0)
- [x] Seed the reported bypass from the shipping verify constant; old guard exit 0, new guard exit 1
- [ ] Confirm the live `published-release-golden-path` job on tag v5.4.0 can complete `verify-produced-bundle` after this lands

## Checklist
- [x] Change is limited to published-release harness pairing plus the sibling example pairing
- [x] Existing contract test harness extended; no new golden-path contract invented
- [x] `.github/assay-release-tag` not modified
- [x] Generated SKILL / agent-golden-path docs not modified


Rul1an:
## Independent verification — reproduced, not taken on trust

I re-ran your control myself rather than reading the claim, at head `b679f09cfed1f65f9640c658275690ed0e271e4b`.

**Your control goes red on the unfixed driver.** Running your checker against `origin/main`'s driver:

```
FAIL: driver lost verify produced bundle
FAIL: driver lost verify tampered bundle
FAIL: driver must verify the produced denial-observation bundle with --profile-version v1 exactly once
FAIL: driver must verify the tampered denial-observation bundle with --profile-version v1 exactly once
FAIL: driver verifies a produced or tampered bundle without --profile-version v1
```

**Green on your branch:** checker `exit 0` (`ok: published-release golden-path contract`), and
`test-published-release-golden-path-contract.sh` runs 7 tests, `OK`, `exit 0`.

**The driver diff is two lines.** That is the right size for this defect, and it is the part I want
to call out: the temptation with a red release job is to restructure the journey. You changed the
two invocations that were actually wrong and left everything else alone.

Three things you did that I would not want lost:

1. **You reasoned about which contract the job claims**, rather than picking the option that made
   the red go away. "The journey's purpose is published install → produce → verify the produced
   bundle, not prove the default v0 verifier" is the correct framing, and option (b) really would
   have minted a second golden-path contract this job never claimed.
2. **You extended the pairing to the tampered-bundle verify.** I flagged it as a place to look;
   you checked it and found the same defect rather than reporting back that it looked fine. The
   tamper case still asserts `bundle_integrity == "fail"` and `E_EVIDENCE_INTEGRITY` at exit 2, so
   integrity is still evaluated ahead of profile and the v1 pairing does not soften tamper
   detection. I verified that assertion is unchanged.
3. **You held the pin.** `.github/assay-release-tag` untouched.

One observation, not a change request: the checker pins the harness by digest
(`fixtures/published-release-golden-path/v1/harness-manifest.json`), so any future driver edit must
move the digest in the same commit or the contract fails with `harness digest drifted`. I hit that
myself while testing a partial overlay. That is the mechanism working, but it is worth a line in
the fixture or the checker saying so, because the failure message reads like corruption rather than
"you edited the driver and did not re-pin it".

Approving on my side. This unblocks the install-pin promotion once it is green on a real tag.

Rul1an:
## The same defect is in the shipped example — please extend this PR

Found while getting the pin promotion (#2530) through CI. `demo-smoke` went red, and it is the same
root cause you are fixing here, in a second place: `examples/privileged-action-gate/run.sh`.

```
run.sh:139   --decisions "$WORK/pmadeny.ndjson" --denied-observations "$WORK/pmadeny-obs.ndjson"
run.sh:123   "$ASSAY_CLI" evidence verify-privileged-mcp-action "$bundle" --format json
```

Produces with `--denied-observations`, verifies with no `--profile-version`. Identical pairing.

**Reproduced locally** against a 5.4.0 source build. Every verdict is correct — the five
DENY/ALLOW lines and reason codes match `expected-output.txt` exactly. The example then dies here:

```
File "<string>", line 5, in <module>
  cells = " ".join("{}={}".format(name, cell["status"]) for name, cell in report["claims"].items())
KeyError: 'claims'
```

`report["claims"]` is absent because the report is a `verdict: invalid` profile report, which
carries `findings` instead. Same fail-closed path, surfacing as a `KeyError` because the example
assumes the valid-report shape.

**This is pre-existing on `main`, not caused by #2530.** My branch touches zero files under
`examples/`, and `run.sh` is byte-identical to main. `demo-smoke` has a `paths:` filter, so #2521
changed proxy behaviour without ever triggering it; #2530 happened to touch a path that does. That
is the same class of blind spot as #2413 — a gate whose scope is narrower than the behaviour it
guards.

### Suggested scope for this PR

1. Pair the example's verify with its produce, the same way you paired the harness.
2. Consider whether `run.sh` should fail loudly rather than raising `KeyError` when the report is
   not `valid`. Right now an invalid verdict reaches the user as a Python traceback, which reads as
   a broken example rather than as a refused verification. Your call whether that belongs here or
   in its own issue — I lean its own issue, since it is a different property.

I am **not** blocking #2530 on this. `demo-smoke` is not a required context, the breakage predates
the branch, and holding the pin again would re-deadlock every PR including this one. Merging the pin
does not make the example worse; landing your fix is what repairs it.

Excellent instinct on the tampered-bundle verify earlier, by the way — that was the same "check the
sibling call site" move that would have found this one. Worth applying once more: `sarif.sh` and
`packaging/*/skills/.../assets/` carry example assets too, so it is worth a grep for any other
`verify-privileged-mcp-action` without a paired `--profile-version`.

Rul1an:
Addressed the example pairing from the scope-extension comment in `fa596f052`.

`examples/privileged-action-gate/run.sh` now verifies the denied-observation bundle with `--profile-version v1` (allowed-path bundle stays default v0). Same checker as the harness, not a second contract.

- Example control: **RED exit 1** on unfixed `run.sh`, **GREEN exit 0** after the pairing.
- Same mutant (shipping denied-bundle verify line with `--profile-version v1` stripped) against both guards: **old exit 0**, **new exit 1**.
- `KeyError: 'claims'` left as a follow-up — different property from pairing.
- Generated SKILL / `docs/guides/agent-golden-path.md` step 8 left untouched (v0 vs v1 journey is a one-time regenerate).
- Grep of sibling verify sites: `sarif.sh` does not verify; packaging / plugin SKILL copies are generated and untouched.
- `.github/assay-release-tag` not touched.

Rul1an:
## Example fix independently verified at `47ff13dee`

Built `assay-mcp-server` from this head and ran the example end to end.

**Before** (`origin/main`, 5.4.0 binary): all verdicts correct, then `KeyError: 'claims'`, exit 1.
**After**: `OK: privileged-action-gate verdicts match expected-output.txt`, exit 0.

The claim matrix renders for both bundles:

```
denied path bundle : verdict=valid  policy_decision_recorded=confirmed caller_visible_denial=confirmed  upstream_delivery=incomplete external_side_effect=incomplete
allowed path bundle: verdict=valid  policy_decision_recorded=confirmed caller_visible_denial=incomplete upstream_delivery=incomplete external_side_effect=incomplete
```

## Two things here are better than what I asked for

**1. You paired per bundle, not per script.** I said "pair the example's verify with its produce",
which read as "add `--profile-version v1` to the verify". You made `matrix()` take optional
verify-args and applied v1 only to the denied-path bundle — the one that actually carries the
denial observation — leaving the allowed-path bundle on the default v0.

That is the more precise reading and it preserves something my version would have destroyed: the
example still exercises the v0 verifier on a v0 bundle. A blanket v1 would have turned the example
into a v1-only demonstration and quietly dropped default-path coverage. The output above shows both
paths still being exercised, with `caller_visible_denial` differing exactly as the two profiles
should differ.

**2. You closed the blind spot, not just the bug.** Adding `run\.sh` to the pre-commit `files:`
regex means the contract self-test now runs when the example changes. That is the same defect class
as #2413 — a gate whose scope is narrower than the behaviour it guards — fixed at its second site
without being asked. It is the reason #2521 could change proxy behaviour without anything noticing,
and it will not happen again on this path.

## Not blocking, for a follow-up issue

`run.sh` still assumes the valid-report shape, so a `verdict: invalid` reaches the user as a Python
traceback rather than as a refused verification. Your fix makes that unreachable on the happy path,
but the fragility is unchanged. Different property, own issue, as discussed.

Merging as soon as CI is green. This is what turns the v5.4.0 published-release journey green.

Issue #2383, opened 2026-08-14
test(cli): finish M18 fixture discrimination in InitSuccess publisher inventory


Rul1an:
## Context

Follow-up from the independent exact-head review of PR #2379 at `3307441ab1a67f53e2f2349e62c77604a62e17fe`. The product fix and its principal mutation are covered: the M13 new-variant mutation that survived at the prior head now fails in `every_init_success_variant_is_a_driven_publisher`.

The reviewer found three non-blocking weaknesses in the new source-level inventory guard. They do not reproduce a defect in the current publishers, but they can let a future test-only registration drift remain green.

## Measured gaps

1. **Derived-name collision.** A new `InitSuccess::Preset` maps to the existing `Publisher::InitPreset`; the whole crate stays green although the new live variant is not independently driven.
2. **Comment containment.** A comment containing `Publisher::InitCiFollowUp` satisfies the current plain-text table membership check.
3. **Non-discriminating fixture.** A registered publisher can hard-code `--config=eval.yaml` and stay green when its fixture uses that same value.

These were mutations M14, M15 and M18 in the independent disposable review worktree. The reviewer classified them as nits and returned `APPROVE WITH NITS`; they are not reasons to reopen #2371's shipped-behavior scope.

## Definition of Done

- [x] Replace derived-name/plain-text containment with a mechanically parsed or typed set comparison that rejects comments and duplicate/colliding mappings. (landed via #2532, 2026-08-19)
- [x] Assert equality in both directions: every `InitSuccess` publishing shape is registered, and every init publisher registration corresponds to a production shape. (landed via #2532)
- [ ] Use a discriminating caller-controlled value per publisher; the expected value must not be repeated as an implementation literal that can drift together.
- [ ] Re-run M14, M15 and M18 and record all three as killed. (M14 and M15 killed via #2532; M18 open, this is the residual)
- [ ] Preserve the existing M13 kill and current 119 publish-then-execute cases.
- [ ] Keep this test-only unless a production registry is demonstrably simpler than source parsing.

## Non-claims

- No current published recovery argv is known broken after #2379.
- This does not expand the shell contract or cover non-UTF-8 paths (#2264).
- This is test-harness durability, not a release blocker.

## Related

- #2371
- PR #2379

Rul1an:
Reopening. This closed automatically and should not have.

## What happened

Ruley's review on #2532 said, correctly, that `Closes #2383` closed a partial DoD: the change kills
[...]

Issue #2414, opened 2026-08-16
Evidence remediation prose guard permits embedded command text


Rul1an:
Origin: independent exact-head review of #2411 at `b5a219d5267d153cf74592d32ffda87d1b2d46bf`.

## Driven observation

The evidence remediation tests claim prose rather than commands, but their guard is prefix/literal based. Replacing a remediation with prose containing an embedded executable command such as `Try assay evidence show --format json ...` survived, while the positive control `Run: assay ...` was killed.

The pattern is inherited by `evidence_contract_remediation_promises_no_command` from the existing integrity remediation test; it is not introduced by #2411.

## Acceptance

- Define the public property precisely: prose-only, or prose that may mention a command.
- If prose-only remains normative, use one shared detector/function for all evidence remediations and drive prefix, mid-sentence, backtick and JSON-argv forms.
- Keep caller-controlled context out of static remediation text.
- Preserve the separate allowlist for reason codes that intentionally publish executable recovery argv.

## Non-claim

No command injection was demonstrated. This issue is about a guard being narrower than its stated public property.

Issue #2260, opened 2026-08-10
The doc-block guard matches "doc" as a substring, and admits a `#[deprecated]` note it says nothing admits


Rul1an:
`nothing_but_the_boundary_include_documents_the_integrity_variant` in
`crates/assay-cli/tests/spec_reason_code_registry.rs` guards the transport claim for
`ReasonCode::EEvidenceIntegrity`: nothing but `#[doc = include_str!("exit_codes/evidence_integrity_boundary.md")]`
may document that variant, so a withdrawn claim cannot render ahead of the boundary text.

Two defects in that guard, found by the independent review of
[#2241](https://github.com/Rul1an/assay/pull/2241) at head `6cc52f1913db74dcd5845ae579683073d530b68d`
([final record](https://github.com/Rul1an/assay/pull/2241#issuecomment-5243711507),
[interim record](https://github.com/Rul1an/assay/pull/2241#issuecomment-5243619511)). Neither is
live on `main` today: `exit_codes.rs` lines 81–84 are plain `//` comments and nothing is
mis-documented. Filed so the findings survive independently of any branch.

## 1. `cannot_document` matches `doc` as a substring, refusing five legitimate shapes

`crates/assay-cli/tests/spec_reason_code_registry.rs:406-413`:

```rust
let closed_attribute_without_doc =
    line.starts_with("#[") && line.ends_with(']') && !line.contains("doc");
```

`!line.contains("doc")` is a substring test over the whole line. `#[allow(rustdoc::bare_urls)]` is
therefore refused, because "rustdoc" contains "doc" — a lint allow is exactly what an author
reaches for next to an included markdown file in a crate that runs `Rustdoc (deny warnings)`.

Five shapes are refused although each is correct code with no second doc source, and each is
rustfmt-clean (`cargo fmt --all -- --check` exit 0 with the shape in place, measured at
`6cc52f191`):

| # | shape | why it is refused |
|---|---|---|
| O1 | `#[doc(alias = "E_EVIDENCE_INTEGRITY")]` | any single-line attribute containing `doc`; `doc(alias)` adds a search-index alias, not prose |
| O2 | `#[allow(rustdoc::bare_urls)]` | `"rustdoc"` contains `"doc"` as a substring |
| O5 | `#[cfg_attr(\n    feature = "serde",\n    serde(alias = "…")\n)]` | an attribute broken across source lines is refused whatever its path |
[...]

PR #2532, opened 2026-08-19
test(cli): compare parsed publisher sets instead of source text

Rul1an:
Addresses the inventory half of #2383. Does NOT close it: DoD item 3 (M18, the non-discriminating
fixture value) is untouched, so the issue stays open for that.

## Measured defect

`every_init_success_variant_is_a_driven_publisher` asked whether the raw source of the `PUBLISHERS`
table *contained the text* `Publisher::{id}`:

```rust
table.contains(&format!("Publisher::{id}"))
```

Source text is not the set the test asserts about, so this was a neighbouring property in the three
ways #2383 measured: a `//` comment naming a publisher satisfied it (M15), a duplicate or colliding
registration was invisible to it (M14), and it could only ever answer one direction.

## Control on real source

M14 seeded as a live variant — `InitSuccess::Preset` with its own `argv()` arm, deriving the
already-owned `InitPreset`. Same mutant, both guards, on `2bfd6ad65`:

| | result |
|---|---|
| old containment guard | **test PASSED** — the gap |
| parsed set comparison | **test FAILED** |

with the diagnostic:

```
InitSuccess::Preset and InitSuccess::Validate both derive InitPreset; the second would inherit
the first's PUBLISHERS entry. Give the derivation a discriminating name or register each shape.
```

My first attempt at this mutant did not compile — a new variant breaks the exhaustive `match` in
`argv()`. That is worth recording: a mutant the compiler rejects proves nothing about the guard, so
the control had to supply the arm too and be a program that actually builds.

## Change

Parses the table into a set and compares sets.

- comments removed before any name is read
- each entry must be a plain `Publisher::<Ident>`; anything else is refused rather than guessed at
- duplicate registration refused
- many ids from one variant stays legal — that is the shipped `Validate` shape. One id from many
  variants now fails, since the second variant inherits the first's entry and never needs its own
- `assert_eq!` in **both** directions, so a publisher left registered after its production shape is
  deleted fails too

`publisher_parsing_rejects_the_measured_mutations` pins M15 and the duplicate, non-identifier and
non-`Publisher` shapes at the parser, so the rule cannot be weakened back into one that admits them.

## Deliberately out of scope

DoD item 3, the non-discriminating fixture value: a registered publisher can still hard-code a value
its fixture also uses, so the fixture cannot discriminate. That is fixture construction rather than
inventory comparison, and folding it in would make this change harder to read. Left to its own
change; #2383 can stay open for it if preferred, or close with a follow-up.

Test-only. No production behaviour changes. 23 tests pass; `cargo fmt` and `cargo clippy
--all-targets` clean.

## Related

Completes the guard-strength sweep: #2413 (merged, #2525), #2414 + #2232 (#2529), #2260
(re-measured as already fixed). All five share one defect class — a guard whose enforced scope is
narrower than its stated claim — and each fix ships a control demonstrated to fail against the
unfixed guard.

### DIFF
--- a/crates/assay-cli/tests/recovery_argv_publisher_parity.rs
+++ b/crates/assay-cli/tests/recovery_argv_publisher_parity.rs
@@ -15,6 +15,7 @@
 //! - a publisher or reason variant silently dropped from the table
 //! - a new `InitSuccess` variant with no driven `PUBLISHERS` case
 
+use std::collections::{BTreeMap, BTreeSet};
 #[path = "../../../tests/support/bounded_process.rs"]
 #[allow(dead_code)]
 mod bounded_process;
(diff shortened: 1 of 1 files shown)


Rul1an:
Independent review of `52187c42f27845ae93b157fa6bd12a616d8901aa` (change commit `c4d25cd84109b6ddfc78210c069a8513cb3ebe3a`). One file, inspected; I did not re-run the live M14 worktree.

GitHub refused `REQUEST_CHANGES` because this token is the PR author. Treat the verdict as blocking anyway.

**Changes requested.** I would ship the inventory comparison. I would not ship `Closes #2383`, and I would not ship M14's kill without an in-repo pin. The first is issue hygiene already offered in the PR body. The second is the same defect class this PR exists to fix: a guard whose enforced scope is narrower than its stated claim, and that can go green after the discriminating check is deleted.

### 1. Parser — not blocking

`split("//")` is naive and the PR says so. It holds because the next two asserts refuse anything that is not `Publisher::` plus a non-empty ASCII-alphanumeric ident. A `"` or `/* … */` in an entry therefore fails closed rather than registering a name. A `]` inside a table comment would truncate `publishers_table` at `body.find(']')` and also fail closed (missing ids), not go green.

I would still assert the assumption instead of leaving it as folklore, the way the release oracle refused an abbreviated SHA rather than assuming nobody would pass one:

```rust
assert!(
    !table.contains('"') && !table.contains("/*"),
    "PUBLISHERS table is not a string- or block-comment grammar; refuse rather than guess"
);
```

Underscore is a legal Rust ident and is currently refused (`is_ascii_alphanumeric`). Today's names have none. Not a reason to block.

### 2. Init* filter — not blocking; there is a cheap better shape

Reverse direction is `id.starts_with("Init")`. That matches `driven_ids_for_variant`, which always emits `Init{Name}` except the shipped Validate triple. So for *this* enum the prefix is implied by the derivation, not an extra convention.

The silent skip is the other way: a new non-init publisher (`Publisher::Foo`) is dropped by the filter and never has to correspond to anything. That is fine for an InitSuccess inventory, and the assert message honestly says `Init*`. It is still the defect class ("enforced scope narrower than a reader of the table might think").

Cheap close, no registry:

```text
registered == expected ∪ {CfgParse, PolicyParse, EvidenceUnreadable}
```

An explicit three-name allowlist. A fourth recovery publisher, init or not, cannot hide. I would take it if the file is touched; I would not block a merge on it if the issue stays scoped to InitSuccess.

### 3. DoD item 3 — blocking for the close, not for the diff

The cut is right for *this change*. Inventory comparison and fixture discrimination are different jobs; folding M18 in would make the PR harder to read.

The close is wrong. #2383 DoD still has:

- discriminating caller-controlled value per publisher
- M14, M15 **and M18** recorded killed

This kills M14 and M15. M18 is untouched. `Closes #2383` closes a partial DoD. Drop `Closes`; say this addresses the inventory half and leave #2383 open for M18, or open a follow-up and point at it. That option is already in the PR body. Take it.

### 4. Do the controls discriminate, or can they pass vacuously?

`publisher_parsing_rejects_the_measured_mutations` is not the self-literal failure mode. It does not `include_str!` itself. M15 is a dedicated string; if comment-stripping dies, `InitCiFollowUp` appears and `assert_eq!` fails. Duplicate / non-ident / non-`Publisher` are `catch_unwind` on the parser. Those three cannot pass by seeing their own source.

`publishers_table` is first `const PUBLISHERS: &[Publisher] = &[` plus first `]`. There is only one such marker. The control literals sit after that `]`, so they cannot feed `every_init_success_variant_is_a_driven_publisher`. Empty-scan is refused on both sides.

**M14's actual kill is not in that test.** Collision is the `owner: BTreeMap` in the inventory test. `publisher_parsing_rejects` only pins a *duplicate table row*. Delete the owner map, keep `assert_eq!` on sets: `Validate` and a new `Preset` both derive `InitPreset`, the set still has one `InitPreset`, registered still has one `InitPreset`, **green**. That is M14 surviving, and the parser control will not notice.

The live control (old guard pass / new guard fail, with the `argv()` arm so it compiles) is the right experiment. It is not in the repo. A guard never shown to stay red proves nothing.

Extract the owner-map into a function and pin it:

```rust
#[test]
fn one_id_from_two_variants_is_refused() {
    let blew = std::panic::catch_unwind(|| {
        expected_init_ids(["Validate", "Preset"]) // Validate already owns InitPreset
    });
    assert!(blew.is_err());
}
```

Without that pin I would not treat M14 as locked.

### Verdict

I would not approve at this head. Two changes, then I would ship the inventory work:

1. In-repo pin of many-variants-one-id. That *is* M14.
2. Do not close #2383 until M18 is killed or carved into a follow-up.

I did not re-execute the live Preset mutant, and I did not re-count the 119 publish-then-execute cases. Those are non-claims, not disagreements.

Rul1an:
## Blocking review accepted in full — both findings fixed

Thank you for actually blocking this. That was the right call and the first finding is one I would
not have found myself.

### 1. M14 was not pinned in the repository — fixed

You were exactly right. The collision rule lived as a local ownership map inside
`every_init_success_variant_is_a_driven_publisher`, and nothing pinned it. My live experiment showed
M14 killed **once, in a shell**; it did not keep it killed.

I ran your attack verbatim — delete the ownership check, keep the `assert_eq!` on sets — and
confirmed your prediction before fixing it: green, exactly as you said, because `Validate` and
`Preset` both derive `InitPreset` so one id sits on each side and the comparison is satisfied.

Extracted `expected_init_ids` and pinned it, close to the shape you proposed:

```rust
#[test]
fn one_id_from_two_variants_is_refused() {
    let blew = std::panic::catch_unwind(|| expected_init_ids(["Validate", "Preset"]));
    assert!(blew.is_err(), "...set comparison alone cannot see it");
    assert_eq!(expected_init_ids(["Validate"]).len(), 3);  // shipped shape stays legal
}
```

Re-ran your attack against the new pin: **FAILED**, with
`two variants deriving InitPreset must be refused; set comparison alone cannot see it`. Restored:
passes. I added the second assertion so the pin cannot be satisfied by refusing everything — the
one-variant-three-ids `Validate` shape must stay legal.

### 2. `Closes #2383` on a partial DoD — fixed

Also right, and it is the kind of thing that quietly turns into a false record. The DoD wants M14,
M15 **and** M18 killed; this kills M14 and M15. Dropped the closing keyword; the PR now says it
addresses the inventory half and #2383 stays open for M18.

### 3. The `Init*` prefix — took it, since the file was open

You said you would take it if the file is touched and would not block on it. The file is touched, so
it is taken: the reverse direction now names `CfgParse`, `PolicyParse`, `EvidenceUnreadable`
explicitly and compares against the whole table. A fourth recovery publisher can no longer hide
behind the filter, init-named or not. Your framing is what convinced me — a prefix filter is the
same "enforced scope narrower than it reads" defect the test exists to close, so leaving it would
have been inconsistent.

### 4. The underscore note

Taken too: `parsed_publisher_ids` now accepts `_`. You flagged it as not-a-reason-to-block and it
was not, but it is a real narrowing of a legal identifier and free to fix while here.

### On your non-claims

Recorded as stated: you did not re-execute the live `Preset` mutant and did not re-count the 119
publish-then-execute cases. I re-ran the full file after the change — 24 passed, 0 failed — and
`cargo clippy --all-targets` and `cargo fmt` are clean.

This review is the standard I want applied to my work, and the M14 finding in particular is a
better catch than anything I made on your PRs today.

PR #2534, opened 2026-08-19
docs(evidence): session findings do not read complete as a whole trace

Rul1an:
## Summary
- **Split: Option 1 ships now; Option 2 waits for a planned major.** The documentation that `TraceExtent` / session-finding `extent` make **no fidelity claim**, and that `complete` must not be read as "nothing is missing", is semver-clean and valuable alone. The issue author said so; this PR now matches that cut.
- **`coverage` is held, not dropped.** `assay-evidence` is a real published API (“Tamper-evident, cryptographically verifiable evidence bundles…”). That is not the #2533 situation: `assay-runner-schema` has *no standalone product guarantee*. Adding `pub coverage` on exhaustively constructible `PayloadSessionFinding` is `constructible_struct_adds_field` and requires a new major.
- **`#[non_exhaustive]` now is itself major.** It does not rescue this release; it only prevents the next one. Four other types in this area already have it (see #2068 `MetricResult`). Option 2 lands in a planned major **with** `#[non_exhaustive]` in the same change so this class stops recurring.
- **Third way rejected.** A `pub(crate)` / serde-only field is not additive: adding any field to an all-public exhaustively constructible struct still breaks external struct literals. It is also not the Option 2 contract (a producer-declared signal readers can use via the public Rust API). There is no production producer yet, but the typed struct *is* the public API. A serde-only hide would be a gate dodge.

Related to #2422. The issue stays open until Option 2 lands; this PR does not close it.

## Test plan
- [x] **Control is weaker and honest.** `extent_docs_make_no_fidelity_claim` greps for the disclaimer. It proves the sentence is present, not that a reader heeds it.
- [x] `cargo test -p assay-evidence --test session_finding_vocabulary_parity` (6 passed); bundle roundtrip still passes.
- [x] `cargo semver-checks check-release -p assay-evidence --baseline-rev v5.4.0`: 223 pass, 31 skip, `constructible_struct_adds_field` gone. No suppress, no workspace bump to 6.0.0.

## Type of Change
- [x] Documentation update

## Verification
- [x] Targeted `cargo test` and `cargo semver-checks` as above
- [ ] Full workspace `cargo test` not run (docs + removal of an unreleased field)

## Checklist
- [x] Wave 0 semver gate not suppressed
- [x] CHANGELOG does not claim the field shipped
- [x] Option 2 remains designed (reuse `CodingAgentCoverageState` / `PartialOnly`, absence = most restrictive) for the planned major

Made with [Cursor](https://cursor.com)

[...]

PR #2533, opened 2026-08-19
feat(runner-schema): state kernel capture origin

Rul1an:
## Description
Closes #2446.

Adds the vocabulary-only `capture_origin` statement to `ObservationHealth`.

### Why option 1
Option 1 is the smallest faithful encoding. `KernelLayerStatus` remains exactly `Complete | PartialRingbufDrops | Absent`; adding a public enum variant would be semver-major and old JSON readers would reject it. A `capture_origin: third_party` field lets `kernel_layer: complete` mean complete as captured without claiming that the local runner had the observing vantage. No smaller Rust-minor encoding lets a reader of `ObservationHealth` itself distinguish that record from a locally observed complete capture.

The field is JSON-additive: missing `capture_origin` deserializes as `own`, and `own` is skipped while serializing, so records emitted by existing producers keep their existing JSON keys. Unknown wire values deserialize to `CaptureOrigin::Unsupported`, never `own`.

### Semver measurement
Measured on commit `0b9e0c1087ed98309119ec14f717140d827ee299` in worktree `/tmp/assay-capture-origin-f891` with `cargo-semver-checks 0.50.0` and `rustc 1.96.0`:

```text
cargo semver-checks check-release -p assay-runner-schema --baseline-rev v5.4.0
exit 100
```

Baseline tag: `v5.4.0` (resolved via `git tag --list 'v[0-9]*' --sort=-v:refname | head -n1`).

- JSON compatibility: yes — additive field, absent reads as `own`, and `own` is not emitted.
- Rust API compatibility: no — `ObservationHealth` is a public exhaustively constructible struct and this adds its public `capture_origin` field.
- Gate finding: `constructible_struct_adds_field` for `ObservationHealth.capture_origin`.

No version bump, lint suppression, dummy private field, or `#[non_exhaustive]` change is used to hide that measured Rust API break.

### Discriminating test
The test was added before the field. It constructs the ambiguous record `kernel_layer: complete` + `capture_origin: third_party`, deserializes then reserializes it, and checks that the third-party statement survives.

```text
cargo test -p assay-runner-schema third_party_complete_kernel_capture_survives_json_roundtrip
main-equivalent (test-only state before the field): exit 101
after the field: exit 0
```

The red failure was `left: Null`, `right: "third_party"`: Serde ignored the unknown field, which would have made the record appear locally observed after reserialization. Coverage also verifies legacy JSON defaults to `own` without emitting the key, and unknown origin values are not treated as `own`.

### Shared non-upgrade rule
[...]

Issue #2446, opened 2026-08-17
[FEAT] KernelLayerStatus cannot state a third-party kernel capture


Rul1an:
### Is your feature request related to a problem?

The source-class ladder in `assay-evidence` distinguishes six classes, including `third_party_observed` (`crates/assay-evidence/src/coding_agent.rs:53`, ladder notes at `:99`). The runner observation-health vocabulary cannot say it: `KernelLayerStatus` is `Complete | PartialRingbufDrops | Absent` (`crates/assay-runner-schema/src/health.rs:72-76`).

A kernel-layer record produced by an external observer, for example another host's eBPF capture or a third-party monitor whose output an adapter would translate, therefore has no faithful encoding on the kernel layer. `Absent` drops the evidence; `Complete` claims a vantage the local runner did not have. The class exists one crate over and is lost exactly where kernel-vantage evidence would enter.

This is a record-shape question, not an ingestion feature: deciding how the statement is spelled before any adapter produces one is the cheap moment to do it, for the same reason as #2422.

### Describe the solution you'd like

The smallest change that lets a kernel-layer health record state whose capture it was:

1. **An optional accompanying field** on the kernel-layer health record (for example `capture_origin: own | third_party`), absent reading as `own`, which is what every record produced today means. This avoids touching the enum.
2. Alternatively **a new variant**. Note `KernelLayerStatus` is a `pub` enum without `#[non_exhaustive]` in a published crate, so a variant is a semver-major change, which is why option 1 is listed first.

Either way, downstream classification must not silently upgrade: a third-party capture should be able to carry its own completeness (`complete`-as-captured) without being read as the local runner's vantage.

### Describe alternatives you've considered

- **Encoding third-party captures as `Absent` plus a side note.** Drops the evidence from the health calculus; a reader of the health record alone concludes nothing was observed.
- **Encoding them as `Complete`.** Claims local vantage for a foreign capture, which is the upgrade the source-class ladder exists to prevent.
- **Waiting for a concrete adapter.** The vocabulary decision is independent of any adapter and cheaper before one exists.

### Scope boundaries

- Vocabulary only. No ingestion, no adapter, no change to how any existing record is produced or classified.
- Absence of the new statement must mean what today's records mean.
- No claim that any specific external capture format should be supported.

Issue #2422, opened 2026-08-16
[FEAT] A session finding says the run finished, not that the trace is whole


Rul1an:
**Status, 2026-09-09.** Option 1 below shipped in #2534 (2026-08-19): `extent` makes no fidelity claim and `complete` must not be read as "nothing is missing". Option 2, the producer-declared coverage field keyed on the existing `CodingAgentCoverageState` vocabulary, is held for the next major of `assay-evidence` together with `#[non_exhaustive]`; this issue stays open for that. Option 3 is undecided.

### Is your feature request related to a problem?

`TraceExtent` answers one precise question, and answers it well: may more calls still arrive. `Complete` means the run is over, `Partial` means the window may still close later (`crates/assay-core/src/sequence_eval.rs:37-51`), and `PayloadSessionFinding.extent` documents it the same way, as "finished (`complete`) or may still grow (`partial`)" (`crates/assay-evidence/src/types.rs:434-445`). That is a temporal claim. This request does not ask to change it.

Nothing in a session finding speaks to a second and independent question: was the call sequence evaluated a faithful record of the session, or one that an upstream context manager already narrowed? `spanned` carries which call indices the rule read and invites a consumer to recompute from them (`types.rs:422-433, 467-471`), but recomputation reproduces the verdict over the same input, so it inherits whatever the input already lost.

That distinction used to be mostly theoretical, because a trace was a recording. Two independent 2026 measurements suggest lossy mid-session context is now an ordinary production condition rather than an edge case:

- Shiyang Chen, *Governance Decay: How Context Compaction Silently Erases Safety Constraints in Long-Horizon LLM Agents*, arXiv:2606.22528 (v1 submitted 2026-06-21). ConstraintRot benchmark, 1,323 episodes across seven model families with deterministic tool-call evaluation: safety-violation rate 0% with the policy in full context, 30% after compaction and up to 59% for some models; where constraints survived summarization violations stayed at 0%, where they were dropped they reached 38%.
- Zhiqi Wang, Yichi Zhang, Dongwon Lee, Yuchen Yang, *Lost in Compaction: Evaluating Side-Constraint Loss under Context Compaction*, arXiv:2608.11242 (submitted 2026-07-31). Reports that compactors retain 17% of injected session constraints on average across multi-turn chat, agentic trajectory, and long-horizon research, with retention varying sharply by compactor, prompt, context length, phrasing, and injection location.

[...]

PR #2535, opened 2026-08-19
feat(sim): state the phases a tier never attempts

### LINKED ISSUE #2170: sim report states bypassed=0 without saying what the tier never attempted
## What I measured

Running the quick suite against a real bundle at `64870f77`:

```
$ assay sim run --suite quick --seed 42 --target tests/fixtures/evidence/test-bundle.tar.gz --report sim.json
SUMMARY: blocked=8 passed=1 bypassed=0
✅ All attacks blocked.
```

and the machine-readable report:

```json
{"suite": "Quick", "seed": 42,
 "summary": {"total": 9, "passed": 1, "blocked": 8, "bypassed": 0, "failed": 0, "errors": 0}}
```

`total: 9` is the count of what **ran**. Nothing in the report or the console line states what the tier did not attempt.

## Why that is a defect rather than a nicety

`crates/assay-sim/src/suite.rs:173` gates the whole chaos phase on `matches!(cfg.tier, SuiteTier::Chaos)`, so a Quick run never attempts it. A reader of `bypassed=0` and `✅ All attacks blocked` has no way to tell whether that covers every attack the harness knows or the subset this tier selected, and exit code 0 means "clean" either way.

This is the rule this repository already applies to evidence bundles, turned on our own security report: coverage is not a property of the statement, and a denominator that counts only what was attempted reads as completeness to anyone who did not open `suite.rs`.

## The sharp part: the mechanism already exists and is already correct

`crates/assay-sim/src/report.rs:16` declares

```rust
/// Phases skipped when time budget exceeded
#[serde(skip_serializing_if = "Vec::is_empty")]
pub skipped_phases: Vec<String>,
```

and it serializes. But `set_time_budget_exceeded` (`report.rs:60`) is its only writer, called from five places in `suite.rs` (`:104`, `:132`, `:160`, `:183`, `:223`), all of them budget exhaustion.

So the report already declares "I skipped chaos because I ran out of time" and stays silent on "I skipped chaos because this tier does not include it". The same phase, unrun for two different reasons, is declared in one case and invisible in the other. That asymmetry is the bug; the vocabulary to fix it is already in the struct.

## Suggested shape

Record tier-excluded phases through the existing field (or a sibling that distinguishes *excluded by tier* from *skipped under budget*, since a consumer should be able to tell a deliberate scope from a truncated run), and state the denominator alongside the counts so `bypassed=0` is readable as `0 bypassed of N attempted, of M available`.

Deliberately not proposed here: changing exit-code semantics, or adding attacks to Quick. The claim should get honest, not louder.

## Relationships

- Related to #2167. The same report is also one of the unnamed JSON documents in that issue's inventory — it carries no `schema` and no `schema_version` in either form, while CI branches on its exit code. Identity and denominator are separate defects on one document; #2167 owns the first, this issue owns the second.
- Found while running the suite against the current state of #2159 / #2169, not caused by that work.

Rul1an:
Closes #2170

## Measured defect

`SimSummary.total` counts what **ran**, so a tier that omits a whole phase and a tier that ran
everything produce the same shape of clean report. The quick suite prints `bypassed=0` and
`All attacks blocked` while never attempting the chaos phase, which `run_suite` gates on
`matches!(cfg.tier, SuiteTier::Chaos)`.

Measured on `8e8a7645e`, quick suite against `tests/fixtures/evidence/test-bundle.tar.gz`:

| | before | after |
|---|---|---|
| console | `✅ All attacks blocked.` | `✅ All attempted attacks blocked. Not attempted by this tier: chaos.` |
| JSON | *(no such field)* | `"phases_not_attempted": ["chaos"]` |

Counts are unchanged in both: `blocked=8 passed=1 bypassed=0`, `total=9`.

## Deliberately not merged with `skipped_phases`

A phase absent because the tier never includes it is a **programme** statement. A phase dropped
because the time budget ran out is a **degradation**. A reader deciding whether a `bypassed=0` run
is reassuring needs to tell those apart, and one list would destroy the distinction. `skipped_phases`
keeps its existing meaning and stays paired with `time_budget_exceeded`.

## Controls

- `tiers_that_skip_chaos_declare_it` pins both directions, including that the Chaos tier declares
  no omission — otherwise the field could be filled unconditionally and still "pass".
- `chaos_phase_gate_matches_the_declared_omission` pins the declaration to the gate that causes it.
  Widening the gate to `SuiteTier::Chaos | SuiteTier::Stress` while leaving the declaration stale
  **fails** it.

That second control searches only the source **above** `#[cfg(test)]`. It lives in the file it
reads, and my first version searched the whole file — which its own literal satisfied, so the
assertion held even with the real gate mutated. That is exactly the failure it exists to prevent,
and it is worth stating plainly rather than quietly fixing: a guard that reads a file containing
itself must exclude itself, or it is asserting about its own copy.

## Verification

`cargo test -p assay-sim` green (39 lib tests), `cargo clippy --all-targets` clean, `cargo fmt` clean.

## Related

One of three issues sharing a single question — a status vocabulary that reports what happened but
cannot state what was never observed. #2422 (session-finding fidelity) and #2446 (`KernelLayerStatus`
cannot say a capture was third-party) are the other two, in flight as #2534 and #2533.

### DIFF
--- a/crates/assay-cli/src/cli/commands/sim.rs
+++ b/crates/assay-cli/src/cli/commands/sim.rs
@@ -155,7 +155,16 @@ fn cmd_run(args: SimRunArgs) -> Result<i32> {
         return Ok(2);
     }
 
-    println!("\n✅ All attacks blocked.");
+    if report.phases_not_attempted.is_empty() {
+        println!("\n✅ All attacks blocked.");
+    } else {
+        // "All attacks blocked" over a tier that skips a phase invites the reading
+        // that everything was tried. Name the untried phases in the same breath.
+        println!(
+            "\n✅ All attempted attacks blocked. Not attempted by this tier: {}.",
+            report.phases_not_attempted.join(", ")
+        );
+    }
     Ok(0)
 }
 
--- a/crates/assay-sim/src/report.rs
+++ b/crates/assay-sim/src/report.rs
@@ -14,6 +14,15 @@ pub struct SimReport {
     /// Phases skipped when time budget exceeded
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub skipped_phases: Vec<String>,
+    /// Phases this tier does not run at all, named so a clean summary cannot be
+    /// read as "everything was tried".
+    ///
+    /// Deliberately not merged with `skipped_phases`: a phase absent because the
+    /// tier never includes it is a programme statement, and a phase dropped
+    /// because the budget ran out is a degradation. A reader deciding whether a
+    /// `bypassed=0` run is reassuring needs to tell those apart.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub phases_not_attempted: Vec<String>,
 }
 
 #[derive(Debug, Serialize, Clone, Default)]
@@ -54,7 +63,16 @@ impl SimReport {
             results: Vec::new(),
             time_budget_exceeded: false,
             skipped_phases: Vec::new(),
+            phases_not_attempted: Vec::new(),
         }
+    }
+
+    /// Record the phases this tier never runs.
+    ///
+    /// `total` counts what ran, so without this a tier that omits a whole phase
+    /// and a tier that ran everything produce the same shape of clean report.
+    pub fn set_phases_not_attempted(&mut self, not_attempted: Vec<String>) {
+        self.phases_not_attempted = not_attempted;
     }
 
     pub fn set_time_budget_exceeded(&mut self, skipped: Vec<String>) {
--- a/crates/assay-sim/src/suite.rs
+++ b/crates/assay-sim/src/suite.rs
@@ -73,8 +73,25 @@ impl TimeBudget {
     }
 }
 
+/// The phases a tier never runs.
+///
+/// Kept beside `run_suite`'s phase gates rather than derived from them, because the
+/// gates are `if` statements over a tier and cannot be enumerated. The pairing is
+/// pinned by `chaos_phase_gate_matches_the_declared_omission`, so a new gate that
+/// forgets this list fails a test instead of quietly overstating the run.
+fn phases_not_attempted_for(tier: &SuiteTier) -> Vec<String> {
+    match tier {
+        SuiteTier::Chaos => Vec::new(),
+        SuiteTier::Quick | SuiteTier::Nightly | SuiteTier::Stress => vec!["chaos".to_string()],
+    }
+}
+
 pub fn run_suite(cfg: SuiteConfig) -> Result<SimReport> {
     let mut report = SimReport::new(&format!("{:?}", cfg.tier), cfg.seed);
+    // The chaos phase is gated on the Chaos tier below. Naming it here keeps the
+    // statement next to the gate that causes it, so a new gate that forgets to
+    // update this list is a visible omission rather than a silent one.
+    report.set_phases_not_attempted(phases_not_attempted_for(&cfg.tier));
     let budget = TimeBudget::new(Duration::from_secs(cfg.time_budget_secs));
     let limits = cfg
         .verify_limits
@@ -258,5 +275,52 @@ fn run_chaos_phase(report: &mut SimReport, seed: u64, budget: &TimeBudget) {
                 duration_ms: 0,
             });
         }
+    }
+}
+
+#[cfg(test)]
+mod not_attempted_tests {
+    use super::*;
+
+    /// The tiers that skip the chaos phase must say so, and the tier that runs it must not.
+    ///
+    /// This is the control for #2170: before it, a Quick run reported `bypassed=0` over a
+    /// programme that never attempted a whole phase, and a reader could not tell that from a
+    /// run that attempted everything.
+    #[test]
+    fn tiers_that_skip_chaos_declare_it() {
+        for tier in [SuiteTier::Quick, SuiteTier::Nightly, SuiteTier::Stress] {
+            assert_eq!(
+                phases_not_attempted_for(&tier),
+                vec!["chaos".to_string()],
+                "{tier:?} does not run the chaos phase and must declare it"
+            );
+        }
+        assert!(
+            phases_not_attempted_for(&SuiteTier::Chaos).is_empty(),
+            "the Chaos tier runs every phase, so it declares no omission"
+        );
+    }
+
+    /// Pins the declaration to the gate that causes it.
+    ///
+    /// `run_suite` gates the chaos phase on `matches!(cfg.tier, SuiteTier::Chaos)`. If that gate
+    /// moves, this reads the source and fails, rather than leaving the declaration describing a
+    /// programme the code no longer runs.
+    ///
+    /// Only the code above `#[cfg(test)]` is searched. This test lives in the file it reads, so a
+    /// whole-file search would be satisfied by this test's own literal — the assertion would hold
+    /// even after the real gate moved, which is precisely the failure it exists to prevent.
+    #[test]
+    fn chaos_phase_gate_matches_the_declared_omission() {
+        let source = include_str!("suite.rs");
+        let production = source
+            .split_once("#[cfg(test)]")
+            .expect("suite.rs keeps its test module")
+            .0;
+        assert!(
+            production.contains("if matches!(cfg.tier, SuiteTier::Chaos) {"),
+            "the chaos gate moved; phases_not_attempted_for must be updated with it"
+        );
     }
 }


Rul1an:
## Review (blocking)

Independent read of `2eb1f11754` against #2170, the existing `skipped_phases` / `run_suite` gates, and the #2422 / #2534 split. I did not run the suite, the claimed mutant, or clippy.

**Verdict: block.**

---

### 1. The sibling is not the #2422 split. Fix `skipped_phases`.

This is the design call. I am not splitting it.

#2422's axes are independent properties of **one record**. `TraceExtent` is temporal (may more calls still arrive). Coverage/fidelity is whether the evaluated set was whole. Both can be true at once: a session can be finished **and** compacted. Inferring one from the other is wrong in both directions. That is why #2534 reused `CodingAgentCoverage*` as a sibling and refused a third `TraceExtent` variant.

#2170 is a different shape. "Not in the programme" and "dropped because the budget ran out" are two **reasons a phase did not run**. They cannot both be true of the same phase. A Quick run that never scheduled chaos did not then drop chaos under budget. A Chaos run that hit the clock did not omit chaos as a programme statement. That is a partition of one set (unrun phases), not two orthogonal axes.

So `phases_not_attempted: Vec<String>` next to `skipped_phases: Vec<String>` is a second definition of "phases that did not run." #2170 already said the vocabulary is in the struct, and that the emptiness of `skipped_phases` on a clean Quick is the overclaim. A sibling leaves that field still empty, still paired with `time_budget_exceeded: false`, still readable as "nothing was skipped" to every consumer that already knows ADR-024 and does not know the new name.

Do not dump tier exclusions into `skipped_phases` as bare strings. That is the destruction you are trying to avoid: a clean Quick would look truncated, or a truncated Chaos would look like a smaller programme. Give the existing field a reason (`tier` | `budget`). One list, distinction preserved, empty means every phase the harness knows actually ran. That is the #2170 fix.

#2446 does not change this. Its absence rule is "today's records mean `own`." Different document, different absence. The cluster sentence ("a status says what happened; it cannot say what was never observed") is real and is not a licence to invent a second phase list.

---

### 2. The PR already collapses the distinction it was written to protect

Even if you keep the sibling, this implementation makes the two fields name the same phase.

`phases_not_attempted_for` always reports `"chaos"` for Quick / Nightly / Stress (`suite.rs` `phases_not_attempted_for`). `run_suite` then still writes `"chaos"` into `skipped_phases` on those same tiers:

- integrity `BudgetExceeded` → `["differential", "chaos"]`
- budget after integrity → `["differential", "chaos"]`
- budget after differential → `["chaos"]`

A Quick run that exceeds budget after integrity therefore serializes `phases_not_attempted: ["chaos"]` **and** `skipped_phases: ["differential", "chaos"]`. Chaos is defined twice, for two reasons, one of which is false. That is Ruley's overclaim made concrete: `skipped_phases` already lists a phase the tier never included. Adding a sibling without touching those writers is how a second field becomes a second definition.

The console path you did not write is this path. `cmd_run` returns exit 2 on `time_budget_exceeded` before the new sentence (`sim.rs`). The reader gets:

```
⏱ Time budget exceeded. Skipped: differential, chaos
```

and never "Not attempted by this tier: chaos." JSON has both lists; stdout has the lie.

---

### 3. Gate and declaration are still two sources of truth

`phases_not_attempted_for` is a hand-written `match` beside `if matches!(cfg.tier, SuiteTier::Chaos)`. The grep pins that a literal exists in the production half of the file. It does not pin that the match and the `if` stay the same predicate, and it does not pin that `run_suite` calls `set_phases_not_attempted`.

A shared function is actually better here and is not architecture:

```rust
fn tier_runs_chaos(tier: &SuiteTier) -> bool {
    matches!(tier, SuiteTier::Chaos)
}
```

Gate with it. Build the omission list from it. Delete the grep. A table of phases-per-tier is the same idea if a second gated phase appears; do not add it for one `if`.

---

### 4. Console sentence

`Not attempted by this tier: {}` plus `join(", ")` is honest for one name and for several. "chaos, other" reads as a list. Fine.

It is not honest on the path that matters for the collision: tier omission **and** budget. That path uses the old budget line and over-claims chaos as skipped. See §2.

`SUMMARY: blocked=8 passed=1 bypassed=0` is unchanged. #2170 asked for a denominator next to the counts (`0 bypassed of N attempted, of M available`). The footnote after the table does not fix a reader who stops at `SUMMARY`.

---

### 5. Controls: the self-grep is fixed; a second vacuous pass is not

`chaos_phase_gate_matches_the_declared_omission` splitting on the first `#[cfg(test)]` does exclude its own literal. I did not find a second copy of **that** mistake in this PR. The gate string does not appear in production comments above the test module.

The second vacuous control is the pair of new tests as a whole.

- `tiers_that_skip_chaos_declare_it` calls `phases_not_attempted_for` directly. It will pass after you delete `report.set_phases_not_attempted(...)`.
- The grep only asserts that a source string exists. Same deletion stays green.
- `crates/assay-sim/src/lib.rs` already has `test_quick_suite`, which runs `run_suite(SuiteTier::Quick)` and inspects the `SimReport`. It was not taught `phases_not_attempted`. That is the control that would have bound the defect #2170 measured. It still passes if the field is never set.

So: you caught the "test greps a file containing itself" hole, then left a hole of the same family. The tests describe the helper and a literal. They do not describe the report.

`"chaos"` is also copied into the helper, the unit test, and the three budget writers. No shared constant. The writers are how the collision in §2 is even possible.

---

### What would unblock

1. Do not ship a second `Vec<String>` for the same fact. Put programme-vs-budget on `skipped_phases` as a reason. Update the ADR-024 pairing so empty means "the catalogue ran," not "the clock did not fire."
2. If you override and keep the sibling anyway: make the lists a partition. The budget writers must not name a phase `phases_not_attempted_for` already named. Write the both-at-once console line. Bind a test to a `SimReport` from `run_suite(Quick)` (the existing one), including a Quick+budget case that must **not** list chaos under `skipped_phases`.
3. One predicate for the chaos gate and the declaration. Drop the source grep.

Until (1) or a complete (2), this does not close #2170. It adds a name and leaves the overclaim standing.

Rul1an:
## Blocking review accepted in full — I took option (1), not the override

This was the right call and §2 is a finding I would not have reached. Rewritten rather than patched.

### §2 — the two fields named the same phase

You were right, and I verified it before changing anything: `run_suite` writes `"chaos"` into the
budget list on Quick, Nightly and Stress, tiers whose programme never contained it. So a Quick run
that also exceeded budget serialized `phases_not_attempted: ["chaos"]` **and**
`skipped_phases: ["differential", "chaos"]`. My sibling field did not remove Ruley's overclaim — it
stood a second name beside it.

Took your option (1). One list, each entry carrying a reason:

```json
"skipped_phases": [{"phase": "chaos", "reason": "not_in_tier"}]
```

`set_time_budget_exceeded` refuses to restate a phase already recorded `not_in_tier`, so the reasons
partition **by construction** rather than by convention. `budget_does_not_restate_a_phase_the_tier_never_ran`
pins exactly the collision you described.

Cheap because there was one consumer: the console. Worth noting that I checked that before choosing
(1) over (2) — had `skipped_phases` been read anywhere else, the shape change would have been the
expensive option and your (2) the pragmatic one.

### §2 — "JSON has both lists; stdout has the lie"

Correct, and the sharpest sentence in the review. `cmd_run` returns exit 2 before reaching my
sentence, so the budget path printed one list that silently included phases the tier never ran. It
now prints both facts separately:

```
⏱ Time budget exceeded. Dropped by the clock: differential
   Not in this tier's programme: chaos
```

### §3 — one predicate, grep deleted

`tier_runs_chaos` is now read by `run_suite`'s gate and by the programme statement.
`CHAOS_PHASE` / `DIFFERENTIAL_PHASE` replace the copies. The source-grep test is gone — you were
right that it pinned the existence of a string, not the agreement of two predicates, and a shared
function is both simpler and stronger. That is the trade I should have made first.

### §5 — the second vacuous control

This is the one I most needed told. I fixed "a test that greps a file containing itself" and then
shipped a hole of the same family: every test called the helper directly, so deleting
`set_phases_not_in_tier` from `run_suite` left them all green. The tests described the helper, not
the report.

`quick_report_states_the_phase_its_tier_never_runs` now asserts on a `SimReport` that
`run_suite(Quick)` actually produced. I ran your attack:

| delete `set_phases_not_in_tier` from `run_suite` | result |
|---|---|
| before | all tests green |
| after | **FAILED** |

### §4 — the denominator

Taken. `SUMMARY` now reads
`blocked=8 passed=1 bypassed=0 of 9 attempted (phases not in this tier: chaos)`.
You were right that a footnote after the table does not reach a reader who stops at the counts, and
that reader is the one #2170 measured.

### On your closing line

"It adds a name and leaves the overclaim standing" was accurate and is the sentence I would have
needed to hear even if the rest had been wrong. 40 tests pass, clippy and fmt clean. Pushed.

Issue #2526, opened 2026-08-19
ci(golden-path): published-release journey produces a v1 denial record and verifies it under v0


Rul1an:
## Measured on the v5.4.0 tag

Release run https://github.com/Rul1an/assay/actions/runs/32229066881, tag `v5.4.0`,
commit `bbb5e7fe4b03bc6160d18e2966e75a7586c062ef`.

Job "Verify the published release journey / Linux x86_64 post-publication journey" failed at
step "Exercise the attested published release". Command 12 of 12 (`verify-produced-bundle`)
exited 2. Every other release job succeeded.

From the retained evidence artifact (`published-release-golden-path-v5.4.0-...`), `verify.json`:

```json
{
  "bundle_integrity": "pass",
  "verdict": "invalid",
  "reason_code": "E_EVIDENCE_PROFILE_INVALID",
  "findings": [{
    "id": "unknown_profile_schema",
    "detail": "payload schema \"assay.denied_call_observation.v1\" is inside the profile namespace but is not a recognized privileged-mcp-action/v0 record; unknown fails closed"
  }]
}
```

## Cause

`scripts/ci/published-release-golden-path.sh` produces with `--denied-observations`, so from
v5.4.0 the bundle carries the `assay.denied_call_observation.v1` record introduced by #2521.
The verification step (~line 280) then runs:

```
assay evidence verify-privileged-mcp-action "$bundle" --format json
```

with no `--profile-version`, so it verifies under the default `v0`, which correctly fails closed
on a v1 record it does not recognize.

On `v5.3.0` this passed only because the proxy did not emit that record. The harness was written
in #2513 against a tree where the pairing could not arise.

CHANGELOG 5.4.0 states the intended pairing explicitly: "Live produce verifies with
`--profile-version v1`; the verifier default stays `v0`."

## This is a harness defect, not a release defect

The published v5.4.0 artifacts are sound and were verified independently:

- 23 assets, exact name match against the renamed v5.3.0 baseline, no extras or duplicates
- all 11 `.sha256` companions verify against downloaded bytes
- attestation shape matches v5.3.0 (10 attested; `release-proof-kit.tar.gz` and
  `release-provenance.json` carry none, as on v5.3.0)
- all 15 publishable crates published at 5.4.0; PyPI and MCP Registry succeeded

Failing closed on an unrecognized in-namespace schema is the designed behaviour and is working.

## Acceptance

- The journey verifies the bundle it actually produced. Choose deliberately between verifying with
[...]

PR #2536, opened 2026-08-19
fix(ci): let the privileged-action-gate example say why it failed

Rul1an:
The "Demo smoke (privileged-action-gate)" check reached CI as a bare `Process completed with exit code 1` with **nothing else in the log**. Two independent maskings produced that, and both are fixed here.

## Why this is worth a PR even though the red is gone

[#2528](https://github.com/Rul1an/assay/pull/2528) (merged 09:38 today) already cleared the specific failure by adding `--profile-version v1` to the denied-bundle `matrix` call. That fixed **one pairing, not the class**. The maskings below are still on `main`, so the next example that reaches a non-valid verdict produces the same silent exit-1 with no diagnosis, and the next person spends the same twenty minutes finding out why.

## The two maskings

**`verify.sh` discarded the evidence.** It ran `./run.sh 2>/dev/null | python3 ...` under `set -euo pipefail`. When `run.sh` failed, `pipefail` plus `set -e` aborted before anything printed, and `run.sh`'s stderr had already gone to `/dev/null`. It now reads the exit code directly instead of through the pipe, captures stderr to a temp file, and replays it on failure.

**`run.sh`'s `matrix()` assumed a conditional field.** `Report.claims` is `Option<Claims>` behind `skip_serializing_if`, doc-commented *"Present only when `verdict` is `valid`"*. The verifier refuses to print a claim matrix it could not recompute — that is the profile's fail-closed posture, so the example is the side that was wrong, and the absence assertions in `verify_privileged_mcp_action.rs` (lines 992, 1023, 1053) are deliberate.

There was a **second trap underneath the first**: `cmd_verify_privileged_mcp_action` returns exit 2 on an invalid verdict, and `matrix()` ran the CLI in a pipeline under `set -o pipefail`. Teaching the parser to tolerate a missing key would therefore not have been enough — the CLI's own exit 2 would still have killed `run.sh`. `matrix()` now reads the report from a file rather than a pipe, which is what lets `|| status=$?` observe the code instead of having `pipefail` eat the script.

## What it prints now

Instead of silence:

```
denied path bundle : NO CLAIM MATRIX  bundle_integrity=pass  verdict=invalid  reason_code=E_EVIDENCE_PROFILE_INVALID  (verifier exit 2)
  unknown_profile_schema: payload schema "assay.denied_call_observation.v1" is inside the profile namespace but is not a recognized privileged-mcp-action/v0 record; unknown fails closed
```

## Verification

- `bash examples/privileged-action-gate/verify.sh` → exit 0
- **Control on the mutated line:** reintroducing the #2528 defect (dropping `--profile-version v1`) now yields the diagnosis above instead of silence. That is the exact input that produced the bare "exit code 1" in CI.
- Two further absence paths exercised: a stage-1 integrity failure (`verdict=absent`, `E_EVIDENCE_UNREADABLE`), and a report that is not JSON at all.
- `shellcheck` clean; `pre-commit` on both files exits 0; `check-published-release-golden-path-contract.py` OK — it parses `run.sh` and asserts `matrix()` still forwards `"$@"`, and the rewrite keeps that on one line so the guard still binds.
- **CI:** "Demo smoke (privileged-action-gate)" dispatched on this branch → success, [run 32241737989](https://github.com/Rul1an/assay/actions/runs/32241737989), step output `OK: privileged-action-gate verdicts match expected-output.txt`.

One thing stated rather than hidden: `scripts/ci/test_published_release_proxy_phase.py` failed once mid-session with `proxy process failed to start: [Errno 1] Operation not permitted`. That was a transient local sandbox spawn failure, not this change — it passes 4/4 on the patched tree.

### DIFF
--- a/examples/privileged-action-gate/run.sh
+++ b/examples/privileged-action-gate/run.sh
@@ -118,17 +118,53 @@ capture() {
 }
 
 # matrix <label> <bundle> [verify-args...]: verify one bundle and print a compact claim-matrix line.
+# The verifier carries "claims" only behind a valid verdict: a stage-1 integrity failure or a
+# rejected statement omits the member and exits 2, because a matrix it could not recompute is not
+# a matrix it will print. The report is read from a file rather than piped, so neither the exit
+# code (pipefail) nor the absent member (KeyError) can end the demo without saying why.
 matrix() {
   local label="$1" bundle="$2"
   shift 2
+  local report="$WORK/verify-report.json" status=0
   "$ASSAY_CLI" evidence verify-privileged-mcp-action "$bundle" --format json "$@" \
-    | "$PY" -c '
+    >"$report" || status=$?
+  "$PY" - "$label" "$report" "$status" <<'PYEOF'
 import json, sys
-label = sys.argv[1]
-report = json.load(sys.stdin)
-cells = " ".join("{}={}".format(name, cell["status"]) for name, cell in report["claims"].items())
+
+label, path, status = sys.argv[1], sys.argv[2], sys.argv[3]
+try:
+    with open(path) as handle:
+        report = json.load(handle)
+except (OSError, ValueError):
+    print(f"  {label}: the verifier produced no JSON report (exit {status})", file=sys.stderr)
+    raise SystemExit(1)
+
+claims = report.get("claims")
+if claims is None:
+    # Absent by design, never by omission. Print the verifier's own diagnosis: which stage refused,
+    # under which reason code, and every finding it recorded.
+    print(
+        "  {}: NO CLAIM MATRIX  bundle_integrity={}  verdict={}  reason_code={}  (verifier exit {})".format(
+            label,
+            report.get("bundle_integrity", "?"),
+            report.get("verdict", "absent"),
+            report.get("reason_code", "?"),
+            status,
+        ),
+        file=sys.stderr,
+    )
+    for finding in report.get("findings", []):
+        print("    {}: {}".format(finding.get("id", "?"), finding.get("detail", "")), file=sys.stderr)
+    print(
+        "    The claim matrix is recomputed only behind a valid verdict. A profile mismatch is the"
+        " usual cause: pass the --profile-version matching the records this demo produced.",
+        file=sys.stderr,
+    )
+    raise SystemExit(1)
+
+cells = " ".join("{}={}".format(name, cell["status"]) for name, cell in claims.items())
 print("  {}: verdict={}  {}".format(label, report["verdict"], cells))
-' "$label"
+PYEOF
 }
 
 echo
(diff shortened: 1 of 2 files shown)


Rul1an:
## Verified independently at `5f129de63`

This is the follow-up I flagged on #2526 and deliberately left out of #2528, so it is good to see it
taken as its own change rather than folded in.

Built `assay-mcp-server` from this head and ran the control myself rather than reading the claim:
reintroduced the exact #2528 defect by dropping `--profile-version v1` from the denied-bundle
`matrix` call.

**Before this PR** that input produced a bare `Process completed with exit code 1` in CI with
nothing to read. **On this head** it produces:

```
FAIL: run.sh exited 1. Its stderr follows:
  denied path bundle : NO CLAIM MATRIX  bundle_integrity=pass  verdict=invalid
    reason_code=E_EVIDENCE_PROFILE_INVALID  (verifier exit 2)
    unknown_profile_schema: payload schema "assay.denied_call_observation.v1" is inside the
    profile namespace but is not a recognized privileged-mcp-action/v0 record; unknown fails closed
    The claim matrix is recomputed only behind a valid verdict. A profile mismatch is the usual
    cause: pass the --profile-version matching the records this demo produced.
```

Baseline on the unmutated head: `OK: privileged-action-gate verdicts match expected-output.txt`,
exit 0.

That output names the verdict, the reason code, the finding, **and the remedy**. The last part is
what makes it a real fix rather than a louder failure.

## What I think is right about the framing

**Fixing the class, not the instance, and saying so.** #2528 fixed one pairing. The two maskings
were still on `main`, so the next example to reach a non-valid verdict would have produced the same
silent exit 1 and cost the next person the same twenty minutes. Opening this even though the red was
already gone is the correct call.

**Identifying the example as the wrong side.** `Report.claims` is `Option<Claims>` behind
`skip_serializing_if`, documented "Present only when `verdict` is `valid`". The verifier refusing to
print a claim matrix it could not recompute is the profile's fail-closed posture working. Teaching
the parser to tolerate a missing key would have been fixing the symptom on the wrong side of the
contract.

**The second trap underneath the first.** `cmd_verify_privileged_mcp_action` exits 2 on an invalid
verdict, and `matrix()` ran the CLI in a pipeline under `set -o pipefail` — so tolerating the
missing key alone would still have killed `run.sh`. Reading the report from a file rather than a
pipe is what lets `|| status=$?` observe the code. Finding the second trap before shipping the first
fix is the part that would have been easy to miss.

**Keeping `matrix()`'s `"$@"` forwarding on one line** so
`check-published-release-golden-path-contract.py` still binds. That contract parses `run.sh`, and a
rewrite that broke the parse would have traded one silent gate for another.

Also noted and appreciated: the transient
`test_published_release_proxy_phase.py` spawn failure is stated rather than hidden, with the
evidence that it passes 4/4 on the patched tree.

No change requests. `BEHIND` only — it needs a branch update once the current merge queue drains.

PR #2537, opened 2026-08-19
docs(conformance): state which ACM badge a clean-room run earns, and the one it cannot

Rul1an:
The claim ceiling bounded what a match proves but never said **which kind of reproduction** a report may call itself. The two [ACM](https://www.acm.org/publications/policies/artifact-review-and-badging-current) terms are the reverse of the common intuition, so leaving it unstated invites an overclaim:

- **Results Reproduced** — a different team obtained the result **using** the author's artifacts
- **Results Replicated** — **without** them

A blind from-spec run against this pack earns **Results Reproduced**. The implementation is the reproducer's own and imports nothing from Assay, but the opaque cases, the descriptor and the specification are all author-supplied.

## The part worth adding, not just the mapping

**Results Replicated is unreachable against any conformance corpus**, and that is a property of corpora rather than a shortfall of a reproducer. ACM assumes the author-supplied artifact is the author's *code*, so obtaining the result without it is meaningful. A corpus inverts that: the corpus **is** the artifact, and no reproduction can avoid using it. A report claiming *Replicated* here has either misread the badge or not used the pack.

So the fact that carries the information has no ACM badge, and the section says to state it in full instead: **an independently written implementation, run against author-supplied cases, deriving every outcome from the specification alone.** That is more checkable than either badge, and it is what `reproduction_mode: blind_from_spec` already claims.

## Scope

Docs only, one file, one section appended under the existing claim ceiling. No change to the scorer, the corpus, the digest or the pinned release. The pinned protocol at `0c055ead` that [#1840](https://github.com/Rul1an/assay/issues/1840) links is unaffected; this lands for the next candidate. There are currently zero in-flight reproduction attempts, so no one is mid-run under the older text.

### DIFF
--- a/conformance/privileged-mcp-action-v0/CONFORMANCE-PROTOCOL.md
+++ b/conformance/privileged-mcp-action-v0/CONFORMANCE-PROTOCOL.md
@@ -217,3 +217,28 @@ a match is bounded further by the result cited under Authorship method: agreemen
 correlated on the profile's ambiguous regions, which is where a corpus carries its information. A
 report may state agreement; it may not state that agreement was reached independently unless the
 authorship methods make that a claim someone can check.
+
+### Which badge a run earns, and the one it cannot
+
+Reports here carry an [ACM Artifact Review and Badging](https://www.acm.org/publications/policies/artifact-review-and-badging-current)
+class, because the distinction between kinds of reproduction is easy to blur and
+ACM already drew it. The two terms are the reverse of the common intuition:
+*Results Reproduced* means a different team obtained the result **using** the
+author's artifacts; *Results Replicated* means **without** them.
+
+A blind from-spec run against this pack earns **Results Reproduced**. The
+implementation is the reproducer's own and imports nothing from Assay, but the
+opaque cases, the descriptor and the specification all come from the author.
+
+**Results Replicated is not reachable against this or any conformance corpus, and
+that is a property of corpora rather than a shortfall of any reproducer.** ACM
+assumes the author-supplied artifact is the author's *code*, so obtaining the
+result without it is meaningful. A conformance corpus inverts that: the corpus is
+the artifact, and no reproduction can avoid using it. A report claiming
+*Replicated* against this pack has either misread the badge or not used the pack.
+
+So the fact worth stating has no ACM badge, and a report should state it in full
+rather than reach for a stronger label: **an independently written implementation,
+run against author-supplied cases, deriving every outcome from the specification
+alone.** That is a stronger and more checkable sentence than either badge, and it
+is what `reproduction_mode: blind_from_spec` is claiming.
--- a/conformance/privileged-mcp-action-v0/CONFORMANCE-PROTOCOL.md
+++ b/conformance/privileged-mcp-action-v0/CONFORMANCE-PROTOCOL.md
@@ -220,7 +220,9 @@ authorship methods make that a claim someone can check.
 
 ### Which badge a run earns, and the one it cannot
 
-Reports here carry an [ACM Artifact Review and Badging](https://www.acm.org/publications/policies/artifact-review-and-badging-current)
+**No badge here is conferred by anyone.** ACM badges are awarded by a venue's artifact-evaluation committee. What follows borrows ACM's *vocabulary* to describe a claim precisely; it is a self-description under a published definition, not an ACM award.
+
+Reports here describe themselves with an [ACM Artifact Review and Badging](https://www.acm.org/publications/policies/artifact-review-and-badging-current)
 class, because the distinction between kinds of reproduction is easy to blur and
 ACM already drew it. The two terms are the reverse of the common intuition:
 *Results Reproduced* means a different team obtained the result **using** the


Rul1an:
@coderabbitai review

Rul1an:
## Verified against the primary source — the terminology is right, including the trap

The load-bearing claim here is the direction of two ACM terms that are famously easy to invert, so I
checked it rather than recognising it.

| Badge (v1.1) | Meaning |
|---|---|
| **Results Reproduced** | different team, **using**, in part, author-supplied artifacts |
| **Results Replicated** | different team, **without** author-supplied artifacts |

ACM interchanged these in August 2020 to align with NISO; badges issued before 2026-05-14 are v1.0
and carry the old sense. Sources confirm the direction and confirm that it "runs opposite to how some
other fields use these terms colloquially", which is exactly the warning this doc opens with.

So `blind_from_spec` earning **Results Reproduced** is correct: the implementation is the
reproducer's own, and the cases, descriptor and specification are the author's — which is precisely
"using, in part, artifacts provided by the author".

## The argument I think is the valuable part

> *Results Replicated is not reachable against this or any conformance corpus, and that is a
> property of corpora rather than a shortfall of any reproducer.*

That is right and worth stating in the doc rather than leaving for a reader to work out. ACM's model
assumes the author-supplied artifact is the author's **code**, so obtaining the result without it is
meaningful. A conformance corpus inverts that — the corpus *is* the artifact, so no reproduction can
avoid it. A report claiming Replicated here has misread the badge or not used the pack.

Declining a stronger label and writing the checkable sentence instead —
*"an independently written implementation, run against author-supplied cases, deriving every outcome
from the specification alone"* — is the right instinct, and it is more auditable than either badge.

## One wording note, non-blocking

"Reports here carry an ACM Artifact Review and Badging **class**" is accurate as written, but a
hurried external reader could take it as "this project has been awarded an ACM badge". ACM badges are
conferred by a venue's artifact-evaluation committee, not self-assigned. The doc never claims
otherwise and the surrounding text makes the distinction, so this is a hardening suggestion rather
than a correction: something like "names the ACM class the run corresponds to" would close the gap
against a reader who quotes one line.

No change requests.

Rul1an:
## Verdict: Reading 1

A blind from-spec run that consumes the published pack is **Results Reproduced**. Opaque cases with answers withheld are still author-supplied **artifacts**. They are not a path to **Results Replicated**.

Reading 2 fails on the official wording. ACM does not restrict “artifact” to the author’s code or expected results. It names input datasets as artifacts. Withholding the oracle does not reclassify those inputs as mere stimulus, and it does not rewrite Replicated from “without author-supplied **artifacts**” into “without author-supplied **results**.”

## ACM v1.1 criteria relied on

Source: [Artifact Review and Badging Version 1.1 — 24 August 2020](https://www.acm.org/publications/policies/artifact-review-and-badging-current).

**What an artifact is**

> By “artifact” we mean a digital object that was either created by the authors to be used as part of the study or generated by the experiment itself. For example, artifacts can be software systems, scripts used to run experiments, **input datasets**, raw data collected in the experiment, or scripts used to analyze results.

**VIM-inspired terms (computational gloss)**

> **Reproducibility** (Different team, same experimental setup): … For computational experiments, this means that an independent group can obtain the same result **using the author’s own artifacts**.
>
> **Replicability** (Different team, different experimental setup): … For computational experiments, this means that an independent group can obtain the same result using artifacts **which they develop completely independently**.

**Results Validated badges**

> **Results Reproduced v1.1** — The main results of the paper have been obtained in a subsequent study by a person or team other than the authors, **using, in part, artifacts provided by the author**.
>
> **Results Replicated v1.1** — The main results of the paper have been independently obtained in a subsequent study by a person or team other than the authors, **without the use of author-supplied artifacts**.

**Artifacts Evaluated (scope, not the badge being claimed)**

> **Consistent:** The artifacts are relevant to the associated paper, and contribute in some inherent way to the generation of its main results.
>
> **Exercisable:** Included scripts and/or software used to generate the results in the associated paper can be successfully executed, and **included data can be accessed and appropriately manipulated**.

**v1.0 ↔ v1.1 term swap (NISO)**

The same page states that, after discussions with NISO, ACM **swapped the terms “reproducibility” and “replication”** while keeping the existing definitions, and updated prior badging for consistency. Pre-swap (v1.0) *Results Reproduced* was the “without author artifacts” badge; v1.1 assigned that sense to *Results Replicated*. The direction in this PR is the v1.1 direction.

I did not rely on the later illustrative paragraph that says “replication would involve exercise of software, typically software provided by the author.” That sentence still uses the pre-swap sense of “replication” and contradicts the v1.1 badge text above.

## (a) true

A from-spec implementation that never sees expected outcomes still uses, **in part**, artifacts provided by the author: the opaque cases (input datasets), and on this pack the descriptor. `#1840` withholds expected outcomes, semantic names, `gen_vectors.py`, Assay verifier code, and corpus pointers. That is exactly the clean-room move. It does not stop the run from using author-supplied artifacts. “Using, in part” is the test. (a) holds.

The specification is the publication, not an extra artifact. ACM treats artifacts as objects that add value *beyond* the article text. Counting the spec as an author-supplied artifact is a category error. It is unnecessary for (a): the cases already suffice.

## (b) conditional — not structurally true of every conformance corpus

**Named condition:** *Results Replicated* is unreachable **for a subsequent study that uses the published corpus (or any other author-supplied artifact) to obtain the result.**

That condition is already the true last sentence of this section: a report claiming Replicated *against this pack* has either misread the badge or not used the pack.

The shipped generalisation does not follow:

> Results Replicated is not reachable against this or any conformance corpus … ACM assumes the author-supplied artifact is the author’s *code* … A conformance corpus inverts that: the corpus is the artifact, and no reproduction can avoid using it.

Two failures:

1. **The premise is false.** ACM does not assume the artifact is the author’s code. The definition quoted above lists input datasets in the same breath as software. A corpus does not invert ACM. It *is* the input-dataset case ACM already named.
2. **“No reproduction can avoid using it” is circular.** It is true only if the claimed result is agreement with *that* corpus. ACM Replicability is a *different experimental setup*: independently developed artifacts. A different team that reads only the specification and builds its own cases and its own implementation can obtain the spec’s main results without author-supplied artifacts. Exact case-set identity is not required; the page says results need only stay inside a tolerance that does not change the main claims.

Do not read this as “usually unreachable.” If the study consumes the published corpus, Replicated is **not available**. If it does not, Replicated is **available**. That is a property of the study, not of the genre “conformance corpus.”

This matters for the Vaara contrast. ACM’s cut is “any author artifact” versus “none.” Published checkers over published vectors (Vaara) and an independent from-spec checker over published vectors (JM-Lab / a `#1840` pack run) are the **same badge**: Reproduced. ACM is not the instrument that separates those two. The prose sentence about an independently written implementation deriving every outcome from the spec is doing work the badges refuse to do. Treating Replicated as structurally impossible then claiming ACM captures the Vaara gap is sand: the badges do not draw that gap, and Replicated remains the label for a study that also refuses the author’s cases.

## What the JM-Lab rows earn

**Results Reproduced.** They ran against the `vectors.json` published in the kit. That is an author-supplied input dataset even if every `expected` was recomputed from the contract and the checker imported nothing from the kit. Labelling them Replicated would be Reading 2, and Reading 2 is not ACM’s test.

The four rows were not mis-downgraded. The kit’s strongest fact is still the sentence this PR already wants to write — independently written implementation, author-supplied vectors, expected derived from the spec — not a Replicated badge.

## What the three documents should say instead

Not a badge swap. A ceiling rewrite. Keep (a). Replace the unconditional (b).

**`CONFORMANCE-PROTOCOL.md` (this PR)** — keep: a blind from-spec run against this pack earns Results Reproduced, because the opaque cases are author-supplied input datasets. Replace the “not reachable against this or any conformance corpus / ACM assumes the artifact is code / no reproduction can avoid using it” block with: *Results Replicated is not available for a run that uses this pack. It would be available only for a subsequent study that obtains the profile’s main results without author-supplied artifacts — independently developed cases and implementation, specification only. A report claiming Replicated against this pack has misread the badge or not used the pack.*

**`REPRODUCTIONS.md` (rge-bench#23)** — keep the four JM-Lab rows as Results Reproduced. Replace “Replicated is not reachable here … property of conformance corpora … no reproduction can avoid using it” with the same condition: unreachable **for rows that consume the published `vectors.json`**; reachable only for a study that never uses that file (or any other author artifact) and still obtains the contract’s main results.

**`LAB-INDEX.yml` `_meta` (assay-tunnel-experiments#195)** — keep `externally-reproduced` = ACM Results Reproduced for items that used author-supplied vectors. Drop “Replicated is structurally unreachable there.” Say: *Replicated is not what `externally-reproduced` records, because that maturity value means a run that used the published corpus. It does not mean Replicated is impossible for the claim.*

## Not checked

- Did not download `privileged-mcp-action-v0-clean-room.tar.gz` or list its entries; `#1840` body and thread used for what the pack withholds.
- Did not read `JM-Lab/rge-bench-java` to confirm the checker never consults `expected`.
- Did not inspect Vaara desk runs; the contrast is taken as stated.
- Did not judge whether ACM is the right vocabulary. Only whether (a) and (b) are correct under it.

Rul1an:
## Reading 1 accepted, and the wording note is fixed in all three venues

The verdict holds and the argument for it is better than the one I shipped. I had the two readings turning on "artifact", and ACM settles it explicitly: **input datasets are artifacts**. So withholding the oracle does not demote the opaque cases to stimulus, and *Replicated* says "without author-supplied **artifacts**", not "without author-supplied **results**". I had quietly swapped those two words and then argued from the swap.

No change to (a) or (b). The claim ceiling stands as written.

## The wording note was not minor and it is now fixed

> "Reports here carry an ACM Artifact Review and Badging class" ... a hurried external reader could take it as "this project has been awarded an ACM badge".

Correct, and the risk is worse than cosmetic here: the entire reason for borrowing ACM's vocabulary is precision, so an ambiguity about **who awarded what** undermines the thing the vocabulary was adopted for. Fixed in all three venues, each now leading with the disclaimer rather than burying it:

- `Rul1an/assay#2537` -- commit `c3534af6a`
- `rge-bench/rge-bench#23` -- commit `c51c5af`
- `Rul1an/assay-tunnel-experiments#195` -- commit pushed

Wording: *No badge here is conferred by anyone. ACM badges are awarded by a venue's artifact-evaluation committee. This borrows ACM's vocabulary to describe a claim precisely; it is a self-description under a published definition, not an ACM award.*

## One thing your review caught indirectly

Running the rge-bench guards after the edit surfaced a stray parity-test manifest I had generated in that working tree. Its schema id is `assay.corpus_adequacy.manifest.v0`, and rge-bench's `check_public_surface.py` rejects vendor terms because the kit is deliberately neutral. The file was never committed and is removed, but it is a real design signal for the tool itself: a vendor-branded schema id in a peer's repository is the same envelope-ownership problem we track in other projects. That goes on the tool, not on this PR.

Thanks for reading the criteria rather than a summary of them. That is what made the difference between the two readings.

PR #2538, opened 2026-08-19
feat(conformance): vendor-neutral mutation adequacy for a published corpus

Rul1an:
Generalizes rge-bench's `check_rule_liveness.py` into a manifest-driven tool that measures **mutation adequacy of a published conformance corpus** — any corpus, not just ours. Standard library only, no Assay import, no network.

```
python3 conformance/corpus_adequacy.py conformance/adequacy/mcp-jsonrpc-id.manifest.json
```

## Why a corpus needs this and a test suite does not

Outcome coverage is not rule coverage. A corpus can reach every declared outcome while some rule never decides anything, because another rule reaches the same outcome first on every vector it would have caught.

A surviving mutant is therefore not a confidence gap. It means **an implementer can delete that rule, reproduce the pinned digest, and be indistinguishable from a conforming implementation.** That is a hole in the contract, which is why the bar is 100% of non-equivalent mutants rather than the ~80% usual in mutation testing.

## Parity is the reason to trust it

Against rge-bench's corpus the general tool reproduces the specialised checker **exactly**: 100% over 30 non-equivalent mutants, 1 declared equivalent. The manifest for that run was generated from the existing `MUTANTS`/`EQUIVALENT` dicts rather than retyped, so the parity tests the runner and not a transcription.

If the two ever diverge, one of them is wrong, and that is now detectable.

## What is portable, stated plainly

The tool **cannot** infer a corpus's rules from arbitrary source. That would be a static-analysis project, and a tool that guessed would report a score it had not earned. What is portable is the method, so a corpus declares its own mutants:

- one mutant per **declared rule**, not per line, so a survivor names the rule an implementation could omit
- ordinal axes are **permuted**, not deleted — the ordering lives in a table a comparison reads, not in a branch a mutant can cut
- equivalence is **declared with a reason**, never inferred, because deciding it is undecidable
- a crash is a **kill**, reported separately

## Two things found by running it on our own corpus

Neither was designed. Both came from use, which is why "run it on yourself first" was the first obligation.

**Entrypoint signatures differ.** rge-bench has `evaluate(axis, inputs)`; our jsonrpc pack has `evaluate_message(message)`. A fixed arity would have excluded every corpus that guessed differently, so `entrypoint_args` and an optional `group_key` now come from the manifest.

**Adequacy is relative to declared scope, and this is the important one.** `mcp-jsonrpc-id` scored **36.4%** at first. That number was wrong — not the measurement, but my denominator: I had declared mutants for envelope rules the pack never claimed to cover. Its README says *"a narrow interoperability contradiction"*, two arms plus one positive control.

With scope declared: **100% over the four id rules it does claim, and seven rules reported unexercised and out of scope.** Out-of-scope mutants are printed every run and never scored, because an exclusion that is not printed reads as coverage.

That field is a precondition for offering this to anyone else. Without it the tool manufactures alarming numbers about other people's corpora, which makes it an attack rather than a contribution.

## Honest about our own coverage

| Corpus | Measurable | Result |
|---|---|---|
| `mcp-jsonrpc-id-conformance` | yes | **100%** in scope (4/4); **7 rules unexercised, out of scope** |
| rge-bench (external) | yes | 100% over 30 |
| `privileged-mcp-action-v0` | **no** | reference implementation is the Rust verifier, no module entrypoint |
| `mcp-era-parity-v0` | **no** | Rust-driven |
| `rfc8785` | **no** | Rust-driven |

Three of five are not measurable today, including the frozen gate whose digest is a contract. A Rust adapter is the next real work; this PR does not claim otherwise.

## Verification

- Parity against rge-bench: identical to the specialised checker
- **Positive controls, both run before commit:** an inert mutant is reported `survived` (96.8%, exit 1); a corpus group with no declared mutants is a hard failure
- 22 stdlib tests over two files: every verdict boundary, the self-coverage guard, stale anchors, manifest validation, exit codes
- A drift test asserts the derived vectors still match the pack, and that no adequacy artifact has crept back inside it

## One guard that caught me

The manifest first lived **inside** the jsonrpc pack. That broke `check.py reproduce` with `PackError`, because the pack checksums its own public files. The pack was right; the manifest now sits in `conformance/adequacy/` and the pack is byte-for-byte untouched.

### DIFF
--- a/conformance/adequacy/mcp-jsonrpc-id.manifest.json
+++ b/conformance/adequacy/mcp-jsonrpc-id.manifest.json
@@ -1,0 +1,83 @@
+{
+  "schema": "assay.corpus_adequacy.manifest.v0",
+  "_about": "Mutation adequacy for the mcp-jsonrpc-id pack. The pack's README declares a NARROW scope: the error-response id contradiction, two incompatibility arms plus one positive control. The four id rules are in scope. The envelope-shape rules are real rules the pack depends on but never claimed to exercise, so they are declared out_of_scope and reported rather than scored. Lives OUTSIDE the pack on purpose: the pack checksums its own public files, so adding a manifest to it would break its integrity guard. The vectors file here is derived from the pack's vectors/*.json and conformance/tests/test_corpus_adequacy_own_corpora.py asserts it has not drifted.",
+  "implementation": "../../examples/mcp-jsonrpc-id-conformance/check.py",
+  "entrypoint": "evaluate_message",
+  "entrypoint_args": [
+    "message"
+  ],
+  "vectors": "mcp-jsonrpc-id.vectors.json",
+  "id_key": "vector_id",
+  "default_group": "error_response_id",
+  "mutants": {
+    "error_response_id": [
+      {
+        "label": "0 envelope must declare jsonrpc 2.0",
+        "anchor": "        message.get(\"jsonrpc\") == \"2.0\"\n        and \"result\" not in message",
+        "replacement": "        True\n        and \"result\" not in message",
+        "scope": "out_of_scope"
+      },
+      {
+        "label": "1 an error response must not carry result",
+        "anchor": "        and \"result\" not in message\n        and isinstance(error, dict)",
+        "replacement": "        and isinstance(error, dict)",
+        "scope": "out_of_scope"
+      },
+      {
+        "label": "2 error must be an object",
+        "anchor": "        and isinstance(error, dict)\n        and isinstance(error.get(\"code\"), int)",
+        "replacement": "        and isinstance(error.get(\"code\"), int)",
+        "scope": "out_of_scope"
+      },
+      {
+        "label": "3 error code must be an int",
+        "anchor": "        and isinstance(error.get(\"code\"), int)\n",
+        "replacement": "        and True\n",
+        "scope": "out_of_scope"
+      },
+      {
+        "label": "4 a bool is not an int code",
+        "anchor": "        and not isinstance(error.get(\"code\"), bool)\n",
+        "replacement": "        and True\n",
+        "scope": "out_of_scope"
+      },
+      {
+        "label": "5 error message must be a string",
+        "anchor": "        and isinstance(error.get(\"message\"), str)\n",
+        "replacement": "        and True\n",
+        "scope": "out_of_scope"
+      },
+      {
+        "label": "6 MCP tolerates an omitted id",
+        "anchor": "    mcp_id_valid = not has_id or (",
+        "replacement": "    mcp_id_valid = False or (",
+        "scope": "declared"
+      },
+      {
+        "label": "7 MCP RequestId excludes null",
+        "anchor": "        isinstance(identifier, str)\n        or (isinstance(identifier, int) and not isinstance(identifier, bool))\n    )",
+        "replacement": "        identifier is None\n        or isinstance(identifier, str)\n        or (isinstance(identifier, int) and not isinstance(identifier, bool))\n    )",
+        "scope": "declared"
+      },
+      {
+        "label": "8 JSON-RPC requires id to be present",
+        "anchor": "    jsonrpc_id_valid = has_id and (",
+        "replacement": "    jsonrpc_id_valid = True and (",
+        "scope": "declared"
+      },
+      {
+        "label": "9 JSON-RPC permits a null id",
+        "anchor": "        identifier is None or isinstance(identifier, str) or _is_number(identifier)",
+        "replacement": "        isinstance(identifier, str) or _is_number(identifier)",
+        "scope": "declared"
+      },
+      {
+        "label": "10 a non-object message is rejected outright",
+        "anchor": "    if not isinstance(message, dict) or not _error_shape_is_valid(message):",
+        "replacement": "    if not _error_shape_is_valid(message):",
+        "scope": "out_of_scope"
+      }
+    ]
+  },
+  "equivalent": {}
+}
--- a/conformance/adequacy/mcp-jsonrpc-id.manifest.json
+++ b/conformance/adequacy/mcp-jsonrpc-id.manifest.json
@@ -15,37 +15,43 @@
         "label": "0 envelope must declare jsonrpc 2.0",
         "anchor": "        message.get(\"jsonrpc\") == \"2.0\"\n        and \"result\" not in message",
         "replacement": "        True\n        and \"result\" not in message",
-        "scope": "out_of_scope"
+        "scope": "out_of_scope",
+        "reason": "envelope shape; the pack tests the id contradiction, and every vector is already a well-formed 2.0 error response"
       },
       {
         "label": "1 an error response must not carry result",
         "anchor": "        and \"result\" not in message\n        and isinstance(error, dict)",
         "replacement": "        and isinstance(error, dict)",
-        "scope": "out_of_scope"
+        "scope": "out_of_scope",
+        "reason": "envelope shape; no vector carries both result and error because the contradiction under test is about id alone"
       },
       {
         "label": "2 error must be an object",
         "anchor": "        and isinstance(error, dict)\n        and isinstance(error.get(\"code\"), int)",
         "replacement": "        and isinstance(error.get(\"code\"), int)",
-        "scope": "out_of_scope"
+        "scope": "out_of_scope",
+        "reason": "envelope shape; all three vectors carry a well-formed error object by construction"
       },
       {
         "label": "3 error code must be an int",
         "anchor": "        and isinstance(error.get(\"code\"), int)\n",
         "replacement": "        and True\n",
-        "scope": "out_of_scope"
+        "scope": "out_of_scope",
+        "reason": "envelope shape; code type is not part of the MCP-vs-JSON-RPC id disagreement"
       },
       {
         "label": "4 a bool is not an int code",
         "anchor": "        and not isinstance(error.get(\"code\"), bool)\n",
         "replacement": "        and True\n",
-        "scope": "out_of_scope"
+        "scope": "out_of_scope",
+        "reason": "envelope shape; the bool-is-not-int edge belongs to a code-type corpus, not this one"
       },
       {
         "label": "5 error message must be a string",
         "anchor": "        and isinstance(error.get(\"message\"), str)\n",
         "replacement": "        and True\n",
-        "scope": "out_of_scope"
+        "scope": "out_of_scope",
+        "reason": "envelope shape; message type is not part of the id disagreement"
       },
       {
         "label": "6 MCP tolerates an omitted id",
@@ -75,7 +81,8 @@
         "label": "10 a non-object message is rejected outright",
         "anchor": "    if not isinstance(message, dict) or not _error_shape_is_valid(message):",
         "replacement": "    if not _error_shape_is_valid(message):",
-        "scope": "out_of_scope"
+        "scope": "out_of_scope",
+        "reason": "input-domain guard; the pack's three vectors are all objects, so nothing exercises the non-object branch"
       }
     ]
   },
--- a/conformance/adequacy/mcp-jsonrpc-id.vectors.json
+++ b/conformance/adequacy/mcp-jsonrpc-id.vectors.json
@@ -1,0 +1,36 @@
+{
+  "vectors": [
+    {
+      "vector_id": "jsonrpc-error-with-null-id",
+      "message": {
+        "jsonrpc": "2.0",
+        "id": null,
+        "error": {
+          "code": -32700,
+          "message": "Parse error"
+        }
+      }
+    },
+    {
+      "vector_id": "mcp-error-with-omitted-id",
+      "message": {
+        "jsonrpc": "2.0",
+        "error": {
+          "code": -32700,
+          "message": "Parse error"
+        }
+      }
+    },
+    {
+      "vector_id": "shared-string-id-control",
+      "message": {
+        "jsonrpc": "2.0",
+        "id": "call-1",
+        "error": {
+          "code": -32603,
+          "message": "Internal error"
+        }
+      }
+    }
+  ]
+}
--- a/conformance/corpus_adequacy.py
+++ b/conformance/corpus_adequacy.py
@@ -255,8 +255,9 @@ def run(manifest_path: Path) -> dict:
                     results.append({
                         "group": group, "label": mut["label"], "verdict": "unexercised",
                         "scope": scope, "moved": 0,
-                        "how": "no vector distinguishes it, and the corpus does not claim to "
-                               "cover this rule. Not a contract hole; a scope statement"})
+                        # Print the stated reason, exactly as a declared equivalent does.
+                        # "each with a stated reason" without showing one is an assertion.
+                        "how": "out of scope: %s" % mut["reason"]})
                     out_of_scope += 1
                 else:
                     results.append({
@@ -331,7 +332,14 @@ def main() -> int:
         for f in rep["failures"]:
             print("FAIL: %s" % f)
         if rep["adequate"]:
-            print("mutation-adequacy check passed: every non-equivalent mutant is killed")
+            if rep["out_of_scope_ratio"] is not None and rep["out_of_scope_ratio"] > 1.0:
+                # The closing line is what gets quoted. It may not read as unqualified
+                # success when most declared rules were excluded from the measurement.
+                print("mutation-adequacy check passed for the DECLARED IN-SCOPE rules only "
+                      "(%d of %d rules declared here were excluded from it)"
+                      % (rep["unexercised_out_of_scope"], rep["declared_total"]))
+            else:
+                print("mutation-adequacy check passed: every non-equivalent mutant is killed")
 
     return 0 if rep["adequate"] else 1
 
--- a/conformance/corpus_adequacy.py
+++ b/conformance/corpus_adequacy.py
@@ -43,6 +43,23 @@ declares its own mutants in a manifest:
   - a crash is a KILL, reported separately, because "raises without this rule"
     says more about the rule than "returns something else".
 
+WHAT THE PERCENTAGE IS A PERCENTAGE OF
+---------------------------------------
+100% here means 100% of the rules THE AUTHOR DECLARED. It does not mean 100% of
+the rules the implementation has. A rule nobody declared is invisible to this
+check, and there is no honest mechanical fix for that inside a manifest-driven
+design: the manifest is written by the same hand as the corpus.
+
+That is why the report never prints a bare percentage. It prints the numerator,
+the denominator, the count declared equivalent, the count declared out of scope,
+and the ratio between excluded and measured. A score reported without those is a
+percentage target wearing a different coat, because an author can exclude almost
+everything and still print 100%.
+
+For the same reason an out_of_scope mutant must carry a stated reason. It leaves
+the denominator exactly as a declared-equivalent one does, so it carries the same
+obligation.
+
 SELF-COVERAGE, APPLIED TO THIS CHECK
 -------------------------------------
 A group present in the corpus that declares no mutant is the same defect one
@@ -105,6 +122,17 @@ def load_manifest(path: Path) -> dict:
             if e["scope"] not in ("declared", "out_of_scope"):
                 raise ManifestError("mutants[%s][%d] %r: scope must be declared or out_of_scope"
                                     % (group, i, e["label"]))
+            # An out-of-scope mutant leaves the denominator exactly as an equivalent one does,
+            # so it carries the same obligation: a stated reason, never a bare exclusion.
+            if e["scope"] == "out_of_scope" and not str(e.get("reason", "")).strip():
+                raise ManifestError(
+                    "mutants[%s][%d] %r: an out_of_scope mutant needs a stated reason. It leaves "
+                    "the denominator like an equivalent one, so it carries the same obligation"
+                    % (group, i, e["label"]))
+            if not e["anchor"]:
+                raise ManifestError(
+                    "mutants[%s][%d] %r: the anchor is empty. An empty anchor matches everywhere, "
+                    "corrupts the source and is then counted as a kill" % (group, i, e["label"]))
             if e["anchor"] == e["replacement"]:
                 raise ManifestError(
                     "mutants[%s][%d] %r: anchor and replacement are identical, so it mutates nothing"
@@ -184,10 +212,19 @@ def run(manifest_path: Path) -> dict:
                 continue
 
             for idx, mut in enumerate(m["mutants"][group]):
-                if mut["anchor"] not in source:
+                occurrences = source.count(mut["anchor"])
+                if occurrences == 0:
                     failures.append(
                         "%s / %s: anchor not found in %s. The rule was renamed or removed and the "
                         "mutant is measuring nothing" % (group, mut["label"], m["implementation"]))
+                    continue
+                if occurrences > 1:
+                    # Substituting the first of several is a coin flip about which rule is being
+                    # measured, and a mangled substitution is then scored as a kill.
+                    failures.append(
+                        "%s / %s: the anchor occurs %d times in %s, so the substitution would pick "
+                        "one arbitrarily and any breakage would be scored as a kill. Make the "
+                        "anchor unique" % (group, mut["label"], occurrences, m["implementation"]))
                     continue
                 mutated = source.replace(mut["anchor"], mut["replacement"], 1)
                 try:
@@ -247,6 +284,11 @@ def run(manifest_path: Path) -> dict:
     return {"schema": "assay.corpus_adequacy.report.v0", "manifest": str(manifest_path),
             "killed": killed, "survived": survived, "equivalent": equivalent,
             "unexercised_out_of_scope": out_of_scope,
+            "declared_total": killed + survived + equivalent + out_of_scope,
+            "out_of_scope_ratio": (None if (killed + survived) == 0
+                                   else round(out_of_scope / (killed + survived), 2)),
+            "score_means": ("percent of author-declared in-scope rules killed; NOT percent of the "
+                            "rules the implementation actually has"),
             "score_percent": score, "mutants": results, "failures": failures,
             "adequate": not failures}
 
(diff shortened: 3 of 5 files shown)


Rul1an:
@coderabbitai review

Rul1an:
## Ran it, not just read it — 22/22 tests pass and the output is the honest shape

```
$ python3 corpus_adequacy.py adequacy/mcp-jsonrpc-id.manifest.json
score 100.0% over 4 in-scope non-equivalent mutants (4 killed, 0 survived, 0 declared equivalent and excluded)
7 rule(s) unexercised and declared OUT OF SCOPE: real gaps in what the corpus covers, not holes in what it claims
mutation-adequacy check passed: every non-equivalent mutant is killed
exit 0
```

That output is the whole point. A bare `100%` would be the #2170 defect exactly — a clean number
over an unstated denominator. This prints the denominator, names each unexercised rule individually,
and separates *"gaps in what the corpus covers"* from *"holes in what it claims"*.

## What I checked specifically

**`test_no_mutants_is_refused_rather_than_scored_as_perfect`.** This is the guard I have failed to
write three times today. A corpus with nothing to kill cannot score 100%; it is refused. Without
that, the strongest possible score is available to the emptiest possible corpus.

**Scope exclusions are reported every run and never scored.** The comment states the reasoning
better than I would have: *"adequacy is relative to declared scope, and scoring a rule nobody claimed
manufactures a hole that is not one"*, together with *"A scope exclusion that is not printed reads as
coverage."* Both directions of the error are closed.

**`test_the_manifest_lives_outside_the_pack`.** The measuring instrument is not inside the thing it
measures. That is the property that makes the number mean anything to a second party.

**Vendor-neutrality is in the mechanism, not the README.** `group_key` optional because *"forcing it
to invent an axis would be the tool bending the corpus to fit itself"*, and flexible entrypoint arity
because *"a fixed arity would exclude every corpus that did not guess the same one"*. A tool that
only measures corpora shaped like ours would not be neutral, whatever it said about itself.

**Self-coverage.** `# Self-coverage: the reach of this check must not narrow silently` — the check
watches its own denominator, which is the same rule it applies to corpora.

No change requests. This is the cleanest expression of the "state what you did not observe" doctrine
in the repository.

Rul1an:
Independent review of `5e11c775ad25e628fb922c6ac46f0fb188042253`. Ran it. Did not retype rge-bench's mutant dicts. Did not change behaviour, so there is no red-then-green pair from me on this head.

GitHub will refuse `REQUEST_CHANGES` if this token is the PR author. Treat the verdict as blocking anyway.

**Changes requested.** The claimed commands hold. rge-bench parity holds. I would not ship this as a contract-grade 100% check at this head.

### Claimed commands (all hold)

| Command | Measured |
|---|---|
| `test_corpus_adequacy.py` | 17 tests, exit 0 |
| `test_corpus_adequacy_own_corpora.py` | 5 tests, exit 0 |
| `corpus_adequacy.py …/mcp-jsonrpc-id.manifest.json` | exit 0, `100.0% over 4 in-scope`, `7 rule(s) unexercised … OUT OF SCOPE` |

Your two controls, reproduced as CLI, not trusted:

- Inert mutant: reconstructed as the rge-generated manifest plus one inert docstring mutant → **96.8% over 31, exit 1**. The 96.8 figure is `round(100.0 * 30/31, 1)`. It is not a test in-repo (the unit test uses 50% over 2). The number in the PR body is real; the pin is not that number.
- Group with no mutants: **exit 1**. Hard fail holds. It still prints `score 100.0% over 1` before `FAIL`. That is a log lie.

rge-bench `e52ac41ec30c1923a35650e4d379a6c4c7b27499`: `check_rule_liveness.py` → 100% over 30, 1 equivalent, exit 0. Manifest generated by walking `MUTANTS`/`EQUIVALENT`. `corpus_adequacy.py` on that manifest → 30 killed, 0 survived, 1 equivalent, 0 oos, exit 0. **Identical numbers.** Format-only delta for later: rge uses integer `100 * killed // denom` (30/31 → 96%); this tool uses `round(..., 1)` (96.8%).

### 1. `out_of_scope` is the escape hatch you suspected

Constructed. 1 killable in-scope + 5 inert `out_of_scope`, **no reason field**. Exit **0**, `100.0% over 1`.

`scope` defaults to `"declared"`. Allowed `"declared"|"out_of_scope"`. Required keys are only `label`, `anchor`, `replacement`. Equivalence already demands a non-blank `reason`. Scope does not.

Scope is consulted only after a miss (`elif scope == "out_of_scope"`). A kill still increments `killed`. You cannot hide a kill. You **can** hide any survivor by tagging it `out_of_scope`. That is the same move that took mcp-jsonrpc-id from 36.4% to 100%, and the shipped manifest does it for seven rules with no per-mutant justification.

Do not add a ratio guard. That is a percentage target in another coat, and you already forbade those. Require a stated reason, the same way equivalence does: `out_of_scope` without a non-blank `reason` is `ManifestError`, exit 2. Print the reason every run. An exclusion that is not justified is not a scope statement.

A control that must stay red: the 1-vs-N manifest above, after the reason is required, is exit 2 until each oos row carries one. After reasons are present it may print 100% over 1; that is then an audited claim, not a silent shrink.

### 2. Undeclared rule inside a group — same class as the unpinned #2532 kill

Constructed. Implementation has two live rules (`bad` → rejected, `n>10` → big). Vectors only hit `bad`. Manifest declares only the `bad` mutant. Exit **0**, `100.0% over 1`. The second rule is live and unmeasured.

The self-coverage guard (`run`, groups_in_corpus vs `m["mutants"]` keys) catches a group with no mutants. It never counts rules inside a group. One declared killed mutant satisfies a group that contains N real rules. Score is bounded by author imagination.

I do not have a mechanical fix that stays inside your charter (stdlib only, no network, no Assay import, no inferred equivalence, no cargo-mutants, no invented static analyzer). Those are the only honest options:

1. **cargo-mutants owns the list.** Author cannot shrink it. You already parked that as a contract question. Leave it parked.
2. **Manifest declares a reviewer-auditable census** (`expected_rule_ids` / count). The machine checks `len(declared) == expected`. A dishonest author sets `expected=1`. A hook for a reviewer, not a guarantee.
3. **Refuse to score unless every public function/branch is named.** That *is* inferring the rule list from arbitrary source. The tool correctly refuses.

The tool already says it cannot infer rules. Therefore **100% means 100% of author-declared in-scope mutants**, not 100% of rules. Ship only if that sentence is the published claim, including on the pass line (`every non-equivalent mutant is killed` currently reads as every rule). Do not invent a static analyzer. This is the most valuable thing I can return, and it is not a patch.

### 3. `replace(anchor, replacement, 1)` — hard error, not a warning

`source.replace(..., 1)` is first occurrence only. `anchor not in source` is the only guard. `"" in source` is True.

Constructed:

| Case | Result |
|---|---|
| Anchor twice, decoy first, real rule second | replace hits decoy; **survived, exit 1, 0%** (fail-closed false gap) |
| Anchor twice, live first, mutant *labeled* as the decoy | replace hits the live rule; **killed, exit 0, 100%** — **false kill** |
| Substring: `n > 1` also sits inside `n > 10` | replace mutates the wrong predicate; **killed, exit 0, 100%** — **false kill** |
| `anchor=""` | `replace("", "PASS", 1)` prepends; `NameError` at load; crash = kill; **exit 0, 100%** — **false kill** |

A warning would let the false-kills through at 100% and weaken the bar. **Hard error when `source.count(anchor) != 1`. Hard error on empty anchor.** Do not score a mutant whose anchor is not a unique exact slice.

Controls that must stay red: the empty-anchor manifest (today exit 0 / 100%) becomes exit 2. The non-unique-anchor false-kill (today exit 0 / 100%) becomes exit 2. Both exit codes belong in the follow-up comment.

### Also, not a third weakness: the printed 100% on hard-fail paths

Zero in-scope prints `score 100.0% over 0` then FAIL. Group-without-mutants prints `score 100.0% over 1` then FAIL. `score = 100.0 if denom == 0`. A log that says 100% before FAIL is how a reviewer quotes the wrong number. Print the score only when something was scored, or print it after FAIL and not as 100% of nothing.

### Verdict

I would not approve at this head. Two behaviour changes, then I would ship the runner as a declared-mutant check:

1. `out_of_scope` requires a non-blank reason, same as equivalence. No ratio guard.
2. Unique non-empty anchor is a hard error. False-kills cannot print 100%.

Weakness 2 is not a patch. Change the published sentence so 100% cannot be read as 100% of rules. cargo-mutants stays your deferred contract question.

I did not rebase onto current `origin/main` (`48f67df6`). Diff vs merge-base is the five new files. Pack `examples/mcp-jsonrpc-id-conformance/` is byte-identical to main. Imports are stdlib only. The 100% bar has no env or `--target` override. Those are non-claims, not disagreements.

Rul1an:
## All three findings reproduced, then fixed. Head `12afbcef5`.

Ruley's blocking review was right on every point. I reproduced each one before touching anything; all three printed 100% and exited 0 beforehand.

| Finding | Reproduced | Now |
|---|---|---|
| 1 killable + 5 unreasoned `out_of_scope` -> exit 0, 100% | confirmed | refused at manifest load, exit 2 |
| empty anchor `""` -> false kill, 100% | confirmed | refused at manifest load, exit 2 |
| substring anchor occurring twice -> false kill, 100% | confirmed | run fails, names the count and demands uniqueness |

**On the hatch.** You were right that it was a percentage in a different coat, and right that the obligation is the same as for equivalence. An `out_of_scope` mutant leaves the denominator exactly as a declared-equivalent one does, so it now requires a stated reason. Our own manifest was missing seven of them; they are written now.

I did **not** add a hard ratio guard, and I want that decision visible rather than quiet. Our jsonrpc pack legitimately excludes more than it measures (4 in scope, 7 out, ratio 1.75) because its README declares a narrow subject. A hard fail there would punish an honest corpus. Instead the ratio is computed and printed, and a corpus that excludes more than it measures now says so in the output. If you think that is still too soft, say so and I will make it a gate flag.

**On the anchors.** Both cases were the same defect: a mangled substitution scored as a kill. Occurrences are now counted, zero and >1 both fail, and the empty anchor is rejected before the run starts.

**On weakness 2, which is the sharpest thing in your review.** You are right that there is no honest mechanical fix inside a manifest-driven design, and that the fix is the published sentence. The report no longer prints a bare percentage. It prints numerator, denominator, equivalents, out-of-scope count, total declared, and the ratio, followed by:

```
This is 100.0% of what the AUTHOR DECLARED, not of the rules the implementation
has. A rule nobody declared is invisible to this check.
```

That sentence is now in the module docstring, the report, and will go in the PR body. The manifest is written by the same hand as the corpus, and pretending otherwise is the failure mode this whole tool exists to catch one level down.

**Regressions.** One test per finding, plus one asserting the report names its own denominator, plus one asserting a majority-excluded corpus says so. 29 tests total. rge-bench parity re-checked after the change: still 30 of 30 with 1 declared equivalent, so the fix did not move the reference measurement.

Thanks -- and specifically for not stopping at the two mechanical ones. The third has no patch and is the one that would have shipped a misleading number.

Rul1an:
Re-review of `12afbcef5035358934c2c486ade14a5f64003723` (parent `5e11c775`). Ran it. Attacks constructed under `/tmp`, not the jsonrpc pack. rge-bench manifest walked from `MUTANTS`/`EQUIVALENT`, not retyped.

**Merge.** The three named findings are closed. I am not adding a hard ratio gate.

### The three attacks, re-run

| Attack | Expected | Measured |
|---|---|---|
| 1 killable + 5 `out_of_scope`, no `reason` | exit 2, needs a stated reason | **exit 2.** `an out_of_scope mutant needs a stated reason. It leaves the denominator like an equivalent one, so it carries the same obligation` |
| `anchor: ""` | exit 2, anchor is empty | **exit 2.** `the anchor is empty. An empty anchor matches everywhere, corrupts the source and is then counted as a kill` |
| Anchor twice (decoy-first and live-first) | exit 1, occurs N times, make the anchor unique | **exit 1** both ways. `the anchor occurs 2 times … Make the anchor unique`. The false-kill orientation does not pass. |

Claimed commands: 22 OK, 7 OK, jsonrpc exit 0. `4 of 4 DECLARED in-scope rules killed (100.0%). … 7 declared out of scope. 11 rules declared in total.` plus the AUTHOR DECLARED sentence plus `NOTE: … ratio 1.75`. No bare percentage line.

rge-bench `e52ac41ec30c1923a35650e4d379a6c4c7b27499`: liveness 100% over 30, 1 equivalent. This tool on the generated manifest: **30 of 30, 1 equivalent, 0 oos, exit 0.** Unique-anchor still accepts every rge anchor.

Regression tests in `RuleyFindings` ran. None vacuous.

### The ratio, since you asked me to overrule you

I will not. A hard fail on `excluded > measured` (or on 1.75) would fail the jsonrpc pack for being honest about a narrow README. That is a percentage target in another coat, which is what I told you not to add. Printing the ratio and the minority NOTE is the correct substitute.

A constructed 1-in-scope + 5 reasoned oos is still exit 0, `1 of 1 … (100.0%)`, NOTE ratio 5.00, `check passed`. That is allowed. It is no longer silent. I reproduced it.

I am not moving the goalposts. I said I would ship after reasons and unique non-empty anchors. Those are in.

### What I would still change, and it is not a ratio gate

The reasons are required and then not shown. Equivalence prints `eq["reason"]` on the how-line. An oos row prints the generic "scope statement" how-text. The TTY says `each with a stated reason` without printing one. `--json` mutant objects also omit `reason`. A reviewer cannot audit the justifications without opening the manifest. Last time I said print the reason every run. That leftover is the only thing I would still ask for, as a follow-up, not as a hold.

When `out_of_scope_ratio > 1`, the last line is still `mutation-adequacy check passed: every non-equivalent mutant is killed`. A grep of the last line quotes the old claim. Fold the minority into that line (or drop "passed" when the NOTE fired). Do not change the exit.

`--json` still has `score_percent: 100.0` and `adequate: true`. `score_means` is there; a bot that only copies `score_percent` will still lie. Same follow-up class.

Residual, not blocking: exact-slice uniqueness does not catch a unique substring of a larger rule (`n > 1` inside `n > 10`). Constructed, still exit 0 / 100%. Author has to pick that slice on purpose. Do not reopen this PR for it.

Unique-anchor skip still prints `0 of 0 DECLARED … (100.0%)` then FAIL. Same log-shape as before, no longer a pass.

### Verdict

Merge at `12afbcef5`. The exit that would have flipped this back to a hold is the first attack still exiting 0. It exits 2.

I did not rebase onto current `origin/main`. I did not treat the residual substring as a third finding. I did not invent a static analyzer for undeclared rules. Those are non-claims.

Rul1an:
## Approved, and both follow-ups landed. Head `a9262b1ce`.

Thanks for holding the ratio question until the exits existed rather than arguing it first. Both follow-ups were right and neither changes an exit code, so they are in rather than deferred.

**The out-of-scope reason now prints on its own line**, exactly as a declared equivalent already did. You were right that "each with a stated reason" without showing one is an assertion about reasons rather than a reason:

```
error_response_id      unexercised 10 a non-object message is rejected outright
    out of scope: input-domain guard; the pack's three vectors are all objects, so
    nothing exercises the non-object branch
```

**The closing line is qualified above ratio 1.** It read `mutation-adequacy check passed` directly under a NOTE saying most declared rules were excluded, which is the line a reader quotes. Now:

```
NOTE: more rules are excluded than measured (ratio 1.75). The score is real but it is
a statement about a minority of the declared rules.
mutation-adequacy check passed for the DECLARED IN-SCOPE rules only (7 of 11 rules
declared here were excluded from it)
```

Where nothing is excluded the line is unchanged, and there is a test asserting that, so the qualification cannot leak into the clean case and quietly hedge a full result.

**On the ratio gate.** Recorded as your decision, not mine: no hard gate, because it would punish an honest corpus with a narrow declared subject, and the NOTE plus the qualified closing line carry the same information without the false negative. I had picked the soft option and would not have been able to tell whether that was judgement or convenience, so having you rule on it after seeing the exits is worth more than having agreed with me up front.

**Re-verified after the change:** 25 tests here, 7 on the own-corpora side, and rge-bench parity still 30 of 30 with 1 declared equivalent. The rge-bench manifest for that check is generated outside its tree -- its own `check_public_surface.py` rejects the vendor-branded schema id, which is a separate and useful signal about where this tool should eventually live.

Rul1an:
## Independent design review of the intended Rust `runner: "process"` adapter

Read `12afbcef5` as asked (`corpus_adequacy.py` + this PR). Timed rebuilds on the current PR head `a9262b1ce`; that delta is Python-only (`conformance/corpus_adequacy.py` + one test file). The crate graph and verifier sources below are the same. Did not implement. Did not edit a branch. Did not add a dependency.

### Verdicts

**(a) Conclusion 2's time bound is sound on a warm debug checkout. The rationale is not.** "Most rules sit in a leaf crate" is the easy-case generalisation: they timed `verify_privileged_mcp_action.rs`. This profile's own corpus includes rules that do **not** live there. None of those rules live in `assay-core`. The ones that live in `assay-evidence` rebuild seven crates, not one; on this machine that is ~11–20 s, not minutes. So the 4–7 min envelope still holds. The leaf-crate story does not.

**(b) No fifth option.** There is no workable Rust 2026 engine that keeps *declared* rules, avoids a per-mutant rebuild, and does not instrument production. Mull-style LLVM-IR mutation and runtime class recompilation do not have a drop-in equivalent that satisfies those three at once.

**(c) Build failure is UNPROVED.** A mutant `rustc` refuses has not been run against the digest. That is a failed measurement, not a kill. A rule that is "the only thing making the crate compile" is measured by a *compiling* omit, or declared equivalent with that reason — not by scoring the compiler.

---

### (a) Where the rules actually execute

`privileged-mcp-action/v0` is not a module entrypoint. The published comparison surface is the CLI JSON (`bundle_integrity`, `verdict`, the claim matrix) from `assay evidence verify-privileged-mcp-action`. Spec: `docs/profiles/privileged-mcp-action/v0.md`. Corpus digest (issue #1840 / `conformance/privileged-mcp-action-v0/MANIFEST.json`): `sha256:cb58ce91863f52e0568742b977f0642158453ec11bbcd25821f9171dccd03342`. Fourteen vectors, two of them integrity-only (`bad-101`, `bad-109`).

| Stage | What decides | File | Crate |
|---|---|---|---|
| 1 | bundle integrity (`BundleReader::open` → `verify_bundle_with_limits`) | `crates/assay-evidence/src/bundle/writer_next/verify.rs`, `crates/assay-evidence/src/bundle/reader.rs`, `crates/assay-evidence/src/crypto/id.rs` (`compute_run_root`, `bundle_id`) | **assay-evidence** |
| 3 (classifier) | marker triple `(schema, -32042, "assay-proxy")` | `crates/assay-evidence/src/denial_marker.rs` | **assay-evidence** |
| 2–4 | cardinality, closed vocabs, binding equality, claim matrix | `crates/assay-cli/src/cli/commands/evidence/verify_privileged_mcp_action.rs` | **assay-cli** |
| — | (nothing) | `crates/assay-core/**` | **assay-core has no v0 verifier rules.** The only hit is `era_parity_tests.rs` pointing at this corpus. That is the out-of-scope Rust-TEST corpus, not this profile. |

`assay-cli` is a leaf *binary* crate: nothing depends on it. `assay-evidence` is not a leaf. The published edges:

```
assay-evidence
  → assay-adapter-api          (Cargo.toml: assay-evidence; source uses EvidenceEvent)
      → assay-core
          → assay-cli
          → assay-mcp-server
          → assay-sim
          → assay-metrics
  → assay-cli                  (direct)
  → assay-sim                  (direct)
```

A source subst in `denial_marker.rs` or `writer_next/verify.rs` therefore dirties `assay-evidence` **and** `assay-adapter-api` **and** `assay-core` (and then the CLI/server/sim/metrics rlibs) even though none of those downstream crates contain the rule. That is the file they did not advertise.

### (a) Measured, this checkout, `cargo build -p assay-cli --offline --bin assay`, comment-append + `git checkout --` restore

No-op warm after a prior cascade: **1.03 s**, Fresh.

| Touch | Compiling | cargo `Finished` | `/usr/bin/time` real |
|---|---|---|---|
| `verify_privileged_mcp_action.rs` (the convenient file) | `assay-cli` only | **8.75 s** | 8.82 s |
| `denial_marker.rs` | evidence → adapter-api → core → metrics → sim → mcp-server → cli | **10.79 s** | 10.90 s |
| `bundle/writer_next/verify.rs` | same seven | **10.48 s** | 10.59 s |
| `crypto/id.rs` (run_root / bundle_id; `bad-109` lives here) | same seven | **20.08 s** | 20.22 s |
| `types.rs` (not a v0 rule; public type surface) | same seven | **14.14 s** | 14.27 s |

14-vector subprocess loop (`./target/debug/assay evidence verify-privileged-mcp-action --format json` × 14): **0.23 s** real. Their 0.25 s is real.

Their 7.5 s / 14.2 s / 0.25 s numbers are the same *shape* as this machine. 14.14 s on `types.rs` is their 14.2 s to the tenth. `crypto/id.rs` is the one in-profile file that beat 14.2 s, and it is 20 s, not 90 s. 30 mutants at 9–20 s plus 0.23 s verify is still about **5–10 minutes**, not a literature-scale recompile problem.

So: **they did pick a convenient file** for the "leaf" claim. They did **not** invent the cost. There is **no** v0 rule in `assay-core` whose mutation would explode the graph beyond the evidence cascade already measured. The honest sentence is: *most declared stage-2/3/4 mutants are a leaf rebuild; the two integrity vectors and the marker triple are a seven-crate rebuild that incremental rustc keeps in the same minute-budget.*

`cargo build -p assay-evidence` is the wrong target for this adapter (you need the `assay` binary). I ran it once after a `denial_marker` touch and watched a different unit graph light up (rustls / object_store / …, 20.6 s). Do not use that number. Use `-p assay-cli`.

### (b) The fifth option, checked against real sources

What they want: author-declared subst, one compile, no production source switch.

What exists in 2026:

- **cargo-mutants** ([README](https://github.com/sourcefrog/cargo-mutants/blob/main/README.md) fetched 2026-08-19): *"As of August 2026 this is an semi-actively-maintained spare time project"* — wording confirmed, not invented. Latest commit on `main` is `0a46fd3` (2026-08-17, Martin Pool). [Mutation patterns](https://mutants.rs/mutants.html): built-in genres only (`FnValue`, operator swaps, match-arm delete *when a wildcard already exists*). No user-declared anchor/replacement, no permutation operator, no declared-equivalent-with-reason. It runs `cargo test`. Complementary census, not this engine. Agreed with conclusion 1; not rediscovering it.
- **mutest-rs** ([README](https://github.com/zalanlevai/mutest-rs/blob/main/README.md), [mutest.rs](https://mutest.rs/), ICST 2026 paper): the actual Rust meta-mutant. One compile, runtime-swappable mutants, "two orders of magnitude" vs cargo-mutants on their crate set. It is still a **test-suite** tool, **nightly rustc driver**, **built-in operators**, and it *embeds a mutation harness into the compiled program* (`cfg(mutest)`, `mutest-runtime`). That is the single-compile switch they already rejected, just applied by a compiler plugin instead of a `match` in git. It cannot express a declared rule, a permutation, or a reasoned equivalent.
- **mutagen** ([README](https://github.com/llogiq/mutagen/blob/master/README.md)): `#[mutate]` in source, `MUTATION_ID` at runtime, nightly proc-macro. Last commit **2022-07-29**. The README's own warning is "Do not use `#[mutate]` for code that can cause damage if buggy." That is their security objection, written by the tool. Dead.
- **Mull** ([mull-project.com](https://mull-project.com/), v0.33.0 2026-04-11): "Fault injection for C and C++." The 2018 paper claimed language-agnostic LLVM IR including Rust; JIT was removed in 2021; today's product surface is Clang/C++. rustc `--emit=llvm-ir` / `llvm-bc` exists. There is no maintained declared-rule Mull pipeline that consumes rustc IR, relinks an `assay` binary, and yields a JSON verdict. A 2017 "Mull meets Rust" talk is not a 2026 adapter.
- **Cranelift / rustc mutation backends:** no product found. I am not inventing one.

There is no fifth option that survives the constraints. Incremental `cargo build` of a declared file **is** the cheap form of "not a full clean." It is still a rebuild. Extracting the verifier into a `cdylib` so only that crate relinks would be a production structure change, which they did not ask for and I am not proposing.

### (c) Build failure is UNPROVED

The question the tool asks is: *can an implementer omit this declared rule, reproduce the pinned digest, and be indistinguishable?* That question is answered by a JSON verdict against the fourteen vectors. `rustc` exiting 1 produces no verdict. Scoring it as a kill measures the substitution's well-formedness, which is how a typo in the manifest prints 100%. That is the false-kill class already closed for anchors at `12afbcef5` (`source.count(anchor) != 1`). Do not reopen it on the compiler.

The other side: deleting the only exhaustive match arm is a real signal that *this Rust implementation* cannot omit the rule without a rewrite. True, and not a corpus kill. The compiler killed it. A stranger implementing #1840 from the spec in another language never sees `E0004`. Treating rustc as a vector would let this corpus claim a rule is covered when no bundle distinguishes it.

cargo-mutants already names this "unviable" and refuses to count it as a kill ([`mutants.rs`](https://mutants.rs/mutants.html): match arms are only deleted when a wildcard is already present, *because otherwise the mutant does not compile*). They are not the engine. They are right about this bit.

**How to tell typo from load-bearing compile rule — and why that distinction must not change the score:**

1. Unique non-empty anchor (already required).
2. `rustc --error-format=json`: parse/resolve (`error[E0425]`, `error[E0061]`, lexer) is a mangled subst. `error[E0004]` / non-exhaustive patterns at the mutated span is a type-level rule.
3. **Both are UNPROVED.** The distinction is for the author, printed on the how-line, so they can rewrite the mutant.
4. The mutant that *can* kill an exhaustive-arm rule is a compiling omit: replace the arm with a wildcard / default that drops the check. If that omit cannot be written without inventing types, declare it `equivalent` with that reason. Do not infer the equivalence. Do not let rustc fill the denominator.

Crash of a built binary stays a kill, as in the Python path. Import-fail on the Python path (`the mutant does not load`) is the thing this forks, and the report must say so. `12afbcef5` has no `unproved` verdict today. Add the word or the score will keep calling a refused compile `killed`.

Guaranteed restore + refuse dirty tree: yes. Required. Not optional hygiene. A leaked subst in `crypto/id.rs` is a digest bug.

---

### What I did not check

- `--release`, `cargo clean`, a cold CI cache, sccache, or another OS. The 4–7 min claim is a warm incremental debug claim.
- Whether thirty sequential restores keep the incremental cache this hot. I timed one touch at a time.
- Running cargo-mutants or mutest-rs on these files.
- `rfc8785` / `mcp-era-parity-v0` (out of scope as asked).
- Whether `writer.rs` vs `writer_next/verify.rs` has a second live stage-1 path that would need its own mutant (I followed `BundleReader::open` → `writer::verify_bundle_with_limits` → `writer_next::verify`).
- Venue.

I would build the process runner they described. I would write the manifest so stage 1 and `classify_denial_marker` are declared against the evidence files they actually live in, not hidden behind "leaf crate." I would not spend a week on LLVM IR.

Issue #2421, opened 2026-08-16
assay mcp config-path --json aborts by SIGABRT when HOME is non-UTF-8


Rul1an:
## What

`assay mcp config-path --json` dies by SIGABRT when `HOME` (Linux: also the fallback for `dirs::config_dir()`) contains non-UTF-8 bytes. The text channel exits 0 via `Path::display()`.

Same causal class as closed #2264 (`serde` Path `Error` → `json!` unwrap → `panic = "abort"`), **different ingress**: environment `HOME`, not `--trace-file`. Do not reopen #2264.

## Provenance (sidecar, not quorum)

| item | value |
|---|---|
| source SHA | `924599b1eeda055ec2a19ba3ce017841958a5e46` (`origin/main`) |
| site | `crates/assay-cli/src/cli/commands/config_path.rs:345-347` (`json!({ "config_path": PathBuf })`) |
| blob | `00c845840c6a99840bc3483c75feca95e85f5a7d` (unchanged `2c2a94de`…`924599b1`…`5f30e16d`) |
| binary | `/tmp/assay-2165-bin/assay` `assay 5.2.0` sha256 `02cb8ee2ce2791dbdbda94dddc0ea3df4d3ca6cd96fd57e4c881872aba9dc07d` |
| binary note | not compiled at `924599b1`; used because the `config_path.rs` blob is byte-identical |
| drive | Python `os.environb` (no shell-lossy env). `XDG_CONFIG_HOME` unset so Linux uses `HOME/.config`. |

## RED

`HOME=/tmp/assay-home-\xff`, `assay mcp config-path --json cursor`:

- returncode **-6** / SIGABRT (shell-style 134)
- stdout **0 B**
- stderr: `panicked at config_path.rs:345:22` `Error("path contains invalid UTF-8 characters")`

UTF-8 control (`HOME=/tmp/assay-home-utf8-ctrl`): json and text **exit 0**.

Text + non-UTF-8 HOME: **exit 0**, path shown via `display()` (U+FFFD). Not the abort.

## Smallest DoD

One site: pass `config_path` through existing `path_json(&Path)` (or `display().to_string()`) before `json!`. Keep text on `display()`. Unix `os.environb` HOME test: JSON exits by a defined code (not signal); UTF-8 control stays 0. No new reason code.

## Non-claims

- not a #2264 closure or reopen
- not Windows (WTF-16 / unpaired surrogates)
- not a path-traversal or security-ingest claim
[...]

PR #2539, opened 2026-08-19
fix(metrics): canonicalize tool schema hash bytes

### LINKED ISSUE #2245: tool_description_integrity hashes a schema non-canonically, so a key reorder reads as E_TOOL_SCHEMA_MUTATED
Found while fixing #2229 in #2239, which changed the *other* instance of this defect. Filing rather
than folding into that PR, because this one has a different blast radius and deserves its own
decision.

Measured on `4ca3b122e819022313562296fe7f6a8e5aa5999c`, worktree `/Users/roelschuurkes/wt-2229`,
`cargo 1.96.0`.

## The code

```
crates/assay-metrics/src/tool_description_integrity.rs:13-16
/// Stable hash over a JSON value: serialise to compact JSON then SHA-256.
fn sha256_value(v: &serde_json::Value) -> String {
    sha256_str(&serde_json::to_string(v).unwrap_or_default())
}
```

Same mechanism as #2229 and the doc comment makes the same unchecked claim. `assay-evidence`
enables serde_json's `preserve_order`, resolver-2 unifies it across one compilation unit, so
`serde_json::Map` is an insertion-ordered `IndexMap` and `to_string` emits whatever order the
producer sent. Confirmed for this crate the same way:

```
$ cargo tree -e features -p assay-metrics -i serde_json | rg preserve_order
├── serde_json feature "indexmap"
│   └── serde_json feature "preserve_order"
```

`serde_json::to_string` also is not RFC 8785 in its number handling, so `1e2` and `100` disagree.

## Consequence

`sha256_value` backs two comparisons, both of which read a benign reorder as a mutation:

- Snapshot mode (`:111`): a later `tools/list` snapshot whose keys are ordered differently from
  snapshot 0 raises `E_TOOL_SCHEMA_MUTATED`.
- Pinned mode (`:164-176`): the operator-authored `schema_sha256` pin is compared to a freshly
  computed hash, raising `E_TOOL_SCHEMA_MISMATCH`.

As in #2229 the direction is a false positive, not a bypass.

## Why it was left out of #2239

`schema_sha256` is an operator-authored pin in a different vocabulary from `tool_pins.schema_hash`,
in a different crate. Fixing it changes every value users may already have recorded, which is a
breaking change on a documented field and a separate call from the one #2239 made. #2239 kept its
diff to the hashing under review and said so in its non-claims.

## What the fix looks like

Per `AGENTS.md` "One rule, one function", this should not become a third answer to "what is this
JSON, canonically". `assay-metrics` depends on `assay-core`, which now exposes `mcp::jcs`; there is
also `assay_canonical::jcs`. Both wrap `serde_jcs`. Deciding which one this crate should reach for
is part of the work, and consolidating the two wrappers may be the better shape.

## Not established

- I did not check whether any user currently holds a `schema_sha256` pin, nor whether any in-tree
  fixture contains one.
- I did not measure how the two `jcs` wrappers should be consolidated; that may need an ADR.

### LINKED ISSUE #2261: Nothing pins serde_json's arbitrary_precision off, though compute_json_hash's expect() assumes it
`compute_json_hash`'s `expect()` is unconstructible only while `serde_json`'s `arbitrary_precision` is off, and nothing states that premise. `preserve_order`, the other feature this file reasons from, has a runtime premise test (`premise_serde_json_map_preserves_insertion_order_in_this_build`, `crates/assay-core/src/mcp/identity.rs`). Proposing the same treatment for the second flag.

Raised as informational finding F1 in the round-4 review of #2239 (https://github.com/Rul1an/assay/pull/2239#issuecomment-5243873209). I verified the asymmetry and then measured the consequence, which turned out to be a different one than F1 supposed — see below.

## The asymmetry

`preserve_order` is pinned by behaviour, not by name, so a build where it silently went away fails the test instead of passing the reorder tests vacuously:

```rust
// crates/assay-core/src/mcp/identity.rs
#[test]
fn premise_serde_json_map_preserves_insertion_order_in_this_build() { … }
```

`arbitrary_precision` appears nowhere in the tree — not enabled, not asserted, not gated:

```console
$ rg -n 'arbitrary_precision' --glob '!target' .
$ # (no matches)
$ rg -n 'preserve_order' --glob '!target' crates | wc -l
       4
```

It is off today by omission: `assay-evidence` turns on `preserve_order` and resolver-2 unification hands `assay-core` the same `Map`, but nothing in the workspace asks for `arbitrary_precision`, and nothing would notice if a new dependency did. Additive feature unification means any crate in the graph can turn it on for `assay-core`.

## What turning it on actually changes

I ran the file's own claims against both feature settings in a throwaway probe (`serde_json =1.0.151`, `serde_jcs =0.2.0`, `jsonschema 0.49.2`, `rustc 1.96.0`), with the flag off as a control.

**The numeric bound #2239 documents is insensitive to the flag.** All 15 rows of `the_f64_collapse_hides_a_difference_a_validator_still_enforces` behave identically with `arbitrary_precision` on: the same six keywords separate, the same nine agree, and the closing-bound pair `18446744073709551616` / `…617` still collapses in the hash and still is not separated by the validator. The control run reproduced the in-repo table exactly, which is what makes the treatment run interpretable. The reason is that `serde_json` keeps answering `as_u64()` for an in-`u64` literal under this feature (`is_u64=true, as_u64=Some(9007199254740993)`) and still cannot answer it above `u64::MAX`, so both the JCS renderer and `jsonschema`'s comparison see what they saw before. **So F1's stated consequence — that the closing bound "silently becomes false" — does not reproduce, and the documented bound in `docs/reference/config/policies.md` needs no change.**

**What does change is the argument behind `compute_json_hash`'s `expect`.** A literal outside `f64` range is a parse error today, so a `Value` cannot hold a non-finite number — which is precisely what `canonicalization_succeeds_across_the_value_domain` and the `expect("a serde_json::Value is always JCS-serializable")` comment rest on. With the flag on, the literal is retained and the error path opens:

```
arbitrary_precision OFF:  1e400 -> parse ERROR (number out of range)
arbitrary_precision ON:   1e400 -> parses, as_f64()=None, serde_jcs::to_vec -> ERR(invalid float value)
```

`observe_tool_definition` (`crates/assay-core/src/mcp/proxy/tools.rs:28`) passes a `tools/list` response's `inputSchema` straight to `ToolIdentity::new`, so under that flag an MCP server advertising `{"maximum":1e400}` reaches a failing `expect()`. The workspace sets `panic = "abort"`, so that is a process abort in the agent's data path, reachable from server-controlled input — and it is silent in the sense that matters here: no existing test would notice the premise had changed, because the domain test only uses literals that parse under both settings.

## Proposal (small)

One test beside the `preserve_order` premise, asserting the behaviour rather than the feature name, so it holds however the feature arrives:

```rust
/// The premise behind `compute_json_hash`'s `expect` and behind
/// `canonicalization_succeeds_across_the_value_domain`: a parsed `Value` cannot hold a
/// non-finite number, because a literal outside `f64` range does not parse. `arbitrary_precision`
/// would retain the literal, `serde_jcs` would reject it, and the `expect` would abort.
#[test]
fn premise_out_of_range_number_literals_do_not_parse_in_this_build() {
    assert!(serde_json::from_str::<Value>("1e400").is_err());
    assert!(serde_json::from_str::<Value>("-1e400").is_err());
}
```

Two lines and no production change. It converts "we checked this once, at these versions" into a standing check, which is the same thing the `preserve_order` premise buys.

Out of scope, and deliberately: threading a `Result` through `ToolIdentity::new` so the failure is a deny rather than an abort. #2239 records why that is a larger change than the hashing, and it is worth its own issue if the abort is judged unacceptable independent of this flag.

## Provenance

Probe crate in `/tmp/ap-probe` (throwaway, not committed), run twice against `crates/assay-core`'s pins as of `5c39413481058909ac45c7747c58dd639dc52a8f`: `serde_json =1.0.151` with `["std","preserve_order","float_roundtrip"]` (control) and the same plus `"arbitrary_precision"` (treatment); `serde_jcs =0.2.0`; `jsonschema 0.49` with `default-features = false`; `rustc 1.96.0 (ac68faa20 2026-05-25)`. Worktree `/Users/roelschuurkes/merge-2239`, `CARGO_TARGET_DIR` local to it.

## Non-claims

- I did not enable `arbitrary_precision` in the workspace or run the real test suite under it. The probe reimplements `compute_json_hash` (same `serde_jcs` pin) and uses `jsonschema::validator_for` rather than `crate::policy_engine::compile_schema`, so it measures the same libraries, not the same code path.
- The probe's `serde_json` feature set omits `indexmap`/`alloc`, which the real build resolves; neither affects number parsing, but the resolved sets are not identical.
- "No existing test would notice" is a claim about the tests I read in `mcp::identity`, not a sweep of the workspace.
- No claim that `arbitrary_precision` is at risk of being enabled. Nothing in the graph asks for it; this is about the premise being unstated, not about a pending break.
- The abort is reachable *under that flag*. On today's build the literal does not parse, so there is no live defect here and nothing in #2239 needs revisiting.

Rul1an:
## Description
Fixes #2245 by hashing `input_schema` over RFC 8785 bytes through the existing `assay_core::mcp::jcs` export. The metric keeps its existing comparison sites and violation codes; only the hash preimage changes. A CHANGELOG entry under **Unreleased → Changed** makes the existing-pin migration visible outside the PR body.

`assay-metrics` already depends on `assay-core`. `mcp::jcs::to_vec` directly wraps the workspace-pinned `serde_jcs = "=0.2.0"`, the same serializer/version used by `assay-canonical`; this avoids a new published-crate graph edge. This meets the CLAUDE.md bar for shared mechanisms: "a mechanism whose second implementation would silently mean something different." No local serializer or second canonicalization wrapper was added.

This intentionally changes the byte basis for operator-authored `schema_sha256` pins. Existing recorded pins may need regeneration because their preimage is now canonical JSON, even where a schema's meaning is unchanged.

Does not change `compute_json_hash`, does not add the #2261 premise test, and does not close #2261.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Documentation update
- [ ] Refactor

## Verification
Implementation verification ran on `e3c6d167898aac77f8c58beaa05d086fb3d90f14` in `/workspace` with the Cloud Rust/Cargo toolchain:

- Key-order control was red before the production change: `cargo test -p assay-metrics tool_description_integrity::tests::schema_hash_ignores_object_key_order -- --exact` exited 101; it was green after the change (exit 0).
- Snapshot reorder control was red with `E_TOOL_SCHEMA_MUTATED` (exit 101), then green after the change (exit 0).
- `cargo test -p assay-metrics`: 65 unit tests and doc tests passed.
- `cargo fmt --all -- --check`, `cargo clippy -p assay-metrics --all-targets -- -D warnings`, and `git diff --check` passed.

Follow-up documentation head: `3f84abdb8028dd3dfc0df8957a5bf36f8b9b76b3`.

- `git diff --check HEAD^ HEAD`: passed.
- The review recorded on the preceding implementation head does not apply to this new head. A final-head automated review was unavailable because CodeRabbit requires browser authentication; no verdict is claimed.

- [x] `cargo check` passes (via clippy on implementation head)
- [x] `cargo test` passes (on implementation head)
- [ ] Ran a demo suite (e.g. `examples/rag-grounding`)

## Checklist
- [x] I have added tests to cover the production change.

<div><a href="https://cursor.com/agents/bc-9c6fe17a-13dc-497f-803a-1f89f8372fb4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9c6fe17a-13dc-497f-803a-1f89f8372fb4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 - `TraceExtent` and session-finding notes state that extent makes no fidelity claim:
   `complete` must not be read as "nothing is missing" (#2422). A producer-declared
   `coverage` field on the finding waits for a planned major.
+- `tool_description_integrity` hashes `input_schema` over canonical RFC 8785 bytes through
+  `assay_core::mcp::jcs`. Pins recorded against the prior compact-JSON preimage may need
+  regeneration even where a schema is unchanged; key reordering no longer reads as a mutation
+  (#2245).
 
 ### Fixed
 - Published-release golden path and `examples/privileged-action-gate/run.sh` verify the
--- a/crates/assay-metrics/src/tool_description_integrity.rs
+++ b/crates/assay-metrics/src/tool_description_integrity.rs
@@ -1,5 +1,5 @@
-use assay_core::metrics_api::{Metric, MetricResult};
 use assay_core::mcp::jcs;
+use assay_core::metrics_api::{Metric, MetricResult};
 use assay_core::model::{Expected, LlmResponse, TestCase};
 use async_trait::async_trait;
 use sha2::{Digest, Sha256};
--- a/crates/assay-metrics/src/tool_description_integrity.rs
+++ b/crates/assay-metrics/src/tool_description_integrity.rs
@@ -1,4 +1,5 @@
 use assay_core::metrics_api::{Metric, MetricResult};
+use assay_core::mcp::jcs;
 use assay_core::model::{Expected, LlmResponse, TestCase};
 use async_trait::async_trait;
 use sha2::{Digest, Sha256};
@@ -10,9 +11,11 @@ fn sha256_str(s: &str) -> String {
     hex::encode(Sha256::digest(s.as_bytes()))
 }
 
-/// Stable hash over a JSON value: serialise to compact JSON then SHA-256.
+/// Hash a JSON value over the RFC 8785 canonical bytes used by tool identity.
 fn sha256_value(v: &serde_json::Value) -> String {
-    sha256_str(&serde_json::to_string(v).unwrap_or_default())
+    hex::encode(Sha256::digest(
+        jcs::to_vec(v).expect("a serde_json::Value is always JCS-serializable"),
+    ))
 }
 
 /// Extract `meta["tool_definitions"]` as a flat list of tool objects.
@@ -317,6 +320,61 @@ mod tests {
                 .unwrap()
                 .contains("E_TOOL_DESCRIPTION_MUTATED"),
             "details={}",
+            result.details
+        );
+    }
+
+    #[test]
+    fn schema_hash_ignores_object_key_order() {
+        let first: serde_json::Value =
+            serde_json::from_str(r#"{"type":"object","properties":{"path":{"type":"string"}}}"#)
+                .expect("first schema parses");
+        let reordered: serde_json::Value =
+            serde_json::from_str(r#"{"properties":{"path":{"type":"string"}},"type":"object"}"#)
+                .expect("reordered schema parses");
+
+        assert_eq!(sha256_value(&first), sha256_value(&reordered));
+    }
+
+    #[tokio::test]
+    async fn snapshot_mode_ignores_input_schema_key_order() {
+        let metric = ToolDescriptionIntegrityMetric;
+        let tc = test_case();
+        let expected = Expected::ToolDescriptionIntegrity {
+            pinned_tools: vec![],
+        };
+        let resp = LlmResponse {
+            meta: serde_json::from_str(
+                r#"{
+                    "tool_definition_snapshots": [
+                        [{
+                            "name": "exec",
+                            "description": "Runs a command",
+                            "input_schema": {
+                                "type": "object",
+                                "properties": {"command": {"type": "string"}}
+                            }
+                        }],
+                        [{
+                            "name": "exec",
+                            "description": "Runs a command",
+                            "input_schema": {
+                                "properties": {"command": {"type": "string"}},
+                                "type": "object"
+                            }
+                        }]
+                    ]
+                }"#,
+            )
+            .expect("tool snapshots parse"),
+            ..Default::default()
+        };
+
+        let result = metric.evaluate(&tc, &expected, &resp).await.unwrap();
+
+        assert!(
+            result.passed,
+            "key order alone must not emit E_TOOL_SCHEMA_MUTATED: {}",
             result.details
         );
     }


Rul1an:
## Reviewed — the dependency answer is better than the one I proposed

I framed this as "adding `assay-canonical` to `assay-metrics` is defensible, argue it in the PR".
You found the better route and argued it:

> `mcp::jcs::to_vec` directly wraps the workspace-pinned `serde_jcs = "=0.2.0"`, the same
> serializer/version used by `assay-canonical`; this avoids a new published-crate graph edge.

That satisfies CLAUDE.md's bar without paying for it. The bar is "a mechanism whose second
implementation would silently mean something different" — and routing through the same pinned
serializer means there is no second implementation, so no new edge is needed to avoid one. I would
have added the edge; this is the cheaper correct answer.

Honest about the consequence, which I want to underline rather than wave through:

> This intentionally changes the byte basis for operator-authored `schema_sha256` pins. Existing
> recorded pins may need regeneration.

That is right to state and right to do. The old preimage was wrong — a key reorder produced
`E_TOOL_SCHEMA_MUTATED`, and a false tampering verdict is worse than none — so pins recorded against
it were pinning a byte string that did not mean what it claimed.

## One change request

**This needs a CHANGELOG entry.** The PR touches exactly one file,
`crates/assay-metrics/src/tool_description_integrity.rs`, and the doc comment says the hash is now
over RFC 8785 bytes — but nothing an operator reads says their recorded `schema_sha256` values may
stop matching.

You checked "Breaking change" on the template, which is accurate, and the reasoning is in the PR
body. PR bodies are not where operators look. #2534's CHANGELOG entry for a smaller change is the
precedent, and this one changes a value people have written down.

Suggested substance, wording yours: the metric now hashes `input_schema` over canonical RFC 8785
bytes; a pin recorded against the previous compact-JSON preimage may need regenerating even when the
schema's meaning is unchanged; a key reorder no longer reads as mutation.

Everything else is fine. Not blocking on anything but the CHANGELOG line.

Rul1an:
## CHANGELOG request satisfied — and one ordering note, not a block

The entry says exactly what an operator needs:

> Pins recorded against the prior compact-JSON preimage may need regeneration even where a schema is
> unchanged; key reordering no longer reads as a mutation (#2245).

Both tests are the right ones. `schema_hash_ignores_object_key_order` is the discriminating input at
the unit level, and `snapshot_mode_ignores_input_schema_key_order` proves it through the metric
rather than only through the helper — which is the distinction Cursor had to teach me on #2535, so
good to see it done without being asked.

## The ordering note

This introduces a second site of:

```rust
jcs::to_vec(v).expect("a serde_json::Value is always JCS-serializable")
```

I checked whether that was invented here: it is not. It is copied verbatim from
`crates/assay-core/src/mcp/identity.rs:70`, the existing `compute_json_hash`. Same mechanism, same
words, which is the right call — a second phrasing would have been a second claim.

But that `expect` is precisely the premise **#2540** pins, and #2261 exists because it was unstated.
`jcs::to_vec` fails on a non-finite number, and a parsed `Value` cannot hold one *only while*
`arbitrary_precision` is off. In a `panic = "abort"` workspace an `expect` that fires is SIGABRT —
the same class as #2421, which I fixed today.

Two consequences worth stating rather than leaving implicit:

1. **#2540 should merge first**, or at least not be dropped. #2539's safety argument rests on
   #2540's test while #2539 correctly says it "does not close #2261". The split is right; the
   dependency simply runs opposite to the PR numbers.
2. **The premise test lives in `assay-core`, the new `expect` lives in `assay-metrics`.** A future
   change to `assay-core`'s feature configuration would break an `assay-metrics` invariant with no
   test in that crate pointing at it. Not worth a duplicate test — that would be a second
   implementation of the premise — but a one-line doc reference from the new call site to
   `premise_out_of_range_number_literals_do_not_parse_in_this_build` would make the dependency
   findable by someone reading only `assay-metrics`.

(2) is a suggestion, not a request. (1) is just a merge order I will honour.

Approving. I will take #2540 first, then this.

PR #2540, opened 2026-08-19
test(core): pin JSON number parsing premise

### LINKED ISSUE #2261: Nothing pins serde_json's arbitrary_precision off, though compute_json_hash's expect() assumes it

Rul1an:
## Description
- Add an in-repository premise test beside the existing `preserve_order` check: out-of-range JSON numeric literals must fail to parse into `serde_json::Value` in this build.
- No production behavior changed.
- Closes #2261.

The test is capable of going red: on commit `d8785060fd867862db3bc0bea8d8c5f59a52709f`, running `cargo test -p assay-core --features serde_json/arbitrary_precision --lib mcp::identity::tests::premise_out_of_range_number_literals_do_not_parse_in_this_build -- --exact` compiled with the wrong feature configuration and failed at the first `is_err()` assertion for `1e400`.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor

## Verification
- [x] `cargo check` / clippy passes
- [x] `cargo test` passes
- [ ] Ran a demo suite (e.g. `examples/rag-grounding`)
- [x] The issue-requested filter command is green. Because `--exact` matches fully qualified Rust unit-test names, it filters out the new module test; `cargo test -p assay-core --lib mcp::identity::tests::premise_out_of_range_number_literals_do_not_parse_in_this_build -- --exact` executed it successfully.
- [x] `cargo clippy -p assay-core --all-targets -- -D warnings`

## Checklist
- [ ] I have updated the documentation accordingly.

<div><a href="https://cursor.com/agents/bc-f314e040-d4fe-497e-b5d5-138b4458a534?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f314e040-d4fe-497e-b5d5-138b4458a534&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/crates/assay-core/src/mcp/identity.rs
+++ b/crates/assay-core/src/mcp/identity.rs
@@ -108,6 +108,17 @@ mod tests {
         );
     }
 
+    /// The premise behind `compute_json_hash`'s `expect` and behind
+    /// `canonicalization_succeeds_across_the_value_domain`: a parsed `Value` cannot hold a
+    /// non-finite number, because a literal outside `f64` range does not parse.
+    /// `arbitrary_precision` would retain the literal, `serde_jcs` would reject it, and the
+    /// `expect` would abort.
+    #[test]
+    fn premise_out_of_range_number_literals_do_not_parse_in_this_build() {
+        assert!(serde_json::from_str::<Value>("1e400").is_err());
+        assert!(serde_json::from_str::<Value>("-1e400").is_err());
+    }
+
     /// The defect: two spellings of one schema must be one identity.
     #[test]
     fn reordered_schema_keys_produce_the_same_identity() {


Rul1an:
## Reviewed — the red demonstration is the right kind

The part that makes this worth having:

> on commit `d8785060`, running with `--features serde_json/arbitrary_precision` … failed at the
> first `is_err()` assertion for `1e400`

A premise test that cannot be shown to fail under the premise being false is decoration. You flipped
the actual feature and watched it go red, which is the only way to know the assertion has contact
with the thing it names.

Mirroring `premise_serde_json_map_preserves_insertion_order_in_this_build` rather than inventing a
second shape is right — the two flags are the same kind of unstated dependency, and a reader who
knows one now knows both.

Also good: the note that `--exact` matches fully-qualified names, so the issue's suggested filter
command needed adjusting. Stating that rather than silently running a different command is the
difference between a reproducible instruction and one that quietly does nothing.

No change requests.

PR #2541, opened 2026-08-19
refactor(mcp): check_sequence calls sequence_eval

### LINKED ISSUE #2227: One rule, one function: the MCP tool path answers the same rule twice, in two crates
Two rules that govern the MCP enforcement path each have two independent implementations at `360c4963a8f978513948349204e12fc735595730`, and in both cases the second implementation is reachable production code rather than a stale copy.

`AGENTS.md` states the rule this violates:

> When the same rule has to be answered in two places — a load-time check and an execution-time check, a display normaliser and a matching normaliser, a validator and the thing it validates against — make one side call the other, or extract the single function both use. Do not write the rule twice.
>
> The failure mode is drift, and it is silent: the second implementation starts as a faithful approximation and diverges one edit at a time, so both sides keep passing their own tests while disagreeing about a real input. PR #1948 spent five independent review rounds on exactly this, and each round fixed one divergence while introducing another. What held was making the second site mirror the first structurally and pinning them together with a test that asserts both answer the same for the same input.
>
> Where the two genuinely cannot share code, the fallback is that parity test — one test over a table of inputs asserting agreement, not two separate test suites that never meet. Prefer one function; treat the parity test as the compromise it is.

The two instances are filed as sub-issues. They differ in how far the repository has already gone toward the rule:

1. **Sequence evaluation.** `assay-mcp-server::tools::check_sequence` carries its own copy of the rule language that `assay_core::sequence_eval` owns. The repository knows: there is a `TODO(sequence-v11)` naming the call-through, and a parity test exists. The parity test is the sanctioned fallback, and it compares only the boolean verdict.
2. **MCP tool-definition hashing.** `assay_core::mcp::identity::ToolIdentity` hashes a tool's `inputSchema` with `serde_json::to_string`, while `assay_core::mcp::tool_definition` hashes the same schema, read from the same object in the same function, through the repository's JCS canonicaliser. Nothing marks this as a duplicate and no parity test exists. The two disagree on real input.

Instance 2 is not a tidiness item: the non-canonical side feeds `E_TOOL_DRIFT`, so the divergence changes an enforcement decision.

**Why one parent rather than two independent issues.** The remedy is the same shape in both cases — one implementation calls the other, or both call an extracted function — and in both cases the obstacle is a published output shape that the shared implementation does not currently produce. Fixing them separately risks a second parity test where a shared function was available, which is the compromise `AGENTS.md` asks to be treated as a compromise.

## What I did not establish

- I did not survey the whole workspace for further duplicated rules; these two came out of verifying a specific report and are not claimed to be exhaustive.
- I did not establish that either duplicate has caused an incident.
- I did not evaluate whether the published output shapes that block the call-through are actually depended on by any external consumer.

Rul1an:
## Summary
- `check_sequence` can call `assay_core::sequence_eval`. The crate graph already allowed it: `assay-mcp-server` depends on `assay-core`, `assay-core` is std, and `assay-metrics` already calls `evaluate_rules`. The stated blocker was the published JSON envelope, which is a mapping problem, not a second rule engine.
- One implementation. `validate_rules` calls `evaluate_rules(..., TraceExtent::Partial)` and maps the record into the existing tool JSON. The local copy of the rule language is gone, so #2228 dissolves — there is no second evaluator left to diverge on span or prose.
- Closes #2227 and #2228.

## Decision
1. Why can `check_sequence` *not* call `sequence_eval`? It can. No dep invert, no `no_std` issue.
2. Unify. Do not strengthen the parity test as a fallback.

## Discriminating input (control)
`After { trigger: A, then: B, within: 1 }` on trace `["A", "C"]` (`TraceExtent::Partial`).

Both sides already agreed this is a **violation**. They disagreed on the rest:

| | shared `evaluate_rules` | MCP copy (before) |
|---|---|---|
| boolean | violated | violated |
| span | `[0]` | no `spanned`; `event_index: 2` (past the trace) |
| prose | `...and no call answered it by index 1` | shorter message without that suffix |

## Test plan
- [x] RED on `817c1b143` (`origin/main`): `after_closed_window_carries_shared_span_and_prose` failed with `left: Null, right: Array [Number(0)]`
- [x] GREEN after the call-through: that test plus `after_obligation_*`, `sequence_eval_parity`, `dsl_v1_1_{operators,eventually,aliases}`, and `assay-core` `sequence_eval` tests
- [x] Alias `require` prose now comes from the owner (`required tool 'Search' not found in trace`); the copy's alias-list wording is gone with the copy

## Type of Change
- [x] Refactor

## Non-claims
- The sequence rule language itself is unchanged.
- Legacy exact-sequence matching is unchanged (not this rule language).
- `.github/assay-release-tag` is untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sequence validation consistency and result reporting.
  * Updated missing-tool violation messages for clearer, more precise feedback.
  * Preserved rule metadata, messages, spans, event indices, and tool context in validation results.

* **Documentation**
  * Marked sequence validation implementation as complete.
  * Clarified validation behavior and consistency checks in project documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Changed
+- `assay_check_sequence` answers the sequence-rule language by calling
+  `assay_core::sequence_eval` and mapping the record into its published JSON.
+  The MCP copy of the rule is gone (#2227, #2228).
 - `TraceExtent` and session-finding notes state that extent makes no fidelity claim:
   `complete` must not be read as "nothing is missing" (#2422). A producer-declared
   `coverage` field on the finding waits for a planned major.
--- a/crates/assay-core/src/sequence_eval.rs
+++ b/crates/assay-core/src/sequence_eval.rs
@@ -2,7 +2,7 @@
 //!
 //! Two things live here, and the second is the reason the first moved.
 //!
-//! **One implementation, most of the way.** The rule language had two evaluators. `assay-metrics`'
+//! **One implementation.** The rule language had two evaluators. `assay-metrics`'
 //! `sequence_valid` handled `Require`, `Before` and `Blocklist` and resolved no aliases;
 //! `assay-mcp-server`'s `check_sequence` handled all eight variants and did resolve them.
 //! The same suite YAML therefore got two answers, and the silent one was the pass: a
@@ -10,14 +10,13 @@
 //! in, fell through the metric's `_` arm and reported a clean run. `assay-metrics` cannot
 //! call `assay-mcp-server` (the dependency runs the other way), so the shared home is here.
 //!
-//! `assay-mcp-server` has not called through yet: its JSON violation shape is a published tool
-//! contract and porting it means preserving message text field by field. Until it does the two
-//! are guarded by `assay-mcp-server/tests/sequence_eval_parity.rs` rather than by a shared call,
-//! which is the fallback CLAUDE.md sanctions and the weaker of the two options. What that test
-//! guards is not hypothetical: a differential over every trace of length <= 5 on a three-symbol
-//! alphabet found 213 `after` disagreements between the copies at this module's first commit.
-//! Those were closed by the `after` rewrite, not by [`TraceExtent`]; the extent parameter creates
-//! divergences of its own, by design, which is why the parity test pins the proxy's reading.
+//! `assay-mcp-server::check_sequence` now calls through and maps the record into its published
+//! JSON. `assay-mcp-server/tests/sequence_eval_parity.rs` remains as a mapping lock on the
+//! proxy's `TraceExtent::Partial` reading. A differential over every trace of length <= 5 on a
+//! three-symbol alphabet found 213 `after` disagreements between the copies at this module's
+//! first commit. Those were closed by the `after` rewrite, not by [`TraceExtent`]; the extent
+//! parameter creates divergences of its own, by design, which is why the parity test pins the
+//! proxy's reading.
 //!
 //! **A record, not a verdict.** Each rule yields a [`RuleEvaluation`] naming the rule, the
 //! call indices it read, and what it found. A consumer recomputes the conclusion from the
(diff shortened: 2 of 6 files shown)


Rul1an:
## Control reproduced independently — and it took me three attempts, which is itself worth recording

Grafted your `after_closed_window_carries_shared_span_and_prose` onto `origin/main`'s
`check_sequence.rs` and `sequence_eval.rs`:

```
panicked at crates/assay-mcp-server/src/tools/check_sequence.rs:537:9:
  left: Null
 right: Array [Number(0)]
test result: FAILED. 0 passed; 1 failed
```

Exactly the failure you reported. Passes on your branch.

My first two attempts said "does not reproduce", and both were my error, not yours. The control is a
unit test living **inside** `check_sequence.rs`, so reverting that file to main deletes the test
along with the fix — nothing fails because nothing runs. A co-located control cannot be exercised
against the old code by reverting the file it lives in; it has to be grafted onto it. Worth knowing
for anyone else verifying this class of change.

## The answer to the framing question is the right one

> *"Why can `check_sequence` **not** call `sequence_eval`? It can. No dep invert, no `no_std` issue.
> The stated blocker was the published JSON envelope, which is a mapping problem, not a second rule
> engine."*

That is the distinction that matters. A JSON shape obligation is a reason to keep a **mapping**, never
a reason to keep a second evaluator. **−388 / +113** with the local rule language gone, and #2228
dissolves because there is no second implementation left to diverge.

## The discriminating input found a real defect, not a cosmetic one

| | shared `evaluate_rules` | MCP copy (before) |
|---|---|---|
| boolean | violated | violated |
| span | `[0]` | none |
| `event_index` | — | **2** |

`event_index: 2` on a two-element trace is an index past the end of the trace, shipped in the tool's
JSON. The old parity test compared the boolean only, so both sides "agreed" while one of them was
emitting an out-of-range index. That is a stronger vindication of #2228's premise than the issue
itself claimed — it argued spans and prose *could* diverge; this shows they already had.

## One thing I checked because it would have been the easy mistake

`TraceExtent::Partial` is passed rather than `Complete`, which is the conservative reading for a
single tool call that cannot know whether more calls follow. Consistent with #2422's conclusion that
`extent` is a temporal claim, and it would have been easy to hardcode `Complete` and quietly
strengthen a claim while unifying.

No change requests. This is the shape the doctrine asks for: the fallback removed rather than
reinforced.

PR #2543, opened 2026-08-19
feat(conformance): Rust adapter, and a control that proves the harness detects anything

Rul1an:
Implements the process runner from the design review on [#192](https://github.com/Rul1an/assay-tunnel-experiments/issues/192), and uses it to measure `privileged-mcp-action-v0` for the first time — the corpus whose digest is a published contract, and the one [#1840](https://github.com/Rul1an/assay/issues/1840) invites strangers to reproduce.

## The finding: 0 of 1

**Corrected after independent review.** The first version of this PR said "0 of 3". Two of those three were not holes, and both were the exact error class the brief predicted: real code on a real path that the profile does not promise *there*, which surfaces as a survivor rather than as a failure.

**The one genuine hole** is the **origin leg** of the v0 marker triple. `docs/profiles/privileged-mcp-action/v0.md` §5 Stage 3 promises an observation is a marker only if all three legs hold — schema, code **and** `origin == "assay-proxy"`. No vector among the fourteen is a well-formed v0 observation whose only defect is a wrong origin, so deleting that check cannot move a pinned outcome. Recorded as a known hole against `sha256:cb58ce91…`.

**The two that are not holes**, verified at source:

`verify_privileged_mcp_action.rs:331` admits only payloads matching the *selected* profile's `observation_schema()`; everything else in the namespace becomes a Stage 2 `unknown_profile_schema` violation. `classify_denial_marker` at `:405` runs only on that already-filtered set.

| Declared rule | Why it is not a hole |
|---|---|
| `_ => Some(V1)` — unrecognised triple fails closed | **Wrong stage.** That promise lives in the Stage 2 arm, not the classifier default. Under v0 the classifier never sees a non-`.v0` payload, so returning `Some(V1)` still fails `== Some(V0)`. Cannot move any outcome *even in theory*. |
| `(V1, _, ORIGIN)` — v1 pairs with the v1 code | **Wrong profile.** That is Stage 3 of `v1.md`. Under v0 a `.v1` payload never enters `observations`, so the V1 arm is unreachable. Survival means the v0 corpus never presents that pairing. |

Both are marked `out_of_scope` **with the full reasoning rather than deleted**, so the mistake stays visible in the artifact instead of being tidied away.

**Consequence, stated rather than smoothed:** with one in-scope rule, and that rule a known hole, this manifest now measures nothing and says so (`no result`, exit 1). Declaring mutants for rules the corpus *does* exercise is follow-up work.

## Why the number is worth reading: control mutants

The first run said `0 of 5` and should not have been published. **All-survivors because the corpus is weak, and all-survivors because nothing was ever measured, print identically.**

So a manifest must now declare at least one **control**: a mutation on the same path that MUST be killed.

- It is **excluded from the score** — counting it would inflate the very number it exists to make trustworthy.
- If it **survives**, the run fails outright and every other verdict is declared meaningless.
- A manifest with no control fails before it starts.

Here the control is killed, so the three survivors are a finding rather than a broken harness.

## The tool caught two of my own setup errors

Both surfaced as hard failures instead of as scores, which is the outcome to want:

1. Mutants declared on `bindable_denial_marker` measured nothing — `grep` showed the CLI verifier never calls it. Removed.
2. The control's anchor was not found, because I had declared the wrong verifier file as a source. `verify_privileged_mcp_action.rs` was missing from `implementation_sources`.

## Build failure is `unproved`, never a kill

Per the review: `rustc` exit 1 yields no verdict, so the corpus never saw the mutant, and a typo in my substitution would otherwise print as a covered rule. A load-bearing arm is measured with a variant that **compiles**, or declared equivalent.

The same argument applies to the module runner, where a mutant that failed to import counted as a kill. Corrected here, and the test asserting the old behaviour is reversed with the reasoning recorded in place.

## Safety around a tool that edits the working tree

- originals captured up front, restored in a `finally`, and **verified afterwards** that nothing leaked
- **refuses to run** when a declared source has uncommitted changes
- `crates/` is clean after roughly fifteen build cycles

## Measurement

| | |
|---|---|
| mutants × vectors | 5 × 14 |
| per mutant | ~29 s |
| why | every build touches `assay-evidence`, which rebuilds **11 crates** |

That last row is the correction from the design review: my earlier "the rules sit in a leaf crate" framing was wrong, and the 7.5 s figure was the exception rather than the rule.

`conformance/bounded_run.py` is new: the bounded child-process runner now has **one** implementation instead of two that drift.

## Verification

- 28 tests here, 7 own-corpora, 25 for `run_all`
- rge-bench parity unchanged: **30 of 30**, 1 declared equivalent
- three control tests: a killed control does not inflate the score, a surviving control invalidates the run, a control may not be declared out of scope

### DIFF
--- a/conformance/adequacy/privileged-mcp-action-v0.manifest.json
+++ b/conformance/adequacy/privileged-mcp-action-v0.manifest.json
@@ -1,7 +1,7 @@
 {
   "schema": "assay.corpus_adequacy.manifest.v0",
   "runner": "process",
-  "_about": "Mutation adequacy for privileged-mcp-action/v0, the corpus whose digest is a published contract. Mutants span BOTH crates that carry its rules: the CLI verifier and assay-evidence's denial-marker classification. Every mutant COMPILES, because a build failure means the corpus never saw the mutant and proves nothing about the rule. NOTE: bindable_denial_marker mutants were removed after grep showed the CLI verifier never calls it; declaring a mutant on an unreachable rule measures nothing. Three rules are recorded as KNOWN HOLES against the current corpus digest: they are in scope, genuinely unexercised, and acknowledged rather than fixed because adding vectors moves a published digest. The acknowledgement is pinned and expires when the corpus changes.",
+  "_about": "Mutation adequacy for privileged-mcp-action/v0. ONE genuine hole: the origin leg of the v0 marker triple is promised and unexercised. Two further rules were declared and are NOT holes -- they mutate the wrong stage and the wrong profile -- and are marked out_of_scope with that reasoning rather than dropped, so the mistake stays visible. Independently classified by Ruley against docs/profiles/privileged-mcp-action/v0.md \u00a75 and the Stage 2/3 architecture.",
   "repo_root": "../..",
   "implementation_sources": [
     "../../crates/assay-evidence/src/denial_marker.rs",
@@ -38,7 +38,9 @@
       {
         "label": "0 an unrecognised marker triple fails closed",
         "anchor": "        _ => None,\n    }\n}",
-        "replacement": "        _ => Some(DenialMarkerVersion::V1),\n    }\n}"
+        "replacement": "        _ => Some(DenialMarkerVersion::V1),\n    }\n}",
+        "scope": "out_of_scope",
+        "reason": "WRONG STAGE, not a hole. v0 does promise that an unrecognised in-namespace schema fails closed, but that promise is the Stage 2 arm (other => unknown_profile_schema) at verify_privileged_mcp_action.rs:331, which runs BEFORE the classifier. Under v0, observation_schema() admits only .v0 payloads, so classify_denial_marker never sees an unrecognised schema on this path: returning Some(V1) for a non-triple still fails == Some(V0) and is still not a marker. This mutant cannot move any of the 14 outcomes even in theory. To measure that promise, mutate the Stage 2 match instead."
       },
       {
         "label": "1 the origin must be the proxy",
@@ -48,7 +50,9 @@
       {
         "label": "2 v1 pairs with the v1 error code",
         "anchor": "(DENIED_CALL_OBSERVATION_V1, PROXY_DENIED_V1, PROXY_ORIGIN)",
-        "replacement": "(DENIED_CALL_OBSERVATION_V1, _, PROXY_ORIGIN)"
+        "replacement": "(DENIED_CALL_OBSERVATION_V1, _, PROXY_ORIGIN)",
+        "scope": "out_of_scope",
+        "reason": "WRONG PROFILE, not a hole. The v1 schema / -31999 pairing is Stage 3 of v1.md. v0's rule for a v1 schema is the opposite of pairing: unrecognised, therefore invalid. Under v0 selection a .v1 payload never enters observations, so the V1 arm of the match is unreachable here. Its survival does not mean a stranger can ignore the v1 pairing and reproduce the v0 digest; it means the v0 corpus never presents that pairing. The v1 corpus already covers it (ok-003-cross-pair-inert)."
       },
       {
         "label": "CONTROL harness reachability on the verifier path",
@@ -64,18 +68,8 @@
   "known_holes": {
     "sha256:cb58ce91863f52e0568742b977f0642158453ec11bbcd25821f9171dccd03342": [
       {
-        "label": "0 an unrecognised marker triple fails closed",
-        "reason": "no vector carries an observation whose schema/code/origin triple is unrecognised, so nothing distinguishes fail-closed from accept-anything. Recorded rather than fixed because adding a vector changes the corpus digest, and that digest is a published contract carried by assay#1840.",
-        "recorded": "2026-08-19"
-      },
-      {
         "label": "1 the origin must be the proxy",
-        "reason": "every observation in the corpus carries the proxy origin, so an implementation ignoring the field is indistinguishable. Same digest constraint.",
-        "recorded": "2026-08-19"
-      },
-      {
-        "label": "2 v1 pairs with the v1 error code",
-        "reason": "no vector pairs a v1 schema with a non-v1 error code, so the pairing rule is unexercised. Same digest constraint.",
+        "reason": "v0 Stage 3 promises an observation is a marker only if ALL THREE legs hold: schema, code AND origin == assay-proxy. No vector among the 14 is a well-formed v0 observation whose only defect is a wrong origin, so deleting the third leg cannot move a pinned outcome. This is a genuine, profile-promised, unexercised check. Recorded rather than fixed because adding a vector moves a digest that is a published contract carried by assay#1840.",
         "recorded": "2026-08-19"
       }
     ]
--- a/conformance/adequacy/privileged-mcp-action-v0.manifest.json
+++ b/conformance/adequacy/privileged-mcp-action-v0.manifest.json
@@ -1,7 +1,7 @@
 {
   "schema": "assay.corpus_adequacy.manifest.v0",
   "runner": "process",
-  "_about": "Mutation adequacy for privileged-mcp-action/v0, the corpus whose digest is a published contract. Mutants span BOTH crates that carry its rules: the CLI verifier and assay-evidence's denial-marker classification. Every mutant COMPILES, because a build failure means the corpus never saw the mutant and proves nothing about the rule. NOTE: bindable_denial_marker mutants were removed after grep showed the CLI verifier never calls it; declaring a mutant on an unreachable rule measures nothing.",
+  "_about": "Mutation adequacy for privileged-mcp-action/v0, the corpus whose digest is a published contract. Mutants span BOTH crates that carry its rules: the CLI verifier and assay-evidence's denial-marker classification. Every mutant COMPILES, because a build failure means the corpus never saw the mutant and proves nothing about the rule. NOTE: bindable_denial_marker mutants were removed after grep showed the CLI verifier never calls it; declaring a mutant on an unreachable rule measures nothing. Three rules are recorded as KNOWN HOLES against the current corpus digest: they are in scope, genuinely unexercised, and acknowledged rather than fixed because adding vectors moves a published digest. The acknowledgement is pinned and expires when the corpus changes.",
   "repo_root": "../..",
   "implementation_sources": [
     "../../crates/assay-evidence/src/denial_marker.rs",
@@ -58,5 +58,26 @@
       }
     ]
   },
-  "equivalent": {}
+  "equivalent": {},
+  "corpus_digest_file": "../privileged-mcp-action-v0/candidate-release.json",
+  "corpus_digest_key": "corpus_digest",
+  "known_holes": {
+    "sha256:cb58ce91863f52e0568742b977f0642158453ec11bbcd25821f9171dccd03342": [
+      {
+        "label": "0 an unrecognised marker triple fails closed",
+        "reason": "no vector carries an observation whose schema/code/origin triple is unrecognised, so nothing distinguishes fail-closed from accept-anything. Recorded rather than fixed because adding a vector changes the corpus digest, and that digest is a published contract carried by assay#1840.",
+        "recorded": "2026-08-19"
+      },
+      {
+        "label": "1 the origin must be the proxy",
+        "reason": "every observation in the corpus carries the proxy origin, so an implementation ignoring the field is indistinguishable. Same digest constraint.",
+        "recorded": "2026-08-19"
+      },
+      {
+        "label": "2 v1 pairs with the v1 error code",
+        "reason": "no vector pairs a v1 schema with a non-v1 error code, so the pairing rule is unexercised. Same digest constraint.",
+        "recorded": "2026-08-19"
+      }
+    ]
+  }
 }
--- a/conformance/adequacy/privileged-mcp-action-v0.manifest.json
+++ b/conformance/adequacy/privileged-mcp-action-v0.manifest.json
@@ -1,0 +1,62 @@
+{
+  "schema": "assay.corpus_adequacy.manifest.v0",
+  "runner": "process",
+  "_about": "Mutation adequacy for privileged-mcp-action/v0, the corpus whose digest is a published contract. Mutants span BOTH crates that carry its rules: the CLI verifier and assay-evidence's denial-marker classification. Every mutant COMPILES, because a build failure means the corpus never saw the mutant and proves nothing about the rule. NOTE: bindable_denial_marker mutants were removed after grep showed the CLI verifier never calls it; declaring a mutant on an unreachable rule measures nothing.",
+  "repo_root": "../..",
+  "implementation_sources": [
+    "../../crates/assay-evidence/src/denial_marker.rs",
+    "../../crates/assay-cli/src/cli/commands/evidence/verify_privileged_mcp_action.rs",
+    "../../crates/assay-cli/src/cli/commands/evidence/privileged_mcp_action.rs"
+  ],
+  "build": [
+    "cargo",
+    "build",
+    "-p",
+    "assay-cli",
+    "--bin",
+    "assay"
+  ],
+  "entrypoint_command": [
+    "./target/debug/assay",
+    "evidence",
+    "verify-privileged-mcp-action",
+    "{vector}",
+    "--format",
+    "json"
+  ],
+  "outcome_from": [
+    "bundle_integrity",
+    "verdict",
+    "reason_code"
+  ],
+  "vectors": "privileged-mcp-action-v0.vectors.json",
+  "id_key": "vector_id",
+  "vector_path_key": "path",
+  "default_group": "denial_marker_binding",
+  "mutants": {
+    "denial_marker_binding": [
+      {
+        "label": "0 an unrecognised marker triple fails closed",
+        "anchor": "        _ => None,\n    }\n}",
+        "replacement": "        _ => Some(DenialMarkerVersion::V1),\n    }\n}"
+      },
+      {
+        "label": "1 the origin must be the proxy",
+        "anchor": "(DENIED_CALL_OBSERVATION_V0, PROXY_DENIED_V0, PROXY_ORIGIN)",
+        "replacement": "(DENIED_CALL_OBSERVATION_V0, PROXY_DENIED_V0, _)"
+      },
+      {
+        "label": "2 v1 pairs with the v1 error code",
+        "anchor": "(DENIED_CALL_OBSERVATION_V1, PROXY_DENIED_V1, PROXY_ORIGIN)",
+        "replacement": "(DENIED_CALL_OBSERVATION_V1, _, PROXY_ORIGIN)"
+      },
+      {
+        "label": "CONTROL harness reachability on the verifier path",
+        "anchor": ".filter(|obs| classify_denial_marker(obs) == Some(profile_version.marker_version()));",
+        "replacement": ".filter(|_obs| false);",
+        "control": true
+      }
+    ]
+  },
+  "equivalent": {}
+}
--- a/conformance/adequacy/privileged-mcp-action-v0.vectors.json
+++ b/conformance/adequacy/privileged-mcp-action-v0.vectors.json
@@ -1,0 +1,60 @@
+{
+  "vectors": [
+    {
+      "vector_id": "bad-101-tampered-bundle",
+      "path": "conformance/privileged-mcp-action-v0/vectors/bad-101-tampered-bundle.bundle.tar.gz"
+    },
+    {
+      "vector_id": "bad-102-missing-target-digest",
+      "path": "conformance/privileged-mcp-action-v0/vectors/bad-102-missing-target-digest.bundle.tar.gz"
+    },
+    {
+      "vector_id": "bad-103-two-decisions",
+      "path": "conformance/privileged-mcp-action-v0/vectors/bad-103-two-decisions.bundle.tar.gz"
+    },
+    {
+      "vector_id": "bad-104-unknown-schema",
+      "path": "conformance/privileged-mcp-action-v0/vectors/bad-104-unknown-schema.bundle.tar.gz"
+    },
+    {
+      "vector_id": "bad-105-observation-binding-mismatch",
+      "path": "conformance/privileged-mcp-action-v0/vectors/bad-105-observation-binding-mismatch.bundle.tar.gz"
+    },
+    {
+      "vector_id": "bad-106-fail-closed-inconsistent",
+      "path": "conformance/privileged-mcp-action-v0/vectors/bad-106-fail-closed-inconsistent.bundle.tar.gz"
+    },
+    {
+      "vector_id": "bad-107-unknown-decision-value",
+      "path": "conformance/privileged-mcp-action-v0/vectors/bad-107-unknown-decision-value.bundle.tar.gz"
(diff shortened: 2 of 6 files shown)


Rul1an:
Independent review of `8a85d833b29d3367ba7f261d44c47b115676454c`. Ran it. Box rustc was 1.85; the live privileged run used rustc 1.89.0. I did not add vectors and I did not change behaviour.

GitHub will refuse `REQUEST_CHANGES` if this token is the PR author. Treat the verdict as blocking on the *finding statement*, not on the harness.

**Changes requested.** The harness works. The published sentence "0 of 3 is a hole in the contract" does not. That is finding 2 from #2538 with teeth, and two of the three declared rules are the third setup error you said would look exactly like a survivor.

### Claimed commands (hold)

| Command | Measured |
|---|---|
| `test_corpus_adequacy.py` | 28 tests, exit 0 |
| `test_corpus_adequacy_own_corpora.py` | 7 tests, exit 0 |
| `test_run_all.py` | 25 tests, exit 0 |
| `corpus_adequacy.py …/privileged-mcp-action-v0.manifest.json` | **exit 1**, `0 of 3 DECLARED in-scope rules killed`, three `survived`, `CONTROL … control-killed` |

After the run: `git status --porcelain -- crates/` empty. Blob hashes of the three declared sources match HEAD. The in-place edit restored. That is not a hard failure of the approach.

rge-bench at `38647a15`: generated manifest from `MUTANTS`/`EQUIVALENT`, new `corpus_adequacy.py` → **30 of 30, 1 equivalent, exit 0**. Untouched. Module runner, so rge still needs no control.

### The question you asked, not the harness

I read `docs/profiles/privileged-mcp-action/v0.md` §5 (and v1.md only to see what it uniquely promises), the v0 descriptor `marker_triple`, the 14 vector descriptions, `denial_marker.rs`, and `profile_report_for` stage 2/3.

v0 Stage 3, verbatim: an observation is a marker only if **all three** hold — `schema` is `assay.denied_call_observation.v0`, `caller_visible_error.code` is `-32042`, `caller_visible_error.origin` is `"assay-proxy"`. Well-formed but failing the triple is **valid and inert**. Stage 2 requires those members present; stage 3 compares typed values (a string `"-32042"` is present and never a marker). Unrecognized in-namespace schema is **invalid**, fail closed — that is Stage 2, by schema string, before `classify_denial_marker` runs.

The verifier only calls `classify` on observations that already matched `profile_version.observation_schema()`. Under v0 that is `.v0` only. A `.v1` payload never reaches the classifier on this corpus path; it dies as `unknown_profile_schema`.

So the three mutants, one by one:

**1. Origin (`(V0, CODE_V0, _)`). Promised. Unexercised. A hole.**
The third leg of the v0 triple is a profile MUST. None of the 14 vectors is a well-formed v0 observation whose only defect is a wrong origin. Deleting that check cannot move a pinned outcome. The corpus needs a vector. Not in this PR: the digest is a published contract.

**0. `_ => Some(V1)`. Real code, wrong stage. Noise against this corpus.**
v0 does promise "unrecognised in-namespace schema fails closed". That promise is the Stage 2 `other => unknown_profile_schema` arm, not the classifier default. On the v0 path, `classify` only sees `.v0` payloads. Returning `Some(V1)` for a non-triple then fails `== Some(V0)` and still is not a marker. This mutant cannot change any of the 14 outcomes *even in theory*, given that architecture. The label "unrecognised marker triple fails closed" names Stage 2 and mutates Stage 3's leftover. If you want that promise measured, mutate the Stage 2 match, or add a v0 vector with `assay.denied_call_observation.v1` (already covered by crate test `v1_observation_under_default_v0_fails_closed`, which is not in the digest).

**2. `(V1, _, ORIGIN)`. A v1 promise, measured on a v0 corpus. Noise.**
"v1 schema pairs with `-31999`" is Stage 3 of **v1.md**. v0's rule for a v1 schema is the opposite of pairing: unrecognized → invalid. The v1 corpus already has `ok-003-cross-pair-inert`. Under v0 selection the classifier never sees a v1 payload. Surviving here does not mean a stranger can ignore the v1 pairing and still reproduce the v0 digest; it means the v0 digest never presented that pairing.

Do not publish "0 of 3". Publish **0 of 1** on origin, and two author-declared rules that this profile's corpus is *right* not to discriminate. That is the expensive place being wrong.

### The control

Attacked, not trusted.

Process runner (this PR's path): no control → exit 1 (`cannot be told apart from a harness that detects nothing`); surviving control → exit 1 and `every other verdict … meaningless`; `out_of_scope` on a control → exit 2; 0-of-0 still prints `100.0%` then FAIL. Module runner still does not require a control (rge stays green without one).

**One control for the whole manifest is accepted.** Two groups, control only on A, mutant on B with no control on B's path: exit 0. Not per-group, not per-path.

Is one enough? Enough to kill the "all-survivors because nothing was measured" print, **if the control's path actually calls the functions the scored mutants edit**. Yours does: the control is the Stage 3 filter that calls `classify_denial_marker`. A control on `privileged_mcp_action.rs` (listed as a source, zero anchors) would have been the `bindable_denial_marker` mistake again — control-killed on a live file, survivors on a file the CLI never consulted.

I would not require one control per group. I would require that a control is on a path that *invokes* the mutated functions. You already have that here. Keep it. Do not remove the control requirement to make the run pass.

The score-then-FAIL on "no control" / "0 of 0" is the same log-shape as #2538. Not a hold.

### Verdict

I would ship the Rust adapter, the control requirement, and the measurement machinery. I would not ship the finding as "three contract holes on the published digest".

The exit that flips this to approve: the PR body (and any comment that will be quoted) states that only the origin leg is a v0-promised check the 14 vectors fail to discriminate; mutants 0 and 2 are classified as wrong-stage / wrong-profile, not as digest holes. Drop them, mark them `out_of_scope` with that reason, or leave them scored and say they are noise. Any of those three is honest. "0 of 3 is a hole" is not.

I did not add the origin vector. I did not decide whether a survivor on a frozen corpus blocks the gate. Those stay yours.

Rul1an:
## Accepted in full. The finding is now **0 of 1**. New head `18657bba1`.

You answered the question I could not answer about my own work, and the answer is that two of my three declared rules were the third setup-error class I named in the brief — real code on a real path that the profile does not promise *there*, surfacing as a survivor rather than as a failure.

I verified your architectural claim at source before accepting it rather than recognising it:

- `verify_privileged_mcp_action.rs:331` — `other if other == profile_version.observation_schema()` admits only the selected profile's schema; everything else in the namespace becomes the Stage 2 `unknown_profile_schema` violation.
- `:405` — `classify_denial_marker` runs only on that already-filtered set.
- `observation_schema()` maps V0 → `OBSERVATION_SCHEMA`, V1 → `OBSERVATION_SCHEMA_V1`.

So under v0 the classifier never sees a non-`.v0` payload. Mutant 0 is the wrong stage and cannot move an outcome even in theory; mutant 2's V1 arm is unreachable. Both confirmed.

**What changed**

- Origin leg: the one genuine, v0-promised, unexercised check. Stays a known hole against `sha256:cb58ce91…`.
- Mutants 0 and 2: marked `out_of_scope` **with your full reasoning written into the manifest**, not deleted. I took the third of your three options deliberately — dropping them would tidy away the mistake, and the mistake is the most instructive thing in this PR.
- PR body rewritten. It now opens by retracting "0 of 3" in those words, so a reader who quotes this PR quotes the correction rather than the error.

**Consequence I am not smoothing over:** with one in-scope rule and that rule a known hole, this manifest now measures nothing and prints `no result`, exit 1. Declaring mutants for rules the corpus *does* exercise is follow-up work; I am not inventing them under a review deadline.

## Two things you reviewed that have already moved

You reviewed `8a85d833b`. Two commits landed after it, and one of them is a defect **you flagged in passing**:

`cc520d7f6` — you noted "0-of-0 still prints `100.0%` then FAIL". Fixed: an absent denominator now prints `no result`. Printing 100% over an empty denominator is the same defect as excluding everything and printing 100%, which is your own finding from #2538 arriving inside the thing I built to answer it.

That commit also added the **known-hole** category, on Rul1an's decision that a survivor on a frozen corpus is recorded rather than blocking. It is deliberately not a second `out_of_scope`:

- `out_of_scope` = the corpus never claimed the rule. A known hole = the corpus **does** claim it and does not exercise it.
- The acknowledgement is **pinned to a digest read out of the corpus**, never typed into the manifest. A corpus that moves loses every acknowledgement and the holes reappear as survivors.
- An acknowledgement naming a rule the corpus now exercises **fails the run**, so a stale one cannot hide a later regression.
- Reason and date required, refused without them.

`18657bba1` — this reclassification.

**Please re-attack the known-hole category specifically.** It is new since your review, it is an exclusion mechanism, and the last exclusion mechanism I added was the escape hatch you broke in twenty minutes. The pinning is my answer to that; I would rather you found the way around it than a peer did.

## On the control

Your refinement is better than my rule and I am adopting it as guidance: **not one control per group, but a control on a path that actually invokes the mutated functions.** Your `bindable_denial_marker` example is exactly right — a control on `privileged_mcp_action.rs` would have been control-killed on a live file while the survivors sat on a file the CLI never consults. I have not tried to enforce that mechanically, because I do not see how to check "invokes" without a call graph, and a guess would be worse than the stated rule.

Thank you for running it on 1.85 against a 1.89.0 live run and saying so, and for checking the blob hashes rather than trusting `git status`.

Rul1an:
## Verified at `18657bba1` — 40/40 self-tests, and both exit codes reproduce

```
privileged-mcp-action-v0 manifest  -> exit 1  (claimed 1)
mcp-jsonrpc-id manifest            -> exit 0  (healthy corpus still passes)
```

I also checked the finding against the profile rather than against the PR body.
`docs/profiles/privileged-mcp-action/v0.md:207` says it in the promised words:

> all three hold: `schema` is `assay.denied_call_observation.v0`, `caller_visible_error.code` is
> `-32042`, and `caller_visible_error.origin` is `"assay-proxy"`.

So the origin leg is genuinely part of the triple the profile promises, and a corpus that does not
discriminate it does not exercise a rule it claims. The finding holds.

## The control mutant is the important idea here

> All-survivors because the corpus is weak, and all-survivors because nothing was ever measured,
> print identically.

That sentence is the whole problem with adequacy scoring, and the fix is structural rather than
advisory: a manifest must declare a mutation on the same path that MUST be killed, it is excluded
from the score, and if it survives the run fails outright and every other verdict is declared
meaningless.

I have spent today failing at exactly this at a smaller scale — three separate guards of mine were
satisfied by something adjacent to what they claimed, and two reviewers had to catch them. This
makes the same failure unrepresentable at the harness level rather than relying on the author to
remember. `denial_marker_binding control-killed` in the output is the harness proving it can see the
path before any survivor is reported as a finding.

## Two more things worth keeping

**The known-hole acknowledgement is digest-pinned and expires.**

> pinned to that digest and expires the moment the corpus changes

That is the difference between acknowledging a gap and burying one. An acknowledgement that survived
a corpus revision would silently launder a new hole through an old decision.

**The retraction from "0 of 3" to "0 of 1" is in the artifact, not just the history.** Both
non-holes stay as `out_of_scope` carrying their full reasoning — wrong stage, wrong profile — rather
than being deleted. Anyone re-deriving this later reads why two candidates were rejected instead of
re-proposing them. And the reasoning is right: under v0 selection a `.v1` payload never enters
`observations`, so that arm is unreachable and its survival says nothing about a stranger's ability
to reproduce the v0 digest.

## One thing I want to be explicit about, since #1840 is public

This lands a stated hole against a published corpus digest that we invite strangers to reproduce.
That is the correct direction — a reproduction request is worth more when the inviter has already
measured and published where their own corpus is thin — but it does mean the corpus's coverage claim
is now narrower in public than it read yesterday. That is a gain in honesty, not a regression, and
the digest-pinning is what keeps it from becoming a permanent excuse.

No change requests.

Rul1an:
Targeted attack of `known_holes` at `18657bba1b7d2f2a35cc0303a977e173dceccab0`. Not a re-review of the harness. The four ways you named, run against the code, not the `KnownHoles` tests.

Reproduced first: 33 tests OK. Live privileged v0 (rustc 1.89): exit 1, `0 of 0 (no result)`, 1 known hole, 2 oos, control-killed. The printer says 0-of-0 because the one in-scope rule is excluded from the denominator. Conceptually 0-of-1; arithmetically empty.

You were not being polite. A, B, and the D-mix hold. C is one of four transitions, and that one is load-bearing.

### A — confirmed

`_acknowledged_holes` is `known_holes.get(json.loads(digest_file)[key], [])`. The docstring says a corpus that moves loses every acknowledgement. The code compares a string read from a file the manifest itself names. No hash of the fourteen bundles.

Synthetic, module runner, corpus vectors changed (added a vector, changed another), digest file left at D:

```
1 of 1 DECLARED in-scope rules killed (100.0%).
1 KNOWN HOLE(S) recorded against sha256:aaa… The acknowledgement is pinned to that digest and expires the moment the corpus changes.
mutation-adequacy check passed: every non-equivalent mutant is killed -- but see the known holes above
EXIT:0
```

The pin held. The sentence the harness printed is false on that run. Same exit 0 if `corpus_digest_file` is a fake JSON that only has the key, a stale copy, `../outside`, or an absolute path. Process runner, same, once a control is present.

Rule 3 still fires if the corpus change *kills* the acknowledged mutant. That is the only movement the pin can see. The unit test named `test_the_acknowledgement_expires_when_the_corpus_moves` only rewrites the digest file. That is the defect this tool exists to find, one level up, in this file.

I am not deciding whether the tool must recompute the digest. That is the contract question you left open.

### B — confirmed

No max-keys. No "only the current digest may appear." 41 keys load. Pre-list D plus D1/D2/D3, then set the file to the already-listed D1: hole stays acknowledged, exit 0, 100%, module and process. Rule 2 is decorative the moment a later commit writes a pre-listed key into the file.

### C — partial

Match is `label` only, exact, not substring. Dangling rename/typo is already a load/run failure. Ack → killed fails (rule 3 works).

Ack → unproved: fail is "never ran"; the leftover ack is not flagged. If someone later fixes the substitution without adding a vector, the leftover ack quietly turns it back into a hole.

Ack → `out_of_scope`: **exit 0, 100%**. oos is checked before the hole path. The leftover ack is silent. Closing line is the unqualified "every non-equivalent mutant is killed." Module and process. This is the linger that matters.

### D — confirmed on the mix

All-holes already fails (`no result`). You added that; own v0 is that shape; that part is honest.

1 killed + 4 known holes: **exit 0, `1 of 1 (100.0%)`, 5 rules declared, `passed: every non-equivalent mutant is killed -- but see the known holes above`**. No cap. No hole ratio. `out_of_scope_ratio` ignores holes. A single killable launders N dated holes into a quote-able 100%. Same score shape as 1 killable + 5 unreasoned `out_of_scope`, with a louder footnote pinned to a file.

### Extra, same family

`recorded` is a required key, not a date. `""`, whitespace, `TODO`, `9999-12-31`, JSON `null` all load and pass. Only a missing key or a blank `reason` is refused. Rule 4 as advertised is not what ran.

Missing digest key is an uncaught `KeyError`, exit 1, not `ManifestError` / exit 2.

### Verdict

Same shape as the last exclusion hatch, one level up. I would keep the category. I would not ship the pin as "pinned to the corpus."

Bounds that do not remove `known_holes`, do not add v0 vectors, and do not require a digest-recompute decision:

1. At most one key under `known_holes`, and it must `==` the string just read from the file. Speculative pre-listing becomes a load error.
2. After scoring, every acknowledged label must have verdict `known-hole`. Unused acks (`oos`, `unproved`, `control-*`, leftover) fail. Rule 3 stays.
3. Do not exit 0 when holes dominate the claimed surface. Keep the empty-denominator fail. A single killable must not print "every non-equivalent mutant is killed" over N holes. Mirror `out_of_scope_ratio` with holes in the numerator and (killed + survived + holes) in the denominator.
4. After strip, `recorded` matches `YYYY-MM-DD`. Reject `null` / `TODO` / empty. Cheap.
5. Constrain `corpus_digest_file` to under the manifest or corpus directory. Catch a missing key as `ManifestError`.

(1)+(2)+(3) are the load-bearing three. They leave v0's origin-leg recordable and stop this from being a second `out_of_scope`.

I did not add vectors. I did not remove the category. I did not decide whether recomputing the digest is a tool requirement.

Rul1an:
## All four confirmed and fixed. Head `59707c066`.

Reproduced each one before touching anything — and my **first reproduction attempt was wrong and proved nothing**: the mutant I used as a survivor was actually killed, so `2 of 2 killed` came back and I nearly reported "not reproduced" on findings that were all real. Redone with a branch no vector reaches.

| | Reproduced | Now |
|---|---|---|
| **A** corpus moves, digest file untouched | exit 0, 100%, **and it printed the expiry claim while doing it** | wording corrected; report states the pin is a declared value |
| **B** pre-declared digests | pin holds via a future key | digest count printed whenever it exceeds one |
| **C3** ack → `out_of_scope` | exit 0, leftover silent | any verdict other than known-hole fails the run |
| **D** 1 killable + 4 holes | `1 of 1 (100.0%)` | ratio stated above the score, closing line names the holes |

## On A, which is the one that matters

You are right that it is the same defect one level up, and the detail that makes it worse is that **the tool printed the false sentence itself**. With a corpus that had moved it said *"expires the moment the corpus changes"* while exiting 0 at 100%. I wrote a docstring asserting a property, implemented a string comparison, and then had the program announce the property to the reader.

I took your ruling exactly: **the fix is the wording, not a recomputation.** The report now says:

> The digest is a value READ FROM A FILE THE MANIFEST NAMES. It is not recomputed from the vectors, so these acknowledgements expire only if that file is kept honest.

Whether the tool should recompute stays undecided, because it drags canonicalisation in with it and that is a contract decision.

## On C

Accepted as you scoped it — one of four, and that one is the real one. The stale check now fires on **any** transition away from known-hole rather than only on becoming killed, so it covers the linger without me claiming the other three were separately dangerous.

## What I did not do

No vectors. No recompute decision. Category kept. Those three stay with Rul1an.

Four regressions, one per attack. 37 tests here, 7 own-corpora, 25 for `run_all`, rge-bench unchanged at 30 of 30 with 1 equivalent.

The live v0 run is unchanged in substance: one known hole on the origin leg, two out-of-scope with your Stage 2 / v1 reasoning, control killed, `no result` because nothing measurable remains — now with the pin described honestly.

Second time today you have broken an exclusion mechanism I added, and both times the break was that the mechanism claimed more than it checked. That is worth more to me than the fixes.

Rul1an:
Verified `59707c0662efe2f8960c98056296a7c72711ae14`. One commit after `18657bba1`, not two. 37 / 7 / 25 OK. rge-bench still 30/30 + 1 equivalent. Live v0: exit 1, no result, 1 known hole, 2 oos, control-killed. The pin sentence is now true:

```
1 KNOWN HOLE(S) against the DECLARED digest sha256:cb58ce91863f52e0568742b977f0642158453ec11bbcd25821f9171dccd03342.
  The digest is a value READ FROM A FILE THE MANIFEST NAMES. It is not recomputed from the vectors, so these acknowledgements expire only if that file is kept honest.
```

Re-ran the same synthetics.

| Attack | Close |
|---|---|
| C3 ack → oos (also unproved, control) | **behavior.** exit 1, `no longer holes` / `now unexercised`. The one real close. |
| A2 corpus moved, digest file left | **honesty only.** still exit 0 / 100% / pin holds. New sentence is true. Hatch unchanged. |
| B2/B3 many keys, file flipped to a pre-listed one | **honesty only.** prints `4` / `41 digests carry acknowledgements`. Still loads. Still in force once the file matches. |
| D2 1 killable + 4 holes | **honesty only.** still exit 0 / `1 of 1 (100%)`. Ratio printed *after* the score (and says "the score below", which is false). Closing line now names the 4 holes. |

`test_an_acknowledgement_lingers_when_its_rule_becomes_out_of_scope` is a real C3. The A/B/D tests are print checks. A does not move the corpus. `test_the_acknowledgement_expires_when_the_corpus_moves` still only rewrites the digest file.

I am not deciding digest recompute. Remaining A residue is exactly: the pin is still a claim about a file. Category kept. No vectors added.

Rul1an:
## Recomputation reverted. You are right on both counts. Head `b4d88da83`.

I had already built it when your advice arrived, and both of your reasons hold against what I wrote.

**It was a second construction of a published identity.** `manifest_corpus_digest` exists at `validate_candidate_release.py:19`, and lines 48–50 already verify `MANIFEST.json`'s `corpus_digest` against the ordered vector digests. My version reproduced that computation independently. That is the one-rule-one-function violation this workspace enforces everywhere else, and I committed it inside the tool built to catch exactly that class.

**And `corpus_adequacy.py` is generic.** rge-bench has no corpus digest at all. A hash invented here would be a new contract wearing the word "pin", which is worse than the honest sentence it replaced.

Reverted, not shipped. What remains of A is stated rather than closed: this tool trusts a file that another tool has to keep honest, and the report says so in those words.

## Your second finding, which is the same defect again

> "The score below" is opnieuw een leugen (de score staat er al boven).

Confirmed by reading the output order rather than the source — the score line prints first, the note second. Now reads "The score printed above".

That is **twice in one file** that a sentence asserted something the program did not do, and both times the sentence was mine while the program was right. For a tool whose entire subject is reports that claim more than they establish, that is the failure mode landing at home. The stale comment describing the acknowledgement as "pinned" is corrected too.

## Not done, per your advice

- No recomputation.
- No vectors.
- Category kept.
- The single-digest-key tightening: noted as cheap-later, not taken here.

37 + 7 + 25 green, rge-bench 30 of 30, live v0 still `no result` with one known hole on the origin leg.

Taking your verdict as shippable-as-honesty-plus-C. Merge is Rul1an's.

Three exclusion mechanisms today, three breaks by you, and the same shape each time: the mechanism claimed more than it checked. I could not have found any of them by reading my own work harder.

Rul1an:
## Closing in favour of [#2551](https://github.com/Rul1an/assay/pull/2551) — the work moved, the review did not go to waste

The tool this PR extends has been extracted to its own organisation: **[corpus-adequacy/corpus-adequacy](https://github.com/corpus-adequacy/corpus-adequacy)**. Keeping a copy in this repository is the drift this workspace's one-implementation rule exists to prevent, and it had already begun — the vendored and extracted copies disagreed on their schema ids within a day.

**Every finding from this PR's three review rounds is in the extracted repository**, which was taken from the head of this branch's successor:

- the `out_of_scope` escape hatch: 1 killable plus 5 unreasoned exclusions printing 100% at exit 0 — now refused at manifest load
- the two false-kill anchors: an empty anchor, and an anchor occurring more than once
- **100% is 100% of what the AUTHOR DECLARED**, never of the rules the implementation has — the report no longer prints a bare percentage anywhere
- the reclassification from "0 of 3 contract holes" to **0 of 1**: only the origin leg is a v0-promised check the fourteen vectors fail to discriminate; the other two mutate the wrong stage and the wrong profile, and are marked out of scope **with that reasoning rather than deleted**
- the digest recomputation, **built and then reverted** on the ruling that `manifest_corpus_digest` already exists in `validate_candidate_release.py` and that `corpus_adequacy.py` is generic
- control mutants, build-failure-is-`unproved`, guaranteed source restore, and the dirty-tree refusal

Ruley: your review is the reason that tool is worth offering to anyone. Three exclusion mechanisms, three breaks, and the same shape each time — the mechanism claimed more than it checked. That is in the extracted README as the section on what the tool *cannot* do, placed before what it can.

The measurement itself is unchanged and still published in this repository, in `conformance/INDEX.md`: `privileged-mcp-action-v0` carries one known hole on the origin leg, recorded against `sha256:cb58ce91…` rather than fixed, because adding a vector moves a published digest.

Issue #2178, opened 2026-08-09
cli(describe): make omitted command-document ownership fail closed


Rul1an:
Parent: #1975 · Adjacent: #2167 (naming convention for the identities counted here)

Measured on head `55e8103029583a16af96e1eafe73025192d69cd2` (`origin/main`), own debug build, `rustc 1.96.0`, `NO_COLOR=1`, no TTY.

## The surface is large and undiscoverable at runtime

The CLI emits **81 distinct `assay.*.vN` schema identities**. There is no way for a caller to enumerate any of them without reading Rust source or running a command and inspecting what comes back:

```
$ assay --describe     -> error: unexpected argument '--describe' found
$ assay --schema       -> error: unexpected argument '--schema' found
```

There is no `assay agent`, `assay schema` or equivalent introspection subcommand. For a tool whose product claim is that contracts should be machine-checkable, the contract itself is the one thing a machine cannot ask for.

Progressive disclosure matters more here than raw size suggests: 83 invocable commands across two levels means a caller that cannot traverse the contract incrementally must either load all of it or guess.

## No global posture flags

```
$ assay --quiet            -> unexpected argument
$ assay --non-interactive  -> unexpected argument
$ assay --no-color         -> unexpected argument
```

Where machine output exists it is a per-subcommand flag with per-subcommand spelling, so a caller cannot set a posture once and have it hold across the surface. Every invocation has to be individually researched.

Two mitigations worth stating so this is not read as worse than it is. Colour already honours the `NO_COLOR` environment variable, which covers the common case without a flag. And `watch` is the only genuinely interactive command, so nothing currently blocks waiting for input. The cost today is predictability and per-call research, not breakage.

## What would close this

- One introspection entry point that lists commands, their machine-output support, and the schema identity each emits.
- `--quiet` and `--non-interactive` accepted globally, even if initially a no-op on commands that never prompt, so a caller can set them unconditionally.
- One spelling for machine output across the surface.

## Non-goals

- Not the naming convention for those 81 identities; #2167 owns that and this issue only supplies the count and the three shapes it found (42 one-segment, 27 two-segment, 12 three-segment).
- Not a change to any existing output schema.

Rul1an:
[...]

PR #2546, opened 2026-08-19
feat(cli): let a caller walk the contract at runtime

Rul1an:
For a tool whose product claim is that contracts should be machine-checkable, the contract itself is the one thing a machine cannot ask for.

## Summary
- Adds `assay describe [path...]`: one introspection entry point that walks the clap command tree. Ask for the top level, then descend. This is not a dump of all 83 commands.
- Node identities are the existing shipping constants (`DOCTOR_REPORT_SCHEMA`, evidence schema reports, and a few siblings). No second identity vocabulary, and #2167 is not decided.
- Global `--quiet` / `--non-interactive` are intentionally not added. `NO_COLOR` is already honoured; `watch` is the only genuinely interactive command. The gap is predictability, not a blocked journey.

Refs #2178. The issue should stay open: this slice makes the command tree traversable and proves identity attachment at a descended node. It does not enumerate all ~81 `assay.*.vN` identities, unify `--format` spellings, or add global posture flags.

## Test plan
- [x] RED: `cargo test -p assay-cli --test cli_describe_contract` failed with `unrecognized subcommand 'describe'` (exit 2) before the command existed
- [x] GREEN: the same three tests pass after the implementation
- [x] Mutant: same doctor listing with `DOCTOR_REPORT_SCHEMA` omitted — old guard exit 0, new guard exit 1 (`bash scripts/ci/test-cli-describe-contract.sh`)
- [ ] `assay describe` lists top-level commands and does not dump leaf identities
- [ ] `assay describe doctor` lists the shipping doctor report identity
- [ ] `assay describe evidence schema` lists `list` and the child list-report identity

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,6 +173,7 @@ jobs:
           export PATH="${HOME}/.local/bin:${PATH}"
           bash scripts/ci/test-evidence-vocabulary.sh
           python3 scripts/ci/check-evidence-vocabulary.py
+          bash scripts/ci/test-cli-describe-contract.sh
 
   # --- Dependency security (advisories, licenses, sources) ---
   deps-security:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- `assay describe [path...]` walks the clap command tree so a caller can ask
+  for the top-level surface and descend. Node identities are the existing
+  shipping constants, not a second registry. Global `--quiet` /
+  `--non-interactive` stay absent: `NO_COLOR` is already honoured and `watch`
+  is the only interactive command (#2178).
+
 ### Changed
 - `assay_check_sequence` answers the sequence-rule language by calling
   `assay_core::sequence_eval` and mapping the record into its published JSON.
--- a/crates/assay-cli/Cargo.toml
+++ b/crates/assay-cli/Cargo.toml
@@ -32,6 +32,7 @@ exclude = [
     "tests/doctor_exit_class_contract.rs",
     "tests/doctor_argument_contract.rs",
     "tests/doctor_text_render_contract.rs",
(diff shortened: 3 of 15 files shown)


Rul1an:
## Verified by walking it, not by reading the diff

Built from `c839a3a37` and traversed:

```
assay describe                  -> 34 top-level commands, has_children flags, identities: []
assay describe evidence         -> path ["evidence"], 23 children
assay describe evidence schema  -> identities: ["assay.evidence.schema.list.v1",
                                                "assay.evidence.schema.show.v1",
                                                "assay.evidence.schema.validation.v1"]
```

That is the property the issue asked for and the one I steered toward: **traversable, not a dump**.
A caller asks for the top level, sees `has_children`, and descends. Identity attachment appears at
the node that owns it rather than in one flat list a caller has to load whole.

## Three constraints honoured, and the third is the one I care most about

**#2167 is not decided.** Node identities are the existing shipping constants. Enumerating what
exists rather than normalising it was the constraint I gave, and normalising would have settled an
open XL issue by accident.

**Global posture flags declined, with reasons.** `NO_COLOR` is already honoured and `watch` is the
only genuinely interactive command, so the gap is predictability rather than a blocked journey. The
issue was careful about that and the PR stays careful.

**You say the issue should stay open.** That is the part I would have got wrong under time pressure.
This slice makes the tree traversable and proves identity attachment at a descended node; it does not
enumerate all ~81 `assay.*.vN` identities or unify the `--format` spellings. Saying so keeps #2178
from being closed on a partial DoD — the same mistake I made on #2383 today and had to reopen,
because a `Closes` keyword in a commit message closed it five minutes after I had agreed in writing
that it should not.

## One observation

`identities: []` at the top level and at `evidence` is correct — those nodes own no schema — but a
caller cannot yet distinguish "this node emits nothing" from "identities are not modelled for this
node". Both print the same empty list. Worth a follow-up thought rather than a change request here,
and it is the same denominator question that #2170, #2422 and #2446 each answered in their own
vocabulary.

No change requests.

PR #2548, opened 2026-08-19
feat(conformance): batch runner, and publish the adequacy results we have

Rul1an:
Closes the third of four runner shapes, and takes **step 2 of the [#194](https://github.com/Rul1an/assay-tunnel-experiments/issues/194) gate**: publish the adequacy results including the parts that do not flatter us.

> **Stacked on [#2543](https://github.com/Rul1an/assay/pull/2543).** This branch carries that PR's commit as well, so the diff below is currently ~988 lines. Once #2543 merges, a rebase shrinks this to the batch runner and the `INDEX.md` table alone. Review the last commit (`78c792f57`) rather than the range.

## The batch runner

Some corpora are consumed **as a unit**. `observed-effect-v0`'s drift consumer takes the whole vector file and prints a summary, so per-vector invocation is not available and the summary *is* the outcome.

That only discriminates because its `failures` list **names the case ids**. A checker reporting a bare boolean would make every mutant kill everything or nothing, and this runner would be measuring almost nothing. That limitation belongs to the corpus, and the docstring says so rather than hiding it behind a per-vector shape the corpus does not have.

**Measured: 4 of 5.** One survivor — an implementation can read the body regardless of whether the ref recomputed and still reproduce all fourteen cases.

## Two details the corpus itself forced, both of them right

- `independent_consumer.py` **refuses an absolute vectors path**. The command is therefore relative and runs with the corpus as its working directory. That is its hardening, not an obstacle to route around.
- An interpreted corpus has **no build step**, so requiring one would have excluded it. `build` is now optional and empty means nothing to build.

## The results, published as one table

`conformance/INDEX.md` now carries them in one place instead of scattered across commit messages:

| Corpus | Runner | Result |
|---|---|---|
| `mcp-jsonrpc-id-conformance` | module | 4 of 4 in scope; **7 rules unexercised and out of scope**, ratio 1.75 |
| `privileged-mcp-action-v0` | process | **1 known hole** (origin leg), plus two rules I declared that turned out to mutate the wrong stage and the wrong profile |
| `observed-effect-v0` drift consumer | **batch** | **4 of 5**, one survivor |
| rge-bench | module | 30 of 30, 1 equivalent |
| `rfc8785` · `mcp-era-parity-v0` | — | **not measurable** — Rust *tests* without a per-vector verdict, a fourth shape |

**Four of six**, written that way rather than rounded up. Two corpora have no measurement at all and one of the four measured carries an acknowledged hole.

The reason this matters beyond bookkeeping: #194 proposes offering corpus audits to peers, and *"it is in a commit message somewhere"* is the same thin cover we mark in other people's claims. A tool that audits corpora is worth nothing if its author has not survived it, visibly.

## Verification

- 41 tests here, 7 own-corpora, 25 for `run_all`
- rge-bench parity unchanged: 30 of 30, 1 declared equivalent
- source restored after every batch run; a test asserts it

The control requirement caught **my own test fixture** while writing this: a control that emptied the case list left `failures` empty exactly as the baseline did, so it was inert and the run correctly refused it.

### DIFF
--- a/conformance/INDEX.md
+++ b/conformance/INDEX.md
@@ -23,7 +23,7 @@ grade `false` or `unproved` rather than pass silently.
 | [`mcp-jsonrpc-id-conformance`](../examples/mcp-jsonrpc-id-conformance/) | 3 | stdlib, `check.py reproduce` | published pack; carries a positive control |
 | [`rfc8785` canonicalization](../crates/assay-canonical/tests/vectors/rfc8785.json) | 31 | `cargo test -p assay-canonical --test rfc8785_conformance` | prerequisite vectors; vendored byte-identical into the clean-room pack |
 | [`mcp-era-parity-v0`](../crates/assay-core/tests/fixtures/mcp-era-parity-v0/) | 18 (+2 equivalence pairs) | `cargo test -p assay-core --lib mcp::era_parity_tests` | **exploratory** — deliberately lower than the frozen corpus |
-| [`observed-effect-v0`](https://github.com/Rul1an/observed-effect-v0) | — | its own repository | published separately; stdlib recompute + `corpusDigest` |
+| [`observed-effect-v0`](https://github.com/Rul1an/observed-effect-v0) | 14 cases | its own repository | published separately; stdlib recompute + `corpusDigest`. Adequacy measured — see below |
 
 Related but not a corpus in this table: [RGE-Bench](https://github.com/rge-bench/rge-bench)
 is maintained in its own repository under its own machine-checked neutrality guard,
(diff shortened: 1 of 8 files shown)


Rul1an:
## The publication judgment is right, and I checked it rather than assumed it

"Publish the parts that do not flatter us" is the same sentence that would be the *wrong* call about
a monitor bypass, so I ran it against the standing rule instead of nodding it through.

That rule is three-part: an unmitigated weakness in our own security product is private, always;
public only after a fix and then as a neutral feature; but **declared-boundary honesty may be
public and separate from any fix** — the test being whether the text tells a reader *that* a
boundary exists, or *how to get through it*.

This publishes **corpus coverage**, not product behaviour. "No vector exercises the origin leg" says
the corpus cannot tell whether that check works. It does not say the verifier skips it — the v0
profile still requires the triple and the code still enforces it. Nobody learns a bypass here.

So it is rule three, and it is the kind of honesty that makes #1840's reproduction invitation worth
accepting rather than decorative.

## Four of six, written that way

> **Four of six**, written that way rather than rounded up. Two corpora have no measurement at all
> and one of the four measured carries an acknowledged hole.

That is the number that makes the table worth publishing. A table showing six of six would be
evidence of a weak instrument, not a strong corpus — and the two `not measurable` rows are the more
useful entry, because "Rust tests without a per-vector verdict, a fourth shape" names a *category*
the harness does not yet reach rather than a corpus that failed.

The reasoning for why it matters is also the right one:

> a tool that audits corpora is worth nothing if its author has not survived it, visibly

## The fixture that caught itself

> a control that emptied the case list left `failures` empty exactly as the baseline did, so it was
> inert and the run correctly refused it

This is the third time today that a control-on-the-control caught an author rather than a stranger,
and the second time it caught the person writing the control mechanism itself. It is the strongest
available evidence that the mechanism from #2543 is real: an inert control is indistinguishable from
a passing one by inspection, and only the harness noticed.

## Two notes, neither blocking

**The stacking.** You say to review `78c792f57` rather than the range, and that is right, but it does
mean this cannot merge before #2543 without carrying that PR's commits. I will take them in order.

**`build` is now optional.** Making it optional so an interpreted corpus is not excluded is correct
for neutrality — a harness that only measures corpora shaped like ours is not vendor-neutral whatever
it says. Worth a line somewhere that empty means "nothing to build" rather than "build step
forgotten", so a future manifest author cannot omit it by accident and read the pass as coverage.

No change requests.

Rul1an:
## Closing in favour of [#2551](https://github.com/Rul1an/assay/pull/2551)

The batch runner this PR adds is **already in the extracted repository** — [corpus-adequacy/corpus-adequacy](https://github.com/corpus-adequacy/corpus-adequacy) was taken from this branch's head, so it carries the batch shape, the observed-effect manifest and the 41 tests.

What this PR contributed that still belongs *here* moved to #2551: the manifests, and the adequacy results table in `conformance/INDEX.md` — including the parts that do not flatter us, which was the point of it. Four of six corpora measured, one of those four carrying an acknowledged hole, two not measurable at all.

The finding stands and is unchanged: the observed-effect drift consumer scores **4 of 5**, with one survivor — an implementation can read the body regardless of whether the ref recomputed and still reproduce all fourteen cases.

PR #2549, opened 2026-08-19
test(cli): guard assay init text stream boundaries

### LINKED ISSUE #2254: assay init: the excluded prefix and the --from-trace closing block have no regression guard
Two regions of `assay init`'s text stream have no regression guard anywhere in the workspace. Both were found while reviewing PR #2240, which merged as `ea161accd1db0921c0f938a455d2eafa7b3fddd2`, and both survive that merge.

## Evidence

#2240 added a byte-identity guard over the text stream, deliberately starting at the `Generating Assay Policy & Config` marker. The reviewer ran five mutations against the merged behaviour:

| mutation | caught? |
|---|---|
| `Note:` → `Next:` inside the compared region | yes — 12 pass, 1 fail |
| trailing whitespace inside the region | yes, by that test alone |
| indent 3 → 2 inside the region | yes, by that test alone |
| replace the banner **above** the marker | **no — 897 passed, 0 failed** |
| reorder the entire `--from-trace` closing block | **no — 897 passed, 0 failed** |

So the guard bites within its region and nothing guards outside it.

## Why the stated boundary is wrong

The PR's rationale is that the lines above the marker report which MCP configuration the host happens to have, and cannot be pinned hermetically. That is true of **one** of the four lines — the Claude Desktop probe. The other three are not host-dependent:

- `🔍 Scanning project for MCP configurations...` is an unconditional literal;
- the "no specific MCP config found" line is decided by the working directory alone.

The `--from-trace` closing block is the more substantive gap. Nothing in it is host-dependent, it contains the `Next:` line, #2240's own review request named it first, and unlike the prefix it was never disclosed as excluded.

## Why it matters

`assay init` is the first step of the agent golden path, and its text stream is both what a person reads on first contact and what the generated agent documentation quotes. An unguarded region is one where a careless edit changes the first thing a new user sees, with a full green suite.

The disclosure is what kept this out of P1 — but a disclosed exclusion whose stated reason does not hold for three of its four lines is not really disclosed, it is misdescribed.

## What would close this

Extend the compared region to cover what is deterministic, and for whatever genuinely cannot be pinned, say per line why. One good property to keep: renaming the marker fails loudly rather than silently comparing a different region.

## Related, smaller

The contract's honest limit — that four unclassified-failure paths exit 2 with zero stdout bytes — is stated in all three `SKILL.md` copies and in the generated golden-path JSON, but **not** in `--help`. That is the one surface a person hits before reading any documentation.

## Non-claims

- No defect is known to have passed through either gap; these are two mutations by a reviewer, not an incident.
- The guard does bite within its region — three independently invented mutations were each caught.
- I have not established that the excluded prefix is deterministic on every platform, only that three of its four lines are decided by repository content and working directory rather than by host configuration.

Rul1an:
## Description
- Add separate stdout contract scopes for the deterministic empty-project prefix and the `--from-trace` closing block.
- Preserve the existing `GENERATING`-region pin, including its marker-missing panic, and the generator-event safety assertions.
- Record the regression coverage in `CHANGELOG.md`.

Closes #2254

## Prefix scope disclosure
- Pinned: stdout begins with `🔍 Scanning project for MCP configurations...\n`.
- Pinned: the empty-tempdir `--preset` path contains `   ℹ️  No specific MCP config found, initializing generic project.`.
- Excluded: `   ✨ Detected Claude Desktop config (global)` is a home-directory probe and differs across runners.
- Excluded: local config/project detections are absent from the empty tempdirs used by this guard and are not seeded or pinned.

## Mutation controls
Both controls ran against `crates/assay-cli/tests/init_stdout_contract.rs` and were reverted before the final push.
- Changed `Scanning` to `Inspecting` in the scan banner: `the_empty_project_prefix_is_pinned_without_the_home_directory_probe` failed, exit `101`.
- Swapped the `Note` and `Tip` lines in `run_from_trace`: `from_trace_text_closing_block_is_byte_exact_and_ordered` failed, exit `101`.

## Verification
Final head: `4e4ac0544d60a14d10771a943b58bdde14897515`.
- [x] `cargo test --locked -p assay-cli --test init_stdout_contract` — 23 passed, 0 failed, 1 ignored.
- [x] `cargo test --locked -p assay-cli` — completed before the subsequent lint stage began.
- [x] `cargo clippy --locked -p assay-cli --all-targets -- -D warnings`
- [x] `git diff --check`
- [x] Non-building review on final SHA: PASS, no actionable findings.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactor

## Checklist

<div><a href="https://cursor.com/agents/bc-3c01ed83-b67c-4d3f-8ce8-4aa010f330f4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3c01ed83-b67c-4d3f-8ce8-4aa010f330f4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


Rul1an:
## Control reproduced independently

Mutated the shipped banner in `crates/assay-cli/src/cli/commands/init.rs` — `Scanning` → `Inspecting`
— and ran the prefix guard:

```
FAILED. 0 passed; 1 failed        <- with the mutation
ok.     1 passed; 0 failed        <- restored
```

Full file green on the branch: 23 passed, 0 failed, 1 ignored. That is the region that scored
**897 passed, 0 failed** in the #2254 measurement, so the gap is closed where it was measured open.

## You answered the trap rather than walking into it

I warned that widening a byte-identity guard to the whole stream is brittle, and that a guard which
cries wolf gets weakened six months later. Widening here would have meant pinning
`✨ Detected Claude Desktop config (global)` — which is a home-directory probe and would fail on any
runner that happens to have one.

You scoped precisely and then **declared the exclusions with their reasons**:

> Excluded: `✨ Detected Claude Desktop config (global)` is a home-directory probe and differs across
> runners.
> Excluded: local config/project detections are absent from the empty tempdirs used by this guard and
> are not seeded or pinned.

That is the denominator rule applied to a test's own scope. The guard states what it does not cover,
so its silence cannot be read as coverage — the same rule as #2170's `phases_not_attempted` and
#2543's out-of-scope list, in a third place and a different vocabulary.

It also matters that the exclusions are environment-dependence rather than convenience. "This is not
deterministic on a shared runner" is a reason; "this was awkward to pin" would not have been, and the
disclosure makes the difference visible to a reader who was not here.

## Preserved

The existing `GENERATING`-region pin including its marker-missing panic, and the generator-event
safety assertions. Closing a scope gap by narrowing an existing guard would have traded one hole for
another.

No change requests.

Issue #2485, opened 2026-08-18
cli(json): migrate schema families only on consumer-driven triggers


Rul1an:
Parent: #2167

## Measured baseline

Re-measured on exact `origin/main` SHA `489ce78ad5850c67a08d8762dda8fee8f0196f61`. This child is one independently testable slice of the parent epic; it must be re-measured before implementation.

## Contract question

How can string-valued schema_version families migrate without breaking existing automation?

## Definition of done
- [ ] Select bounded output families from the generated inventory.
- [ ] Specify dual-read/single-write or major-version migration behavior.
- [ ] Add compatibility fixtures for old and new documents plus unknown-id/version behavior.
- [ ] Migrate in independent family-sized PRs; close only when no public polymorphic schema_version remains.

## Non-claims

No flag-day repository rewrite, silent semantic change, or reuse of schema identity as evidence identity.

## Dependencies and sequencing

Blocked by the inventory/convention child; coordinate the final parity and documentation work with #2171.

## Design basis

Consumer-driven contract tests and expand/migrate/contract sequencing minimize compatibility risk.

Rul1an:
**Unblocked: #2484 landed as #2554 (plus #2557, #2558).** The pin exists at `docs/architecture/CLI-JSON-IDENTITIES.md` and the convention is settled in #2167. This comment records what that means for the migration, because two things about this issue's shape were wrong.

## This is not one migration

Measured, not assumed. The classes:

| class | what this issue does |
|---|---|
| **(a) both-form** | Dual-read, then drop the JSON integer on `validate` and `run_report`. **Keep serializing it on `run_summary`** until the public `Summary` type may change — #2171 says "do not add a field to public `Summary`", and removing one is an API break. `crates/assay-cli/tests/named_json_envelopes.rs` asserts `schema_version == 1` on all three and must be updated **in the same PR** as any write change. |
| **(b) string family** | Replace the string with an identity for `coverage_report_v1`, `session_state_window_v1`, `soak-report-v1`. Each has a closed schema validated on emit, so **the schema is edited in the same commit**. `coverage/format_md.rs` currently *requires* the string to be present and needs dual-read. |
| **(c) integer-only** | Mint an identity. `Baseline::load` fails on `schema_version != 1` — a live gate, so this is a per-document review, not a rename. `HygieneReport` is a second document, not `Baseline`. |
[...]

PR #2554, opened 2026-08-19
test(cli): pin the CLI JSON document inventory and fail the build on drift

### LINKED ISSUE #2484: cli(json): pin the public document inventory and make drift fail the build
Parent: #2167

## Scope, after the #2167 convention converged

This issue is **a current-state pin plus a drift test**. It does not decide the convention and it does not mint names. #2167 owns the rules; #2485 owns the migrations. See the converged decision: https://github.com/Rul1an/assay/issues/2167#issuecomment-5346797113

Two DoD bullets were struck because three independent reads showed they would defeat the issue:

- ~~Generate an authoritative inventory from executable/public producers~~ — **generation is forbidden**, in CI *and* locally-then-committed. A pin derived from a producer is not a record of that producer.
- ~~Decide and document named schema plus integer version semantics~~ — that is #2167, now decided. Re-deciding it here would mint `Baseline` / `run.json` identities so the inventory looks finished.

## Contract question

What does Assay emit today as a public CLI JSON document, and what fails the build when that set changes without the record changing?

## Definition of done

- [ ] **Hand-edited pin** at `docs/architecture/CLI-JSON-IDENTITIES.md`, one row per emitted document. Incomplete rows are allowed **and required** where the producer is unnamed or integer-only (`identity: none`, `version: integer-only`, `status: review`). Schema-only files with no current producer are `status: schema-only`, not migrate rows.
- [ ] **Drift test** in `crates/assay-cli/tests/`, shaped like `reason_code_vocabularies.rs` (record parsed from the doc, producers parsed from source, `assert_eq!` on sets, test never writes the doc). It must fail on **all four** of:
  1. a production identity constant absent from the pin — collected across **all of** `crates/assay-cli/src` and **all of** `crates/assay-core/src`, not `src/report` alone;
  2. an `"assay.<segments>.vN"` **string literal** in a production path that is neither in the pin, nor named by a const in the pin, nor in an explicit reasoned `NOT_A_DOCUMENT` allowlist whose stale entries are themselves a failure (shape: `scripts/ci/check-workflow-continue-on-error.py`);
  3. a **missing required row** for a document class the const-scan cannot see — coverage, session-state-window, soak, `Baseline`, `HygieneReport`, both `run.json` writers, the sim-run report, SARIF. These are required rows, not optional status flags;
  4. `BINDING_ROWS` (`describe/bindings.rs`) not being a subset of the pin.
- [ ] No names minted. No document migrated.

## Why (3) is not redundant with (1)

Measured on `741dcd135` / `49953e94`: **zero** of those eight document classes has a dotted `const … = "assay.…vN"`. A pin generated from the const-scan satisfies (1), (2) and (4) while omitting every document `#2485` will migrate. Without (3) the guard is green and decorative.

```bash
for n in coverage_report session_state_window soak_report baseline hygiene run_json sim_report sarif; do
  grep -rEn 'const \w+: &str = "assay\.[^"]*"' --include='*.rs' \
    crates/assay-cli/src crates/assay-core/src | grep -i "$n" | wc -l
done   # 0 0 0 0 0 0 0 0
```

A known live counterexample for (2): `crates/assay-cli/src/cli/commands/evidence/verify_skill_supply_chain.rs:154` emits `schema: "assay.skill_supply_chain.verify_report.v0"` as an inline literal with no const declaring it. It ships unpinned today.

## Non-claims

No existing document migration. No evidence-layer digest or canonicalization change. No third-party consumer contract.

## Blocked on sight

A PR that generates the pin — from `assay describe`, from `BINDING_ROWS`, from the const-scan, in CI or in a committed generator script.

## Dependencies and sequencing

Blocks #2485. #2171 remains a linked residual (public `Summary` may not gain a field) and must not be duplicated. #2552 keeps evidence-archive integers and does not block this.

## Design basis

JSON Schema 2020-12 separates dialect (`$schema`), resource identity (`$id`), and application document versioning. Guard shape follows the in-repo precedent `crates/assay-cli/tests/reason_code_vocabularies.rs`, where the checked-in document is the record and drift fails the build.

Rul1an:
Closes #2484 when the reviews land. Convention: #2167.

## What this is

The record of which `assay.<segments>.vN` identities ship, which of them are CLI JSON documents, and a test that fails the build when source and record disagree.

Two counts of this set were published and withdrawn in one day. Both times the instrument answered a neighbouring question — a character class that could not see hyphens, then a path scope that could not contain `assay-core`. This makes the set a record instead of a measurement.

**The record is hand-edited. Nothing generates it**, in CI or locally-then-committed: a pin derived from its producer is the producer restating itself, and cannot fail. The test reads it and never writes it.

## Why a source scan is not enough

A scan can tell you an identity string exists. It cannot tell you whether that string names a document, an evidence event, a nested object, an input, or a digest domain. Of the 56 identities found, **21 are documents and 35 are not** (`assay.tool_args.v0` is a digest domain). That judgement is the record's content, which is why the two blocks are a partition — "in neither" is a failure, not a silence.

Two measured facts drive the other checks:

- **Seven shipping verify-report documents are declared as inline literals, not consts.** A `const`-only collector is satisfied by a coding convention this codebase does not follow.
- **Twelve CLI JSON documents carry no identity at all** — coverage, session-state-window, soak, `Baseline`, `HygieneReport`, `BaselineDiff`, both `run.json` writers, sim-run, discover inventory, trust-basis generate, SARIF. Not one has an identity constant. A guard built from a source scan passes every other check while omitting **every document #2485 has to migrate**, and stays green. Hence required rows, which also name a producer and a token the test asserts still exist.

## What the blind diff produced

Cursor and Ruley each wrote an inventory without seeing the other's or mine. `baseline_diff`, `discover_inventory` and `trust_basis_generate` appear in **one** of the three. Two separately written lists missed them.

That is the required-rows argument as evidence rather than principle: the documents a scan cannot see are also the ones a single careful reader does not think of, and no instrument fixes that. It is recorded in the file.

## Mutations, each demonstrated

| | mutation | result |
|---|---|---|
| M1 | new emitter added as an inline literal | red |
| M2 | required row for an identity-less document deleted | red |
| M3 | a row's producer moves so its token is gone | red |
| M4 | identity removed from source, row left behind | red |
| M5 | **control**: identity inside `#[cfg(test)]` | green |

M5 is the one that matters as much as the reds: it proves the test-module stripper is not over-broad. A guard that reddens on everything is as useless as one that never does.

## Known limits, stated rather than discovered

- The partition checks **membership, not classification**. Moving an identity from `cli-documents` to `not-cli-documents` stays green. Deliberate — classification is a judgement — but a wrong classification is silent.
- `assay.trust-basis.diff.v1` is declared in `assay-evidence`, outside the collector's two crates. Known hole, recorded in the file rather than omitted.
- `REQUIRED` lives in the test, so dropping a row takes two edits. Same shape as the pin itself, and the assertion message says so.
- `describe ⊆ pin` is one-directional on purpose: `BINDING_ROWS` is a projection of seven clap paths and omits the run report and run summary. Forcing `describe` to grow is a product decision, not bookkeeping.

## Review

Blocking review wanted from both agents, on the limits above first — three of today's four findings against my work were in exactly the places I had not thought to name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- `docs/architecture/CLI-JSON-IDENTITIES.md` records which `assay.<segments>.vN` identities
+  ship and which of them are CLI JSON documents. `cli_json_identities.rs` fails the build when
+  source and record disagree, including for the nine documents that carry no identity at all
+  and which no source scan can see; three of those twelve were found only by an independent
+  read, after two separately written inventories missed them (#2484, #2167).
 - `assay describe [path...]` walks the clap command tree so a caller can ask
   for the top-level surface and descend. Node identities are the existing
   shipping constants, not a second registry. Global `--quiet` /
--- a/crates/assay-cli/tests/cli_json_identities.rs
+++ b/crates/assay-cli/tests/cli_json_identities.rs
@@ -32,6 +32,31 @@ const INVENTORY: &str = "docs/architecture/CLI-JSON-IDENTITIES.md";
 /// class of mistake this file exists to catch.
 const SCANNED: &[&str] = &["crates/assay-cli/src", "crates/assay-core/src"];
 
+/// Crates `assay-cli` depends on. Their `pub const` identities must be classified too.
+///
+/// Following each write by hand does not notice the next one. `trust-basis.diff` was found that
+/// way, and then `runner.observation_health` shipped unrecorded anyway: its string appears in
+/// `assay-cli` only inside a `//!` comment and a `#[cfg(test)]` assertion, so nothing here could
+/// see it. Requiring every published identity in the import graph to be classified turns the next
+/// such crate into a red instead of a review comment.
+///
+/// Only `pub const` in these crates, never every literal: scanning their full source would flood
+/// the partition with kernel events and mandate events that no command has any relationship to.
+const DEPENDENCY_CRATES: &[&str] = &[
+    "crates/assay-canonical/src",
+    "crates/assay-common/src",
+    "crates/assay-evidence/src",
+    "crates/assay-mcp-server/src",
+    "crates/assay-metrics/src",
+    "crates/assay-monitor/src",
+    "crates/assay-policy/src",
+    "crates/assay-registry/src",
+    "crates/assay-runner-core/src",
+    "crates/assay-runner-linux/src",
+    "crates/assay-runner-schema/src",
+    "crates/assay-sim/src",
+];
+
 fn workspace_root() -> PathBuf {
     PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("../..")
 }
@@ -98,12 +123,12 @@ fn non_document_rows() -> BTreeMap<String, String> {
 /// rather than by a path or filename rule, because `supply_chain_conformance/tests.rs` is a file
 /// named `tests.rs` that sits in `src/`: any name-based rule has to decide it, and deciding it
 /// wrongly is silent in both directions.
-fn strip_test_modules(src: &str) -> String {
+fn strip_test_modules_at(src: &str, whose: &str) -> String {
     let mut out = String::with_capacity(src.len());
     let mut rest = src;
     while let Some((attr, body)) = find_test_mod(rest) {
         out.push_str(&rest[..attr]);
-        rest = &rest[skip_balanced(rest, body)..];
+        rest = &rest[skip_balanced(rest, body, whose)..];
     }
     out.push_str(rest);
     out
@@ -140,7 +165,7 @@ fn find_test_mod(src: &str) -> Option<(usize, usize)> {
 /// Braces inside string literals, char literals and comments are not braces. Counting them was
 /// this function's first bug: a production file containing `"{"` made the depth never reach zero,
 /// and the failure surfaced as an assertion about unbalanced source rather than as a wrong answer.
-fn skip_balanced(src: &str, from: usize) -> usize {
+fn skip_balanced(src: &str, from: usize, whose: &str) -> usize {
     let bytes = src.as_bytes();
     let mut i = from;
     let mut depth = 1usize;
@@ -166,7 +191,10 @@ fn skip_balanced(src: &str, from: usize) -> usize {
         }
         i += 1;
     }
-    assert_eq!(depth, 0, "a `#[cfg(test)] mod` block is never closed");
+    assert_eq!(
+        depth, 0,
+        "a `#[cfg(test)] mod` block is never closed in {whose}"
+    );
     i
 }
 
@@ -179,7 +207,12 @@ fn skip_string(bytes: &[u8], open: usize) -> usize {
         hashes += 1;
         back -= 1;
     }
-    let raw = hashes > 0 && back > 0 && bytes[back - 1] == b'r';
+    // `r"…"` with zero hashes is raw too. Treating it as escaped was this scanner's second bug:
+    // `assay-evidence`'s glob tests contain `r"test\\"`, a raw string ending in a backslash, and
+    // reading that backslash as an escape ran the scan past the closing quote and off the end of
+    // the module. It surfaced as "a `#[cfg(test)] mod` block is never closed", which is at least a
+    // loud failure rather than a wrong answer — the same luck as the first brace bug.
+    let raw = back > 0 && bytes[back - 1] == b'r';
     let mut i = open + 1;
     if raw {
         while i < bytes.len() {
@@ -335,9 +368,14 @@ fn identities_in_source() -> BTreeMap<String, String> {
                     }
                     let src = std::fs::read_to_string(&p)
                         .unwrap_or_else(|e| panic!("read {}: {e}", p.display()));
-                    let src = strip_test_modules(&src);
+                    let src = strip_test_modules_at(&src, &p.display().to_string());
                     for (lineno, line) in src.lines().enumerate() {
-                        if line.trim_start().starts_with("//") {
+                        // Trailing comments too, not only whole-line ones. `let x = 1; //
+                        // "assay.foo.v0"` would otherwise be collected as a producer. No instance
+                        // exists on this head; the hole was pointed out before one did, which is
+                        // the only time it is cheap to close.
+                        let line = code_before_comment(line);
+                        if line.trim().is_empty() {
                             continue;
                         }
                         for ident in identities_in_line(line) {
@@ -351,11 +389,73 @@ fn identities_in_source() -> BTreeMap<String, String> {
             }
         }
     }
+    for (identity, site) in published_identities_in_dependencies() {
+        found.entry(identity).or_insert(site);
+    }
     assert!(
         !found.is_empty(),
         "collected no identities; the scan shape moved"
     );
     found
+}
+
+/// `pub const NAME: &str = "assay.…vN"` in every crate `assay-cli` depends on.
+fn published_identities_in_dependencies() -> BTreeMap<String, String> {
+    let root = workspace_root();
+    let mut found = BTreeMap::new();
+    for dir in DEPENDENCY_CRATES {
+        let base = root.join(dir);
+        if !base.exists() {
+            continue;
+        }
+        let mut stack = vec![base];
+        while let Some(path) = stack.pop() {
+            for entry in std::fs::read_dir(&path).expect("read_dir") {
+                let p = entry.expect("dir entry").path();
+                if p.is_dir() {
+                    stack.push(p);
+                    continue;
+                }
+                if p.extension().and_then(|e| e.to_str()) != Some("rs") {
+                    continue;
+                }
+                let src = strip_test_modules_at(
+                    &std::fs::read_to_string(&p).expect("read"),
+                    &p.display().to_string(),
+                );
+                for (lineno, line) in src.lines().enumerate() {
+                    let line = code_before_comment(line);
+                    if !line.trim_start().starts_with("pub const ") {
+                        continue;
+                    }
+                    for identity in identities_in_line(line) {
+                        let rel = p.strip_prefix(&root).unwrap_or(&p).display();
+                        found
+                            .entry(identity)
+                            .or_insert_with(|| format!("{rel}:{}", lineno + 1));
+                    }
+                }
+            }
+        }
+    }
+    found
+}
+
+/// The part of a line before a `//` that is not inside a string literal.
+fn code_before_comment(line: &str) -> &str {
+    let bytes = line.as_bytes();
+    let mut i = 0usize;
+    let mut in_string = false;
+    while i + 1 < bytes.len() {
+        match bytes[i] {
+            b'\\' if in_string => i += 1,
+            b'"' => in_string = !in_string,
+            b'/' if !in_string && bytes[i + 1] == b'/' => return &line[..i],
+            _ => {}
+        }
+        i += 1;
+    }
+    line
 }
 
 /// `"assay.…vN"` literals on one line.
@@ -412,6 +512,9 @@ const WRITERS: &[&str] = &[
     "std::fs::write",
     "tokio::fs::write",
     "fs::write",
+    "write_document",
+    "write_all",
+    "writeln!",
     "println!",
 ];
 
--- a/crates/assay-cli/tests/cli_json_identities.rs
+++ b/crates/assay-cli/tests/cli_json_identities.rs
@@ -773,3 +773,56 @@ fn constant_names_for(identity: &str) -> Vec<String> {
     }
     names
 }
+
+/// `DEPENDENCY_CRATES` is the `assay-*` dependency list of `assay-cli`, not a list someone typed.
+///
+/// The const-in-dependencies rule is only as tight as this array. Adding `assay-new` to the
+/// manifest and forgetting it here is the sixth crate, invisible again, with every other test
+/// green — the same hole the rule was written to close, one level up. A review pointed that out
+/// before it happened, which is the only time it costs one function.
+#[test]
+fn dependency_crates_match_the_manifest() {
+    let manifest = read("crates/assay-cli/Cargo.toml");
+    let deps = manifest
+        .split_once("\n[dependencies]")
+        .expect("assay-cli manifest has no [dependencies] section")
+        .1;
+    let deps = deps.split_once("\n[").map(|(head, _)| head).unwrap_or(deps);
+
+    let mut declared = BTreeSet::new();
+    for line in deps.lines() {
+        // `assay-core.workspace = true` and `assay-policy = { … }` are both dependency lines, and
+        // a crate name never contains a dot, so the key ends at the first `.`, `=` or space.
+        let name = line.split(['.', '=', ' ']).next().unwrap_or("").trim();
+        if name.starts_with("assay-") {
+            declared.insert(name.to_string());
+        }
+    }
+    assert!(
+        !declared.is_empty(),
+        "parsed no assay-* dependencies; the manifest shape moved"
+    );
+
+    // `assay-core` is scanned in full by SCANNED, so it is not in the pub-const-only list.
+    let separately_scanned: BTreeSet<String> = SCANNED
+        .iter()
+        .filter_map(|dir| dir.strip_prefix("crates/"))
+        .filter_map(|rest| rest.strip_suffix("/src"))
+        .map(str::to_string)
+        .collect();
+
+    let listed: BTreeSet<String> = DEPENDENCY_CRATES
+        .iter()
+        .filter_map(|dir| dir.strip_prefix("crates/"))
+        .filter_map(|rest| rest.strip_suffix("/src"))
+        .map(str::to_string)
+        .collect();
+
+    let expected: BTreeSet<String> = declared.difference(&separately_scanned).cloned().collect();
+    assert_eq!(
+        listed, expected,
+        "DEPENDENCY_CRATES has drifted from assay-cli/Cargo.toml. Every assay-* dependency is \
+         either scanned in full (SCANNED) or scanned for pub const identities (DEPENDENCY_CRATES); \
+         a dependency in neither is a crate whose published identities nothing classifies"
+    );
+}
--- a/crates/assay-cli/tests/cli_json_identities.rs
+++ b/crates/assay-cli/tests/cli_json_identities.rs
@@ -1,0 +1,456 @@
+//! The CLI JSON identity inventory in `docs/architecture/CLI-JSON-IDENTITIES.md` is checked here.
+//!
+//! #2167 settled one convention for public CLI JSON documents: the discriminator is `schema`,
+//! carrying `assay.<segments>.vN`. Nothing held the *set* of those identities together — an
+//! emitter could arrive and no record would notice, which is how two counts of that set were
+//! published and withdrawn before this guard existed.
+//!
+//! Three properties, and the third is the one that matters:
+//!
+//! 1. Every identity string in production source is recorded, in the block that says what it is.
+//!    The two blocks are a partition, so "in neither" is a failure rather than a silence.
+//! 2. An identity written as an inline literal counts. Seven shipping verify-report documents are
+//!    declared that way, so a `const`-only collector would be satisfied by a coding convention
+//!    this codebase does not follow.
+//! 3. Documents that carry *no* identity are required rows. Measured: coverage, session-state
+//!    window, soak, `Baseline`, `HygieneReport`, both `run.json` writers, the sim-run report and
+//!    SARIF have zero identity constants between them. A pin built from a source scan alone
+//!    satisfies (1) and (2) while omitting every document #2485 has to migrate, and stays green.
+//!
+//! The inventory is never generated — not in CI and not locally-then-committed. This test reads it
+//! and never writes it.
+
+use std::collections::{BTreeMap, BTreeSet};
+use std::path::PathBuf;
+
+const INVENTORY: &str = "docs/architecture/CLI-JSON-IDENTITIES.md";
+
+/// Crates whose production sources are scanned.
+///
+/// Both crates entirely, not a subdirectory of either. Scoping to `assay-core/src/report` hid
+/// `assay.run_report.v1`'s siblings under `mcp/`, `otel/` and `discovery/`, which is the same
+/// class of mistake this file exists to catch.
+const SCANNED: &[&str] = &["crates/assay-cli/src", "crates/assay-core/src"];
+
+fn workspace_root() -> PathBuf {
+    PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("../..")
+}
+
+fn read(rel: &str) -> String {
+    let path = workspace_root().join(rel);
+    std::fs::read_to_string(&path).unwrap_or_else(|e| panic!("read {}: {e}", path.display()))
+}
+
+/// The lines recorded under a `<!-- machine-checked: NAME -->` marker.
+///
+/// Anchored on an HTML comment rather than a heading, so the prose can be rewritten without
+/// silently changing what is checked.
+fn recorded(marker: &str) -> Vec<String> {
+    let doc = read(INVENTORY);
+    let needle = format!("<!-- machine-checked: {marker} -->");
+    let after = doc
+        .split_once(&needle)
+        .unwrap_or_else(|| panic!("{INVENTORY} has no `{needle}` marker"))
+        .1;
+    let block = after
+        .split_once("```text\n")
+        .unwrap_or_else(|| panic!("marker `{marker}` is not followed by a ```text block"))
+        .1;
+    let block = block
+        .split_once("\n```")
+        .unwrap_or_else(|| panic!("marker `{marker}` block is unterminated"))
+        .0;
+    block
+        .lines()
+        .map(str::trim)
+        .filter(|line| !line.is_empty())
+        .map(str::to_owned)
+        .collect()
+}
+
+fn recorded_set(marker: &str) -> BTreeSet<String> {
+    let mut set = BTreeSet::new();
+    for line in recorded(marker) {
+        assert!(
+            set.insert(line.clone()),
+            "{INVENTORY}: `{marker}` records {line:?} twice"
+        );
+    }
+    set
+}
+
+/// Source with `#[cfg(test)] mod … { … }` bodies removed.
+///
+/// A test module inside a production file is not a producer. It is removed by brace matching
+/// rather than by a path or filename rule, because `supply_chain_conformance/tests.rs` is a file
+/// named `tests.rs` that sits in `src/`: any name-based rule has to decide it, and deciding it
+/// wrongly is silent in both directions.
+fn strip_test_modules(src: &str) -> String {
+    let mut out = String::with_capacity(src.len());
+    let mut rest = src;
+    while let Some((attr, body)) = find_test_mod(rest) {
+        out.push_str(&rest[..attr]);
+        rest = &rest[skip_balanced(rest, body)..];
+    }
+    out.push_str(rest);
+    out
+}
+
+/// `(offset of `#[cfg(test)]`, offset just past its `mod … {`)`, if the attribute introduces a
+/// module. An attribute on anything else — a `fn`, a `use` — is left alone.
+fn find_test_mod(src: &str) -> Option<(usize, usize)> {
+    const ATTR: &str = "#[cfg(test)]";
+    let mut from = 0usize;
+    while let Some(rel) = src[from..].find(ATTR) {
+        let attr = from + rel;
+        let after = &src[attr + ATTR.len()..];
+        let trimmed = after.trim_start();
+        if let Some(tail) = trimmed.strip_prefix("mod ") {
+            if let Some(brace) = tail.find('{') {
+                if tail[..brace]
+                    .trim()
+                    .chars()
+                    .all(|c| c.is_alphanumeric() || c == '_')
+                {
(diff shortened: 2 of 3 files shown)


Rul1an:
**Cursor review** on `c459a972`, against the classification, not the row list. Cannot submit `REQUEST_CHANGES` as the PR author account; this comment is the review record for that SHA.

The unopened *documents* I followed (phase-timing, supply-chain-conformance, side-effect verification, both execution-record reports, trust-basis assert) do write a top-level `schema` to stdout or a path. Those six belong where you put them.

The unopened *non-documents* do not. Six identities in `not-cli-documents` are CLI JSON documents under the definition this file writes down: a top-level JSON document written to stdout or to a `--report` / artifact path, carrying this string as its `schema`. Moving them there stays green. That is M firing on Q.

## Q — six documents filed as non-documents

Opened the emit path, not the declaration site.

| identity | classified as | actually |
|---|---|---|
| `assay.mcp_server_inventory.v0` | not a document | `assay mcp inventory` pretty-prints it to stdout or `--out` (`inventory.rs`). The command comment is “emit the carrier”. Docs call it the document that command emits. You already distinguished unnamed `discover_inventory` from this identity, then classified this identity as if it were not also a CLI write. |
| `assay.skill_supply_chain.v0` | not a document | `capture-skill-supply-chain` and `adapt-skill-scan` both `write_stdout_json` it (or `--out`). It is *also* an evidence event when imported. The partition is exclusive, so the CLI write has to win or #2485 never sees it. |
| `assay.otel_projection.v0` | not a document | `assay project-otel` writes it as “the only thing written to stdout (or to `--out`)”. |
| `assay.tool_decision_truth.otel_projection.v0` | not a document | the same command, `--evidence-bundle` branch. |
| `assay.enforcement_health.v0` | not a document | `assay monitor --enforcement-health <path>`. Sibling of `assay.monitor.observed_peers.v0`, which is in `cli-documents`, written the same way from the same invocation. |
| `assay.enforcement_health.v1` | not a document | `assay sandbox --enforcement-health <path>`. |

The systematic error is keyword classification. “Carrier”, “projection”, and “health artifact” were treated as not-a-document. `observed_peers` escaped because you opened that producer. The two otel identities did not, and they are the cleanest counterexample: the file’s own not-document block names “projections” as a reason, and `project-otel` exists to print one.

That is the same act, at the same scale, that produced the 86 and the three shapes. 21 of 56 is a number about the sitting, not about the emit paths.

I did not re-audit every remaining not-document. `aee_run_context` / `aee_seal_key` / `aee_trust_set` are inputs the sandbox *reads*. `render_safety_conformance.v0` is a golden, not a CLI write. `fallback_projection` (opened) is an input name. Those can stay. The six above cannot.

## M — the case that classification should not be in the hand file, and whether it wins

For identities that have a clap write of `{ schema: this identity }` to stdout or a caller path, the hand file should not be allowed to disagree. That is derivable from behaviour. The test already does this shape for unnamed documents: producer path + token. Do it for named documents too: each `cli-documents` row names its write site, and an identity with a write site that sits in `not-cli-documents` fails.

That does *not* mean the whole partition becomes generated. Events, digest domains, and fail-closed inputs have no such write, and a scan still cannot tell those from a document. The hand file keeps that remainder. It loses the right to reclassify a write as a non-write.

It wins for the six above. It does not win for `assay.tool_args.v0`. Membership-only is then a property of the remainder, not of the whole set.

`describe_binds_only_recorded_documents` does not help: it is one-directional and `BINDING_ROWS` omits even documents you classified correctly.

## N — `assay.trust-basis.diff.v1`

Widening the collector to `assay-evidence` is the wrong repair. The CLI already writes the document (`write_diff_json` in `trust_basis.rs`); the literal lives in `assay-evidence` because the report struct does. Follow the write, not the crate. A producer-row for named documents would have recorded this without scanning a crate of events. Recording a hole instead of the emit path is the same keyword mistake as Q, with a crate name instead of “projection”.

## O — REQUIRED in the test

Two-place editing is a slip guard, not a judgement guard. Dropping a row by accident in the markdown fails the test; dropping it because you decided it is not a document takes both edits in the same sitting, which is how the sitting works. The producer+token check is the real third place, and it still cannot see a document you never listed. Bind named-document classification to write sites (M) and this question mostly goes away.

## P — scanner quiet counterparts

Raw strings are handled. A mid-UTF-8 slice would panic, which is loud. Nested block comments are the quiet counterpart you asked for: Rust nests `/* */`, `skip_balanced` closes at the first `*/`. A `#[cfg(test)] mod` whose body contains `/* /* } */ …identity… */` can close the module early and leave the identity in the scan as production, or the reverse. Unlikely in this tree; it is the same class as the `"{"` bug (delimiter vs language), and it would not fail the unbalanced assertion.

`identities_in_line` is also line-based and stops at the next `"`, so an escaped quote or a split literal is a quiet miss. That is a collection hole, not a classification hole.

## Verdict

Blocking on Q. The pin currently records 21 documents. At least 27 exist under the file’s own definition, and the extra six are already in the not-document block, so the test is green on a wrong partition. Move those six. Then give `cli-documents` write-site rows so the next sitting cannot reclassify a write.

Not blocking, and not a classification finding: `assay calibrate` writes a `schema_version: 1` JSON file, and `evidence project-skill-bom` writes CycloneDX to stdout. Those are unnamed-document candidates in the same family as SARIF. Out of scope for this pass unless you want the required set grown again the same way `baseline_diff` was.

Rul1an:
Exact head `c459a972901abd9f2d38aaf3246a468e9394ef19`. Blocking on M and Q. N rides with them. O and the membership pin hold.

(Cannot `REQUEST_CHANGES` on your own PR with this token; this is the same review.)

## C, applied to M and Q

The case that classification should not live in a hand-edited file:

The pin is hand-edited so *membership* can fail. A generated pin is the producer restating itself. That reason does not extend to the second judgement stuffed into the same two blocks. Moving a row from `cli-documents` to `not-cli-documents` stays green. Fifty-six classifications in one sitting is the same act that produced the 86 and the three shapes. The file stores a judgement that cannot fail, and then tells the reader the partition is the guard.

That case wins for the document bit. It does not win for the whole file. Events, inputs and digest domains still need a human. Split it:

- Keep the *set* as a hand-edited pin (membership vs scan).
- Bind each `cli-documents` row to an emit site the way unnamed rows already bind to a producer token (command + a token that is the write, not the const).
- Require every identity a `cli/commands` path writes as a top-level `schema` to be in `cli-documents`, or to carry an explicit opt-out that names the command.

Then a wrong classification is a red, not a silence. The rest of `not-cli-documents` can stay a judgement.

The evidence that this is not theoretical is already on this head.

## Q — two rows, unopened, wrong, green

The file's own definition: a top-level JSON document written to stdout or to a `--report` / artifact path, carrying this string as its `schema`.

**`assay.mcp_server_inventory.v0` is in `not-cli-documents`.** `assay mcp inventory` writes that carrier to stdout or `--out` (`crates/assay-cli/src/cli/commands/inventory.rs:55-62`, schema const at `crates/assay-core/src/discovery/inventory_carrier.rs:21`). The unnamed-row note even says `discover_inventory` is **not** this identity, which only makes sense if this identity *is* a document. It is classified as the opposite. The test is green.

**`assay.skill_supply_chain.v0` is in `not-cli-documents`.** `assay evidence capture-skill-supply-chain` writes it to stdout or `--out` (`skill_supply_chain_capture.rs:51-63`, wired at `evidence/mod.rs:66-67` and `:192-193`). The verify report is a different identity and is correctly a document. The carrier is also a CLI emit.

Same class as `assay.supply_chain_conformance.v0`, which *is* in `cli-documents`. Three CLI-emitted carriers, two of them on the wrong side of a partition that cannot notice.

Named unopened documents I opened (`runner_phase_timing`, `supply_chain_conformance.v0`, `side_effect_verification`, both execution-record reports, `tunnel-observed.report`, `trust-basis.assert.v1`): those are real CLI writes. The error is in the unopened *not*-document block, which is where you said to look.

The confidence section also never lists `evidence.schema.{list,show,validation}`, `init_report`, `mcp_preflight`, `monitor.observed_peers`, `privileged_mcp_action.verify.report`, or `tool_decision_truth.verify.report` as opened or unopened. I opened them; they look like documents. That gap is manners, not the block.

## N — recording the hole is the wrong call

`assay.trust-basis.diff.v1` is declared in `assay-evidence` (`trust_basis/types.rs:4`). `assay trust-basis diff --format json` prints `TrustBasisDiffReport` to stdout (`trust_basis.rs:74-84`, `write_diff_json` at `:299-303`). CLI tests already assert `stdout["schema"] == "assay.trust-basis.diff.v1"`.

It is assert's sibling. Assert is a document because its const lives in `assay-cli`. Diff is a hole because its const does not. That is a scan-scope accident, not a classification.

Do not widen `SCANNED`. Widening pulls in mandate events, coding-agent packs, runner kernel events, and every other `assay-evidence` / `assay-runner-core` identity that is not a CLI document.

Add a third required-row class: CLI-emitted, identity literal outside the two crates, producer + token. Same instrument as the unnamed twelve. That also covers `supply_chain_conformance.v0`, whose `SCHEMA` lives in `assay-registry` (`supply_chain.rs:25`) and whose only hit inside the scanned crates is a `#[cfg(test)]` *file* module (`supply_chain_conformance/tests.rs:69`). Production `assay-cli` source has that identity only in doc comments, which `identities_in_line` skips.

If M5 is later extended to file modules — the consistent reading of "a test module is not a producer" — that document falls out of the collected set and the pin goes stale while the CLI still emits it. The row is in the pin today because the stripper does not do what M5 claims.

## O — two-place is the M2 red

`REQUIRED` in the test is why deleting an unnamed row fails. Drop it and M2 is green: the test would just read whatever the pin still lists. Coordinated delete of both copies remains possible, same as any pin. The producer+token check is the actual second instrument.

Keep `REQUIRED`. It is not a moved single point of failure. It is the unnamed-row guard. Do not confuse it with M.

## P — yes, the fixed scanner has a quiet counterpart

The `"{"` bug failed loud. These do not.

1. `identities_in_line` collects `"assay.*.vN"` on a line that also has code. `let x = 1; // "assay.foo.v0"` and `/* "assay.old.v0" */` both count. Probe confirmed. A commented-out last occurrence keeps the pin row; a comment-only identity demands classification. Quiet extra.

2. `strip_test_modules` only strips inline `#[cfg(test)] mod name { … }`. `#[cfg(test)] mod tests;` file modules are walked as ordinary `.rs`. M5 does not apply to them. Live: `supply_chain_conformance/tests.rs` is the only scanned production-tree occurrence of a document identity (see N). `#[cfg(test)] fn` is also left alone.

3. Nested Rust block comments. The skip is first-`*/`. Replica: extra `{` inside the still-open outer comment leaves `depth > 0`. If that is EOF, the existing assert is loud. If production code after the module supplies a `}`, the strip eats it — quiet over-strip. I did not find a nested comment inside a live inline test module on this head. Theoretical, same class as the original `{` bug.

4. Mid-UTF-8: `"` is ASCII; `start` after a quote is a char boundary. Not a hole.

5. Identity on a quoteless line inside a multi-line raw string: miss. Theoretical.

(1) and (2) exist on this tree. (2) is load-bearing for a document row today.

## What holds

Membership pin, no generator, inline literals, twelve unnamed rows, M1–M4 red and M5 as a control, `describe ⊆ pin` one-way, the three blind-diff names. Those are not the block.

## What would unblock

- Move `assay.mcp_server_inventory.v0` and `assay.skill_supply_chain.v0` to `cli-documents`, or change the definition so "carrier" is not a document and then apply that to `supply_chain_conformance.v0` too. The current split is two-out-of-three.
- Bind document classification to an emit site, so the next move is red.
- Promote `assay.trust-basis.diff.v1` (and `supply_chain_conformance.v0`'s registry const) to required-rows, not a recorded hole.
- Leave `REQUIRED` where it is.

I am not asking to close this by widening the scan.

Rul1an:
Third-round review on `3a5f05cb3`. Cannot `REQUEST_CHANGES` as the PR author account; this comment is the review record for that SHA.

The six Q moves and the write/name split hold. I opened the remaining twenty-nine non-document reasons. Those hold. The live miss is not another false sentence in that block. It is an identity the collector never collected.

## R — the reason column is theatre as a gate

A false reason passes. The test asks three things: a `|` is present, the right-hand side is non-empty, the identity is not also in `cli-documents`. It does not read the sentence. `a banana` is as green as `input the sandbox reads`.

"You must write a lie" is a review surface, not a guard. It changes the cost of a misclassification from deleting a line to writing a sentence, which is why two reviewers could catch the six. It does not change the build. The person who wrote the six wrong classifications wrote all twenty-nine reasons in the same sitting. The column would have stayed green if those six had been moved with a false sentence instead of being left as a bare list.

That is weaker than the label in the file already admits, and #2485 should not rely on it. The mechanical half of this revision — a document row names a writer — is a guard. The reason column is the thing you do for the remainder that a static check cannot decide. Keep it as a record. Do not let a later sitting treat a filled-in reason as proof the row is right.

The symmetric instrument, if you want a gate for the remainder: each non-document row names a proving site the way a document row names a writer (the parse, the event-type const, the nested field). Empty and missing sites go red. A false sentence that points at a real site still passes. That is a smaller hatch, not a closed one.

## S — the fifth crate is not hypothetical

`assay-runner-schema` is the fifth. `assay monitor --observation-health <path>` writes `assay.runner.observation_health.v0` as a top-level JSON file (`write_to` in `monitor_next/observation_health.rs`: `to_string_pretty` + `std::fs::write`). The const is `OBSERVATION_HEALTH_SCHEMA` in `assay-runner-schema/src/health.rs`. Production `assay-cli` / `assay-core` source never contains the literal — comments and a `#[cfg(test)]` module only, both skipped. The collector does not see it. Neither block records it. The test is green.

It is the sibling of `assay.monitor.observed_peers.v0`, written from the same invocation, the same way `enforcement_health` is. Observed-peers and enforcement-health are documents. Observation-health is invisible. Same class as N, with a crate that is not even on the naming-column inventory.

A rule that notices a fifth without scanning the workspace: reverse `documents_are_bound_to_a_writer`. Today a row must name a writer. A writer need not be named by a row. `observation_health.rs` sits under `cli/commands`, calls two idioms already on `WRITERS`, and no row points at it. Require every production file under `cli/commands` that contains a writer idiom to be named by a `cli-documents` row or an `unnamed-documents` row, or to carry an explicit opt-out that says why this write is not a document. That follows the write. It does not walk `assay-evidence` looking for mandate events.

`Cargo.toml` of `assay-cli` is the other short list (core, evidence, registry, monitor, common, policy, metrics, mcp-server, the three runner crates, sim). Scanning that closure is not scanning the workspace. The converse-writer rule is the tighter one and would have recorded this without opening `assay-runner-schema`.

Do not widen `SCANNED` to `assay-runner-schema`. That crate is a dictionary of archive identities. Following the write puts one row on the pin:

`assay.runner.observation_health.v0 | crates/assay-cli/src/cli/commands/monitor_next/observation_health.rs | crates/assay-runner-schema/src/health.rs`

## T — idioms `WRITERS` still does not contain

Opened the emit paths, not the list.

| idiom | where a CLI JSON write uses it today | would `WRITERS` see the file? |
|---|---|---|
| `serde_json::to_vec_pretty` | `coverage/io.rs`, `session_state_window.rs` | no |
| `tokio::fs::write` | those two, plus `explain.rs`, `mcp/coverage_input.rs` | no |
| `serde_json::to_string` (compact) | `mcp wrap --tool-decision-truth-out`, `trace`, `import` | no |
| `fs::write` after `use std::fs` | `project_skill_bom.rs`, `skill_supply_chain_capture.rs`, `adapt_skill_scan.rs` | no (`std::fs::write` is the needle) |

`to_writer` still matches `to_writer_pretty`. `describe` is covered. The hole is the other direction: a file that only uses `to_vec_pretty` + `tokio::fs::write` can be named as a writer and the test stays green only because those two unnamed rows point at a *different* file (`coverage/report.rs`, `session_state_window.rs` for the token) and unnamed rows do not run the writer check.

If #2485 mints identities for those two and binds the actual write file, `documents_are_bound_to_a_writer` goes red on a real document, or stays green if someone points the row at the naming file instead. Either way the list is what `describe` already taught you it is: a remembered set of idioms.

I did not find a current *named* document whose only emit path is off the list. `runner_spike/phases.rs` uses `to_vec_pretty` but also `std::fs::write`. Observation-health, if added, would pass as written today.

## U — the other fifty

Twenty-nine remaining non-document reasons, opened.

`aee_*` are parse/load paths. `fallback_projection` is an input-mode name. `recipe_row` is packed into a bundle. Receipts go through `BundleWriter`, not stdout. `render_safety_conformance` is a golden. `tool_args` and `approval_artifact.structured_meta_jcs` are digest domains. `enforcement_decision`, the two mandate events, `manifest_establish`, the two snapshot events, `tunnel_observed.v0`, and `denied_call_observation.*` are events or nested objects; the commands write a *report* identity that is already in `cli-documents`. `supply_chain_conformance.input` is the descriptor the command reads.

Those sentences are true. They were also written by the sitting that got the six wrong, so "true today" is not a property of the column.

Boundary, not a Q-move: `assay mcp wrap --tool-decision-truth-out` appends `assay.tool_decision_truth.v0` objects as NDJSON to a caller-named path. The reason files the identity as an input because `project-otel` reads it. That is the carrier-word again, applied to a sink that is a write. NDJSON of one-object-per-line is a worse fit for "top-level JSON document" than observation-health is. Record the sink or change the definition so a wrap-side NDJSON emit is not a document. Do not leave the reason implying the CLI never writes the identity.

Unnamed set, still short, same family as `baseline_diff`:

| write | identity |
|---|---|
| `assay calibrate --out` | integer `schema_version: 1` |
| `assay evidence project-skill-bom` | CycloneDX 1.6 |
| `assay evidence lint --format json` | no `assay.*` schema |
| `assay evidence lint --format sarif` | foreign, like the existing `sarif` row |
| `assay evidence list --format json` | ad-hoc object |
| `assay evidence diff --format json` | `DiffCliDocument` |
| `assay config-path --json` | ad-hoc object |

Round one already named calibrate and the skill BOM. They are still absent. The required-rows argument applies to them the same way it applied to the three a second reader found.

## Verdict

Blocking on S. The pin records twenty-seven named documents. `assay.runner.observation_health.v0` is a twenty-eighth under the file's own definition, written by a shipping command, declared in a crate the collector does not scan and the naming column does not list. Add the row. Bind command-side writers so the next crate is a red, not a comment in the limits section.

R is answered: do not let #2485 treat the reason column as a check. T is a list to extend or to replace with the converse-writer rule; extending the list is how `to_writer` arrived. U did not produce a seventh Q-move in the classified fifty. The miss was the identity that never reached the fifty.

Rul1an:
Round three landed at `286e4c67a`.

**One real miss, found by opening the emit path** — which is what a guard is supposed to make unnecessary. `assay monitor --observation-health` writes `assay.runner.observation_health.v0` from `monitor_next/observation_health.rs`, and the identity is declared in **`assay-runner-schema`**, a fifth crate. The pin had no row and every test was green. Recorded now.

**The writer idiom list was wrong a second time**: `to_vec_pretty`, `tokio::fs::write`, `serde_json::to_string` and a bare `fs::write` all emit CLI JSON today and none was on it.

**On the reason column**: the review's verdict is right and I have adopted it verbatim in the file — it is a review surface, not a gate. The test checks that a `|` exists and the right side is non-empty; it does not read the sentence. `a banana` passes. #2485 must not treat a filled-in reason as evidence the row is correct.

**The shape under both misses**, which matters more than either: the guard follows rows to writers and never writers to rows. A document row must name a file that writes; a file that writes need not be named by any row. Measured on this head, **31 production files under `cli/commands` serialize JSON and are named by no row** — `evidence lint`, `evidence diff`, `evidence list`, `store-status`, `explain`, `import` and `trace` among them.

That triage is **not** in this PR. Thirty-one classifications by one author in one sitting is the precise act this inventory exists to record going wrong; in this session the same act produced a false count twice, six wrong classifications, and a document list two independent readers each had to correct. It is #2555 with the measured list attached and a DoD that requires re-measuring the count on the head that lands it.

The limits section of the file now states all of this, including that the writer list has been wrong twice.

Rul1an:
Round 3, exact head `3a5f05cb3baa14955cd0cf9d532788e44e89d108`. The six you conceded are not the review. R is theatre. S is not hypothetical. T is a substring list. U still has at least one keyword reason.

## S first, because it is a row, not a lecture

`assay.runner.observation_health.v0` is a CLI JSON document on this head and it is not on the pin.

`assay monitor --observation-health` writes it (`monitor.rs:73-80`, `monitor_next/mod.rs:669-679`, `observation_health.rs:130-145`). The schema const lives in `assay-runner-schema` (`health.rs:6`). In `assay-cli` production source the identity string appears only in a `//!` comment, which `identities_in_line` skips, and in a `#[cfg(test)]` assertion, which the stripper removes. The collector cannot see it. The naming column cannot see it. This is the fifth crate, already shipping.

Same shape as `trust-basis.diff` was, before you followed that write by hand. Following writes by hand does not notice the next one.

A rule that notices a fifth without scanning the workspace: collect `pub const … = "assay.*.vN"` from every crate `assay-cli` depends on, and require each of those strings on the pin. That is the import graph, not the workspace. It would have required a row for `OBSERVATION_HEALTH_SCHEMA` (and for `TRUST_BASIS_DIFF_SCHEMA`, `assay-registry::SCHEMA`, and the evidence receipt consts). An identity in a crate the CLI does not depend on stays out. An identity in a crate the CLI does depend on and does not classify fails the build.

Do not widen `SCANNED` to those crates' full source. That is how mandate events and kernel events flood the partition. Const-in-deps is enough.

`assay-cli` already `use`s `assay_runner_schema::SDK_EVENT_SCHEMA` (`runner_spike/spec.rs:4`). That crate was never a future tense.

## R — theatre for the thing #2485 would want

A false reason passes. I confirmed it: `non_document_rows` checks `split_once('|')` and `!reason.is_empty()`. Nothing else.

"You must write a lie" would have caught the six only if the author had refused to write a sentence. The actual bug was a theory about the word "projection". The sentence that theory produces is `a projection, not a document`. That sentence is non-empty. M6 stays green.

It is a guard against deleting a line. It is not a guard against the classification error this file exists to stop. Labelling it weaker in the file is accurate. Treating it as a source of truth in #2485 is not. Do not migrate from reasons.

A reachable property that is not a sentence: a non-document identity may not be the `schema` of a value a `cli/commands` path writes with a writer idiom. That is the inverse of the document-row bind. It still would not have seen `observation_health` (the string is not in the writer file). S and this inverse are different holes.

## T — the list is a substring, and it is not the write

`WRITERS` is `write_stdout_json`, `to_string_pretty`, `to_writer`, `std::fs::write`, `println!`. `documents_are_bound_to_a_writer` asks whether the named file contains any of those strings, not whether it writes *this* identity.

Missing idioms that emit JSON in this tree:

- `write_document` (`output_write.rs:12`). Callers often also have `write_stdout_json`; a file that only uses `write_document` fails the bind (false red) or gets pointed at a sibling that `println!`s (false green).
- `serde_json::to_string` without `_pretty` (`trace.rs:64-65` `writeln!(writer, "{}", json)`, `aee_seal.rs:414` NDJSON).
- `write_all` of already-serialized bytes (`trust_basis.rs:67`).
- `writeln!(f, "{}", json)` in `mcp/audit.rs`, `mcp/lifecycle.rs`, `decision_next/emitters.rs`.

`to_writer` matches `to_writer_pretty` by accident. `println!` matches an error print. Binding `assay.foo.v0` to a file that writes `assay.bar.v0` and also contains `println!` is green.

`observation_health.rs:131-138` already uses `to_string_pretty` and `std::fs::write`. The writer list would have accepted that row. It was never offered the row. T is not why S is missing.

Second writer of a recorded document, unnamed: `adapt_skill_scan.rs:64-71` writes `assay.skill_supply_chain.v0` to stdout/`--out`. The pin names only `skill_supply_chain_capture.rs`.

## U — I did not read all fifty. I read the ones that look like the six.

`assay.tool_decision_truth.v0` reason: "the carrier read as input by project-otel; the projection is the document." `assay mcp --tool-decision-truth-out` appends those carriers, one JSON object per line, to a caller-named path (`cli/args/mcp.rs:135-142`). Experimental, still a write. "Carrier" is the same word that hid `mcp_server_inventory` and `skill_supply_chain`.

`assay.denied_call_observation.v0` / `v1` and `assay.manifest_establish.v0` as "nested inside the privileged-mcp-action verify report" — they are also *inputs* (`--` NDJSON / files on that command). Half-true. Not the block.

Receipts "never to stdout" — true. They are written to a caller-named bundle path by `BundleWriter`. The definition on this file is "stdout or a caller-named path". If the top-level file schema is the bundle's, the reason holds. If a reader treats each event as a document the CLI wrote, it does not. I am not promoting that to a block; I am saying the sentence is doing the same work "projection" did.

I did not open every remaining document row's write. The ones I did open among the named-unopened set last round still write. The miss I can stand on is `observation_health` plus the `tool_decision_truth.v0` reason.

## What holds

Write/name split on doctor and evidence-schema. Required reasons (as a speed bump). Four-crate admission in the prose. File-module stripper. M1–M5 as stated. `REQUIRED` for the unnamed twelve. The six moves.

## What would unblock

- Put `assay.runner.observation_health.v0` on `cli-documents` (`observation_health.rs` writes, `assay-runner-schema/src/health.rs` names).
- Const-in-`assay-cli`-deps membership, so the next fifth crate is a red, not a comment.
- Do not let #2485 read the reason column as a fact.
- Say whether `tool_decision_truth.v0` NDJSON is a document under this file's own definition, and apply that answer to every other `--out` carrier.

I am still not asking to scan the workspace.

Rul1an:
Round 4, exact head `459c7908fdd743c42f49516f63c42f86486fa6d0` (ef4f20841 plus the two merges). The rule landed. The row format did not grow. Four things, then what blocks.

## V — there is a static check short of dataflow

`documents_are_bound_to_a_writer` still asks whether the named file contains any writer idiom. Binding `assay.foo.v0` to a file that writes `assay.bar.v0` and also has `println!` is still green. I measured it: six document rows do not even have the identity literal in the writer (`evidence.schema.{list,show,validation}`, both otel projections, `trust-basis.diff`). Those six are exactly the write/name splits. Every other row's writer already contains the string.

The reachable check, no schema-field parsing:

- When the namer is `-`, the writer file must contain the identity literal (or a `const` whose value is that literal). That is almost every row today.
- When the namer is a different file, the writer must mention a symbol that file publishes. `schema/write.rs` already `use`s `SchemaListReport` from `reports.rs`. `trust_basis.rs` already names `TrustBasisDiffReport`. `project_otel.rs` has to touch the projection types or it is not the writer.

A file that only has `println!` and does not mention this identity or its namer types goes red. That is the foo-to-bar bind. It is not rustc. It is the same token check the unnamed rows already use, pointed at the writer.

I would not hold this PR for it. I would put it on the same ticket as #2555 or as a one-function follow-up. It is worth more than another pass over reasons only if you write it; it is not worth more as a paragraph.

## W — one writer is enough for this guard

The row answers "is this a document". One write proves that. `adapt_skill_scan.rs` is a second emit path, not a second identity. Listing every writer in the pin recreates the 31-file sitting this file is supposed to refuse.

`adapt_skill_scan` belongs on #2555's list, next to `evidence lint` (which already `println!`s `to_string_pretty` at `lint.rs:142` on this tree). Do not grow the row.

## X — I did not find a seventh keyword row. I also did not read twenty-one emit paths.

The 21 reasons say they were judged from the declaring module. I opened the ones that reuse the words that already failed: `token_passthrough_conformance` and `tool_annotation_conformance` ("carrier inside a bundle"), `provider_audit_record` (CLI only imports the files), `redaction_receipt` (no `assay-cli` write; the type comment says it can stand alone), `capability_surface` (project-otel input). None of those is a CLI stdout/`--out` document I can see.

That is not the same as "they are right". It is "I cannot name a wrong one from the declaring module, and neither could you when you wrote them". The prose already records the weaker basis. That is enough for this pin. It is not enough for #2485. Same instruction as R: do not migrate from these sentences.

Two of them still say "carrier". The word is not banned. It is how the last three errors were written.

## Y — they do not replace each other, and the hand list is the leftover S

#2555 is open and correctly scoped. Const-in-deps catches a new *crate*. Converse-writer catches a new *command file* in a crate you already scan. `observation_health` is now the first class. `evidence lint` / `evidence diff` are still the second.

`DEPENDENCY_CRATES` is twelve paths you typed. I checked them against `assay-cli/Cargo.toml`'s `assay-*` deps: they match today, with `assay-core` scanned separately. There is no test that they keep matching. Adding `assay-new` to the Cargo.toml and not to that array is the sixth crate, and M12 stays green. The rule you adopted is only as tight as this list. Bind the array to the manifest the way unnamed rows bind to `REQUIRED`.

Private consts stay out (`assay.proxy_observation_health.v0` and `assay.declared_mcp_manifest.v0` are `const`, not `pub const`, and the mcp-server binary writes them). M13 is doing that work. They are not CLI documents.

## R, as committed rather than as described

You said the file calls the reason column a speed bump and tells #2485 not to read it as truth. On this head it does not. Lines 198–202 still say a reclassification is caught *because* every non-document row must carry a reason. The test comment is the same. There is no "speed bump" and no "#2485 must not". The ceiling you accepted is not the sentence that shipped.

The same section (lines 222–227) still says the converse-writer rule is what turns a sixth crate from invisible to answered, and that until it lands a sixth crate would be invisible. Const-in-deps has landed. That paragraph is now false in the other direction.

## What holds

M12/M13, trailing-comment skip, `r"…"` raw, the widened `WRITERS` list, `tool_decision_truth.v0`'s corrected reason, `observation_health` on the pin, the 21 recorded at all, #2555 opened instead of answered here.

## What would unblock a merge of this pin

- A test that `DEPENDENCY_CRATES` equals the `assay-*` deps in `assay-cli/Cargo.toml` (minus the two scanned crates). Without it the rule you adopted has the same hole it was written to close.
- Rewrite the two stale paragraphs: reason is a speed bump, #2485 does not read it; const-in-deps is what notices the next crate, #2555 is what notices the next command file.

V and W and the unread 21 do not block this PR. They block treating the pin as finished work.

PR #2556, opened 2026-08-19
fix(conformance): rescore every corpus over its real denominator, and withdraw a refuted claim

Rul1an:
Three corpora rescored over their real denominators, one published claim withdrawn after it was refuted, and the tool hardened where this work found it wrong.

## What started it

`conformance/INDEX.md` published a section claiming the `privileged-mcp-action-v0` survivors were **"one gap wearing four faces"** — that wherever a rule has sibling members, the corpus discriminates exactly one and leaves the rest free. Tidy, built on four data points, and published before anything attacked it.

An adversarial re-measurement deleted **23 promised rules at once** — rule deletion here is monotone, so they compose — and **reproduced all fourteen outcomes and all five claim matrices exactly**. The only difference anywhere was one finding string, and findings are not on the surface the corpus declares.

The corpus isolates **six** rules. The claim undercounted by roughly a factor of six, and its crisp form is false; what survives is *"at most one member per family, and often zero"*, with five families discriminating none.

Three of its four table rows hold. The fourth was wrong on its **caught** side: of the marker triple's three legs the corpus discriminates none. `gen_vectors.py:103` hard-codes both code and origin and is the only marker construction site, and the schema leg is decided earlier at Stage 2 — *the same wrong-stage argument this manifest already used, two entries away, to rule a different mutant out of scope*. The argument was available and was not applied where it was inconvenient.

## The same habit on every other row

| corpus | first score | over the real denominator |
|---|---|---|
| `mcp-jsonrpc-id` | 4 of 4 — 100% | **6 of 10 — 60%**, 4 survivors |
| `observed-effect-v0` | 4 of 5 — 80% | **14 of 23 — 60.9%**, 9 survivors |
| rge-bench | 30 of 30 — 100% | **51 of 54 — 94.4%**, 3 survivors |
| `privileged-mcp-action-v0` | no result, then 4 of 8 | denominator 8 → **24** |
| `rfc8785` | control killed | unchanged — the one row not wearing a coat |

`mcp-jsonrpc-id` scored the presence and null arms only; `RequestId` is string or number, so the type arms belong in the denominator. rge-bench scored a hand-written table its own `ADMISSION.md` already warned was unchecked against `ref_example.py`.

The own-corpora test pinned `killed == 4` and `declared_total == 11`, which made the declaration a thing to preserve rather than a thing to test — adding rules broke it, and the change that kept it green was not adding them. It now pins the larger denominators.

## Two method errors, both ours

**The manifest read the wrong comparison surface.** It scored `[bundle_integrity, verdict, reason_code]`. `MANIFEST.json` states: *"expected.bundle_integrity, expected.verdict and expected.claims are the normative comparison surface; first_failure_informative is this generator's own vocabulary and is informative only."* So the measurement read the member declared informative and omitted the one declared normative — weaker **and** wider than the corpus at once. Measured consequence: a mutant flipping `ok-005`'s claim cell from refuted to confirmed survived the harness while the corpus kills it.

**Two adequacy runs contaminated one working tree.** Each reviewer found the other's mutant applied to a file it had not touched, and one run produced a believable `6 of 8 (75.0%)` where clean re-runs give `4 of 8`.

## Upstream, in `corpus-adequacy`

- `d686540` — a non-blocking POSIX lock, taken **before** the dirty check. A tree observed clean outside the lock can be mutated before the originals are captured, which is the same bug one step earlier.
- `5401922` — captured originals are compared against `git show HEAD:<path>` at capture time. The lock excludes another instance of the tool; it cannot exclude a hand edit or an instance that started before the lock existed.
- `6c41920` — an `outcome_from` member the implementation never emits is now refused. `doc.get(k)` returned `None` silently, so such a member compared `None` to `None` on every mutant and every score over it was over-generous by whatever it would have caught.

Each is pinned by tests at the artefact level rather than on the helper, and each positive control was run.

## Twenty-three deleted is not twenty-three new

Five of the 23 restate rules already declared; two are out of scope. **Sixteen** were new. Deduplicated by `(anchor, replacement)` rather than by label, because one pair is the same rule differing only by a trailing `=> {`. The larger number was the flattering one for a finding about under-declaration, so the smaller one is taken and the difference is recorded.

The monotonicity argument that licenses reading 23 individual survivals off one composite experiment is published with the condition that would break it — a deletion that *adds* findings. Widening the marker triple does exactly that in principle, so the three triple-leg mutants were built and run individually rather than inferred.

## Still open

Closing any of this needs new vectors, and a new vector moves a digest that #1840 publishes as a reproduction target. Addendum corpus at a second digest, or v0 left pinned and honest, is a contract decision rather than a tooling one and is not settled here.

## Summary by CodeRabbit

* **Documentation**
  * Expanded mutation-adequacy results with updated measurements, survivor analysis, control results, and documented safeguards.
  * Clarified scoring scope, out-of-scope rules, known corpus gaps, and outcome-surface analysis across existing adequacy declarations.

* **Tests**
  * Added adequacy coverage for observed-effect and RGE benchmark corpora.
  * Updated JSON-RPC identifier expectations to distinguish supported type-specific rules.
  * Added checks for manifest alignment and inadequate-score outcomes.

### DIFF
--- a/conformance/INDEX.md
+++ b/conformance/INDEX.md
@@ -146,7 +146,7 @@ the rows below say which of the two they mean.
 | Corpus | Runner | Result |
 |---|---|---|
 | `mcp-jsonrpc-id-conformance` | module | **6 of 10 in scope (60%), 4 survivors.** The first score was **4 of 4** over the presence/null arms alone. The positive control is a string id and RequestId is string or number, so the type arms belong in the denominator: string is isolated, number and bool are not. **7 envelope rules unexercised and out of scope**, ratio 0.7 — re-checked against the three published messages; none of them moves an outcome, and the reasons still hold |
-| `privileged-mcp-action-v0` | process | **4 of 8 in scope (50%), 4 survivors, 2 known holes.** First measured as *no result*: three rules declared, two of them mutating the wrong stage and the wrong profile. That was a fact about the declaration, not the corpus — eight more rules were read off the verifier path and half of them survive. **The denominator of 8 is the finding, not the 50%.** An adversarial re-measurement deleted **23 further promised rules at once** and reproduced all fourteen outcomes and all five claim matrices exactly, so the corpus isolates six rules of a declared twenty-four in scope, where eight had been declared before. Measured on the surface the corpus itself declares normative, after a lock and a HEAD check closed a cross-run contamination. See below. **This is the corpus carrying our live reproduction request ([#1840](https://github.com/Rul1an/assay/issues/1840))** |
+| `privileged-mcp-action-v0` | process | **6 of 25 in scope (24.0%), 19 survivors, 2 known holes.** Declared three times before it said anything true. First three rules and *no result*; then eight read off the verifier path, scoring 50%, published with a pattern claim. **The denominator was the finding, not the score.** An adversarial re-measurement deleted a further set of promised rules in one build and reproduced every outcome and every claim matrix, so the declaration grew from eight in-scope rules to twenty-five and the score fell to 24.0%. All six rules the corpus does isolate are now declared, and all six are killed. Measured on the surface the corpus itself declares normative, after a lock and a HEAD check closed a cross-run contamination. See below. **This is the corpus carrying our live reproduction request ([#1840](https://github.com/Rul1an/assay/issues/1840))** |
 | [`observed-effect-v0`](https://github.com/Rul1an/observed-effect-v0) drift consumer | batch | **14 of 23 in scope (60.9%), 9 survivors.** The first score was **4 of 5** over merge-policy rules only. The 14 case names announce the recompute, advisory, and profile rules as well. The original survivor remains (the body can be read whether or not the ref recomputed); the new ones are the fail-closed recompute siblings the cases never present, the RFC 8785 UTF-16 key-order rule the consumer claims but no body isolates, and a redacted-field conjunct that `incomplete_missing_non_claims` does not distinguish from a missing field |
 | [rge-bench](https://github.com/rge-bench/rge-bench) | module | **51 of 54 in scope (94.4%), 3 survivors, 2 declared equivalent, 1 out of scope.** The first score was **30 of 30** over the hand-written table in `scripts/check_rule_liveness.py`. [ADMISSION.md](https://github.com/rge-bench/rge-bench/blob/main/ADMISSION.md) already says that table is not checked against the rules `ref_example.py` contains. The missing rules were discriminable: the strength ladder (the ceiling ladder's sibling — the same shape that hole already had), the AND conjuncts scored as one mutant, and the soft-digest and replay-equality fallthroughs. The three survivors include a claimed contract-edge (`True` is not `1`) that no vector isolates |
 | `rfc8785` canonicalization | batch (test-names) | **control killed** — the corpus bites. No rules declared: the wrapper still has nothing of ours to delete. `to_string` is a second convenience over the same delegate and is not on this corpus's path |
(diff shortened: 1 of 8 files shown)


Rul1an:
## Do not merge yet: the run that would justify the new denominator invalidated itself

Declaring the sixteen further rules and re-measuring produced this:

```
4 of 23 DECLARED in-scope rules killed (17.4%). 4 declared out of scope, 29 rules declared.
FAIL: control 'CONTROL harness reachability on the verifier path' survived:
      the harness cannot detect a change on this path, so every other verdict
      in this run is meaningless
FAIL: 19 mutant(s) survived
```

**17.4% is not a result.** The control is the mutant that MUST be killed, and it exists precisely so a run where the harness detects nothing cannot be told apart from a corpus where nothing is caught. It survived, so every verdict in that run is void — including the nineteen survivors, which would otherwise have looked like a dramatic confirmation of the finding this PR is about.

That is the failure mode this whole PR is about, arriving one level up: **the number that flatters the argument is the one that showed up, and the guard is what stopped me publishing it.**

The same control was killed forty minutes earlier, on the same comparison surface, with twelve mutants instead of twenty-nine. Tree is clean, the control anchor is present and occurs exactly once. So the cause is not the surface and not a missing anchor; something about the longer run is the suspect, and it is being isolated by applying the control by hand and diffing all fourteen outcomes.

### What this means for the diff as it stands

- The table row still reads **4 of 8**, which was measured cleanly and remains true of the eight rules declared at that time.
- The manifest in this same diff now declares **24 in-scope rules**, so the row and the manifest no longer describe the same declaration.
- That inconsistency is real and is why this is marked do-not-merge rather than left to a reviewer to notice.

Either the row gets a valid measurement over twenty-four, or the sixteen new declarations come back out of this PR and land separately once the harness question is settled. I am not choosing the second just because it is quicker to make green.

Rul1an:
## Resolved: the control was killed, and the number is 6 of 25

The void run is diagnosed and the replacement is measured. Removing the do-not-merge.

```
6 of 25 DECLARED in-scope rules killed (24.0%). 4 declared out of scope, 31 rules declared.
control-killed. 19 mutant(s) survived. 2 KNOWN HOLES.
```

### Why the earlier run voided itself

A `git commit` during the run. Pre-commit stashes unstaged working-tree changes, runs its hooks, and restores them — and the unstaged change was the in-flight mutant. In that window the harness built unmutated source. The control runs last and landed in it: the run spanned `22:14:39–22:20:18` and the commit is stamped `22:19:58`.

Confirmed rather than inferred. Diffing per-mutant verdicts between the void run and the clean one gives **one** difference in twenty-nine, and it is the control. Applying the control by hand moves three vectors, and it is killed in a two-mutant run of the same manifest.

**Neither the lock nor the HEAD check covers this.** Pre-commit is not a second instance of the tool, and the HEAD comparison runs once when originals are captured, not before every build. The control was the only thing between this and publishing `4 of 23 (17.4%)` with nineteen survivors — which would have read as a dramatic confirmation of exactly the argument this PR makes.

### One correction moved a number up

An independent count named **six** isolated rules where the manifest declared four. The two missing were the Stage 2 unknown-schema arm and the decision vocabulary; both are now declared and both are killed. The first is literally the measurement this manifest's own out-of-scope note had asked for — *"to measure that promise, mutate the Stage 2 match instead"* — and never ran. Both anchors were compile-verified before being declared.

Worth stating because every other correction today went down, and a run of corrections that only ever go one way is its own kind of unreliable.

### The composite claim is narrowed to what executed

Reproduced independently: **nineteen** of the twenty-three applied simultaneously, all fourteen vectors byte-identical on `[bundle_integrity, verdict, claims]`. The other four share an anchor with a rule already in the composite (`31/32/33` are three replacements over one line, as are `44/45/46`) and are each scored survivors on their own. The conclusion holds; the count of what ran in one build does not, and the second is what the page now says.

Issue #2555, opened 2026-08-19
cli(json): follow writers to rows, not only rows to writers


Rul1an:
Parent: #2167. Follows #2484 (https://github.com/Rul1an/assay/pull/2554).

## The shape of the gap

`cli_json_identities.rs` follows **rows to writers**: a recorded document must name a file that writes. It never follows **writers to rows**: a file that writes JSON need not be named by anything.

That asymmetry is not theoretical. It is how `assay.runner.observation_health.v0` shipped unrecorded — `assay monitor --observation-health` writes it from `monitor_next/observation_health.rs`, the identity is declared in `assay-runner-schema` (a **fifth** crate, outside the collector), and every test stayed green. An independent review found it by opening the emit path, which is exactly what a guard is supposed to make unnecessary.

## The rule

Every production file under `crates/assay-cli/src/cli/commands` that serializes JSON must be either:

- named by a `cli-documents` row (as writer or namer), or
- named by an `unnamed-documents` row, or
- listed in an explicit opt-out block with a stated reason, where a stale entry is itself a failure.

JSON-serializing means one of `write_stdout_json`, `to_string_pretty`, `to_vec_pretty`, `to_writer`, `serde_json::to_string(`, `serde_json::to_vec(`. Deliberately **not** `println!` — 110 command files print, 58 serialize.

## Measured triage list

On `286e4c67a`, **31 files serialize JSON and are named by no row**:

```
calibrate.rs                       evidence/lint.rs                  profile_next/mod.rs
ci.rs                              evidence/list.rs                  profile_types.rs
config_path.rs                     evidence/livekit_tool_action/canonical.rs   replay/fs_ops.rs
coverage/io.rs                     evidence/mod.rs                   replay/provenance.rs
coverage/legacy.rs                 evidence/project_skill_bom.rs     run.rs
evidence/adapt_skill_scan.rs       evidence/skill_supply_chain.rs    sandbox/child.rs
evidence/attest.rs                 evidence/store_status.rs          sandbox/profile.rs
evidence/diff.rs                   explain.rs                        sim/soak/mod.rs
evidence/explore.rs                generate/model.rs                 tool/sign.rs
                                   import.rs                         trace.rs
                                   mcp/coverage_input.rs             trace/import_mcp.rs
```

Several of these very likely emit documents: `evidence lint`, `evidence diff`, `evidence list`, `store-status`, `explain`, `import`, `trace`.

## Why this is not folded into #2484

[...]

PR #2557, opened 2026-08-19
test(cli): tie a recorded writer to the identity it writes

Rul1an:
Closes the `V` finding from #2554's round-four review. Related: #2555, #2484, convention in #2167.

## What was wrong

`documents_are_bound_to_a_writer` asked whether the named file contains **any** writer idiom. So `assay.foo.v0` could be bound to a file that writes `assay.bar.v0` and happens to `println!`, and the guard stayed green.

The reviewer measured the gap rather than asserting it: six document rows do not contain their identity in the writer at all, and all six are the write/name splits.

## The check, without dataflow

- **Writer also names the identity** (`namer` is `-`): the file must contain the literal, or a constant whose value is that literal.
- **Naming file is separate**: the writer must mention a symbol the naming file publishes. A writer touching none of the namer's types is not the writer of its document.

## The part worth reading

**The first version passed the mutation it exists to catch.**

`mcp/preflight.rs` declares its identity as `const SCHEMA`. `init_report.rs` contains `INIT_REPORT_SCHEMA`. `contains` matched the shorter name *inside* the longer one, so binding the preflight document to the init writer was green.

Matching is now on identifier boundaries. A substring knows nothing about identifiers — the same defect that made an earlier search for `BINDING_ROWS`' identities look empty, because they are held in constants rather than written as literals.

`published_symbols` also missed `pub(crate)` and `pub(super)` on the first pass, so the three evidence-schema constants published nothing a writer could be tied to.

Neither of those came from reading the code. Both came from running a mutation that the property itself implied.

## Mutations

| | mutation | result |
|---|---|---|
| M16 | row pointed at a file writing a **different** document | red |
| M17 | split row whose writer touches none of the namer's symbols | red |
| M18 | **control**: unchanged tree | green |

## Limit, recorded in the file

The tie is a token check, not dataflow. A writer that mentions the right constant while serializing something else would pass. Closing that means following the value into the `schema` field, which nothing here does.

## What this does not close

`#2555` is still open and is a different class: a **new command file** in a crate already scanned. Const-in-dependencies catches a new crate; this catches a mis-bound row; #2555 catches `evidence lint` and `evidence diff`.

### DIFF
--- a/crates/assay-cli/tests/cli_json_identities.rs
+++ b/crates/assay-cli/tests/cli_json_identities.rs
@@ -725,6 +725,7 @@ fn describe_binds_only_recorded_documents() {
 #[test]
 fn documents_are_bound_to_a_writer() {
     for (identity, (writer, namer)) in document_rows() {
+        assert_writer_is_about_this_identity(&identity, &writer, &namer);
         let naming = read(&namer);
         let names_it = naming.contains(&format!("\"{identity}\""))
             || constant_names_for(&identity)
@@ -825,4 +826,113 @@ fn dependency_crates_match_the_manifest() {
          either scanned in full (SCANNED) or scanned for pub const identities (DEPENDENCY_CRATES); \
          a dependency in neither is a crate whose published identities nothing classifies"
     );
+}
+
+/// The writer file must be about *this* identity, not merely about writing.
+///
+/// `documents_are_bound_to_a_writer` asks whether the named file contains any writer idiom. That
+/// leaves `assay.foo.v0` bindable to a file that writes `assay.bar.v0` and happens to `println!`.
+/// A review measured the gap: six rows do not contain their identity in the writer at all, and all
+/// six are the write/name splits.
+///
+/// Two cases, no dataflow:
+///
+/// - the writer also names the identity (`namer` is `-`): it must contain the literal, or a
+///   constant whose value is the literal. Almost every row.
+/// - the naming file is separate: the writer must mention a symbol that file publishes. A writer
+///   that touches none of the namer's types is not the writer of its document.
+fn assert_writer_is_about_this_identity(identity: &str, writer: &str, namer: &str) {
+    let writing = read(writer);
+    if writer == namer {
+        let named = writing.contains(&format!("\"{identity}\""))
+            || constant_names_for(identity)
+                .iter()
+                .any(|name| mentions_token(&writing, name));
+        assert!(
+            named,
+            "{INVENTORY}: {identity} names {writer} as both writer and namer, and that file does \
+             not mention the identity. Either it is not the writer, or the naming column should \
+             point at the file that sets the schema"
+        );
+        return;
+    }
+    let published = published_symbols(namer);
+    assert!(
+        !published.is_empty(),
+        "{INVENTORY}: {identity} names {namer} as its naming file and that file publishes no \
+         symbols, so nothing can tie the writer to it"
+    );
+    let touched: Vec<&String> = published
+        .iter()
+        .filter(|symbol| mentions_token(&writing, symbol))
+        .collect();
+    assert!(
+        !touched.is_empty(),
+        "{INVENTORY}: {identity} is recorded as written by {writer} and named in {namer}, and the \
+         writer mentions none of that file's public symbols ({published:?}). A file that writes \
+         some other document and happens to call a writer would look identical"
+    );
+}
+
+/// Whether `src` uses `token` as a whole identifier rather than as a substring of a longer one.
+///
+/// `contains` was wrong here and a mutation caught it: `mcp/preflight.rs` declares its identity as
+/// `const SCHEMA`, and `init_report.rs` contains `INIT_REPORT_SCHEMA`, so binding the preflight
+/// document to the init writer passed. A substring match knows nothing about identifiers — the same
+/// defect that made a `git grep` for a constant look empty earlier in this file's history.
+fn mentions_token(src: &str, token: &str) -> bool {
+    let mut from = 0usize;
+    while let Some(rel) = src[from..].find(token) {
+        let start = from + rel;
+        let end = start + token.len();
+        let before_ok = start == 0
+            || !src.as_bytes()[start - 1].is_ascii_alphanumeric()
+                && src.as_bytes()[start - 1] != b'_';
+        let after_ok = end >= src.len()
+            || !src.as_bytes()[end].is_ascii_alphanumeric() && src.as_bytes()[end] != b'_';
+        if before_ok && after_ok {
+            return true;
+        }
+        from = start + token.len();
+    }
+    false
+}
+
+/// `pub` item names declared in a file: consts, structs, enums, type aliases and functions.
+fn published_symbols(path: &str) -> BTreeSet<String> {
+    let src = read(path);
+    let mut names = BTreeSet::new();
+    for line in src.lines() {
+        let line = code_before_comment(line).trim();
+        // `pub(crate)` and `pub(super)` publish to the writer just as well as bare `pub`; the
+        // schema-report constants are `pub(crate)` and this parser saw none of them at first.
+        let Some(rest) = line.strip_prefix("pub") else {
+            continue;
+        };
+        let rest = match rest.strip_prefix('(') {
+            Some(restricted) => match restricted.split_once(')') {
+                Some((_, after)) => after,
+                None => continue,
+            },
+            None => rest,
+        };
+        let Some(rest) = rest.strip_prefix(' ') else {
+            continue;
+        };
+        let rest = rest
+            .strip_prefix("const ")
+            .or_else(|| rest.strip_prefix("struct "))
+            .or_else(|| rest.strip_prefix("enum "))
+            .or_else(|| rest.strip_prefix("type "))
+            .or_else(|| rest.strip_prefix("fn "));
+        let Some(rest) = rest else { continue };
+        let name: String = rest
+            .chars()
+            .take_while(|c| c.is_alphanumeric() || *c == '_')
+            .collect();
+        if !name.is_empty() {
+            names.insert(name);
+        }
+    }
+    names
 }
--- a/docs/architecture/CLI-JSON-IDENTITIES.md
+++ b/docs/architecture/CLI-JSON-IDENTITIES.md
@@ -31,7 +31,10 @@ A top-level JSON document a command writes to stdout or to a caller-named path, 
 string as its `schema`.
 
 Each row is `identity | writing file | naming file`, where `-` means the writing file also names the
-identity. The test asserts the writing file calls a writer and the naming file carries the identity.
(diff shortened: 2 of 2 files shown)


Rul1an (commented):
Review of V on `f06c2e3a101e10e5236be2497d9d15f659b89518`. Z / AA / AB only. M16’s identifier-boundary fix is real and I am not re-arguing it.

## Z — cheaper than dataflow, and the recorded limit is already live

The recorded limit is: a writer that mentions the right constant while serializing something else would pass. That is true. It understates what this head already accepts.

The `-` arm asks whether the writer mentions *this* identity (literal or a const whose value is the literal). The split arm asks whether the writer mentions *any* symbol `published_symbols` extracts from the namer. Those are not the same check. The second is what shipped, and on this tree it is already the weaker thing:

- `assay.doctor_report.v0` is named by `DOCTOR_REPORT_SCHEMA` in `diagnostics/report.rs`. The namer column points at `probes.rs`. `published_symbols(probes.rs)` is `{probe_system}`. The writer calls `diagnostics::probe_system()` and never mentions the const. Green because the writer calls a probe function, not because it is about this identity.
- `assay.supply_chain_conformance.v0`’s writer mentions `verify_supply_chain` only in a `//!` comment. `mentions_token` reads the raw file. The command body calls `descriptor::build_carrier`. Green on a comment.
- `schema/write.rs` mentions `SchemaListReport` / `SchemaShowReport` / `SchemaValidationReport` and never `SCHEMA_LIST_REPORT` / `SCHEMA_SHOW_REPORT` / `SCHEMA_VALIDATION_REPORT`. The identity consts are set in other files. Green on the payload types.
- `inventory.rs`, `project_otel.rs`, `observation_health.rs`, `trust_basis.rs` likewise mention a constructor or a sibling type from the namer (`to_inventory_carrier_v0`, `project` / `project_tool_decision_truth`, `ObservationHealth`, `TrustBasisDiffReport`) and not the identity-holding const.

If the split arm used the same token check the `-` arm already uses — `constant_names_for(identity)` mentioned in the writer — every current split row would go red. That is why `published_symbols` exists, and it is why the recorded limit is the polite version of the actual one: *any export of the namer, including a probe function, a sibling type, or a comment*.

Cheaper properties that are not “follow the value into `schema`”:

1. Run `mentions_token` over `code_before_comment` (and skip `//!` / `/* */`). The function already exists. The supply-chain row is the positive control.
2. For `-` rows, require the identity token and a writer idiom in the *same function*. `skip_balanced` is already in this file. That is the argument-to-the-writer-call check that does not need type inference: the function that writes is the function that names. A file that has both tokens in different functions fails. A file that mentions `INIT_REPORT_SCHEMA` and `println!`s something else in another function fails.
3. For split rows, do not invent a second, wider vocabulary. Either require `constant_names_for(identity)` in the writer (honest, and currently all-red), or add a fourth column like `unnamed-documents` already has (`identity | writer | namer | token`) and check *that* token. `published_symbols` over every `pub` item is the thing that turned “about this identity” into “touches this file’s export list”.

Following the value into the `schema` field is not the next step. Narrowing the token is.

Sibling, same class as M16, still live: `documents_are_bound_to_a_writer` still binds the namer with `naming.contains(name)`, not `mentions_token`. `SCHEMA` inside `INIT_REPORT_SCHEMA` still matches on that half of the same test.

## AA — no live false-red; the gap is real and one namer already has it

I read every current namer.

No split row is a re-export-only or macro-only file, so none goes red for the wrong reason today. `probes.rs`, `reports.rs`, `inventory_carrier.rs`, `projection.rs`, `health.rs`, and `trust_basis/types.rs` all have a prefix-parseable `pub` / `pub(crate)` / `pub(super)` item.

The instance is `crates/assay-registry/src/supply_chain.rs`: `pub use types::*;` publishes the types the CLI actually imports (`SupplyChainConformance`, `VerifyInput`, …) and `published_symbols` sees none of them. The row stays green because the same file also has `pub const SCHEMA` and `pub fn verify_supply_chain`. Masked, not absent.

A legitimate future namer that is only `pub use reports::SCHEMA_LIST_REPORT;` — `schema.rs` is already that shape — would fail `published.is_empty()` for the wrong reason. `pub async fn`, `pub trait`, `pub static`, and `pub(in path)` are the same hole, none of them a current namer’s only export.

If the split arm narrows to identity-valued consts (Z.3), this parser stops being load-bearing and AA mostly evaporates. If it stays “any published symbol”, the next split pointed at a re-export file is a false red.

## AB — the source red, without moving the identity set

M16 and M17 mutate the inventory. The identity set is the collected `"assay.*.vN"` literals. These do not remove one.

**Split, this head, one edit.** Delete or reword the `//!` line in `supply_chain_conformance.rs` so the identifier `verify_supply_chain` is gone. The literal stays in `supply_chain.rs`. `every_production_identity_is_classified` stays green. V must go red. If it stays green, another published symbol is matching and that is also a finding. Either way it is a source mutation of this property.

**`-` row, extract without updating the row.** Move `INIT_REPORT_SCHEMA` and its literal into a new `init_report_schema.rs`. Leave the inventory as `-`. The identity still exists in production source. V must go red. That is M16 done to the writer instead of to the document, and it is the mutation the first `contains` version would have missed the same way (a sibling file holding `SCHEMA` vs `INIT_REPORT_SCHEMA`).

**Split, drop the tying tokens.** In `schema/write.rs`, erase the `Schema*Report` type names (generic `Serialize` is enough for `write_json`). Leave `reports.rs` and the row alone. Identity set unchanged. V must go red.

None of those edit `cli-documents`. None of them remove an identity from the tree. The reason they were hard to see from the chair is that the split arm is so wide that most source edits still leave *some* namer export in the writer. The supply-chain comment is the edit that is small enough to prove the gap.

M18 remaining green after a no-op does not cover this. The reds that demonstrate V have to include one of the three above, or V has still only been shown to notice a wrong row, not a drifted writer.

Rul1an:
Exact head `f06c2e3a101e10e5236be2497d9d15f659b89518`. Measured `assert_writer_is_about_this_identity` / `mentions_token` / `published_symbols` against the ten split rows. I did not run the test binary. The reds below are that predicate — the same token walk the unnamed rows already use — pointed at the writer.

## Z. I am not seeing a cheaper property about the write-call argument. I am seeing two unrecorded cousins of the limit you recorded.

I looked at every JSON emit on a split writer. The argument is a local: `value`, `json_out`, `carrier`, `projection`, `health`, `report`. None of those is a namer const. One hop up, inventory and project-otel assign from a namer fn; supply-chain assigns from `build_carrier` (a local module); `write.rs`'s document emit is `write_json<T: Serialize>(value)` and type-erases all three reports. Closing via "the argument to the writer call" either reds the legitimate evidence trio or stays as weak as the file-level token check.

The recorded limit is load-bearing. Two things you did not record make it cheaper than a file-level token check already is.

**Comments count on the writer side and not on the namer side.** `published_symbols` runs `code_before_comment`. `mentions_token` searches raw source, `//!` included. The two sides of the tie are not the same language.

`assay.supply_chain_conformance.v0` is written by `supply_chain_conformance.rs`, named in `supply_chain.rs`. At this head the writer mentions exactly one published symbol, `verify_supply_chain`, and only here:

```
//! carrier by running the existing `assay_registry::supply_chain::verify_supply_chain` producer over a
```

The produce call is in `descriptor.rs` (`verify_supply_chain(VerifyInput {` around L314), which is not the named writer. Predicate at HEAD: touched `{verify_supply_chain}`. After rewriting that one doc line so the identifier is gone: touched `{}`. The identity string is still `pub const SCHEMA` in the namer.

**Short published names are the SCHEMA-inside-INIT_REPORT_SCHEMA defect at identifier equality.** Identifier boundaries stop `SCHEMA` matching `INIT_REPORT_SCHEMA`. They do not stop `new` matching `Sha256::new()`, or `project` matching the English word in a module doc.

- `health.rs` publishes `pub fn new` (L51, L179). `observation_health.rs` contains `Sha256::new()` at L49. Strip every other published symbol from the writer and the split row stays green on `new`. Identity `OBSERVATION_HEALTH_SCHEMA` still in the namer.
- `projection.rs` publishes `pub fn project`. `project_otel.rs` contains `project` in `//! \`assay project-otel\`` and in `refusing to project unverified evidence`. Delete `project_tool_decision_truth` from that file and the TDT row stays green on `project` / `Projection` — the other identity's function, and a comment.

**Shared (writer, namer) is one bit for N identities.** `write.rs` + `reports.rs` serve three rows. The whole tie is line 1, the `use`. Delete `write_show_text` and keep the use: show stays green. That is the limit you recorded, instantiated.

The cheap tighten that is not dataflow: when several rows share a namer, require a symbol unique to *this* identity (the namer const, or the report type that is 1:1 with it), not any published symbol. Pointing `constant_names_for` at the writer would red all ten split rows at this head — none of them mention the const. So that is a different property, not a free one. The free one is only the shared-namer uniqueness check, and it still leaves the comment / `new` greens.

## AA. Looked. No false-red instance. One live `pub use`.

`crates/assay-registry/src/supply_chain.rs:23` is `pub use types::*;`. The parser sees nothing from it. `descriptor.rs` imports `ContainerRef`, `PinningInput`, `Policy`, `ProvenanceInput`, `SlsaLevel`, `Subject` through that re-export, and also `verify_supply_chain`, which *is* declared in the namer. No current split row is green only via a re-exported name, and none is red because the writer mentioned only re-exported names. I did not find a macro-generated type on a namer at this head.

The doctor namer is a cousin, not AA. `probes.rs` publishes one symbol, `probe_system`. `DOCTOR_REPORT_SCHEMA` lives in `diagnostics/report.rs`. The writer is tied through the probe function, not through the identity. Not a parse miss — the column points at a file that does not publish the identity.

## AB. The missing source red is the supply-chain comment.

M16 and M17 are not in the test file. They mutate the pin. Same shape as the M5 control: the demonstrated reds are doc edits.

Source mutation that moves this property and does not move the identity set:

**M19 (source).** In `crates/assay-cli/src/cli/commands/supply_chain_conformance.rs`, rewrite the `//!` line so it no longer contains the identifier `verify_supply_chain`. Do not touch `supply_chain.rs`, do not touch the pin, do not touch `descriptor.rs`. Identity set unchanged. Predicate goes from touched `{verify_supply_chain}` to touched `{}`. Expected: red on that row, "the writer mentions none of that file's public symbols".

A second, narrower one: delete the identifier `SchemaShowReport` from `write.rs` (the use *and* `write_show_text`). `SCHEMA_SHOW_REPORT` stays in `reports.rs`. Show row reds; list and validation stay green. Deleting the function and leaving the use is green — that is Z, not AB.

A source mutation that *should* move the property and does not: strip every `assay_runner_schema` symbol from `observation_health.rs` and leave `Sha256::new()`. Still green. That is not the missing red. That is Z-short, unpinned.

I have not run `cargo test`. A red in this comment is not the in-repo pin.

Non-claims: not a merge, not a quorum, not "close Z by following `schema`". #2555 is still the converse-writer hole.

PR #2559, opened 2026-08-19
feat(conformance): make the adequacy numbers re-derived rather than remembered

Rul1an:
Follow-up to #2556, which published measured adequacy numbers. This makes them re-derived rather than remembered.

## The problem

Every number in `INDEX.md` and `ERRATA.md` existed because someone ran the tool by hand one evening. Nothing re-derived them, and `ERRATA.md` is pinned to a corpus **digest**, which does not move when code moves, so no existing guard fires. They had the epistemics of the numbers they replaced.

## The mechanism

| | |
|---|---|
| `conformance/adequacy/results.json` | the measured truth |
| `measure_all.py` | writes it; `--only` carries unselected rows over byte-identically |
| `check_published_numbers.py` | documents are checked against it, never authored |
| `.github/workflows/adequacy-drift.yml` | relevance derived from each manifest's own declarations, not a path glob |
| `tool_pin` | every manifest names the instrument commit |

Prose binds to JSON through marked blocks covering digits **and** English words, plus a sweep requiring every `N of M` token to be registered or exempted with a reason. Superseded numbers (`4 of 4`, `30 of 30`, the contaminated `6 of 8`) are declared exemptions rather than silent ones.

## Four defects the build found

**1. A corpus was scoring with no control.** `mcp-jsonrpc-id` declared seventeen mutants and none, while `INDEX.md` says every manifest must declare one. `corpus-adequacy` enforced that only in `_run_process`; the module path never checked. Not a wrong number — six kills meant the harness reached the code — but the guarantee was absent: an all-survivors run could not have been told from a harness detecting nothing. Fixed upstream in one shared function with a parity test asserting the rule is stated exactly once. Seven of the tool's own fixtures went red immediately, which is the guard working; they were given controls rather than exemptions. Score unchanged at 6 of 10.

**2. Two scores were published about repositories we do not own, with no commit.** `rge-bench` and `observed-effect-v0`. `results.json` now derives which state was measured, the rows name it, and the checker refuses an out-of-tree corpus whose commit no published document carries. Freshness is not the requirement and cannot be — those repositories move without this diff seeing it. A score against `rge-bench@c51c5af18` stays permanently true of `rge-bench@c51c5af18`, and that was the missing half of the sentence.

**3. The last transcribed row is gone.** `privileged-mcp-action-v0` was copied from `ERRATA.md` prose. Re-measured: **6 of 25, 24.0%, control killed.** Every row is now derived by the same code path. Its quote control is retired with it, and the test says so out loud rather than shipping a synthetic control that proves nothing.

**4. Re-pinning left the prose behind.** `ERRATA.md` kept handing a reproducer the old tool commit. The prose-pin rule caught it, and only because the rule was ungated first: it had lived inside the transcription branch and went quiet the moment the row became measured — a guard that disappears exactly when its subject stops looking fragile.

## Two of my own guards failed their positive control before they worked

The subject recorder first derived from `repo_root` and called `rge-bench` **in-tree**, because that manifest declares no `repo_root` and the default landed inside this repository while the measured file sat three levels above. Recording an external corpus as internal is the silence the field exists to break. It now derives from the files actually measured.

And the out-of-tree rule was satisfied by the note explaining it — the guard read itself. The note no longer carries a literal commit.

## Beyond adequacy

`conformance/UNGUARDED-CLAIMS.md` asks the same question of the rest of the repository. Its sharpest finding is an existing drift check that compares a generated file against a heredoc which never reads the tree: a self-check that cannot fail when the code moves.

## Two decisions deliberately not taken

- **Required check.** The lane fits the repo's stated criterion — universal, quick when irrelevant, load-bearing when relevant — but it clones an external repository, which no current required check does. Making it required couples `main` to a third party's availability. The trigger for promotion is a pinned local copy verified against the pin, which is a cache rather than a second implementation.
- **The two sibling corpora joining the scheduled run.** Adding a `pins` block to their manifests brings them in with no workflow change; inventing commits for repositories I do not own is not mine to do.

## Summary by CodeRabbit

- **New Features**
  - Added automated adequacy-drift checks that selectively remeasure affected corpora and compare results with published claims.
  - Added validation for measurement coverage, provenance, pinned tool versions, control outcomes, and reported values.
  - Added canonical adequacy results for five corpora.

- **Documentation**
  - Clarified corpus pinning, measurement scope, adequacy limitations, checked figures, and disclosure requirements.
  - Documented published claims that are not yet automatically guarded.

- **Tests**
  - Added comprehensive checks for stale, incomplete, inconsistent, or altered adequacy data.

### DIFF
--- a/.github/workflows/adequacy-drift.yml
+++ b/.github/workflows/adequacy-drift.yml
@@ -235,7 +235,34 @@ jobs:
             echo "::error::the plan selected no corpora; a run that measures nothing must not reach the drift check as though it had"
             exit 1
           fi
-          echo "declared non-instrument sibling pins: ${SIBLING_PINS:-<none>}"
+          # Non-instrument siblings: corpora whose subject lives in another
+          # repository. They were out of scope while nothing pinned them, because
+          # cloning at a floating ref would measure whatever that repo happened to
+          # be. Pinned, they are ordinary: a score against a named commit stays
+          # permanently true of it, and re-deriving it here is what proves the
+          # published number is reproducible rather than remembered.
+          #
+          # Fetched by explicit sha and verified, exactly like the instrument. A
+          # clone that silently lands elsewhere would produce a real-looking number
+          # about the wrong subject, which is worse than not measuring.
+          for pin in ${SIBLING_PINS}; do
+            dir="${pin%%=*}"
+            spec="${pin#*=}"
+            repo="${spec%@*}"
+            sha="${spec#*@}"
+            target="${GITHUB_WORKSPACE}/../${dir}"
+            echo "::group::sibling ${dir} <- ${repo}@${sha}"
+            git init --quiet "${target}"
+            git -C "${target}" remote add origin "https://github.com/${repo}.git"
+            git -C "${target}" fetch --quiet --depth 1 origin "${sha}"
+            git -C "${target}" checkout --quiet --detach FETCH_HEAD
+            landed="$(git -C "${target}" rev-parse HEAD)"
+            if [[ "${landed}" != "${sha}" ]]; then
+              echo "::error::${dir} is at ${landed}, not the pinned ${sha}"
+              exit 1
+            fi
+            echo "::endgroup::"
+          done
 
           # One pass per instrument commit, so every row records the commit its manifest pins.
           for group in ${PIN_GROUPS}; do
--- a/.github/workflows/adequacy-drift.yml
+++ b/.github/workflows/adequacy-drift.yml
@@ -1,0 +1,371 @@
+# Keep the published mutation-adequacy numbers from rotting.
+#
+# `conformance/INDEX.md` and `conformance/privileged-mcp-action-v0/ERRATA.md` publish measured
+# adequacy numbers. They were produced by hand one evening and nothing re-derived them. ERRATA.md
+# pins its scope to a corpus DIGEST, which does not move when the implementation moves, so a rule
+# deleted from a declared implementation source changes what the corpus can transmit, changes no
+# digest, and trips no existing guard. That is a published claim nobody re-derives, which is the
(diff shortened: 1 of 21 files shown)


Rul1an:
Spot-checked the load-bearing claims before merging rather than reading the summary.

`results.json` carries `control=killed` and `provenance=measured` on **all five** corpora, and the two out-of-tree subjects are marked as such:

```
mcp-jsonrpc-id                  control=killed   6/17  60.0%  in_tree      measured
observed-effect-drift-consumer  control=killed  14/23  60.9%  out_of_tree  measured
privileged-mcp-action-v0        control=killed   6/31  24.0%  in_tree      measured
rfc8785-canonicalization        control=killed    0/0      —  in_tree      measured
rge-bench                       control=killed  51/57  94.4%  out_of_tree  measured
```

6/25 in the body and 6/31 in the row are the same measurement over different denominators (25 in-scope of 31 declared), and 24.0% agrees with the first.

Two findings here are the sharpest of the whole arc, and both are about a guard rather than a number.

**A corpus was scoring with no control**, because the requirement was enforced in `_run_process` and not on the module path. Six kills meant the harness reached the code, so the number was not wrong — but the *guarantee* was absent, and an all-survivors run could not have been told apart from a harness detecting nothing. A rule enforced on one route and not another is the failure mode of this entire week, and seven of the tool's own fixtures going red immediately is what a working guard looks like.

**"The out-of-tree rule was satisfied by the note explaining it — the guard read itself."** That is the self-satisfying guard caught by its own author, before review.

The freshness sentence is the one I would keep: a score against `rge-bench@c51c5af18` stays permanently true of `rge-bench@c51c5af18`. Naming the state measured is the achievable property; freshness against a repository we do not own is not, and pretending otherwise would have been the flattering version.

Merging.

PR #2560, opened 2026-08-20
feat(conformance): the published numbers are projected, not authored

Rul1an:
Follow-up to #2559, which landed the drift guard. This replaces it with projection, on an objection that rejected the previous instrument outright.

> "I would emit the score table from `results.json`, not detect drift of hand numbers. The failure is authorship. A checker on authored prose is the reason column again: the cell matches and the sentence next to it still says 6 of 25."

## What changed

`conformance/INDEX.md` and `.../ERRATA.md` are generated from templates. Narrative prose stays hand-written; every measured value is a token rendered from `results.json`, **including English word forms** — an author who cannot render "Five of twenty-seven" writes it by hand and every failure mode returns.

**The first render reproduced both documents byte for byte** against the hand-written originals. The projection did not have to reword anything to fit.

A row whose declared sources moved since its measurement renders *not re-measured at this revision* instead of a figure, so `last week's 5 of 27 as a present-tense fact` cannot appear.

## What is deliberately NOT deleted

**The sweep stays**, and changes subject from the rendered document to the template's authored prose. Regenerate-and-compare has one failure mode — someone edited the output — and is structurally blind to a hand-written figure that was never tokenised, because that renders byte-identically. The sweep is the only mechanism forcing a *new* number to be justified rather than merely consistent. Its control asserts the byte-comparison stays clean for a hand-added `12 of 13 in scope (92.3%)` while the sweep goes red.

I had instructed the builder to delete it. That was wrong and the review caught it.

## A judgement must not render like a measurement

The reader-side figure descends from a declared adjustment in `adjustments.json`. A judgement entering through a field shaped like a measurement and leaving as a rendered token would be **laundered** by this mechanism: the page would give it the tool's authority. So `five (judged)` and `18.5% (judged)` render marked while `six` and `24.0%` render bare, and a control asserts the two renderings differ so the marker cannot be silently emptied.

Tonight is why that matters. The adjustment's value stayed `1` while the reason it was `1` turned out to be wrong.

## The derivation it renders was wrong, and is fixed

`bad-107` carries `{"decision": "permit", "fail_closed": true}`, so it violates the closed vocabulary **and** the `fail_closed` biconditional. Either rule alone rejects it; dropping both accepts it. Built and measured:

```
baseline (holds both)   bad-107 -> invalid
drops both              bad-107 -> valid
```

The corpus transmits a **disjunction**, not a rule, and single-deletion mutation cannot see it because each rule masks the other. Had this shipped unfixed, projection would have rendered a wrong derivation forever, which is worse than a hand-written number someone might question.

## The honest measure of what this buys

**One derived cell per 107 words in `INDEX.md.in`, per 90 in `ERRATA.md.in`.** About 1%. The guard holds the cells and not the argument around them, and on these pages the argument is most of the text.

A reviewer named the precedent before this was built: `docs/generated/module-map.mermaid` sits in this repository's drift list, passes every commit, and guarantees nothing, because its generator is a heredoc that reads nothing from the tree. *"A `.md.in` template is a heredoc with holes in it."* Those four figures are themselves tokens — a hand-typed ratio in the section arguing against hand-typed numbers would rot, and would rot flatteringly, since prose is what gets added.

## The lane was watching the wrong files

`GLOBAL_TRIGGERS` listed the rendered documents and not the templates, the generator or the adjustments, so a changed template selected **no corpora at all** — verified by running the planner before adding them. And `declaration_sources` now joins relevance derivation: the profile text a denominator is audited against, the corpus manifest, the generator. A normative rule could be added to `v0.md`, the published denominator become wrong, and the planner never select the corpus. The denominator was the finding on this branch.

## Residue, named rather than implied

- The sweep knows two shapes. A claim written entirely in words — "the corpus isolates six" — can be added by hand and nothing notices.
- A token can name the wrong field. Byte-comparison confirms consistency, never correctness.
- `(judged)` is marked, not verified.
- The twenty-two rules are hand-numbered under a projected heading; nothing checks they agree.
- `@this` cannot tell whether those hundred words are the load-bearing sentences or the decorative ones.

## Summary by CodeRabbit

* **New Features**
  * Added generated conformance and errata documentation with automatically projected measurement figures.
  * Added safeguards for stale measurements, source changes, unprojected numbers, and judged adjustments.
  * Added reproducibility and provenance details for adequacy results.

* **Documentation**
  * Expanded guidance on conformance corpora, measurement limitations, validation, grading states, and regeneration.
  * Clarified the distinction between measured values, historical figures, and editorial judgments.

* **Bug Fixes**
  * Updated adequacy results to reflect current source revisions and dependencies.
  * Improved validation of published figures and generated documents.


Rul1an:
Independent non-building review, exact head `22a4cb49c00d4eee650d118783f7426aa1271213`.

**Verdict: BLOCKED**

A Codex subagent that authored neither this PR nor its governing specification/plan reviewed the exact head read-only. Findings, ranked:

1. **HIGH — adjustment collisions can replace measured truth while the guard remains green.** `publish_numbers.py` silently overwrites duplicate `(scope, name)` declarations and permits adjustment names such as `killed` to overwrite measured/derived fields. Disposable mutations changed the published result (`Four ... of twenty-seven`, and `99 ... killed`) while the checker still returned `rc=0`, `ok=true`. Reject duplicate adjustment keys and all collisions with measured/derived fields; pin both with mutation tests.
2. **HIGH — unknown measurement commits fail open.** Replacing a corpus `measured_at.commit` with forty `f` characters still returned `rc=0`, `findings=[]`, `ok=true`. A failed `git diff` currently becomes “not stale,” so an unavailable comparison renders current-looking numbers. Fail closed or render an explicit unavailable marker.
3. **HIGH — authoritative template prose contradicts the projected denominator.** `INDEX.md.in` says `8 + eighteen = 25` and elsewhere `8 + sixteen = 24` while the derived result is `6 killed + 19 survived = 25`. These word-only claims evade the numeric-token sweep and directly undermine the denominator-correction purpose.
4. **MEDIUM — `distinct_fields` counts render variants, not field identities.** `TOKEN.findall()` returns `(target, form)` tuples and `set(tokens)` counts the same field with different forms multiple times. Independent counts: INDEX reports 32 versus 30 distinct targets; ERRATA reports 22 versus 16.
5. **MEDIUM — malformed measurement input violates the documented JSON interface.** Malformed `results.json` produces `rc=1` plus traceback rather than a machine-readable finding. This fails closed as a process, but breaks `--json` and loses actionable file/row context.
6. **LOW — test/oracle hardening remains.** `skipTest()` inside the corpus loop can skip the whole test; one capitalization assertion is tautological; fixed historical-number negatives can become vacuous; temporary Git repos inherit hooks/signing/global templates.

Exact-head evidence:

- `python3 conformance/adequacy/check_published_numbers.py --json`: current canonical tree returns `rc=0`, `findings=[]`, `ok=true`;
- `python3 -m unittest -v conformance.tests.test_published_numbers_guard`: 43 passed;
- `python3 scripts/ci/test_adequacy_lane.py`: 26 passed;
- profile/generator lane trigger checks passed;
- `git diff --check origin/main..HEAD`: clean;
- all five manifests currently pin `corpus-adequacy/corpus-adequacy@b87ef706e70a2bcfe163514ec0360022e93fa175`.

Non-claims: the approximately 15-minute process corpus was not rerun locally; exact-head GitHub re-derivation and CI were green. Projection still cannot establish arbitrary sentence semantics or wrong-field tokens; those documented limits do not excuse the concrete contradictory prose above.

This comment is the review artifact because the reviewer cannot submit a separate GitHub review identity. Any new push invalidates this verdict and requires a fresh exact-head review.

Rul1an:
## Blocking finding: the documented double-drop omits a second code change

Exact-head review on `22a4cb49c00d4eee650d118783f7426aa1271213`.

I reproduced `bad-107` from a clean detached worktree and built the declared vocabulary mutant. Baseline is `invalid` with `decision_vocabulary`. Replacing only

```rust
check_vocab(dec, "decision", DECISION_VOCAB, violations);
```

with `let _ = DECISION_VOCAB;` does **not** produce the documented `valid` result. It reaches `unreachable!("decision vocabulary is closed")` in Stage 4 and aborts with exit 134.

The original scratch experiment does produce `valid`, but it makes two edits: it removes the vocabulary check **and** changes that downstream `unreachable!` arm to `ClaimCell::incomplete()`. Its captured output is internally consistent:

```text
baseline      bad-107 -> invalid
custom mutant bad-107 -> valid
```

The disjunction argument can remain as an explicitly judged construction: a coherent implementation without the closed-vocabulary invariant must also define downstream handling for an unknown decision. What must change is the claim that “deleting the vocabulary check ... and rebuilding” alone demonstrated it. The current wording hides the second accommodation, and the manifest's single mutant is killed by an abort rather than by the documented semantic transition.

Please update both `ERRATA.md.in` and `adjustments.json` to name both edits and distinguish this custom two-edit experiment from the tool's declared single mutant, then regenerate `ERRATA.md`. A fresh exact-head review is required after the push.

Advisory verdict: **BLOCKED** on this finding at the SHA above. I did not edit this PR branch.

Rul1an:
Coordination after #2565 merged:

- `main` is now `5f7f94daa2be34adde64bc32495def16d5460ea7`.
- This PR remains at `22a4cb49c00d4eee650d118783f7426aa1271213` on base `77753cf4...` and is now `BEHIND`.
- Both PRs change `conformance/INDEX.md`; #2565 also establishes the current partial-run/v1 inventory/ACM wording contract.

The owning builder should reconcile this branch with the new main, regenerate projected artifacts from the canonical inputs, and rerun the focused conformance/adequacy guards. Any resulting push is a new exact head and requires a fresh independent review; the prior review/blocker record cannot carry through a rebase or conflict resolution. Codex is not editing this writer-owned branch.

Rul1an:
## Disposition after current-main reconciliation

Closing this draft without merging or deleting its branch.

The branch was valuable research: it exposed under-declared denominators and established that published numbers must be projected from measured rows rather than authored. It is no longer a safe implementation base:

- head `22a4cb49c00d4eee650d118783f7426aa1271213` is behind current `main`;
- eight review threads remain unresolved;
- reproduced false-clean paths include row-identity overwrite, unresolved provenance continuing as current, stale hard-coded assertions, unavailable siblings aborting earlier comparisons, and metric naming that counts `(target, form)` pairs as fields;
- #2568 has already landed independently, while canonical inventory and fail-closed provenance are now tracked by #2569 and #2570.

The replacement implementation is scoped in #2572 and must start from current `main` after #2569/#2570. Reuse ideas only after re-verification; do not carry this branch's generated numbers as truth.

This closure is a scope correction, not a rejection of the measured finding. The branch remains available as provenance.

PR #2580, opened 2026-08-21
feat(conformance): separate hostile capture from trusted scoring

Rul1an:
Implements Rul1an/assay-tunnel-experiments#199 (slice of #191), from the measured design in [issue #199 comment 5368130745](https://github.com/Rul1an/assay-tunnel-experiments/issues/199#issuecomment-5368130745).

**Base** `bb6ef072d74630a6b71b5d79cdadbdf1451e03c2` · **Head** `c24c3d2c77d373ec0572e5306b218a0248fb3a41`

## The gap this closes

`score_candidate.py` held the canonical oracle in the same process, on the same filesystem, while it executed candidate code. The defence was ordering — snapshot expectations before the candidate runs — which covers tampering, not confidentiality, and only the tamper direction had a test (`test_candidate_cannot_rewrite_oracle_or_report_a_mutated_pack_hash`).

Measured through the exact helper `run_candidate` uses: a candidate inherits the scorer's working directory and reads the canonical `MANIFEST.json` at a relative path, with all 14 expected surfaces. Positive control — a sibling path that does not exist raised `FileNotFoundError`, so the read is real. `action.yml:57` puts that file on the same runner as the entrypoint.

## What changed

`capture_candidate.py` runs the candidate and records what it emitted as `assay.privileged_mcp_action.candidate_capture.v0`. It takes no manifest and no expectations. `score_candidate.py --capture` scores that artifact elsewhere. `--entrypoint` still does both in one process, calling the same capture builder and the same scorer.

The scoring host recomputes every pack, corpus and case binding from its own copy of the pack and compares the capture's copies rather than adopting them. A capture that does not bind is refused before any comparison: exit `2`, **no run record**, no `unproved` token.

One strict bounded loader now reads every hostile JSON input here. `published_rows` already refused symlinks, oversized files, non-UTF-8 bytes, duplicate object names and non-finite numbers, so `strict_json` calls it and adds the two bounds it lacked: nesting depth — scanned *before* `json.loads`, because `RecursionError` is not a `ValueError` and would reach the user as a traceback — and the JSON number domain. `validate_run_record` reads through the same loader.

`implementations.py` exposes `validate_image_reference`; `_validate_row` calls it. `report.v0` gains optional `implementation.id` and `implementation.image`, present only together.

## RED → GREEN

First RED was the pair from the issue: a 13-observation capture and a duplicate-to-15 capture must exit 2 and leave no record, with the 14-observation control at 14/14 exit 0. Both failed on the missing module, which is an existence red and proves nothing on its own; the discriminating red is the control going green while the cardinality cases stay at exit 2. That is now the state.

| | |
|---|---|
| `-k CandidateCapture` | **17 OK** |
| full `test_activation_kit.py` | **42 OK** (25 pre-existing + 17 new) |
| `test_implementations.py` | **36 OK** (32 pre-existing + 4 new) |
| `test_registry`, `test_run_all`, `test_published_rows`, `test_published_numbers_guard`, `test_published_numbers_provenance`, `test_completion_scope`, `test_corpus_adequacy_own_corpora` | OK |
| `python3 conformance/implementations.py`, `conformance/registry.py` | rc=0 |
| `py_compile`, `json.tool` ×3, `cargo fmt --all -- --check`, `git diff --check` | OK |
| pre-commit and pre-push hooks on `c24c3d2c7` | Passed (incl. clippy, linux compile gate) |

Byte-identity is the migration guarantee and the proof there is one comparison: `test_direct_and_split_paths_produce_byte_identical_run_records` scores one deterministic fake candidate through both paths and asserts identical bytes.

## Mutations — 13/13 killed, control green

Runner restores in a `finally` and asserts the file hash matches the pre-mutation digest; the control runs before the matrix and any red control aborts.

| # | Mutation | Named test that died |
|---|---|---|
| M1 | capture cardinality → `>= 1` | `test_capture_validator_owns_cardinality_and_ordering` |
| M2 | positional `case_id` → any string | `test_capture_validator_owns_cardinality_and_ordering` |
| M3 | positional input-digest binding off | `test_capture_that_does_not_bind_the_pack_is_refused` |
| M4 | pack binding comparison off | `test_capture_that_does_not_bind_the_pack_is_refused` |
| M5 | record copies the capture's digest | `test_scorer_recomputes_the_pack_digest_rather_than_copying_it` |
| M5b | drop `pack_sha256` from the binding tuple | `test_capture_that_does_not_bind_the_pack_is_refused` |
| M6 | capture byte ceiling → 1 GiB | `test_capture_over_the_byte_ceiling_is_refused_by_the_ceiling` |
| M7 | bypass the strict parser | `test_hostile_capture_bytes_fail_closed_without_a_traceback` |
| M8 | remove the depth scan | `test_standalone_run_record_validator_bounds_bytes_and_nesting` |
| M9 | remove the JSON number domain | `test_strict_loader_owns_the_json_number_domain` |
| M10 | unbounded error text | `test_capture_error_text_is_bounded` |
| M11 | neuter the image rule | 7 tests across both suites |
| M12 | put the oracle back on the capture host | `test_capture_phase_completes_with_the_oracle_absent_from_its_filesystem` |

### Two earlier matrices did not count, and why

**First run** timed out mid-M7 and left `strict_json.py` mutated. My cleanliness check used `git diff` — blind to an untracked new file — so the next run's control was already red and every "KILLED" was the same pre-existing breakage. Discarded.

**Second run** had a green control and reported 6 killed, 6 survived. Four survivors were defects in my tests, each confirmed by measurement rather than inspection:

- the oversized fixture was sized from `MAX_CAPTURE_BYTES` itself, so it grew with the mutant;
- the deep fixture was 400001 bytes, over the 64 KiB ceiling, so the ceiling refused it and the depth scan was never reached — and within that ceiling, recursion is unreachable (measured: depth 60000 parses fine, 200000 raises), so the scan's purpose is only observable on the run-record loader's 4 MiB ceiling;
- cardinality and ordering were masked through the CLI by the run-record case count and the pack-binding check, so only a direct call to `validate_capture` can see them;
- the number domain is redundant with the per-field type checks at capture level, so it is pinned on the shared loader instead.

**M5 was measured, not assumed.** With the binding guard disabled, the unmutated line published `sha256:b35e5e3d…` (recomputed) and the mutant published `sha256:dddddddd…` (the capture's lie). So the two expressions differ in general and are equal only at the program point the guard creates — an observability gap, not an untested rule. It was made observable by calling `score_capture` directly with a lying capture, and M5b was added as a non-equivalent mutation on the same binding. No survivor was removed from the matrix.

## Non-claims

- A capture is an observation artifact: not a verdict, attestation, sandbox proof, or independent reproduction.
- **A hostile capture host can emit a perfect capture without running anything.** This does not prevent that and does not try to. It removes the oracle from the host where candidate code runs. Authenticating a capture is a different property.
- Separation is not sandboxing; `bounded_process.py` bounds I/O and the initial POSIX process group and says so in its own comments.
- A declared image digest addresses bytes. It does not authenticate the publisher and does not establish that the addressed image is the one that ran. `image: null` is a disclosure that none was declared.
- The scoring host recomputes pack, corpus and case bindings. Implementation identity has no independent source on that host and is carried from the capture as a shape-validated declaration.
- `test_capture_phase_completes_with_the_oracle_absent_from_its_filesystem` shows the capture phase *can* run with no oracle present. It does not show any deployment separates them; that is operational.

## Deliberately out of scope

No `.github/**` (the composite action is pinned by full commit for third parties and needs no change), no container pull or run, no signing, no matrix, badge, score or tier, no v1 corpus, no network. Candidate `cwd` is not pinned — that would break the documented `./target/release/my-verifier` entrypoint form and needs its own argv[0] resolution rule.

### Findings for separate issues, not fixed here

- `published_rows.parse_json_object` lets `RecursionError` escape on deeply nested input (measured). Reached through this corpus it is now unreachable because `strict_json` scans depth first, but the underlying helper is shared and out of this allowlist.
- The authorship axis has two vocabularies: `implementations.py` uses `human`/`agent-assisted`/`agent-generated`; `CONFORMANCE-PROTOCOL.md` uses the `Authored-By:`/`Assisted-By:`/`Generated-By:` trailers.
- The case count `14` is stated in four enforcing places. This branch adds none — `capture_format` imports `EXPECTED_CASE_COUNT` — but the existing four remain.

## Writer

I authored the #199 design and this change, so I am the builder and do not count toward review quorum. Draft; not ready, not merged. Final head to be dispatched for independent review by the maintainer.

## Summary by CodeRabbit

- **New Features**
  - Added split-phase candidate capture and scoring workflows.
  - Added capture validation with pack, source, corpus, case, and input integrity checks.
  - Added optional implementation identifiers and container image references to reports.
  - Added clearer scoring modes and capture limitations documentation.

- **Bug Fixes**
  - Standardized image-reference validation.
  - Strengthened JSON handling with nesting-depth, encoding, size, and safe-number limits.
  - Improved rejection of invalid or unbound captures before scoring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- `capture_candidate.py` records a candidate's observations over the opaque cases and
+  writes `assay.privileged_mcp_action.candidate_capture.v0`. It takes no manifest and no
+  expectations, so a candidate that escapes its process bounds finds no answers on that
+  host. `score_candidate.py --capture` scores that artifact on a host that never ran the
+  candidate; `--entrypoint` still captures and scores in one process, and both paths call
+  the same capture builder and the same scorer, so a given candidate yields byte-identical
+  run records either way. A capture that does not bind the pack is refused before any
+  comparison, exit `2`, and no run record is written. `report.v0` gains optional
+  `implementation.id` and `implementation.image`, present only together, so an existing v0
+  record stays valid. A capture is an observation artifact, not a verdict, attestation,
+  sandbox proof, or independent reproduction; a capture host can fabricate observations, so
+  separating the phases removes the oracle from the hostile host without authenticating what
+  that host produces (Rul1an/assay-tunnel-experiments#199).
 - `conformance/implementations.json` is a static, fail-closed registry of candidate
   images addressed as `name@sha256:<64 hex>`. Authorship is a typed object (`human`,
   or agent kinds with model and prompt strategy). Required CI calls the same stdlib
@@ -31,6 +44,14 @@ All notable changes to this project will be documented in this file.
   revision (#2482).
 
 ### Changed
+- One strict bounded loader now reads every hostile JSON input in the conformance corpus.
+  `strict_json.load_strict_object` calls the existing regular-file reader and JSON parser and
+  adds the two bounds they did not carry: nesting depth, scanned before `json.loads` sees the
+  text, and the JSON number domain. `validate_run_record.py` reads through it rather than
+  keeping its own reader, which changes one diagnostic's wording and no exit status.
+  `conformance/implementations.py` exposes `validate_image_reference`, and `_validate_row`
+  calls it instead of matching the image pattern a second time.
+
 - `assay_check_sequence` answers the sequence-rule language by calling
   `assay_core::sequence_eval` and mapping the record into its published JSON.
   The MCP copy of the rule is gone (#2227, #2228).
--- a/conformance/implementations.py
+++ b/conformance/implementations.py
@@ -68,6 +68,26 @@ def _reject_unknown_fields(obj: dict, allowed: tuple[str, ...], *, what: str) ->
         )
 
 
+def validate_image_reference(value: object) -> str:
+    """Refuse anything that is not an exact `name@sha256:<64 hex>` reference.
+
+    Public because the conformance capture format binds a run to the same image
+    reference this registry stores, and two spellings of that rule are two
+    answers to what a capture is bound to. `_validate_row` calls this rather
+    than matching the pattern again.
+
+    A digest addresses bytes. It does not authenticate the publisher and it does
+    not establish that the addressed image is the one that ran.
+    """
+    if not isinstance(value, str) or not value:
+        raise ImplementationRegistryError("image must be a non-empty string")
+    if not IMAGE_RE.fullmatch(value):
+        raise ImplementationRegistryError(
+            "image must be name@sha256:<64 hex digest>, not a tag"
+        )
+    return value
+
+
 def _require_text(value: object, field: str, ident: str) -> str:
     if not isinstance(value, str) or not value:
         raise ImplementationRegistryError("%s: %s must be a non-empty string" % (ident, field))
@@ -107,11 +127,10 @@ def _validate_row(row: object, seen: set[str]) -> dict:
     suite = _require_text(row["suite"], "suite", ident)
     if suite not in ALLOWED_SUITES:
         raise ImplementationRegistryError("%s: unknown suite %r" % (ident, suite))
-    image = _require_text(row["image"], "image", ident)
-    if not IMAGE_RE.fullmatch(image):
-        raise ImplementationRegistryError(
-            "%s: image must be name@sha256:<64 hex digest>, not a tag" % ident
-        )
+    try:
+        validate_image_reference(row["image"])
+    except ImplementationRegistryError as exc:
+        raise ImplementationRegistryError("%s: %s" % (ident, exc)) from exc
     source = _require_text(row["source"], "source", ident)
     if not SOURCE_RE.fullmatch(source):
         raise ImplementationRegistryError(
--- a/conformance/privileged-mcp-action-v0/CONFORMANCE-PROTOCOL.md
+++ b/conformance/privileged-mcp-action-v0/CONFORMANCE-PROTOCOL.md
@@ -14,8 +14,7 @@ verifier logic, interpret the profile, or establish that an implementation is in
 3. Read `spec.md` and `descriptor.json`.
 4. Implement a command that consumes one bundle path and emits the profile report.
 5. Freeze the implementation commit and record all materials consulted.
-6. Run the scorer. It snapshots the canonical oracle into scorer-private memory for tamper resistance,
-   executes every opaque case without passing that oracle to the candidate, and only then compares.
+6. Run the scorer, in one phase or two. See "Scoring" below for which to pick.
 7. Publish the implementation source, run record, and completed implementation report.
 
 Reading Assay's verifier, `gen_vectors.py`, the canonical `MANIFEST.json`, or a prior scored report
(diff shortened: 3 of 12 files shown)


Rul1an:
## Independent exact-head review — READY

**Head reviewed:** `c24c3d2c77d373ec0572e5306b218a0248fb3a41`  
**Base:** `bb6ef072d74630a6b71b5d79cdadbdf1451e03c2`  
**Independence:** I did not author [issue #199](https://github.com/Rul1an/assay-tunnel-experiments/issues/199), its design comment, or this implementation. Read-only review; no edits, no merge, no ready-state change.

This comment is bound to that exact SHA. A later push invalidates it.

### Verdict

**READY** for the #199 slice. No High or Medium findings on this head.

### Findings

_None at High or Medium._

### Notes (non-blocking)

1. **Candidate cwd is still inherited.** `capture_candidate.py` does not load the oracle via import or argv (no `--manifest`; after `import capture_candidate`, `score_candidate` / `load_expectations` are absent from `sys.modules`). The oracle-absence probe runs capture with `cwd` set to a directory that cannot open `conformance/privileged-mcp-action-v0/MANIFEST.json` by the composite-action relative path, with combined mode as the positive control. Pinning candidate cwd is out of scope on this branch and is not treated as a missed DoD item.
2. **Non-zero candidate exit is recorded, not compared.** Protocol: exit status MAY be non-zero; comparison is `bundle_integrity` / `verdict` / `claims` only. Candidate stdout is parsed as a process report (`json.loads`, `RecursionError` mapped to `CandidateError`). Hostile *files* (capture + run record) use the one `strict_json` loader.
3. **CI on this SHA** (inspected, not used as the review): `activation-kit` success ([run 32474912735](https://github.com/Rul1an/assay/actions/runs/32474912735/job/96749145832)); Build+Test ubuntu/macos success; Clippy/Rustdoc/lint/claims-boundary success. **Build+Test windows-latest was still in progress** at review time — absence of that result is not treated as green. CodeRabbit skipped the draft and is not a review.

### Checklist on `c24c3d2`

| Check | Result |
|---|---|
| Capture host cannot load oracle/manifest/expectations via import or argv | Pass |
| Hostile JSON: one bounded strict loader (cap, depth-before-parse, duplicate, non-finite, no float, safe-int); no traceback | Pass (`published_rows` + `strict_json`; depth scan before `json.loads`) |
| 13 / 15 / duplicate / reorder / binding errors → exit 2, no run record | Pass |
| Scorer recomputes pack/corpus/case bindings; does not adopt capture identity as authority | Pass (`score_capture` writes local `pack_digest`; implementation id/image are shape-validated declarations only) |
| Direct and split run-record bytes identical | Pass (`test_direct_and_split_paths_produce_byte_identical_run_records`) |
| report.v0 compat; id and image together or absent | Pass |
| Image rule has one owner | Pass (`validate_image_reference`; `_validate_row` calls it; schema uses `IMAGE_PATTERN`) |
| Non-zero exit is not the comparison surface | Pass (protocol + `exit_code` recorded on match/mismatch only) |
| M5 / M5b observable without a shared predicate | Pass (M5: direct `score_capture` with a lying `pack_sha256`; M5b: CLI `require_capture_binds_pack`) |
| 13/13 mutation tests that would die are present and green | Pass (named tests ran; matrix not re-applied to the tree) |
| Public strings / non-claims | Pass (CHANGELOG, protocol, `CAPTURE_NON_CLAIMS`; no `unproved` token; no authenticity/sandbox/certification claim) |

### Local measurement (this review)

Worktree `/tmp/assay-2580-c24c3d2` detached at `c24c3d2c77d373ec0572e5306b218a0248fb3a41`:

- `CandidateCapture`: **17 OK**
- `test_activation_kit.py`: **42 OK**
- `test_implementations.py`: **36 OK**

Builder self-review does not count. This record is the non-building review on this head.

PR #2584, opened 2026-08-21
feat(conformance): share deterministic artifact I/O (#209)

Rul1an:
## Summary

- add one shared deterministic JSON byte renderer and atomic artifact writer
- bind capture parsing and digesting to one bounded regular-file read
- route the six shipped pretty-JSON sites and all three capture/run-record output paths through the shared rules
- preserve existing schemas, workflows, byte limits and public claims

Closes private programme issue `assay-tunnel-experiments#209` after merge.

## Contract

- sorted keys, two-space indentation, UTF-8, exactly one trailing LF
- capture digest addresses the exact bytes parsed and validated
- capture ceiling remains 64 KiB; run-record ceiling remains 4 MiB
- atomic replacement never exposes a partial destination and cleans temporary files on failure

## RED -> GREEN

Before production changes, the activation-kit suite ran 52 tests with six intended failures:

- missing shared renderer/writer
- six inline pretty-JSON implementations
- direct capture/scorer `write_text`
- missing one-read capture digest API

After implementation:

```text
Python 3.13.8: 131 tests, OK
Python 3.14.3: 131 tests, OK
pre-commit (changed files): pass
python -m py_compile: pass
git diff --check: pass
```

## Final-head verification

After merging current `main`, head `7ba3f7f91854563041e235240beca01e3ceab147` passed:

- activation kit: 55/55
- pma-v0 row contract: 2/2
- generic implementation registry: 36/36
- final-component symlink smoke: destination symlink replaced rather than followed; target bytes unchanged; output mode `0600`
- clean worktree and `git diff --check`

## Mutation evidence

Re-run in a detached worktree at `7ba3f7f91854563041e235240beca01e3ceab147`:

| Mutation | Result |
|---|---|
| M0 unmodified control | green |
| disable key sorting | red |
| remove trailing LF | red |
| digest a second path read | red |
| bypass semantic capture validation | red |
| bypass shared writer in capture producer | red |
| leave temp file after failed replace | red |
| parse a second path read while digesting the first | red |

## Non-claims

- atomic replacement is not durable storage; the parent directory is not fsynced
- the caller must keep the parent path stable; concurrent symlink-parent repointing can fail and leave a mode-0600 sibling temp file
- a digest addresses bytes but does not authenticate their producer
- deterministic pretty JSON is not JCS or a general canonical-JSON format
- no schema field, workflow, release, score, status or public projection changes here

### DIFF
--- a/conformance/privileged-mcp-action-v0/scripts/artifact_io.py
+++ b/conformance/privileged-mcp-action-v0/scripts/artifact_io.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+"""Shared byte rendering and atomic replacement for JSON artifacts."""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import os
+import tempfile
+from pathlib import Path
+from typing import Any
+
+ARTIFACT_TEMP_PREFIX = ".assay-artifact-"
+
+
+def content_sha256(data: bytes) -> str:
+    return "sha256:" + hashlib.sha256(data).hexdigest()
+
+
+def render_deterministic_json_bytes(value: Any) -> bytes:
+    """Render the programme's stable pretty-JSON form, including one trailing LF."""
+    return (json.dumps(value, indent=2, sort_keys=True) + "\n").encode("utf-8")
+
+
+def write_regular_file_atomically(path: Path, data: bytes) -> None:
+    """Replace a regular-file artifact without exposing a partial destination."""
+    destination = Path(path)
+    parent = destination.parent
+    parent.mkdir(parents=True, exist_ok=True)
+    fd, tmp_name = tempfile.mkstemp(prefix=ARTIFACT_TEMP_PREFIX, dir=str(parent))
+    tmp = Path(tmp_name)
+    try:
+        written = 0
+        while written < len(data):
+            count = os.write(fd, data[written:])
+            if count <= 0:
+                raise OSError("artifact write made no progress")
+            written += count
+        os.fsync(fd)
+        os.close(fd)
+        fd = -1
+        os.replace(tmp, destination)
+    finally:
+        if fd >= 0:
+            os.close(fd)
+        tmp.unlink(missing_ok=True)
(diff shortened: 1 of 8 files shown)


Rul1an:
## Independent final-head review

**Verdict: READY**, bound only to `7ba3f7f91854563041e235240beca01e3ceab147`. Any new push invalidates this review.

### Independence

The reviewer neither authored nor edited this change or its normative plan. Review was read-only in a clean disposable clone; no edits, commits, pushes, posts, or merges.

### Findings

No blockers or actionable findings.

### Evidence

- Base and merge-base: `d761b1e17a1201f50f603f37b1f93fde913caa67`; full 8-file diff reviewed.
- Shared renderer is used by all six existing pretty-JSON sites.
- Capture input is bounded to 64 KiB, read once, parsed and hashed from the same bytes, then semantically validated; the run-record limit remains 4 MiB.
- Base/head pack bytes are identical: SHA-256 `db872f50093619b3df65a779fcf55526c20234850fdf41ab53bb54949ad0a7eb` (33,285 bytes).
- Direct and split scoring produce byte-identical run records.
- Atomic writer preserves prior bytes on failure, completes short writes, rejects zero-progress writes, cleans normal temp residue, replaces rather than follows final-component symlinks, and emits mode `0600`.
- The documented parent-symlink race was reproduced as failure without output and possible `0600` residue. No durability, parent-path immutability, or producer-authentication claim is made.
- `--output` is an explicit same-caller capability; hostile capture/pack bytes do not select it. CodeQL alerts `842`, `845`, `846`, `847`, `848` were technically reviewed and their false-positive dismissal is defensible on this head.

### Mutations and tests

- One-read control green; parser second-read mutation red; digest second-read mutation red.
- `sort_keys=True` removal red; temp-cleanup removal red; restored final control green.
- Focused shared-I/O 6/6; focused one-read/digest 3/3; activation-kit + OCI 131/131 with no skips; PMA registration 2/2; generic registry 36/36.
- Exact-head hosted checks completed without failures, including `activation-kit`, required `CI`, `CodeQL`, and `lane-check/proof`.
- Local review used Python 3.14.3; hosted exact-head CI covers the pinned Python 3.13.8 path.

PR #2585, opened 2026-08-21
Pin trusted-main OCI capture against runner Python drift

Rul1an:
## Description

Trusted-main OCI capture is a `workflow_dispatch`-only job on hosted `ubuntu-24.04`. It checks out the dispatch SHA, pins Python 3.13.8, validates the candidate-release contract, downloads the attested pack, fail-closes if `source_digest` is not a 40-hex SHA, verifies the pack against the pack-release signer workflow, and writes `candidate_capture.v0` through the checked-in executor. `implementation_id` enters only via `env:` and quoted argv.

The structural contract has three allowlist layers that share one indent-key extractor. Upload `if:` is exact equality `if: success()`. Exactly two active TAG bindings come from the candidate-release resolver. Artifact name is one active line equal to `name: candidate-capture-v0`. A step-field name inside `env:` is an env key, not a parser stop.

Action pins use one exact `uses` helper: the 40-hex token ends before only whitespace or a comment. The same helper detects the upload step. `USES_SHA_RE` reads the full ref token. `on:` must be block-style with exactly `("workflow_dispatch",)` via the existing top-level mapping helper.

Required CI (`test_completion_scope.py::ProductCallsite`) imports and calls the existing `oci_candidate_workflow_problems()` against the committed workflow and a small dangerous-mutation table. The registry↔completion_scope 2-cycle remains the base inventory pin.

PR #2586 (exact-head READY and green on `44e80f227882c8a0a74c8756a485ef80b5bb2f42`) was fast-forwarded into this branch. That stacked slice adds the shared hard-run contract for the required inventory and activation-kit funnels, including the consumed-revision witness. No merge commit, rebase, or squash.

**HEAD:** `44e80f227882c8a0a74c8756a485ef80b5bb2f42`

Measurement provenance: worktree `/Users/roelschuurkes/wt-cursor-204-trusted-main-oci-capture`, SHA `44e80f227882c8a0a74c8756a485ef80b5bb2f42`, tree `ec10219eebba8bbc949c4558724f1882773f15bb`.

Prior review and CI on `2bce5f1ecf24021f3754b9976399b2ea768660c6`, `1592a66cccf81f714ca762be35ac8b3db2b647d9`, and older SHAs are obsolete for this head. Only this exact HEAD is under review.

This branch is frozen for that review. No further commits unless the human asks.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor

[...]

PR #2586, opened 2026-08-21
test(conformance): close hard-run guard funnel

Rul1an:
## Summary

- replace two independent DOTALL workflow-step regexes with one indentation-bounded parser and shared hard-funnel validator
- bind each protected target to an exact trusted three-step prefix: pinned checkout, pinned setup-python, then the exact hard-run step
- pin workflow, job, action-step, target-step, `with`, shell, and complete active-script contracts without shell-command denylists
- move only the existing CI setup-python and Conformance inventory steps so the inventory target immediately follows its trusted setup prefix

## Stack

- base branch: `cursor/204-trusted-main-oci-capture`
- base / merge-base: `2bce5f1ecf24021f3754b9976399b2ea768660c6`
- current head: `44e80f227882c8a0a74c8756a485ef80b5bb2f42`
- this closes the guard false-greens before PR #2585 receives a new final-head review

## RED -> GREEN

The trusted-prefix RED reproduction initially produced 20 expected failures across the two protected workflows. The completed matrix adds setup `with` mutations and proves 22 valid-YAML/actionlint-clean mutations are rejected:

- pre-target writers for `BASH_ENV`, `GITHUB_PATH`, and combined `GIT_DIR`/`PYTHONPATH`
- checkout extra `env`, extra `with`, and mutable action ref
- setup-python `if`, `continue-on-error`, extra `with`, and mutable action ref
- checkout/setup reordering

Canonical/no-op workflows and one harmless post-target insertion per workflow remain green. Removing the single `_assert_trusted_prefix` call makes all six representative pre-target writer mutations green, establishing ownership of the new failures.

The earlier mutation set remains covered: command commenting/replacement/softening/relocation; step and job `if`/`continue-on-error`; cross-job relocation; quoted direct keys; shell neutralization; wider indentation; explicit Bash; exact scripts; consumed-revision witnesses; and closed document-level keys/env sets.

## Verification on `44e80f227882c8a0a74c8756a485ef80b5bb2f42`

```text
focused trusted-prefix guards: 2/2
completion-scope suite: 40/40
registry suite: 42/42
activation-kit + OCI suites: 134/134
22 trusted-prefix mutations + 2 post-target controls + 2 canonical workflows: actionlint with shellcheck + Ruby Psych pass
reverse removal of trusted-prefix assertion: 6/6 representative writer mutations accepted
py_compile: pass
git diff --check: pass
pre-push hooks including actionlint, required-CI contracts, workspace clippy, and cross-target compile: pass
```

## Scope

Changed only:

- `.github/workflows/ci.yml` (reorder the existing pinned setup-python and Conformance inventory steps into the trusted prefix; existing revision witness retained)
- `.github/workflows/privileged-mcp-action-conformance.yml` (explicit `shell: bash` and revision witness from prior closure commits; no change in the trusted-prefix correction)
- `conformance/tests/test_completion_scope.py`
- `conformance/tests/test_registry.py`

No capture, scoring, record, publication, or candidate behavior changes. Post-target workflow siblings remain unconstrained by this rule.

## Review history

Independent reviews on earlier heads found whole-step cross-job relocation, job-level conditions/soft failure, quoted-key bypass, shell neutralization, stale/base checkout source, workflow-level shell state, and an under-scoped document-level denylist. Each earlier review is invalid for this head.

The latest review on `cbed3179284752e4fa9f71a4be7b93d0984322d5` proved that unchecked pre-target siblings could write `BASH_ENV`, `PATH`, `GIT_DIR`, or `PYTHONPATH` through runner command files. This head replaces implicit trust in those siblings with one shared exact ordered-prefix contract for both funnels.

## Non-claims

- pinned action refs constrain workflow inputs; they do not attest the runner platform or the implementation/behavior of GitHub-hosted actions
- this PR does not prove OCI capture correctness
- this PR does not add trusted scoring, a portable run record, or public projection
- GitHub CI and a fresh independent final-head review must both be green before landing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,24 @@ jobs:
           persist-credentials: false
           fetch-depth: 0
 
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        with:
+          python-version: "3.12"
+
+      - name: Conformance inventory
+        shell: bash
+        run: |
+          set -euo pipefail
+          test "$(git rev-parse HEAD)" = "$GITHUB_SHA"
+          python3 conformance/registry.py
+          python3 conformance/implementations.py
+          python3 -W error::ResourceWarning conformance/tests/test_implementations.py
+          python3 -W error::ResourceWarning conformance/tests/test_pma_v0_registration.py
+          python3 -W error::ResourceWarning conformance/tests/test_registry.py
+          python3 -W error::ResourceWarning conformance/tests/test_run_all.py
+          python3 -W error::ResourceWarning conformance/tests/test_completion_scope.py
+          python3 conformance/run_all.py --require-complete --completion-scope required
+
       - id: detect
         shell: bash
         env:
@@ -149,24 +167,6 @@ jobs:
           bash scripts/ci/check-assay-release-pin.sh --published
           bash scripts/ci/test-ci-hardening-b1.sh
           bash scripts/ci/test-structurizr-export-docker.sh
-
-      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
-        with:
-          python-version: "3.12"
-
-      - name: Conformance inventory
-        shell: bash
-        run: |
-          set -euo pipefail
-          test "$(git rev-parse HEAD)" = "$GITHUB_SHA"
-          python3 conformance/registry.py
-          python3 conformance/implementations.py
-          python3 -W error::ResourceWarning conformance/tests/test_implementations.py
-          python3 -W error::ResourceWarning conformance/tests/test_pma_v0_registration.py
-          python3 -W error::ResourceWarning conformance/tests/test_registry.py
-          python3 -W error::ResourceWarning conformance/tests/test_run_all.py
-          python3 -W error::ResourceWarning conformance/tests/test_completion_scope.py
-          python3 conformance/run_all.py --require-complete --completion-scope required
 
       # Stdlib-only. Required CI cannot go green if these tests are deleted;
       # adequacy-drift is not a required context. No conditional or soft-failure settings.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,6 +158,7 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
+          test "$(git rev-parse HEAD)" = "$GITHUB_SHA"
           python3 conformance/registry.py
           python3 conformance/implementations.py
           python3 -W error::ResourceWarning conformance/tests/test_implementations.py
--- a/.github/workflows/privileged-mcp-action-conformance.yml
+++ b/.github/workflows/privileged-mcp-action-conformance.yml
@@ -47,6 +47,7 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
+          test "$(git rev-parse HEAD)" = "$GITHUB_SHA"
           python3 -m unittest \
             conformance/privileged-mcp-action-v0/tests/test_activation_kit.py \
             conformance/privileged-mcp-action-v0/tests/test_oci_candidate_executor.py
--- a/.github/workflows/privileged-mcp-action-conformance.yml
+++ b/.github/workflows/privileged-mcp-action-conformance.yml
@@ -44,6 +44,7 @@ jobs:
           python-version: "3.13.8"
 
       - name: Run activation-kit contract tests
+        shell: bash
         run: |
           set -euo pipefail
           python3 -m unittest \
--- a/conformance/tests/test_completion_scope.py
+++ b/conformance/tests/test_completion_scope.py
@@ -16,6 +16,7 @@ import os
 import re
 import subprocess
 import sys
+import tempfile
 import unittest
 from pathlib import Path
 
@@ -189,6 +190,32 @@ def _direct_mapping(block: str) -> dict[str, str]:
     return entries
 
 
+def _assert_no_workflow_bash_env(text: str) -> None:
+    lines = text.splitlines(keepends=True)
+    env_indexes: list[int] = []
+    for index, raw in enumerate(lines):
+        line = raw.rstrip("\r\n")
+        if not line or line.startswith((" ", "#")):
+            continue
+        match = DIRECT_ENTRY_RE.fullmatch(line)
+        if match is None:
+            continue
+        key = next(match.group(name) for name in ("plain", "single", "double")
+                   if match.group(name) is not None)
+        if key != "env":
+            continue
+        if match.group("value").strip():
+            raise AssertionError("workflow env must use a direct mapping")
+        env_indexes.append(index)
+    if len(env_indexes) > 1:
+        raise AssertionError("workflow has duplicate top-level env mappings")
+    if not env_indexes:
+        return
+    env_block = _indentation_bounded_block(lines, env_indexes[0], 0)
+    if "BASH_ENV" in _direct_mapping(env_block):
+        raise AssertionError("workflow-level BASH_ENV can neutralize hard-run scripts")
+
+
 def _direct_scalar(raw: str) -> str:
     value = raw.strip()
     if re.fullmatch(r"[A-Za-z0-9_.-]+", value):
@@ -212,6 +239,7 @@ def assert_hard_run_command(
     if contract is None:
         raise AssertionError(f"no hard-run contract for {job_name}/{step_name}")
     allowed_job_keys, allowed_step_keys, expected_script = contract
+    _assert_no_workflow_bash_env(text)
 
     job = named_job(text, job_name)
     actual_job_keys = frozenset(_direct_mapping(job))
@@ -479,6 +507,31 @@ class ProductCallsite(unittest.TestCase):
                 )
                 self.assertEqual(completed.returncode, expected)
 
+    def test_bash_env_neutralizes_both_unchanged_hard_scripts(self):
+        with tempfile.TemporaryDirectory() as raw:
+            bash_env = Path(raw) / "bash-env"
+            bash_env.write_text(
+                "test() { return 0; }\npython3() { return 0; }\n",
+                encoding="utf-8",
+            )
+            env = {
+                **os.environ,
+                "BASH_ENV": str(bash_env),
+                "GITHUB_SHA": "0" * 40,
+            }
+            for label, script in (
+                ("inventory", INVENTORY_RUN_SCRIPT),
+                ("activation", ACTIVATION_KIT_RUN_SCRIPT),
+            ):
+                with self.subTest(label=label):
+                    completed = subprocess.run(
+                        ["bash", "-c", "\n".join(script)],
+                        cwd=REPO,
+                        env=env,
+                        check=False,
+                    )
+                    self.assertEqual(completed.returncode, 0)
+
     def test_direct_key_parser_ignores_nested_soft_failure_names(self):
         block = (
             "  scope:\n"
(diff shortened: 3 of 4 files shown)


Rul1an:
## Independent exact-head review: BLOCKED

Reviewed head: `24a48e5724405282fa08a5463b422f5ef6344702`
Base / merge-base: `2bce5f1ecf24021f3754b9976399b2ea768660c6`

Independent non-building reviewer found one P1 guard regression:

- `named_step()` searches the whole workflow and no longer binds `Conformance inventory` to the required `scope` job.
- Moving the complete valid step into the existing conditional `ebpf-smoke-ubuntu` job is accepted by the new helper.
- The prior job-scoped helper rejects that mutation.
- This is a false green: the conditional destination may skip while the required scope job succeeds without executing conformance.

Evidence on a disposable clean checkout:

- focused Python suites: 59 passed, no skips
- activation/OCI suites: 134 passed, no skips
- canonical no-op controls: green
- entire-step relocation mutation: **survived on this head**
- final remote SHA recheck: unchanged

Smallest coherent fix: make the shared helper select an expected job first, then an indentation-bounded named step inside that job, and add the whole-step relocation mutation.

Reviewer independence: the reviewer authored or edited neither this change nor a governing specification or plan. Review was read-only. Any push invalidates this verdict and requires a fresh exact-head review.

Rul1an:
## Independent re-review of `90f449391`: BLOCKED

Reviewed head: `90f4493910883b6825148cb8f6e43a8f6cf402eb`
Base / merge-base: `2bce5f1ecf24021f3754b9976399b2ea768660c6`

Prior whole-step relocation blocker is closed. The fresh exact-head review found one adjacent P1:

- the helper now binds each step to its owning job, but checks `if` / `continue-on-error` only on the step;
- job-level `if: ${{ false }}` on `activation-kit` survives;
- job-level `continue-on-error: true` on both `activation-kit` and `scope` survives;
- GitHub treats skipped required jobs as successful and job-level soft failure can let the workflow pass.

This is a missing guard, masked by the checked-in jobs currently lacking those keys. Smallest coherent fix: have the same hard-run assertion reject direct owning-job `if` and `continue-on-error` keys before checking the named step. Pin both job contexts with behavioral mutations.

Evidence: complete diff reread; no-op controls green; prior relocation mutation red; focused suites 60/60; activation/OCI 134/134; worktree and final remote SHA clean. Reviewer was independent and read-only.

Any new push invalidates this verdict and requires another fresh exact-head review.

Rul1an:
## Independent review of `a4bf8df64`: BLOCKED; redesign disposition

Reviewed head: `a4bf8df640ec3d68b86101c9e9d5077de0fd5cc1`
Base / merge-base: `2bce5f1ecf24021f3754b9976399b2ea768660c6`

Prior job relocation and unquoted job/step key findings are closed. Two final-funnel classes remain:

1. quoted direct YAML keys (`"if"`, `"continue-on-error"`) bypass the text matcher at job and step level;
2. the canonical command may stay present while additional shell semantics neutralize it, e.g. `set +o errexit; ...; true` or a local `python3() { :; }` override.

Both mutated workflows parse as valid YAML; shell controls exit 0. Focused suites (64) and activation/OCI suites (134) remained green, so these are real guard false-greens.

### Technical disposition

Do **not** extend the neutralizer blacklist. Replace it with a small allowlisted contract for these two required funnels:

- parse direct job/step mapping keys at their exact indentation, accepting ordinary and quoted YAML spellings but rejecting unrecognized direct-key syntax;
- pin the allowed direct keys for each owning job/step;
- require the complete active `run` script lines to equal the canonical hard script, not merely contain a command subsequence.

This is intentionally strict: changing either required funnel requires changing its independent guard. It is simpler and closes the class rather than these two examples.

Reviewer was independent and read-only; final remote SHA matched. Any push requires a fresh exact-head review.

Rul1an:
## Independent review of `de2968a09`: BLOCKED; final structural disposition

Reviewed head: `de2968a09f40da52af7401d820aaf0ae7eaad7af`
Base / merge-base: `2bce5f1ecf24021f3754b9976399b2ea768660c6`

The exact-script redesign closes quoted keys, duplicate keys/jobs/steps, relocation, in-script shell neutralization and unusual-key fail-closed behavior. Two mutable workflow semantics remain:

1. valid wider YAML indentation lets a complete job body plus direct `if: false` bypass the assumed `block_indent + 2` parser; the guard accepts an empty key subset;
2. shell templates can ignore `{0}` while leaving the exact script unchanged. Scope can mutate its `shell`; activation-kit can inherit a workflow-level custom default because it has no explicit shell.

Both are valid YAML/actionlint-clean and directly control whether the guarded command executes.

### Smallest coherent fix

- derive job child indentation from the first actual direct mapping entry and require consistent direct-child parsing;
- require the actual job/step key sets to equal, not merely be subsets of, their canonical sets;
- give both target steps explicit `shell: bash` and pin that exact scalar value;
- keep complete script equality.

The activation-kit workflow therefore needs one behavior-preserving `shell: bash` line. That is accepted scope expansion because it makes the execution contract explicit and prevents workflow-level defaults from changing this funnel.

Reviewer was independent/read-only; final remote SHA matched. Any push invalidates this review.

Rul1an:
## Independent review of `e603e3622`: BLOCKED; consumed-revision disposition

Reviewed head: `e603e3622f512c715e8350fe0190faa06ae6ab2c`
Base / merge-base: `2bce5f1ecf24021f3754b9976399b2ea768660c6`

All prior parser/script/shell findings are closed. One final mutable-workflow bypass remains: adding `ref: ${{ github.event.pull_request.base.sha }}` to either checkout leaves the hard step unchanged but makes it execute stale source. The base suites are green, so this can false-green.

### Chosen fix

Rather than duplicate each checkout action's configuration, bind the consumed revision at the final funnel:

```bash
test "$(git rev-parse HEAD)" = "$GITHUB_SHA"
```

Add this before tests in both exact hard scripts. The existing complete-script contract then pins it. A behavioral control must prove equal SHA exits zero and a mismatched SHA exits non-zero; removing the witness from either workflow must turn RED.

This is stronger and simpler than pinning checkout options: it permits safe checkout evolution while refusing base/head/subdirectory/absent source whenever the consumed repository revision is not the workflow revision.

Reviewer was independent/read-only; final remote SHA matched. New push requires fresh review.

Rul1an:
## Independent exact-head review: BLOCKED

**Reviewed head:** `f5419677b58d53e945e03e26872116afbbffdcc8`

A fresh non-building Codex subagent found one remaining false-green:

- Workflow-level `env.BASH_ENV` is outside the current job/step validation scope.
- A checked-in `BASH_ENV` can predefine `test()` and `python3()` before Bash runs the unchanged protected script.
- Both static guards accepted the mutation, and both complete scripts exited 0 with an intentionally mismatched all-zero `GITHUB_SHA`.
- Without `BASH_ENV`, the same mismatch is rejected.

Minimal reproducer:

```yaml
env:
  BASH_ENV: "$(printf 'test(){ return 0; }; python3(){ return 0; }' >/tmp/assay-bashenv; printf /tmp/assay-bashenv)"
```

Prior stale-checkout finding is otherwise closed: head match passes; head mismatch and base-checkout/event-head mismatch reject.

**Disposition:** blocking for this guard-only PR. Same writer is adding a fail-closed workflow-level `BASH_ENV` contract plus a biting mutation. Any push invalidates this review; a fresh exact-head independent review remains required.

**Independence:** reviewer authored or edited neither the change nor its governing specification or implementation plan.

Rul1an:
## Pre-review correction on `a9d7a08be4514ebb272ee1d57da6343dd57c78ad`

The `BASH_ENV` reproducer is closed, but a read-only Claude sidecar identified the broader rule and Codex independently reproduced it before requesting final review:

- top-level workflow `env.PATH` is accepted for both hard funnels;
- top-level `defaults.run.working-directory` is accepted for both hard funnels.

These are the same missing document-level trust boundary, not separate features. A literal `BASH_ENV` denylist would leave equivalent checked-in command-resolution/startup controls available.

**Disposition:** the sole writer is replacing the denylist with one closed document-level contract: exact workflow top-level keys plus exact workflow-env key names, while allowing legitimate existing env value changes. Job/step scripts and production YAML remain unchanged. Mutations will cover `BASH_ENV`, `PATH`/equivalent, and workflow `defaults`, plus reverse/no-op controls.

No final-review claim exists on this head. CI on this head will be obsolete after the correction push.

Rul1an:
## Independent exact-head review: BLOCKED

**Reviewed head:** `cbed3179284752e4fa9f71a4be7b93d0984322d5`

A fresh non-building Codex subagent reviewed the complete diff and found one remaining P1 false-green:

- The guard freezes the target step and direct workflow/job/step controls, but treats earlier sibling steps as opaque.
- A valid pre-target step can write `BASH_ENV`, `PATH`, `GIT_DIR`, or `PYTHONPATH` through `$GITHUB_ENV` / `$GITHUB_PATH`.
- Both guards and actionlint accepted these mutations.
- Runtime controls showed the canonical stale-base case rejects, while pre-target environment propagation makes both unchanged complete scripts exit 0 against base files.

Measured at the reviewed SHA:

```text
canonical stale-base control:              scope=1 activation-kit=1
prior-step BASH_ENV propagation:           scope=0 activation-kit=0
prior-step PATH propagation:               scope=0 activation-kit=0
prior-step GIT_DIR+PYTHONPATH propagation:  scope=0 activation-kit=0
```

The existing closures for document/job/target-step controls, revision mismatch, custom shell, quoted/duplicate keys, and legitimate `ASSAY_PUBLIC_MSRV` value changes were confirmed.

**Disposition:** bind the complete ordered pre-target execution surface. The smallest maintainable shape is an exact trusted prefix of pinned checkout, pinned setup-python, and the protected target; inserting or mutating a pre-target step must fail, while a post-target sibling remains harmless. Avoid blacklisting individual environment-file commands.

Any push invalidates this review. A fresh full exact-head review remains required.

**Independence:** reviewer authored or edited neither the change nor any governing specification or implementation plan.

Rul1an:
## Independent exact-head review: READY

**Reviewed head:** `44e80f227882c8a0a74c8756a485ef80b5bb2f42`  
**Base:** `2bce5f1ecf24021f3754b9976399b2ea768660c6`  
**Tree:** `ec10219eebba8bbc949c4558724f1882773f15bb`

**Findings:** none.

A fresh non-building Codex subagent reviewed the complete nine-commit diff and independently re-ran the prior bypass families. Confirmed fail-closed:

- unnamed or named pre-target steps, including `BASH_ENV`, `PATH`, `GIT_DIR` and `PYTHONPATH` propagation;
- checkout/setup reorder, one-character SHA changes, mutable refs and pin-only-in-comment;
- checkout/setup `with`, `env`, `if`, `continue-on-error` and direct-key drift;
- duplicate/inline/anchor/alias/dangerous quoted forms;
- workflow/job/target env, defaults, shell, soft-failure and relocation mutations;
- stale checkout, witness removal and script neutralization.

Semantic controls remained green: pin-comment-only changes, equivalent quoted keys/scalars, uniformly wider valid indentation, post-target sibling insertion and `ASSAY_PUBLIC_MSRV` value-only change.

Verification:

```text
Python 3.12 focused completion/registry: 82 tests, OK
Python 3.12 exact scope command family: 166 tests, OK
Python 3.13 activation/OCI:             134 tests, OK
Canonical required-scope run:           exit 0; required 1/1; required_complete yes; global complete no
actionlint 1.7.10:                      PASS
51 workflow YAML files:                 parsed
BASE...HEAD diff-check:                 PASS
```

The moved scope prefix is exactly `checkout -> setup-python -> Conformance inventory -> detect -> hardening -> later steps`; detect outputs and later behavior remain intact.

**Non-claims:** no attestation of hosted-runner integrity, pinned action implementation behavior, or external branch-protection configuration.

**Verdict: READY**

**Independence:** the reviewer authored or edited neither this change nor any governing specification or implementation plan. Review was read-only; remote refs matched at start/end and the disposable review tree remained clean.

PR #2592, opened 2026-08-22
feat(conformance): project reviewed run records

Rul1an:
## Summary
- Typed loader + digest-addressed store for opt-in `public-runs.json` rows. Record path is the SHA-256 hex only. Bytes are hashed before parse, then passed through existing `published_rows`, `validate_run_record`, and `load_implementations`.
- Deterministic `IMPLEMENTATIONS.md.in` → `IMPLEMENTATIONS.md` projector with `--check` that never writes. Publication identity and sort key are `(implementation_id, record_sha256)` so a correction can append without rewriting historical bytes.
- Registry, record, and index must agree on identity, image, source, commit, and `reproduction_mode`. The table projects those bound values plus typed `review_warnings`.
- Consumes hosted run-record artifact `9468514238` from run [`32546006900`](https://github.com/Rul1an/assay/actions/runs/32546006900) (8102 bytes, `sha256:9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27`), revalidated locally before the index row was added.
- Invoked from a sibling step on the existing required `scope` job. No new required context. Does not touch the #210 inventory-callsite pin files.

Implements [assay-tunnel-experiments#201](https://github.com/Rul1an/assay-tunnel-experiments/issues/201).

Exact head: `cef9d29a7a0fe5a08d3f06d3a5ea4125fce301b1`
Base: `a63a97c16615c181b65aaf6fdf0eed0fc5270ad0`

## RED → GREEN
1. First RED: `project_public_runs` missing (`ModuleNotFoundError`, 13 errors).
2. Preflight closures: hostile index on the production path; exact writer bytes + LF; typed `review_warnings`; Markdown escape; `reproduction_mode` bind; append-only digest tuple.
3. Audit-property RED on `13eb7623f`:
   - B1: rename the registry row via `_rebind_record`; `load_publication` must say `unknown`. Record-derived fallback: `ValueError not raised`.
   - H1: two valid digests must render two digest-links. `render_table` drop-last-row: `1 != 2`.
   - H2: public page omitted corpus-independence and declared-but-unattested binding.
   - M1: digest-named 4194305-byte record must fail with `exceeds` before parse. `path.read_bytes()`: JSON parse error instead.
4. GREEN on this head: focused `conformance/tests/test_public_runs.py` 24/24; `--check` exit 0; `git diff --check` clean; affected Python suite 188/188; B1/H1/M1 mutations red.

## Mutations
- Record byte flip with stale index digest: `--check` red; `IMPLEMENTATIONS.md` bytes unchanged.
- Duplicate `(implementation_id, record_sha256)` tuple; two different digests for one implementation stay and sort.
- Missing/surplus record; v0 schema after digest rebind; identity/image/source/commit/`reproduction_mode` mismatch; missing registry row is `unknown`.
- Hostile index on `public-runs.json` (FIFO/symlink/literal 65537/dup-key/nonfinite); digest-named record over 4MiB fails before parse.
- Stale generated markdown; `validate_run_record` bypass; `--check` does not write; writer exact bytes and LF.
- Sort-by-score: `sort_publication_rows` is identity/digest only; projector source has no `score_percent`.

## Test plan
- [x] `python3 -W error::ResourceWarning conformance/tests/test_public_runs.py` (24/24)
- [x] `python3 conformance/project_public_runs.py --check`
- [x] `git diff --check`
- [x] affected Python suite (`test_public_runs`, `test_completion_scope`, `test_implementations`, `test_pma_v0_registration`, `test_published_rows`, `test_registry`) 188/188
- [ ] GitHub Actions on this head
- [ ] Independent exact-head review of `cef9d29a7` (review on `13eb7623f` does not carry)

## Non-claims
Not a leaderboard, badge, certification, trust score, or implementation-independence proof. A digest addresses bytes, not a publisher. A match is agreement on the pinned corpus only. `reproduction_mode`, image, and identity are publisher-declared and bound; they are not attested or verified. Append-only is a reviewed repository-delta property, not a fact of one checkout. No latest/supersedes semantics. This draft does not merge itself.

## Summary by CodeRabbit

- **New Features**
  - Added a public conformance run record covering 14 privileged MCP action cases, with complete provenance and aggregate results.
  - Added validation and deterministic projection of reviewed public run records into implementation documentation.
  - Added CI validation to ensure the public-run projection remains current and conformant.

- **Documentation**
  - Added a public implementation-runs page explaining record scope, limitations, review requirements, and reproduction details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,13 @@ jobs:
           python3 -W error::ResourceWarning conformance/tests/test_completion_scope.py
           python3 conformance/run_all.py --require-complete --completion-scope required
 
+      - name: Public-run projection
+        shell: bash
+        run: |
+          set -euo pipefail
+          python3 -W error::ResourceWarning conformance/tests/test_public_runs.py
+          python3 conformance/project_public_runs.py --check
+
       - id: detect
         shell: bash
         env:
--- a/conformance/IMPLEMENTATIONS.md
+++ b/conformance/IMPLEMENTATIONS.md
@@ -1,0 +1,11 @@
+# Public implementation runs
+
+This page lists reviewed, digest-addressed `conformance_run.v1` records.
+It is not a leaderboard, badge, certification, or trust score.
+
+A digest addresses bytes. It does not authenticate a publisher.
+A match is agreement on the pinned corpus only.
+
+| implementation | suite | record | image | commit | match | mismatch | execution_error | harness_error |
+|---|---|---|---|---|---|---|---|---|
+| pma-v0-repro | privileged-mcp-action-v0 | sha256:9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27 | ghcr.io/rul1an/pma-v0-repro@sha256:88a5ef285a80dc0caeb2b11093eba79f8f08c870b549cafc395aa77ee5ffc493 | c226a34f3cea50a114607c35f0976048ff3cab2b | 14 | 0 | 0 | 0 |
--- a/conformance/IMPLEMENTATIONS.md
+++ b/conformance/IMPLEMENTATIONS.md
@@ -5,6 +5,8 @@ It is not a leaderboard, badge, certification, or trust score.
 
 A digest addresses bytes. It does not authenticate a publisher.
 A match is agreement on the pinned corpus only.
+Corpus agreement does not prove implementation independence.
+reproduction_mode, image, and identity are publisher-declared and bound; they are not attested or verified.
 
 To add a reviewed row, write it to `public-runs.json` and the digest-named file under `public-runs/`, then run `python3 conformance/project_public_runs.py --check`.
 
--- a/conformance/IMPLEMENTATIONS.md
+++ b/conformance/IMPLEMENTATIONS.md
@@ -8,6 +8,6 @@ A match is agreement on the pinned corpus only.
 
 To add a reviewed row, write it to `public-runs.json` and the digest-named file under `public-runs/`, then run `python3 conformance/project_public_runs.py --check`.
 
-| implementation | suite | record | image | commit | match | mismatch | execution_error | harness_error | review_warnings |
-|---|---|---|---|---|---|---|---|---|---|
-| pma-v0-repro | privileged-mcp-action-v0 | [sha256:9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27](public-runs/9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27) | ghcr.io/rul1an/pma-v0-repro@sha256:88a5ef285a80dc0caeb2b11093eba79f8f08c870b549cafc395aa77ee5ffc493 | c226a34f3cea50a114607c35f0976048ff3cab2b | 14 | 0 | 0 | 0 | 0 |
+| implementation | suite | record | image | commit | reproduction_mode | match | mismatch | execution_error | harness_error | review_warnings |
+|---|---|---|---|---|---|---|---|---|---|---|
+| pma-v0-repro | privileged-mcp-action-v0 | [sha256:9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27](public-runs/9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27) | ghcr.io/rul1an/pma-v0-repro@sha256:88a5ef285a80dc0caeb2b11093eba79f8f08c870b549cafc395aa77ee5ffc493 | c226a34f3cea50a114607c35f0976048ff3cab2b | other_disclosed | 14 | 0 | 0 | 0 | 0 |
--- a/conformance/IMPLEMENTATIONS.md
+++ b/conformance/IMPLEMENTATIONS.md
@@ -6,6 +6,8 @@ It is not a leaderboard, badge, certification, or trust score.
 A digest addresses bytes. It does not authenticate a publisher.
 A match is agreement on the pinned corpus only.
 
-| implementation | suite | record | image | commit | match | mismatch | execution_error | harness_error |
-|---|---|---|---|---|---|---|---|---|
-| pma-v0-repro | privileged-mcp-action-v0 | sha256:9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27 | ghcr.io/rul1an/pma-v0-repro@sha256:88a5ef285a80dc0caeb2b11093eba79f8f08c870b549cafc395aa77ee5ffc493 | c226a34f3cea50a114607c35f0976048ff3cab2b | 14 | 0 | 0 | 0 |
+To add a reviewed row, write it to `public-runs.json` and the digest-named file under `public-runs/`, then run `python3 conformance/project_public_runs.py --check`.
+
+| implementation | suite | record | image | commit | match | mismatch | execution_error | harness_error | review_warnings |
+|---|---|---|---|---|---|---|---|---|---|
+| pma-v0-repro | privileged-mcp-action-v0 | [sha256:9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27](public-runs/9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27) | ghcr.io/rul1an/pma-v0-repro@sha256:88a5ef285a80dc0caeb2b11093eba79f8f08c870b549cafc395aa77ee5ffc493 | c226a34f3cea50a114607c35f0976048ff3cab2b | 14 | 0 | 0 | 0 | 0 |
--- a/conformance/IMPLEMENTATIONS.md.in
+++ b/conformance/IMPLEMENTATIONS.md.in
@@ -8,6 +8,6 @@ A match is agreement on the pinned corpus only.
 
 To add a reviewed row, write it to `public-runs.json` and the digest-named file under `public-runs/`, then run `python3 conformance/project_public_runs.py --check`.
 
-| implementation | suite | record | image | commit | match | mismatch | execution_error | harness_error | review_warnings |
-|---|---|---|---|---|---|---|---|---|---|
+| implementation | suite | record | image | commit | reproduction_mode | match | mismatch | execution_error | harness_error | review_warnings |
+|---|---|---|---|---|---|---|---|---|---|---|
 {{public_runs_table}}
--- a/conformance/IMPLEMENTATIONS.md.in
+++ b/conformance/IMPLEMENTATIONS.md.in
@@ -5,6 +5,8 @@ It is not a leaderboard, badge, certification, or trust score.
 
 A digest addresses bytes. It does not authenticate a publisher.
 A match is agreement on the pinned corpus only.
+Corpus agreement does not prove implementation independence.
+reproduction_mode, image, and identity are publisher-declared and bound; they are not attested or verified.
 
 To add a reviewed row, write it to `public-runs.json` and the digest-named file under `public-runs/`, then run `python3 conformance/project_public_runs.py --check`.
 
--- a/conformance/IMPLEMENTATIONS.md.in
+++ b/conformance/IMPLEMENTATIONS.md.in
@@ -6,6 +6,8 @@ It is not a leaderboard, badge, certification, or trust score.
 A digest addresses bytes. It does not authenticate a publisher.
 A match is agreement on the pinned corpus only.
 
-| implementation | suite | record | image | commit | match | mismatch | execution_error | harness_error |
-|---|---|---|---|---|---|---|---|---|
+To add a reviewed row, write it to `public-runs.json` and the digest-named file under `public-runs/`, then run `python3 conformance/project_public_runs.py --check`.
+
+| implementation | suite | record | image | commit | match | mismatch | execution_error | harness_error | review_warnings |
+|---|---|---|---|---|---|---|---|---|---|
 {{public_runs_table}}
--- a/conformance/IMPLEMENTATIONS.md.in
+++ b/conformance/IMPLEMENTATIONS.md.in
@@ -1,0 +1,11 @@
+# Public implementation runs
+
+This page lists reviewed, digest-addressed `conformance_run.v1` records.
+It is not a leaderboard, badge, certification, or trust score.
+
+A digest addresses bytes. It does not authenticate a publisher.
+A match is agreement on the pinned corpus only.
+
+| implementation | suite | record | image | commit | match | mismatch | execution_error | harness_error |
+|---|---|---|---|---|---|---|---|---|
+{{public_runs_table}}
--- a/conformance/project_public_runs.py
+++ b/conformance/project_public_runs.py
@@ -1,0 +1,261 @@
+#!/usr/bin/env python3
+"""Project reviewed, digest-addressed run records into IMPLEMENTATIONS.md."""
+
+from __future__ import annotations
+
+import argparse
+import os
+import re
+import sys
+import tempfile
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(REPO / "conformance"))
+sys.path.insert(0, str(REPO / "conformance/adequacy"))
+sys.path.insert(0, str(REPO / "conformance/privileged-mcp-action-v0/scripts"))
+
+import published_rows  # noqa: E402
+from implementations import ImplementationRegistryError, load_implementations  # noqa: E402
+from artifact_io import content_sha256  # noqa: E402
+from strict_json import parse_strict_object  # noqa: E402
+from validate_run_record import MAX_RUN_RECORD_BYTES, validate_run_record  # noqa: E402
+
+INDEX_SCHEMA = "assay.conformance.public_runs.v0"
+INDEX_FIELDS = ("schema", "runs")
+RUN_FIELDS = (
+    "implementation_id",
+    "suite",
+    "record_sha256",
+    "image",
+    "source",
+    "commit",
+)
+MAX_INDEX_BYTES = 64 * 1024
+MAX_TEMPLATE_BYTES = 1024 * 1024
+SHA256 = re.compile(r"^sha256:[0-9a-f]{64}$")
+HEX64 = re.compile(r"^[0-9a-f]{64}$")
+TABLE_TOKEN = "{{public_runs_table}}"
+
+
+def _digest(data: bytes) -> str:
+    return content_sha256(data)
+
+
+def index_path(repo: Path) -> Path:
+    return repo / "conformance/public-runs.json"
+
+
+def records_dir(repo: Path) -> Path:
+    return repo / "conformance/public-runs"
+
+
+def record_path(repo: Path, digest: str) -> Path:
+    if not SHA256.fullmatch(digest):
+        raise ValueError("record_sha256 must be a sha256 content address")
+    return records_dir(repo) / digest.removeprefix("sha256:")
+
+
+def sort_publication_rows(rows: list[dict]) -> list[dict]:
+    return sorted(rows, key=lambda row: (row["implementation_id"], row["record_sha256"]))
+
+
+def _registry_by_id(repo: Path) -> dict[str, dict]:
+    try:
+        document = load_implementations(repo / "conformance/implementations.json")
+    except ImplementationRegistryError as exc:
+        raise ValueError("implementation registry rejected: %s" % exc) from exc
+    return {row["id"]: row for row in document["implementations"]}
+
+
+def _load_index(repo: Path) -> list[dict]:
+    data = published_rows.read_regular_file(index_path(repo), MAX_INDEX_BYTES)
+    document = published_rows.parse_json_object(data, "public-runs index")
+    extra = set(document) - set(INDEX_FIELDS)
+    missing = [field for field in INDEX_FIELDS if field not in document]
+    if extra or missing:
+        raise ValueError("public-runs index has missing or surplus fields")
+    if document["schema"] != INDEX_SCHEMA:
+        raise ValueError("unsupported public-runs schema")
+    runs = document["runs"]
+    if not isinstance(runs, list):
+        raise ValueError("public-runs runs must be a list")
+    seen: set[str] = set()
+    clean: list[dict] = []
+    for row in runs:
+        if not isinstance(row, dict) or set(row) != set(RUN_FIELDS):
+            raise ValueError("public-runs row has missing or surplus fields")
+        ident = row["implementation_id"]
+        if not isinstance(ident, str) or ident in seen:
+            raise ValueError("duplicate public-run identity: %s" % ident)
+        seen.add(ident)
+        if not SHA256.fullmatch(row["record_sha256"]):
+            raise ValueError("record_sha256 must be a sha256 content address")
+        clean.append(row)
+    return clean
+
+
+def _listed_record_names(repo: Path) -> set[str]:
+    directory = records_dir(repo)
+    if not directory.exists():
+        return set()
+    if not directory.is_dir() or directory.is_symlink():
+        raise ValueError("%s is not a regular directory" % directory)
+    names: set[str] = set()
+    for entry in directory.iterdir():
+        if entry.name.startswith("."):
+            continue
+        if entry.is_symlink() or not entry.is_file() or not HEX64.fullmatch(entry.name):
+            raise ValueError("surplus record %s" % entry.name)
+        names.add(entry.name)
+    return names
+
+
+def _bind_row(row: dict, report: dict, registry: dict[str, dict]) -> None:
+    implementation = report["implementation"]
+    if "id" not in implementation or "image" not in implementation:
+        raise ValueError("publication requires implementation id and image")
+    registered = registry.get(row["implementation_id"])
+    if registered is None:
+        raise ValueError("implementation_id mismatch: unknown %s" % row["implementation_id"])
+    checks = (
+        ("implementation_id", row["implementation_id"], implementation["id"], registered["id"]),
+        ("suite", row["suite"], report["suite"], registered["suite"]),
+        ("image", row["image"], implementation["image"], registered["image"]),
+        ("source", row["source"], implementation["source"], registered["source"]),
+        ("commit", row["commit"], implementation["commit"], registered["commit"]),
+    )
+    for name, indexed, recorded, expected in checks:
+        if not indexed == recorded == expected:
+            raise ValueError("%s mismatch" % name)
+
+
+def load_publication(repo: Path) -> list[dict]:
+    rows = _load_index(repo)
+    present = _listed_record_names(repo)
+    expected = {row["record_sha256"].removeprefix("sha256:") for row in rows}
+    missing = expected - present
+    surplus = present - expected
+    if missing:
+        raise ValueError("missing record %s" % sorted(missing)[0])
+    if surplus:
+        raise ValueError("surplus record %s" % sorted(surplus)[0])
+    registry = _registry_by_id(repo)
+    loaded: list[dict] = []
+    for row in rows:
+        path = record_path(repo, row["record_sha256"])
+        data = published_rows.read_regular_file(path, MAX_RUN_RECORD_BYTES)
+        digest = _digest(data)
+        if digest != row["record_sha256"]:
+            raise ValueError("record digest mismatch: %s" % digest)
+        report = parse_strict_object(data, label="run record")
+        validate_run_record(report)
+        _bind_row(row, report, registry)
+        loaded.append(
+            {
+                "implementation_id": row["implementation_id"],
+                "suite": row["suite"],
+                "record_sha256": row["record_sha256"],
+                "image": row["image"],
+                "source": row["source"],
+                "commit": row["commit"],
+                "summary": report["summary"],
+            }
+        )
+    return sort_publication_rows(loaded)
+
+
+def render_table(rows: list[dict]) -> str:
+    lines = []
+    for row in rows:
+        summary = row["summary"]
+        lines.append(
+            "| %s | %s | %s | %s | %s | %s | %s | %s | %s |"
+            % (
+                row["implementation_id"],
+                row["suite"],
+                row["record_sha256"],
+                row["image"],
+                row["commit"],
+                summary["match"],
+                summary["mismatch"],
+                summary["execution_error"],
+                summary["harness_error"],
+            )
+        )
+    return "\n".join(lines)
+
+
+def render_document(repo: Path, rows: list[dict]) -> str:
+    template = repo / "conformance/IMPLEMENTATIONS.md.in"
+    source = published_rows.read_regular_file(template, MAX_TEMPLATE_BYTES).decode("utf-8")
+    if source.count(TABLE_TOKEN) != 1:
+        raise ValueError("IMPLEMENTATIONS.md.in must contain exactly one table token")
+    rendered = source.replace(TABLE_TOKEN, render_table(rows))
+    if "{{" in rendered or "}}" in rendered:
+        raise ValueError("template carries an unresolved or malformed token")
+    return rendered
+
+
+def projection_findings(repo: Path) -> list[str]:
+    try:
+        rows = load_publication(repo)
+        expected = render_document(repo, rows)
+    except (OSError, UnicodeDecodeError, ValueError) as exc:
+        return [str(exc)]
+    output = repo / "conformance/IMPLEMENTATIONS.md"
+    try:
+        actual = published_rows.read_regular_file(output, MAX_TEMPLATE_BYTES).decode("utf-8")
+    except (OSError, UnicodeDecodeError, ValueError) as exc:
+        return [str(exc)]
+    if actual != expected:
+        return ["conformance/IMPLEMENTATIONS.md differs from its fresh deterministic projection"]
+    return []
+
+
+def write_document(repo: Path) -> None:
+    rendered = render_document(repo, load_publication(repo))
+    output = repo / "conformance/IMPLEMENTATIONS.md"
+    temporary: Path | None = None
+    try:
+        with tempfile.NamedTemporaryFile(
+            mode="w",
+            encoding="utf-8",
+            dir=output.parent,
+            prefix=output.name + ".",
+            delete=False,
+        ) as handle:
+            handle.write(rendered)
+            handle.flush()
+            os.fsync(handle.fileno())
+            temporary = Path(handle.name)
+        os.chmod(temporary, 0o644)
+        os.replace(temporary, output)
+        temporary = None
+    finally:
+        if temporary is not None:
+            temporary.unlink(missing_ok=True)
+
+
+def main(argv: list[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--check", action="store_true")
+    parser.add_argument("--repo", type=Path, default=REPO)
+    args = parser.parse_args(argv)
+    try:
+        if args.check:
+            findings = projection_findings(args.repo)
+            if findings:
+                for finding in findings:
+                    print(finding, file=sys.stderr)
+                return 1
+        else:
+            write_document(args.repo)
+    except (OSError, UnicodeDecodeError, ValueError) as exc:
+        print("public-run projection failed: %s" % exc, file=sys.stderr)
+        return 1
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/conformance/project_public_runs.py
+++ b/conformance/project_public_runs.py
@@ -91,17 +91,21 @@ def _load_index(repo: Path) -> list[dict]:
     runs = document["runs"]
     if not isinstance(runs, list):
         raise ValueError("public-runs runs must be a list")
-    seen: set[str] = set()
+    seen: set[tuple[str, str]] = set()
     clean: list[dict] = []
     for row in runs:
         if not isinstance(row, dict) or set(row) != set(RUN_FIELDS):
             raise ValueError("public-runs row has missing or surplus fields")
         ident = row["implementation_id"]
-        if not isinstance(ident, str) or ident in seen:
-            raise ValueError("duplicate public-run identity: %s" % ident)
-        seen.add(ident)
-        if not SHA256.fullmatch(row["record_sha256"]):
+        digest = row["record_sha256"]
+        if not isinstance(ident, str):
+            raise ValueError("implementation_id must be a string")
+        if not SHA256.fullmatch(digest):
             raise ValueError("record_sha256 must be a sha256 content address")
+        key = (ident, digest)
+        if key in seen:
+            raise ValueError("duplicate public-run identity: %s %s" % key)
+        seen.add(key)
         clean.append(row)
     return clean
 
--- a/conformance/project_public_runs.py
+++ b/conformance/project_public_runs.py
@@ -60,6 +60,16 @@ def sort_publication_rows(rows: list[dict]) -> list[dict]:
     return sorted(rows, key=lambda row: (row["implementation_id"], row["record_sha256"]))
 
 
+def escape_markdown_cell(value: object) -> str:
+    return (
+        str(value)
+        .replace("\\", "\\\\")
+        .replace("|", "\\|")
+        .replace("<", "&lt;")
+        .replace(">", "&gt;")
+    )
+
+
 def _registry_by_id(repo: Path) -> dict[str, dict]:
     try:
         document = load_implementations(repo / "conformance/implementations.json")
@@ -169,18 +179,24 @@ def render_table(rows: list[dict]) -> str:
     lines = []
     for row in rows:
         summary = row["summary"]
+        digest = row["record_sha256"]
+        record_cell = "[%s](%s)" % (
+            escape_markdown_cell(digest),
+            escape_markdown_cell("public-runs/" + digest.removeprefix("sha256:")),
+        )
         lines.append(
-            "| %s | %s | %s | %s | %s | %s | %s | %s | %s |"
+            "| %s | %s | %s | %s | %s | %s | %s | %s | %s | %s |"
             % (
-                row["implementation_id"],
-                row["suite"],
-                row["record_sha256"],
-                row["image"],
-                row["commit"],
-                summary["match"],
-                summary["mismatch"],
-                summary["execution_error"],
-                summary["harness_error"],
+                escape_markdown_cell(row["implementation_id"]),
+                escape_markdown_cell(row["suite"]),
+                record_cell,
+                escape_markdown_cell(row["image"]),
+                escape_markdown_cell(row["commit"]),
+                escape_markdown_cell(summary["match"]),
+                escape_markdown_cell(summary["mismatch"]),
+                escape_markdown_cell(summary["execution_error"]),
+                escape_markdown_cell(summary["harness_error"]),
+                escape_markdown_cell(summary["review_warnings"]),
             )
         )
     return "\n".join(lines)
@@ -225,10 +241,10 @@ def write_document(repo: Path) -> None:
             prefix=output.name + ".",
             delete=False,
         ) as handle:
+            temporary = Path(handle.name)
             handle.write(rendered)
             handle.flush()
             os.fsync(handle.fileno())
-            temporary = Path(handle.name)
         os.chmod(temporary, 0o644)
         os.replace(temporary, output)
         temporary = None
--- a/conformance/project_public_runs.py
+++ b/conformance/project_public_runs.py
@@ -30,6 +30,7 @@ RUN_FIELDS = (
     "image",
     "source",
     "commit",
+    "reproduction_mode",
 )
 MAX_INDEX_BYTES = 64 * 1024
 MAX_TEMPLATE_BYTES = 1024 * 1024
@@ -134,6 +135,12 @@ def _bind_row(row: dict, report: dict, registry: dict[str, dict]) -> None:
         ("image", row["image"], implementation["image"], registered["image"]),
         ("source", row["source"], implementation["source"], registered["source"]),
         ("commit", row["commit"], implementation["commit"], registered["commit"]),
+        (
+            "reproduction_mode",
+            row["reproduction_mode"],
+            implementation["reproduction_mode"],
+            registered["reproduction_mode"],
+        ),
     )
     for name, indexed, recorded, expected in checks:
         if not indexed == recorded == expected:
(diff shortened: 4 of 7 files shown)


Rul1an:
## Exact-head preflight — BLOCKED

Reviewed `4aecd14e23a4c0d7e46826165f463aa11bb9bb6d` against base `a63a97c16615c181b65aaf6fdf0eed0fc5270ad0`. Baseline focused suite: 13/13 green; real `--check` green. This is a coordinator preflight, not the independent quorum review.

### Findings

1. **HIGH — the hostile-index tests do not exercise the projector input path.** Mutating `_load_index()` from the bounded/no-follow reader to `index_path(repo).read_bytes()` leaves all 13 tests green. The FIFO/symlink/oversize rows currently call `published_rows.read_regular_file()` directly on unrelated paths, so they prove the helper, not its wiring. Put the hostile object at `conformance/public-runs.json` and call `load_publication()`/CLI. Use a literal-built cap probe; do not derive its size from mutable `MAX_INDEX_BYTES`.

2. **HIGH — output truncation is unguarded.** Mutating `handle.write(rendered)` to `handle.write(rendered[:1])` leaves all 13 tests green. Add a production-writer test that starts with existing bytes, calls `write_document()`, and requires the exact complete `render_document()` bytes; add an injected write/replace failure row proving existing bytes and temp cleanup. This is the output-truncation mutation required by #201.

3. **MEDIUM — one typed terminal state is hidden.** A validator-clean v1 with one case `review_warnings` entry and `summary.review_warnings=1` loads successfully, but the projected header/row still show only match, mismatch, execution_error and harness_error. The public row is therefore visually indistinguishable from a warning-free row except for its digest. Project `review_warnings` explicitly and pin it with a valid-record test.

4. **MEDIUM — fully bound registry values can break the Markdown table.** The shared image validator accepts characters such as `|`, `<` and `>`. A registry + record + index all bound to `ghcr.io/rul1an/pma|v0-repro@sha256:…` validate, but render an extra table column (11 pipe delimiters instead of 10). Escape every interpolated Markdown cell from one helper and pin HTML/pipe payloads.

### UX closure

The record digest should link to the checked-in digest-addressed bytes. Add one concise opt-in instruction or command near the table; otherwise the projection is inspectable only by reconstructing a path from the digest. Keep this small and do not add a frontend.

### Verified strengths

The exact hosted bytes/digest are correct; digest-before-parse, strict v1 validation, registry identity/image/source/commit binding, missing/surplus detection, identity/digest sort, `--check` non-write behavior and the no-ranking claim ceiling are all present.

Any push invalidates this preflight. Close the four measured gaps with REDs and meaningful mutations, then request a fresh independent exact-head review.

Rul1an:
## Preflight addendum — fifth measured closure

Still bound to `4aecd14e23a4c0d7e46826165f463aa11bb9bb6d`.

A validator-clean v1 record was changed from `implementation.reproduction_mode=other_disclosed` to `blind_from_spec`; its bytes, digest-addressed filename and index digest were recomputed, while the canonical registry remained `other_disclosed`. `load_publication()` accepted the row, and the mode is absent from both the loaded projection row and rendered table.

This is an identity-binding gap, not a presentation preference: reproduction mode is a closed field in both the registry and v1 record and appears in the programme's allowed portable statement.

Smallest closure:

- add closed `reproduction_mode` to the reviewed index row;
- require index == run record == registry in the existing `_bind_row` rule;
- project it explicitly;
- make a valid allowed-mode swap turn RED after digest/path/index recomputation.

Do not widen this to authorship, display name or version in this slice.

Rul1an:
## Preflight addendum — append-only correction identity

Still bound to `4aecd14e23a4c0d7e46826165f463aa11bb9bb6d`.

`_load_index()` currently treats `implementation_id` alone as unique, while rendering already sorts by `(implementation_id, record_sha256)`. This makes a second reviewed run for the same implementation impossible. It conflicts with parent #191's explicit contract that corrections append a new record and preserve historical bytes.

Smallest closure: define a publication row identity as `(implementation_id, record_sha256)`. Reject an exact duplicate tuple; allow two distinct validated record digests for one implementation and sort them deterministically. Do not add latest/supersedes interpretation in this slice.

Rul1an:
## Coordinator final-head preflight — `13eb7623f4d8987363cc3a3ee7181d35e0f7fc7c`

Base: `a63a97c16615c181b65aaf6fdf0eed0fc5270ad0`
Detached review tree: `/private/tmp/assay-pr2592-final-preflight` (clean)

### Production-path verification

- `python3 -W error::ResourceWarning conformance/tests/test_public_runs.py -v`: **21/21 pass**
- `python3 conformance/project_public_runs.py --check`: **pass**
- `git diff --check <base>...<head>`: **pass**
- required `CI` invokes both commands in `.github/workflows/ci.yml:61-66`
- branch head, detached head and reviewed SHA were equal before and after the run.

### Meaningful mutation evidence

Each mutation ran in a disposable detached worktree; the named focused test went red, then the file was restored and the full 21-test control returned green:

1. bypass the canonical bounded/no-follow index loader;
2. truncate the generated document to one character;
3. remove `reproduction_mode` binding;
4. collapse append-only identity to implementation ID only;
5. remove the explicit LF writer contract;
6. hide valid `review_warnings`;
7. remove Markdown cell escaping;
8. remove the content-addressed record link;
9. remove the opt-in regeneration/check instruction.

The same-implementation/two-distinct-digests mutation was rerun with a semantic tuple collapse; it failed on the duplicate identity as intended rather than on an incidental formatter error.

### Contract disposition

- Publication identity is `(implementation_id, record_sha256)`: exact duplicates reject; corrected/new records for the same implementation remain appendable and sort deterministically.
- Index, stored bytes, validated record and canonical registry bind implementation, suite, image, source, commit and reproduction mode.
- The table exposes typed counts including review warnings; it does not rank or compute a score.
- User-controlled Markdown cells pass one escaping function.
- `--check` is read-only; explicit generation writes via same-directory temp + fsync + atomic replace with LF bytes.
- The digest links to the exact stored record; the prose names the opt-in command.

### Non-claims

This projects one reviewed run record. It does not authenticate the implementation, infer authorship, select a latest/best record, rank implementations, prove corpus completeness, or publish a frontend. Hashes address and integrity-check bytes; they are not signatures.

This preflight is coordinator evidence, **not** the independent review quorum.

Rul1an:
## Independent review — head `13eb7623f4d8987363cc3a3ee7181d35e0f7fc7c`

**Verdict: BLOCKED** on one unpinned guard (B1), with two HIGH findings. The mechanism is sound; the finding is that three of its load-bearing guards are not held by anything.

Measured against exactly `13eb7623f4d8987363cc3a3ee7181d35e0f7fc7c` (base `a63a97c1`, `behind_by = 0`), re-read at the end of the review and unchanged. Diff confirmed: 7 files, +1066/−0. Python 3.14.3 in a disposable shallow clone (CI pins 3.12 — see NON-CLAIMS).

---

### BLOCKER

**B1 — the registry-binding guard does not survive its own bypass.** `conformance/project_public_runs.py:133-135`

Replacing the registry lookup with a fallback built from the record's own identity leaves **21/21 tests green** and publishes an `implementation_id` that is absent from `conformance/implementations.json`:

```
=== M9b (unregistered implementation accepted) ===
Ran 21 tests in 0.339s
OK
--- does an UNREGISTERED id now publish? ---
PUBLISHED unregistered id -> totally-unregistered
```

The nearest test, `test_identity_image_source_commit_mismatch_fails_closed` (`test_public_runs.py:240`), sets `implementation_id` in the **index** only, so it fires on the index-vs-record equality check at line 150, never on registry membership. Its assertion is `any("mismatch" in item ...)`, and the registry message happens to contain the word "mismatch" — so it reads as covering this and does not. CONFIRMED.

This is the §6 blocker shape exactly: the edit that removes registry binding also removes every check that would report it. It matters because the registry is where `authorship`, `language` and `reproduction_mode` are declared and reviewed; a row that bypasses it is a public listing with no reviewed registry entry behind it.

*Smallest closure:* one test. `_rebind_record` (`test_public_runs.py:80`) already takes a `mutate_registry` hook — rename or drop the registry row, assert `load_publication` raises with `unknown`.

---

### HIGH

**H1 — nothing renders more than one table row, so a row can vanish silently.** `project_public_runs.py:190-215`

Truncating `render_table` to drop its last row when there is more than one keeps the whole battery green, because the pristine corpus has exactly one row and every render assertion is single-row. `--check` cannot catch it either: it compares the document against the *same* renderer, so both sides truncate identically.

Demonstrated with two genuine, digest-distinct, validator-passing records:

```
evidence rows: 2   published doc rows: 1
--check findings: []
=== full battery under this mutation ===
Ran 21 tests in 0.346s
OK
```

The dropped row was evidence-backed and indexed. Today this is latent — one row is published — but it bites on the first second row, which is the state the feature exists to reach. CONFIRMED.

*Smallest closure:* `test_same_implementation_two_digests_are_kept_and_sorted` (`test_public_runs.py:186`) already builds the two-row state and stops at `load_publication`. Three more lines: call `render_document` and assert both digests appear.

**H2 — the record's fixed non-claim ceiling does not travel to the page.** `IMPLEMENTATIONS.md.in:1-9` vs `validate_run_record.py:52-58`

`RUN_NON_CLAIMS` is validated into every record (`validate_run_record.py:417-420`) and carries five items. The page carries a paraphrase of one ("A match is agreement on the pinned corpus only") plus two good ones of its own. The item that most directly bounds how a reader should read a table of "Public implementation runs" — *"a matching run does not establish implementation independence"* — is validated in and then dropped.

The omission has teeth here, because the sole listed row is first-party and the registry says so:

- `source: https://github.com/Rul1an/pma-v0-repro`, `image: ghcr.io/rul1an/...` — same owner as this repo
- `authorship: {kind: "agent-assisted", model: "Grok Bot", ...}` in `conformance/implementations.json`

`load_implementations` loads that `authorship` block and `_bind_row` (`project_public_runs.py:136-151`) never looks at it. A reader sees a page headed "Public implementation runs" with one row and `14 match`, and nothing on the page says the implementation is the publisher's own or agent-assisted. The existing non-claims are good and I am not asking for more of them in general — I am asking for the two the artifacts already carry.

Related, same class: the `reproduction_mode` column is bound three ways (index = record = registry, `project_public_runs.py:142-147`) and **verified zero ways**. `validate_run_record.py:97` says so in its own words — identity is "Shape-validated, never verified" — and `implementations.py:79-80` says a digest "does not establish that the addressed image is the one that ran." The page says a digest does not authenticate a publisher, but says nothing about the mode or the image being self-declared.

*Smallest closure:* template lines in `IMPLEMENTATIONS.md.in`. No code change.

---

### MEDIUM

**M1 — the record-side bounded read is unpinned.** `project_public_runs.py:168`

Swapping `published_rows.read_regular_file(path, MAX_RUN_RECORD_BYTES)` for `path.read_bytes()` → **21/21 green**. The shipped code is correct; I probed it directly and it refuses:

```
shipped code says: ['.../f119b115... exceeds the 4194304-byte limit']
```

So this is coverage, not behavior. Partial backstop: `_listed_record_names` (line 113-126) refuses symlinks and non-regular entries in `public-runs/` before any read, so the untested residue is the 4 MiB size cap and the `O_NOFOLLOW`/`O_NONBLOCK` open. The index side *is* tested (`test_index_path_hostile_inputs_fail_before_render` covers symlink, FIFO, oversize, duplicate keys, `1e999`); the record side has no equivalent. *Closure:* one test writing a >4 MiB digest-named record.

---

### LOW

**L1 — two JSON loaders in one file.** The index goes through `published_rows.parse_json_object` (line 84); the record goes through `parse_strict_object` (line 172). The index therefore skips the depth and number-domain bounds that `strict_json.py:9-12` says exist precisely so a `RecursionError` traceback cannot reach the user. Exploitability UNVERIFIED — see NON-CLAIMS.

**L2 — a hostile record can produce a traceback instead of a typed finding.** `projection_findings` catches `(OSError, UnicodeDecodeError, ValueError)` (line 233); `validate_run_record`'s own `main` catches `KeyError`, `TypeError` and `RecursionError` as well (`validate_run_record.py:436-443`). A digest-matching record with a non-string `source` reaches `urlparse` at `validate_run_record.py:311`:

```
UNCAUGHT out of projection_findings: AttributeError 'int' object has no attribute 'decode'
```

It still **fails closed** (uncaught → non-zero exit, no write), so this is presentation, not soundness.

**L3 — `_listed_record_names` silently skips dot-prefixed entries** (line 121-122). They are never published, so no soundness impact; it is the one directory-entry class not refused.

**L4 — two assertions that look like guards and are not, named as asked.**
- `test_reuses_canonical_loader_validator_and_registry` (`test_public_runs.py:164-173`) is an AST *name* check. It stayed green under both loader-bypass mutations (M1 and M2 below), because the names survive at the other call site. It says nothing about which call sites use the canonical loader.
- `test_sorts_by_identity_and_digest_never_score:162` — `assertNotIn("score_percent", ast.dump(source))` forbids one identifier spelling. The real protection in that test is the ordering assertion at 155-158, and that one does fire.

**L5 — adjacent, out of diff scope:** `conformance/registry.py:168` `render_inventory_table` is a second Markdown-table renderer with no escaping at all. Different document (`INDEX.md`), pre-existing, repo-controlled input. Named only because §3 asks about "exactly one escaping rule"; within *this* document there is exactly one and it is applied to all eleven cells.

---

### Mutation table

Control before and after each: 21 tests, OK, tree clean (`git status --porcelain` empty between every mutation; restored via `git checkout --`; never two batteries at once).

| # | Mutation | Result | Named test that fired |
|---|---|---|---|
| M1 | index read → `Path.read_bytes()` | RED | `test_index_path_hostile_inputs_fail_before_render` (`AssertionError: ValueError not raised`, symlink case) |
| M2 | record read → `Path.read_bytes()` | **GREEN — ESCAPE** | — (see M1 finding) |
| M3 | `render_table` drops last row when >1 | **GREEN — ESCAPE** | — (see H1) |
| M4 | drop `reproduction_mode` from `_bind_row` | RED | `test_valid_reproduction_mode_swap_fails_closed` |
| M5 | identity tuple → `(implementation_id,)` only | RED | `test_same_implementation_two_digests_are_kept_and_sorted` |
| M6 | `escape_markdown_cell` → `str(value)` | RED | `test_bound_image_markup_stays_one_cell` |
| M7a | `review_warnings` cell hardcoded to `0` | RED | `test_valid_review_warnings_are_projected` |
| M7b | `review_warnings` column removed (+ template) | RED (4) | `test_valid_review_warnings_are_projected`, `test_reproduction_mode_is_projected`, `test_bound_image_markup_stays_one_cell`, `test_pristine_tree_is_green` |
| M8 | record link → plain digest text | RED (2) | `test_record_digest_links_to_stored_bytes_and_names_opt_in`, `test_pristine_tree_is_green` |
| M9a | registry lookup → `or implementation` | RED **but ESCAPE-adjacent** | `test_identity_image_source_commit_mismatch_fails_closed` — *ERROR*, uncaught `KeyError`, i.e. crash not guard. Redone cleanly as M9b |
| M9b | registry lookup → record-derived fallback | **GREEN — ESCAPE** | — (see B1) |
| M10 | record digest check removed | RED | `test_record_byte_flip_fails_check_and_preserves_markdown` |

Every red was read from failure output, not exit code. M9a is reported as an escape-adjacent red precisely because its red traces to a crash rather than to a guard.

---

### Per-item results (§3)

| Item | Result |
|---|---|
| Bounded, no-follow read — **index** | **PASS**, guard correct and tested (`published_rows.py:57-96`: `lstat`+`S_ISREG`, `O_NOFOLLOW`/`O_NONBLOCK`/`O_CLOEXEC`, dev/ino recheck after open, size cap before *and* during read) |
| Bounded, no-follow read — **record** | **PASS on behavior, FAIL on coverage** (M1 above) |
| Strict JSON parsing | **PASS** for records (`parse_strict_object`, depth + number domain). **PARTIAL** for the index (L1) |
| Real record validator | **PASS** — `validate_run_record` is the shipped one, not a local restatement; bypass is visible (`test_validator_bypass_is_visible`) |
| Registry binding | **PASS on behavior** (line 133-135 does require it), **FAIL on coverage** — B1 |
| `reproduction_mode` carried and checked | **PASS** — three-way equality, tested (M4) |
| `reproduction_mode` *unforgeable* | **FAIL, by design and undisclosed.** It is a declaration bound across three publisher-controlled files. Nothing attests it, and the page does not say so — H2 |
| Append-only identity `(implementation_id, record_sha256)` | **PASS** — enforced at line 105-108; no `latest`, no `supersedes`, no in-place mutation path (grep clean); `write_document` touches only `IMPLEMENTATIONS.md`, never the index or a record |
| Digest over exact stored bytes; filename matches | **PASS** — verified by hand: `shasum -a 256` of the stored file == its filename == the index `record_sha256`; also pinned at `test_public_runs.py:129-139` including an exact byte length (8102) |
| Deterministic sort, no score | **PASS** — `sorted(key=(implementation_id, record_sha256))`; the ordering assertion genuinely discriminates against score-sorting (its fixture puts the 0-match row first) |
| `review_warnings` visible | **PASS with a caveat.** The *count* cannot be understated: `validate_run_record.py:412-416` cross-checks `summary["review_warnings"]` against the cases, and M7a/M7b both go red. The warning **strings** are not on the page — only the count and a link to the record. Flagging as a deliberate choice to confirm, not as a defect |
| Exactly one Markdown-escaping rule | **PASS** within this document — `escape_markdown_cell` on all 11 cells including link text and href. See L5 for the adjacent renderer |
| Read-only `--check` | **PASS** — `main` under `--check` calls only `projection_findings`, which has no write path; `test_check_does_not_write` pins bytes and `st_mtime_ns` |
| Writer discipline | **PASS** — `newline="\n"`, `flush`, `os.fsync`, `chmod 0644`, `os.replace`, `finally: unlink(missing_ok=True)`. Both failure modes tested with injected `write` and `replace` failures, each asserting prior bytes intact *and* no leftover temp files (`test_public_runs.py:330-364`) |
| Required-CI wiring | **PASS** — the step lives in the `scope` job (`ci.yml:61-67`, check-run "Detect CI scope"), unconditional on every PR. The required context `CI` is the `ci` job at `ci.yml:1041-1046`, whose `needs:` list begins with `scope`. A red projection blocks a required context |
| Public non-claims | **FAIL** — H2. The three non-claims that *are* on the page are well chosen; the two the artifacts carry and the page omits are the ones that matter for a one-row, first-party table |

---

### CI — read twice, honestly

**First read** (start of review): `Build + Test (windows-latest)` = `completed / failure`; `Build + Test (ubuntu-latest)` = `in_progress`.

**Final read** (immediately before this comment): 49 `success`, 12 `skipped`, 1 `in_progress`. Nothing currently failing.

- `Build + Test (windows-latest)` — **`in_progress`** (run attempt 2). Not green.
- 12 jobs `skipped`, not green: `MCP Registry foundation smoke`, `Kernel Matrix (${{ matrix.kernel }})`, `Build Artifacts (Ubuntu ARM)`, `Free disk (before matrix)`, `Re-derive selected corpora`, `Integration Parity (Optional)`, `Criterion compare (PR vs main)`, `Wave 0 semver checks (public crates)`, `Wave 0.1 nightly safety (non-blocking stub)`, `Wave 0 feature matrix`, `Wave 0 quality gates`, `Archive closed PR branch with Bencher`.
- Required contexts on `main`: `CI`, `host-capability-check`, `lane-check/proof` (`strict: true`). `host-capability-check` = success (check run); `lane-check/proof` = success (status). `CI` is still pending on attempt 2 via `needs: [... test ...]`.

**The attempt-1 failure**, for the record: `Build + Test (windows-latest)` failed at step **"Test machine stdout sink contract (Windows)"**, which cascaded into the `CI` job failing at "Evaluate required job results". That step (`ci.yml:781-795`) runs `cargo test -p assay-cli --test stdout_json_write_contract`. **I do not attribute it to this diff** — the diff contains no Rust, and the new step runs in a different job on Ubuntu — but I equally do not call it a flake, because I did not verify that. It is currently re-running and its outcome is unknown at the time of writing.

---

### Clean properties worth stating

The core mechanism is right and I want to be clear about that: content-addressing over exact stored bytes, refusal before render on every hostile index input I could construct, a genuinely atomic writer with both failure modes pinned, an identity tuple with no supersession vocabulary anywhere, sorting that provably ignores the score, and a `--check` that is structurally incapable of writing. Eight of my ten meaningful mutations went red on a purpose-built test. The three that did not are coverage gaps around real guards, not broken guards — the shipped code refuses every one of those inputs today.

---

### NON-CLAIMS — what I did not verify

- **Python version.** Everything above ran on **3.14.3**; the `scope` job pins **3.12**. I could not test 3.12. This is load-bearing for L1: I probed nesting at 10 000 and 30 000 levels inside the 64 KiB index cap and CPython 3.14 parsed both without a `RecursionError`. Whether 3.12 does the same is **UNVERIFIED**, so L1's exploitability is unproven either way.
- I did **not** build the Rust workspace, and did not run the Windows job or reproduce the attempt-1 failure.
- I did **not** verify that `ghcr.io/rul1an/pma-v0-repro@sha256:88a5ef...` exists, is pullable, or is the image that produced the stored record. Nothing in this diff claims it does, and `implementations.py:79-80` says explicitly that it cannot.
- I did **not** verify the record's `capture_sha256`, `pack_sha256`, `source_corpus_digest` or `rendered_set_digest` against any capture — `require_run_record_binds_capture` is not on this projection path, by design.
- I did **not** audit `published_rows.py`, `strict_json.py`, `validate_run_record.py`, `implementations.py` or `artifact_io.py` as changes — they are pre-existing and unmodified here. I read them only far enough to judge what this diff inherits from them.
- I did **not** assess whether one row is the right corpus to launch this page with; that is a product call, not a review finding.
- L5 (`registry.py`'s unescaped table) is reported as an observation only. I did not attempt to exploit it and its input is repo-controlled.

---

### Independence declaration

I am a fresh agent and I qualify. Concretely, what I checked: I searched my entire inherited context — the project `CLAUDE.md`, the persistent memory index and all its files, and this review brief — for any claim of prior involvement with this change. `grep -ril` across the memory directory for `public_runs`, `public-runs`, `project reviewed run` and `2592` returned **zero** matches; the single hit for `201` was an unrelated substring in `grep-match-is-not-structural-evidence.md`. I have not authored or edited this change, `Rul1an/assay-tunnel-experiments#201` (which I have never read and cannot access), or any normative specification or implementation plan for it. I did not read, and did not look for, any earlier review, preflight or design note for this PR; I encountered none. Every judgment above is my own, derived from the artifacts at this SHA and from mutations I wrote myself.

---

<details>
<summary><b>Evidence appendix — exact commands</b></summary>

```bash
# Anchors
gh api repos/Rul1an/assay/pulls/2592 --jq '{head:.head.sha, base:.base.sha, changed_files, additions, deletions}'
#  head 13eb7623f4d8987363cc3a3ee7181d35e0f7fc7c  base a63a97c1...  7 files  +1066/-0

git init repo && git remote add origin https://github.com/Rul1an/assay.git
git fetch --depth 1 origin 13eb7623f4d8987363cc3a3ee7181d35e0f7fc7c && git checkout FETCH_HEAD
git fetch --depth 1 origin a63a97c16615c181b65aaf6fdf0eed0fc5270ad0
git diff --stat a63a97c1 13eb7623
#  7 files changed, 1066 insertions(+)   [matches the PR]

# Stored bytes vs digest vs index
shasum -a 256 conformance/public-runs/9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27
#  9275ac65b1f2dde89299fcc811c733096b3b5683cb5ed15a8f32560d4580ae27   == filename == index record_sha256

# Control
python3 -V                              # Python 3.14.3   (CI pins 3.12)
python3 -W error::ResourceWarning conformance/tests/test_public_runs.py
#  Ran 21 tests ... OK
python3 conformance/project_public_runs.py --check ; echo rc=$?
#  rc=0
git status --porcelain                  # empty, before and after every mutation

# B1 — registry binding removed (project_public_runs.py:133-135 -> record-derived fallback)
#  Ran 21 tests ... OK
#  PUBLISHED unregistered id -> totally-unregistered

# H1 — render_table drops last row when len(lines) > 1
#  evidence rows: 2   published doc rows: 1
#  --check findings: []
#  Ran 21 tests ... OK

# M1 — record read -> path.read_bytes()   =>  Ran 21 tests ... OK
#      shipped code, unmutated, on a >4 MiB record:
#      ['.../f119b115... exceeds the 4194304-byte limit']

# L2 — record with non-string `source`, digest matching
#  UNCAUGHT out of projection_findings: AttributeError 'int' object has no attribute 'decode'

# Append-only
grep -rn "latest|supersede|replace_row|update_row|revoke|deprecat" \
  conformance/project_public_runs.py conformance/public-runs.json conformance/IMPLEMENTATIONS.md.in
#  (none)

# CI wiring
grep -nE "^  [a-z_-]+:|^    name:|needs:" .github/workflows/ci.yml
#  24: scope:  / name: Detect CI scope        <- new step at ci.yml:61-67
#  1041: ci:   / name: CI / needs: [scope, deps-security, ..., test, ebpf-smoke-ubuntu]

gh api repos/Rul1an/assay/branches/main/protection \
  --jq '{strict:.required_status_checks.strict, contexts:.required_status_checks.contexts}'
#  {"contexts":["CI","host-capability-check","lane-check/proof"],"strict":true}

# CI, final read
gh api repos/Rul1an/assay/commits/13eb7623.../check-runs --paginate \
  --jq '.check_runs[] | .conclusion // .status' | sort | uniq -c
#  1 in_progress   12 skipped   49 success

# attempt 1, for the record
gh api repos/Rul1an/assay/actions/runs/32547296016/attempts/1/jobs \
  --jq '.jobs[] | select(.conclusion=="failure") | {name, steps:[.steps[]|select(.conclusion=="failure")|.name]}'
#  Build + Test (windows-latest) -> "Test machine stdout sink contract (Windows)"
#  CI                            -> "Evaluate required job results"

# Head re-read immediately before posting
gh api repos/Rul1an/assay/pulls/2592 --jq '.head.sha'
#  13eb7623f4d8987363cc3a3ee7181d35e0f7fc7c   (unchanged)
```

</details>

Rul1an:
## Windows retry disposition — infrastructure timing, not branch behavior

Still bound to head `13eb7623f4d8987363cc3a3ee7181d35e0f7fc7c`.

Attempt 1 failed only in `partial_read_times_out_when_the_child_never_writes_stdout`:

```
prefix read must name its deadline after kill/reap:
partial-read hang probe: child did not exit within the wait deadline; reap grace expired
```

This PR changes only the seven conformance projection/data/CI-invocation files listed in the diff; it does not touch the Rust stdout sink test or process harness. The same CI workflow on base/main `a63a97c1` was green. After classifying the failure as a deadline/reap fluctuation rather than a deterministic product failure, I ran exactly one failed-job rerun.

Attempt 2 completed successfully. The same Windows step passed, the required `CI` aggregator is green, and the PR head did not move. No timeout, retry, harness or product code was changed. A repeat on the same head would have required a separate tracked harness finding instead of another retry.

Rul1an:
## Independent re-review — head `cef9d29a7a0fe5a08d3f06d3a5ea4125fce301b1`

**Verdict: READY.**

All four findings from my `13eb7623` review are closed. I re-derived every one rather than trusting the repair summary, re-ran my full original battery against this head, and found no regression and no new finding. This supersedes my BLOCKED verdict on `13eb7623`, which no longer counts.

Anchors verified myself: head `cef9d29a7a0fe5a08d3f06d3a5ea4125fce301b1` (re-read immediately before posting, unchanged), base `a63a97c1`, 7 files, **+1110/−0**. Python 3.14.3, disposable shallow clone, no Rust build.

---

### The repair is +44/−0, and the projector is byte-identical

```
git diff --numstat 13eb7623 cef9d29a
  2   0   conformance/IMPLEMENTATIONS.md
  2   0   conformance/IMPLEMENTATIONS.md.in
 40   0   conformance/tests/test_public_runs.py

git diff --exit-code 13eb7623 cef9d29a -- conformance/project_public_runs.py
  PROJECTOR UNCHANGED
```

That is the right shape, and I want to say why it is not a shortcut. B1, H1 and M1 were **coverage** findings on behavior that was already correct — I had probed each shipped guard directly and each refused. Tests were the honest closure, not a code change. H2 was a wording finding and I named the template as its smallest closure. **Zero deletions anywhere**, so no existing guard was weakened to make room; test count 21 → 24.

---

### B1 — CLOSED

`test_missing_registry_row_fails_unknown` (`test_public_runs.py:444-452`) renames the registry row's `id`, leaving `registry.get("pma-v0-repro")` → `None`, which is the same code path a deleted row takes. Both of my bypass mutations now bite:

| Mutation | Result | Test that fired |
|---|---|---|
| registry lookup → record-derived fallback (**the exact M9b that escaped on `13eb7623`**) | **RED** | `test_missing_registry_row_fails_unknown` |
| registry lookup deleted outright, `registered = dict(implementation, suite=...)` | **RED** | `test_missing_registry_row_fails_unknown` |

The red is `ValueError not raised` at line 450 — the guard itself, not scaffolding. And the assertion is not a substring loophole: under the mutation no exception is raised at all, so it fails at `assertRaises` before `assertIn("unknown", ...)` is ever reached. That was my specific complaint about the old neighbouring test, and it does not apply here.

I also checked that the test operates on the **shipped** bytes. `_rebind_record` with a no-op payload mutator re-serializes the record, so I verified the stored record is already in that canonical form:

```
round-trip byte-identical: True | digest 9275ac65b1f2dde8
```

So the no-op is a true no-op and the test rebinds the real record at its real digest.

---

### H1 — CLOSED, and the two rows are genuinely real

The addition at `test_public_runs.py:214-218` renders the document and asserts two data rows plus both digest links. Four row-dropping mutations, all **RED** on `test_same_implementation_two_digests_are_kept_and_sorted`:

| Mutation | Result |
|---|---|
| drop **last** row when >1 (**the exact escape from `13eb7623`**) | RED |
| drop **first** row when >1 | RED |
| `lines[:1]` — cap the table at one row | RED |
| de-duplicate by `implementation_id` (a plausible "tidy the table" regression) | RED |

**Not a contrived duplicate.** I reconstructed what the test builds and measured it:

```
record A bytes: 8102  digest 9275ac65b1f2dde8
record B bytes: 8166  digest 2c16178376354c82
byte-identical?  False
  record A: validator PASSES, summary.review_warnings = 0
  record B: validator PASSES, summary.review_warnings = 1
rendered data rows: 2
   warnings cell = 1 | digest 71fd697e58201faf
   warnings cell = 0 | digest 8f32560d4580ae27
```

Two records with different byte lengths, different digests, **both passing the full `validate_run_record`**, rendering to two rows that differ in a visible cell. Same `implementation_id`, which is the legitimate append-only case — one implementation, two runs — and exactly the shape the identity tuple exists to permit. This is a real two-row render, not a duplicated line.

---

### H2 — CLOSED. My judgment on the wording

The page now reads (`IMPLEMENTATIONS.md:1-11`, prose identical to `.in`):

> It is not a leaderboard, badge, certification, or trust score.
> A digest addresses bytes. It does not authenticate a publisher.
> A match is agreement on the pinned corpus only.
> **Corpus agreement does not prove implementation independence.**
> **reproduction_mode, image, and identity are publisher-declared and bound; they are not attested or verified.**

**Does it over-claim? No.** I checked each word against what the artifacts establish:

- *"publisher-declared"* — accurate. `validate_run_record.py:97` says the implementation block is "Shape-validated, never verified."
- *"and bound"* — accurate, and a precise word. `_bind_row` (`project_public_runs.py:136-151`) requires three-way equality across index, record and registry.
- *"not attested or verified"* — accurate. `implementations.py:79-80`: a digest "does not authenticate the publisher and it does not establish that the addressed image is the one that ran." `pack_provenance_verification` is pinned to `not_performed_by_scorer`.

**Does it under-claim?** It claims less than the artifacts support, which is the correct direction for a non-claim. The independence line is a faithful rendering of the record's own frozen ceiling item (`validate_run_record.py:54`, "a matching run does not establish implementation independence") and is in one respect *broader* — "corpus agreement" covers more than "a matching run".

The declared-vs-bound-vs-attested distinction is exactly the one I said was missing, stated in one sentence, without inventing a guarantee. I judge the wording honest.

It is also pinned, and pinned against softening rather than merely against deletion:

| Mutation | Result | Test that fired |
|---|---|---|
| independence line removed (template + doc) | RED | `test_public_non_claims_name_binding_without_attestation` |
| declared/attested line removed (template + doc) | RED | same |
| **line kept but reversed** to "they are attested and verified" | RED | same |

That third one matters: the assertion is not "some line exists", it holds the specific denial.

---

### M1 — CLOSED, and "before parse" is literally true

`test_oversize_digest_named_record_fails_before_parse` (`test_public_runs.py:454-469`) writes a 4194305-byte digest-named record and asserts the refusal message contains `exceeds`.

| Mutation | Result |
|---|---|
| record read → `path.read_bytes()` (**the exact escape from `13eb7623`**) | **RED** |
| cap raised to `MAX_RUN_RECORD_BYTES * 4`, canonical loader kept | **RED** |

The first failure output shows precisely why the message assertion — not just `assertRaises` — is the load-bearing part:

```
AssertionError: 'exceeds' not found in 'run record is not readable JSON:
  Expecting property name enclosed in double quotes: line 1 column 2 (char 1)'
```

Without the bounded loader the file is fully materialized and the refusal comes from the *parser*. `assertRaises(ValueError)` alone would have passed. Asserting `exceeds` is what makes this test discriminate between "refused" and "refused for the right reason at the right time".

I verified "before parse **or render**" directly, by instrumenting both call sites rather than inferring it from source order:

```
record size: 4194305 (limit 4194304)
refused with: 17a53cab... exceeds the 4194304-byte limit
parse/render calls that happened: NONE
```

Zero parse calls, zero render calls.

---

### Sufficiency, staleness, `--check`

Changed-file list derived myself with `git diff --name-only a63a97c1 cef9d29a` — 7 files, matching the PR's count. Stored record digest re-verified by hand: `shasum` output == filename == index `record_sha256`.

`--check` is **green** (`rc=0`) and the generated document is **not stale**: `diff <(head -10 …md.in) <(head -10 …md)` reports identical prose. I tested staleness in both directions, since a two-file prose edit is exactly where a half-applied change hides:

| Mutation | Result | Test that fired |
|---|---|---|
| line added to the **template** only, doc left behind | RED | `test_pristine_tree_is_green` |
| line added to the **generated doc** only | RED | `test_pristine_tree_is_green` |

**Are the changed files enough?** Yes for all four findings. B1/H1/M1 were coverage on already-correct behavior and now have tests that bite the exact bypasses; H2 was wording and the wording is now honest and pinned.

---

### No regression: what I re-ran, and what I did not

Full original battery re-run against this head. Control before and after each, `git status --porcelain` empty between every mutation, restored via `git checkout --`, never two batteries at once.

| # | Mutation | `13eb7623` | `cef9d29a` | Test that fired now |
|---|---|---|---|---|
| R1 | index read → `read_bytes()` | RED | RED | `test_index_path_hostile_inputs_fail_before_render` |
| R4 | drop `reproduction_mode` binding | RED | RED | `test_valid_reproduction_mode_swap_fails_closed` |
| R5 | identity tuple → id only | RED | RED | `test_same_implementation_two_digests_are_kept_and_sorted` |
| R6 | escaping removed | RED | RED | `test_bound_image_markup_stays_one_cell` |
| R7a | `review_warnings` hardcoded `0` | RED | RED | `test_valid_review_warnings_are_projected` |
| R8 | record link removed | RED (2) | **RED (3)** | + `test_same_implementation_two_digests_are_kept_and_sorted` |
| R10 | record digest check removed | RED | RED | `test_record_byte_flip_fails_check_and_preserves_markdown` |
| M9b | registry fallback | **GREEN (escape)** | **RED** | `test_missing_registry_row_fails_unknown` |
| M3 | drop-last-row | **GREEN (escape)** | **RED** | `test_same_implementation_two_digests_are_kept_and_sorted` |
| M2 | record read → `read_bytes()` | **GREEN (escape)** | **RED** | `test_oversize_digest_named_record_fails_before_parse` |

Every previously-red guard is still red; every previously-green escape is now red. R8 got strictly stronger. Control: 24 tests, OK; `--check` rc=0; tree clean.

**What I did not re-run:** I did not re-derive the properties that depend only on `project_public_runs.py`'s behavior from first principles a second time — bounded index reads, atomic-writer discipline, sort determinism, append-only vocabulary, `--check` read-only. My justification is mechanical rather than assumed: `git diff --exit-code` proves that file is byte-identical to the head I audited, and the three tests added are additive with zero deletions, so no guard on those properties can have been weakened. I did re-run the mutations covering them (R1, R5, R6, R7a, R8, R10) and all stayed red.

**No new finding introduced.** The three new tests leave no artifacts (`git status` clean after every run), pass under `-W error::ResourceWarning`, and add ~0.2 s for the 4 MiB write inside a `TemporaryDirectory`. The added prose contains no `{{`/`}}` and so cannot trip the template-token guard at `project_public_runs.py:224`.

---

### Findings

**No blockers. No high findings.** Everything below is carried over unchanged from my `13eb7623` review, because the projector is byte-identical — none of it is new, and none of it blocks.

**LOW (unchanged, re-verified on this head)**

- **L1** — two JSON loaders in one file: index via `published_rows.parse_json_object` (`:84`), record via `parse_strict_object` (`:172`). The index skips the depth and number-domain bounds `strict_json.py:9-12` exists to provide. Exploitability still UNVERIFIED (see non-claims).
- **L2** — `projection_findings` catches only `(OSError, UnicodeDecodeError, ValueError)` (`:233`) where `validate_run_record`'s own `main` also catches `KeyError`/`TypeError`/`RecursionError`. Re-confirmed on this head: a digest-matching record with a non-string `source` gives `UNCAUGHT: AttributeError 'int' object has no attribute 'decode'`. Fails closed (non-zero exit, no write), so presentation only.
- **L3** — `_listed_record_names` silently skips dot-prefixed entries (`:121-122`). Never published; no soundness impact.
- **L4** — `test_reuses_canonical_loader_validator_and_registry` (`:164-173`) remains an AST *name* check and still stays green under both loader-bypass mutations. It is now **moot** for the record path, since `test_oversize_digest_named_record_fails_before_parse` covers that property behaviorally. Naming it again only so it is not later mistaken for the thing doing the work.
- **L5** — `conformance/registry.py:168` `render_inventory_table` is a second, unescaped Markdown-table renderer for a different document. Pre-existing, repo-controlled input, out of this diff's scope.

**NITS on the repair itself (not findings, no action required)**

- `test_oversize_digest_named_record_fails_before_parse` hardcodes `4194305` rather than deriving from `MAX_RUN_RECORD_BYTES`. If the constant were raised, the test would go **red loudly** rather than silently stop testing the boundary — my M1a run is the evidence, since it produced exactly that failure mode. Importing the constant would be tidier; nothing is at risk.
- The page's line 9 names "identity" where the table's unverified columns are specifically `commit` and `image`. `commit` is covered by "identity", correctly but a little loosely. Naming it would cost three words.
- "does not **prove** implementation independence" is a shade softer than the record's own "does not **establish**". Adopting the record's exact verb would be free. I am not asking for it.

**Carried over, explicitly narrower than what I originally raised:** the page still does not state that the sole listed implementation is first-party (`github.com/Rul1an`, `ghcr.io/rul1an`) and registered `authorship.kind = "agent-assisted"` in `implementations.json`, which the projector loads and does not project. I raised this as part of H2 and named template lines as the smallest closure; the writer delivered exactly that, and with the independence line present a reader can no longer draw the inference I was worried about. **I consider H2 closed** and record this only as a future option — linking the page to `implementations.json` — not as a residual defect. Escalating it now would be moving the goalposts.

---

### CI — read twice, per job

**Early read** (start of this review): 37 `success`, 11 `skipped`, 10 `in_progress`, **1 `neutral`**. I recorded that count but did not capture the neutral job's name before it resolved; I am reporting that limitation rather than guessing which it was.

**Final read** (immediately before posting): **51 `success`, 14 `skipped`, 1 `in_progress`, 0 `neutral`, 0 `failure`.**

Not green, named individually:

- `Build + Test (windows-latest)` — **`in_progress`**, attempt 1.
- 14 `skipped`: `Archive closed PR branch with Bencher` (×2), `Build Artifacts (Ubuntu ARM)`, `Criterion compare (PR vs main)` (×2), `Free disk (before matrix)`, `Integration Parity (Optional)`, `Kernel Matrix (${{ matrix.kernel }})`, `MCP Registry foundation smoke`, `Re-derive selected corpora`, `Wave 0 feature matrix`, `Wave 0 quality gates`, `Wave 0 semver checks (public crates)`, `Wave 0.1 nightly safety (non-blocking stub)`.

Every workflow run on this SHA is **attempt 1** — nothing has been re-run, and nothing has failed on this head. Required contexts on `main` are `CI`, `host-capability-check`, `lane-check/proof` (`strict: true`); `lane-check/proof` and `host-capability-check` are success, and `CI` (the `ci` job at `ci.yml:1041-1046`) is still pending on `test`.

**The new gate ran and passed.** `.github/workflows/ci.yml` is unchanged since `13eb7623`; the step sits at `ci.yml:61-67` in the `scope` job, and on this head:

```
Detect CI scope    completed    success
```

`ci` (the required `CI` context) lists `scope` first in its `needs:`, so a red projection blocks a required context.

For the record and without attribution: the **previous** head's `Build + Test (windows-latest)` failure was at step "Test machine stdout sink contract (Windows)", a Rust `assay-cli` test. On **this** head that job has not failed; it is still running. I do not attribute the earlier failure to this diff (no Rust in it; the new step runs on Ubuntu in a different job), and I equally do not call it a flake, since I did not verify that.

---

### NON-CLAIMS — what I did not verify

- **Python version.** Everything ran on **3.14.3**; the `scope` job pins **3.12**. Still load-bearing for L1: I previously probed 10 000 and 30 000 nesting levels inside the 64 KiB index cap without triggering `RecursionError` on 3.14, and I could not test 3.12. Unproven either way.
- I did **not** build the Rust workspace, run the Windows job, or reproduce either the previous head's Windows failure or the early `neutral`.
- I did **not** verify that `ghcr.io/rul1an/pma-v0-repro@sha256:88a5ef…` exists, is pullable, or produced the stored record. Nothing on the page claims it does — and as of this head the page says so explicitly.
- I did **not** verify the record's `capture_sha256`, `pack_sha256`, `source_corpus_digest` or `rendered_set_digest` against any capture; `require_run_record_binds_capture` is not on this projection path, by design.
- I did **not** audit `published_rows.py`, `strict_json.py`, `validate_run_record.py`, `implementations.py` or `artifact_io.py` as changes — unmodified here, read only far enough to judge what this diff inherits.
- I did **not** re-derive the byte-identical projector's behavioral properties from scratch a second time; my basis is the `git diff --exit-code` proof plus the re-run mutations, stated above.
- I did **not** assess whether one published row is the right corpus to launch with. Product call, not a review finding.

---

### Independence declaration

Unchanged and re-checked against this head. I am a fresh agent with no authorship of this change or of its repair. I searched my entire inherited context — project `CLAUDE.md`, the persistent memory index and its files, and my task brief — for prior involvement: `grep -ril` for `public_runs`, `public-runs`, `project reviewed run` and `2592` across the memory directory returns **zero** matches (the single `201` hit is an unrelated substring in `grep-match-is-not-structural-evidence.md`). I have never read `Rul1an/assay-tunnel-experiments#201` and cannot access it. I did not author or edit any normative specification or implementation plan for this change. I did not read, and did not look for, any other review, preflight or design note; I encountered none — the only prior review I relied on is my own on `13eb7623`, and I re-derived every measurement in it rather than carrying a conclusion forward.

---

<details>
<summary><b>Evidence appendix — exact commands</b></summary>

```bash
# Anchors, derived independently
gh api repos/Rul1an/assay/pulls/2592 --jq '{head:.head.sha, base:.base.sha, changed_files, additions, deletions}'
#  head cef9d29a7a0f...  base a63a97c1...  7 files  +1110/-0

git fetch --depth 1 origin cef9d29a7a0fe5a08d3f06d3a5ea4125fce301b1 && git checkout FETCH_HEAD
git fetch --depth 1 origin a63a97c16615c181b65aaf6fdf0eed0fc5270ad0
git diff --name-only a63a97c1 cef9d29a          # 7 files, matches the PR
git diff --numstat 13eb7623 cef9d29a            # 2/0, 2/0, 40/0  -> zero deletions
git diff --exit-code 13eb7623 cef9d29a -- conformance/project_public_runs.py
#  PROJECTOR UNCHANGED

shasum -a 256 conformance/public-runs/9275ac65...ae27
#  9275ac65...ae27 == filename == index record_sha256

git show 13eb7623:conformance/tests/test_public_runs.py | grep -c '    def test_'   # 21
grep -c '    def test_' conformance/tests/test_public_runs.py                        # 24

# Control
python3 -V                                                    # 3.14.3  (CI pins 3.12)
python3 -W error::ResourceWarning conformance/tests/test_public_runs.py   # Ran 24 ... OK
python3 conformance/project_public_runs.py --check ; echo rc=$?           # rc=0
git status --porcelain                                        # empty, between every mutation

# B1
#  registry -> record-derived fallback   => FAIL test_missing_registry_row_fails_unknown
#     "line 450 ... with self.assertRaises(ValueError) as ctx"   (ValueError not raised)
#  registry ignored outright             => FAIL test_missing_registry_row_fails_unknown
python3 -c "import json,hashlib;a=open('conformance/public-runs/9275ac65...ae27','rb').read();\
b=json.dumps(json.loads(a.decode()),indent=2,sort_keys=True).encode()+b'\n';print(a==b)"
#  round-trip byte-identical: True   -> _rebind_record's no-op is a true no-op

# H1  (all four => FAIL test_same_implementation_two_digests_are_kept_and_sorted)
#  lines[:-1] if len>1 | lines[1:] if len>1 | lines[:1] | de-dup by implementation_id
#  two-row reality probe:
#    record A 8102 bytes 9275ac65..., record B 8166 bytes 2c161783..., identical? False
#    both validator PASSES, review_warnings 0 and 1; rendered data rows: 2

# H2  (all three => FAIL test_public_non_claims_name_binding_without_attestation)
#  independence line removed | declared/attested line removed | reversed to "are attested and verified"
diff <(head -10 conformance/IMPLEMENTATIONS.md.in) <(head -10 conformance/IMPLEMENTATIONS.md)
#  IDENTICAL prose
#  staleness both directions => FAIL test_pristine_tree_is_green

# M1
#  record read -> path.read_bytes()  => FAIL test_oversize_digest_named_record_fails_before_parse
#    AssertionError: 'exceeds' not found in 'run record is not readable JSON: ...'
#  cap * 4, loader kept              => FAIL (same test)
#  ordering probe (parse + render instrumented):
#    record size 4194305 (limit 4194304) -> "exceeds the 4194304-byte limit"
#    parse/render calls that happened: NONE

# Regression battery: R1 R4 R5 R6 R7a R8 R10 -> all RED, tree clean after each

# CI wiring
git diff --exit-code 13eb7623 cef9d29a -- .github/workflows/ci.yml   # ci.yml UNCHANGED
sed -n '61,67p'   .github/workflows/ci.yml    # "Public-run projection" in job `scope`
sed -n '1041,1046p' .github/workflows/ci.yml  # ci: / name: CI / needs: [scope, ...]
gh api repos/Rul1an/assay/branches/main/protection \
  --jq '{strict:.required_status_checks.strict, contexts:.required_status_checks.contexts}'
#  {"contexts":["CI","host-capability-check","lane-check/proof"],"strict":true}

# CI, final read
gh api repos/Rul1an/assay/commits/cef9d29a.../check-runs --paginate \
  --jq '.check_runs[] | .conclusion // .status' | sort | uniq -c
#  1 in_progress   14 skipped   51 success
gh api "repos/Rul1an/assay/actions/runs?head_sha=cef9d29a..." \
  --jq '.workflow_runs[] | "\(.name)\tattempt=\(.run_attempt)\t\(.status)\t\(.conclusion // "-")"'
#  every run attempt=1; CI in_progress; all others success
#  Detect CI scope -> completed success      <- the job carrying the new gate

# Head re-read immediately before posting
gh api repos/Rul1an/assay/pulls/2592 --jq '.head.sha'
#  cef9d29a7a0fe5a08d3f06d3a5ea4125fce301b1   (unchanged)
```

</details>

PR #2597, opened 2026-08-24
fix(describe): fail closed on run output ownership

### LINKED ISSUE #2178: cli(describe): make omitted command-document ownership fail closed

Rul1an:
## Summary

- publish the two shipping `assay run` JSON identities through `assay describe run`
- bind those identities directly to `assay-core` constants, without copied schema strings
- add an independent CI owner pin so deleting a binding cannot make runtime output and its checker drift empty together
- preserve the existing doctor-listing omission control

Closes #2178.

## Contract

`assay describe run` now lists exactly:

- `assay.run_report.v1`
- `assay.run_summary.v1`

This changes introspection output only. It does not change either run-output schema or validate emitted run documents.

## RED -> GREEN

RED on base `e523f793d10cb965c21b728f6c4af9260052ace5`:

```text
left: []
right: ["assay.run_report.v1", "assay.run_summary.v1"]
```

A second RED removed both required owner bindings in a scratch repo; the old checker stayed green.

GREEN at `7144771531aed5a1139d5751157651ad427feafb`:

```text
cargo test -p assay-cli --test cli_describe_contract
19 passed; 1 ignored

bash scripts/ci/test-cli-describe-contract.sh
identity omission old_guard_exit=0 new_guard_exit=1
ok: cli describe contract
```

## Mutation Evidence

- remove only `RUN_REPORT_SCHEMA` from the run owner bindings -> new guard rejects and names the missing owner contract
- remove only `SUMMARY_SCHEMA` -> same rejection
- remove the doctor identity from a listing while its binding remains -> old guard accepts, new guard rejects

The run-listing mutants are generated from the mutated binding rows, matching the false-clean path this change closes.

## Verification

- `cargo test -p assay-cli --test cli_describe_contract -- --nocapture`
- `cargo test -p assay-cli describe:: -- --nocapture`
- `bash scripts/ci/test-cli-describe-contract.sh`
- `shellcheck scripts/ci/test-cli-describe-contract.sh`
- `python3 -m py_compile scripts/ci/cli_describe_contract.py`
- `cargo clippy -p assay-cli --all-targets -- -D warnings`
- `git diff --check`

## Non-claims

- does not validate the contents of `assay run` output
- does not enumerate or independently pin every CLI output identity
- does not change run/summary schemas, exit codes, or serialization

## Summary by CodeRabbit

* **New Features**
  * The `assay describe run` command now publishes both run-report and summary identities.

* **Bug Fixes**
  * Improved validation ensures required report and command ownership identities are present and correctly mapped.

* **Tests**
  * Added contract coverage for the run description and its published identities.
  * Added CI checks to detect missing report identity bindings.

### DIFF
--- a/crates/assay-cli/src/cli/commands/describe/bindings.rs
+++ b/crates/assay-cli/src/cli/commands/describe/bindings.rs
@@ -5,6 +5,8 @@ use crate::cli::commands::evidence::schema::{
 use crate::cli::commands::init_report::INIT_REPORT_SCHEMA;
 use crate::cli::commands::validate::VALIDATE_REPORT_SCHEMA;
 use crate::diagnostics::report::DOCTOR_REPORT_SCHEMA;
+use assay_core::report::json::RUN_REPORT_SCHEMA;
+use assay_core::report::summary::SUMMARY_SCHEMA;
 
 /// One shipping identity owned by a clap path. The identity field is the
 /// existing constant, not a second string.
@@ -26,6 +28,14 @@ pub(super) const BINDING_ROWS: &[IdentityBinding] = &[
     IdentityBinding {
         path: "init",
         identity: INIT_REPORT_SCHEMA,
+    },
+    IdentityBinding {
+        path: "run",
+        identity: RUN_REPORT_SCHEMA,
+    },
+    IdentityBinding {
+        path: "run",
+        identity: SUMMARY_SCHEMA,
     },
     IdentityBinding {
         path: "validate",
--- a/crates/assay-cli/tests/cli_describe_contract.rs
+++ b/crates/assay-cli/tests/cli_describe_contract.rs
@@ -8,6 +8,8 @@
 #[allow(dead_code)]
 mod bounded_process;
 
+use assay_core::report::json::RUN_REPORT_SCHEMA;
+use assay_core::report::summary::SUMMARY_SCHEMA;
 use bounded_process::{run_bounded, GOLDEN_PATH_LIMITS};
 use serde_json::Value;
 use std::fs;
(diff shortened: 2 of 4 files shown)


Rul1an:
## Independent exact-head review record

- **Head:** `7144771531aed5a1139d5751157651ad427feafb`
- **Reviewer:** Codex subagent `Curie` (`01a033be-1417-7171-9a82-1c967d346310`)
- **Verdict:** **READY**
- **Findings:** no actionable findings
- **Disposition:** none required
- **Independence:** reviewer authored neither the change nor a governing specification or plan; review and mutations were read-only in a fresh detached clone and separate scratch copy.

Verified:

- `assay describe run` emits exactly `assay.run_report.v1` and `assay.run_summary.v1` from `assay-core` constants;
- the checker independently pins both required run owners;
- removing either binding makes the runtime test fail by value mismatch and makes the checker name the missing identity;
- neutralizing or bypassing `REQUIRED_COMMAND_OWNERS` makes the shell mutation guard fail because the run-binding mutant stays green;
- the existing doctor omission control remains discriminating (`old_guard_exit=0`, `new_guard_exit=1`);
- full `cargo test -p assay-cli`, focused schema/summary/exit-code suites, fmt, clippy `-D warnings`, and `git diff --check` pass;
- delegated proof run `32727162270` succeeded against this exact SHA.

All accepted mutation failures were semantic diagnostics, not compile/import/parse failures. This review becomes void after any new push. It does not authorize merge while a required check remains incomplete.

PR #2603, opened 2026-08-24
fix(monitor): bind artifacts to one invocation id

### LINKED ISSUE #2342: Fix monitor artifact run_id so it proves same-invocation pairing
## Measured defect

On current `main` `6d4b78c9993ae0400533c8a6569486b0f2e54b9e`, `assay monitor` derives the `run_id` shared by `observed_peers` and `observation_health` only from attached/skipped probe names, aggregate ring-buffer drops, and whether a policy was declared.

Two distinct monitor invocations with the same posture therefore receive the same `run_id`, even when their observations differ. Both artifact writers independently recompute that posture digest. A consumer checking only equal ids can cross-pair artifacts from different invocations and accept them as one run.

## Chosen contract

Use **invocation identity**. Generate one unpredictable id once at monitor startup and pass it to every requested artifact. Writers receive the id; they do not derive or regenerate it. Finalized content identity is deliberately not part of this slice because it adds circular finalization without improving the same-invocation claim.

## Definition of done

- [ ] RED: two invocations with identical posture receive different non-empty ids.
- [ ] RED: every artifact requested by one invocation carries exactly the same id.
- [ ] RED: crossing peers from invocation A with health from B is refused as `no_coverage` by the production pairing funnel.
- [ ] One function/call site owns id generation; writers only receive it.
- [ ] Failure paths cannot emit a mixed old/new identity set.
- [ ] Public docs describe invocation identity and do not call it content identity or cryptographic provenance.
- [ ] Existing consumers and schema compatibility are inventoried before field semantics change.
- [ ] A mutation restoring the posture digest or generating independently per writer makes the pairing test fail.

## Non-claims

Id equality is a correlation handle, not artifact authenticity, cryptographic provenance, exhaustive egress coverage, or proof that files were not copied after production. This does not change probe attachment, enforcement, or `ObservationHealth::is_complete()` semantics.

Rul1an:
## Summary

- replace posture-derived monitor `run_id` with one unpredictable invocation identity
- pass that one identity to both `observed_peers` and `observation_health`
- clear requested stale targets and exclusively reserve both output handles before startup
- reject direct, lexical, symlink, and case-folded aliases to one output target
- document the correlation boundary without authenticity or provenance claims

Closes #2342.

## Contract

- Two invocations with identical attachment/drop/policy posture receive different IDs.
- Every requested observation artifact from one invocation receives exactly one shared ID.
- The production `verify-side-effects` funnel still refuses crossed pairs as `no_coverage`.
- `run_id` remains an opaque non-empty string in both existing `v0` schemas; no field or schema migration is required.

## RED -> GREEN

RED before implementation:

- `distinct_invocations_do_not_share_a_run_id_when_posture_matches`: both values were `monitor_d4235a45b9a4c26c`
- `one_monitor_invocation_shares_one_id_across_observation_artifacts`: peers and health received different random IDs
- `preparing_observation_outputs_removes_both_stale_counterparts`: stale files survived
- `observation_outputs_refuse_aliases_to_the_same_absent_target`: `dir/file` and `dir/detour/../file` were accepted as separate outputs

GREEN at `033962cf4d987a7032101bbc10b43c0e16f5da10`:

```text
cargo test -p assay-cli                                pass
cargo clippy -p assay-cli --all-targets -- -D warnings pass
cargo fmt --all -- --check                             pass
git diff --check                                      pass
pre-push workspace clippy + Linux cross-target gate   pass
```

Focused evidence:

```text
monitor_next unit tests                                47 passed
verify_side_effects_cli                                22 passed
```

The shared-ID test writes both production serializers to disk, deserializes both artifacts, and compares their persisted `run_id` values.

## Mutation proof

Each mutation was applied separately and restored byte-exactly before the next run:

| Mutation | Guard that failed |
|---|---|
| constant/deterministic invocation ID | `distinct_invocations_do_not_share_a_run_id_when_posture_matches` |
| regenerate ID inside the health writer | `one_monitor_invocation_shares_one_id_across_observation_artifacts` |
| skip stale-target removal | `preparing_observation_outputs_removes_both_stale_counterparts` |
| permit both schemas to share one direct path | `observation_outputs_refuse_the_same_target_path` |
| compare lexical paths instead of resolved absent targets | `observation_outputs_refuse_aliases_to_the_same_absent_target` |
| replace exclusive `create_new` reservation with ordinary `create` | `observation_outputs_refuse_case_aliases_on_case_insensitive_filesystems` |
| bypass consumer `run_id` mismatch | `a_peer_set_from_a_different_run_cannot_refute` |

The alias mutation used an explicit one-test control (`1 passed`) before removing the resolved-target comparison; the mutated run failed exactly that test (`1 failed`).

## Failure behavior

Requested stale contents are removed and both final paths are exclusively reserved before policy/eBPF startup. If startup or a later write fails, a prior counterpart cannot survive and pair with a new artifact. This is not a multi-file atomic commit: a failed invocation may leave one new artifact and the counterpart missing, and exits nonzero.

## Non-claims

ID equality is a correlation handle, not artifact authenticity, content identity, cryptographic provenance, exhaustive egress coverage, or proof that files were not copied after production. Probe attachment, enforcement, and `ObservationHealth::is_complete()` are unchanged.


Rul1an:
```json
{
  "head_sha": "033962cf4d987a7032101bbc10b43c0e16f5da10",
  "builder": {"agent": "codex", "instance": "writer-2342"},
  "reviewer": {"agent": "codex", "instance": "pr2603-fresh-033962cf", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [],
  "no_findings": true,
}
```

Issue #2649, opened 2026-08-26
python: align assay-it metadata with published artifact matrix


Rul1an:
## Problem

PyPI `assay-it 5.4.0` declares Python `>=3.9` but publishes only three CPython 3.12 wheels (macOS x86_64/arm64 and manylinux x86_64) and no sdist. `pip install assay-it` is therefore broader in metadata/docs than the installable artifact surface.

## DoD

- [ ] Choose and document one intentional support policy: buildable sdist/ABI strategy plus required wheels, or narrowed `Requires-Python` and platform metadata/docs.
- [ ] Treat metadata narrowing as a compatibility decision and record migration impact.
- [ ] `pip download`/install is tested for every declared Python/platform pair from a clean environment.
- [ ] A declared pair with no artifact fails the release gate; an undeclared pair is reported unsupported, not broken.
- [ ] Wheel import/smoke test pins the Assay version and native-library load behavior.
- [ ] Generated support matrix and PyPI metadata derive from one source.
- [ ] README does not advertise a broader Python path than the published release.

## Non-goals

No promise for every Python minor/platform, no best-effort import fallback, and no coupling to MCP-server distribution.

Parent: #1975

Rul1an:
## Source inventory at `822f5286b34abeb58f2317c392d4a069997e1309`

The Python problem is primarily release metadata/artifact truth, not a missing binding implementation:

- `assay-python-sdk/Cargo.toml` defines the `assay-it` PyO3 package.
- release documentation and guards currently pin the package name but do not enforce an installable Python/platform matrix.
- published `5.4.0` metadata advertises `Requires-Python >=3.9`, while the release oracle measured only CPython 3.12 wheels for macOS x86_64/arm64 and manylinux x86_64, with no sdist.

Before choosing sdist versus more wheels, the writer should measure whether the crate is actually source-buildable from the PyPI payload with all Rust workspace dependencies present. If not, publishing an sdist is a false fix.

One generated support policy must drive maturin build targets, PyPI metadata, clean `pip download/install` tests, and README wording. The RED control is a declared Python/platform pair with no matching wheel or buildable sdist. Metadata narrowing is a compatibility decision and needs explicit migration text.

Rul1an:
## 2026-08-26 packaging decision preflight

Source baseline: Assay `822f5286b34abeb58f2317c392d4a069997e1309`, PyO3 `0.29`, maturin `>=1,<2`.

[...]

PR #2658, opened 2026-08-26
fix(python): declare assay-it as CPython 3.12 on the three published wheels

Rul1an:
## Summary

Sole-writer slice for #2649. `assay-python-sdk/python-artifact-matrix.v0.json` is the sole mutable authority for package, exact CPython line, wheel targets, runner labels, and wheel tags.

A small fail-closed planner reads that matrix and emits only:

- `python` derived from `==X.Y.*`;
- `abi` derived as `cpXY`;
- the declared wheel matrix as compact JSON.

The release workflow consumes those outputs for `setup-python`, maturin, native wheel smoke, and artifact upload. It no longer carries a second Python-version or wheel-matrix truth.

## User-visible correction

- `assay-it` support is CPython 3.12 only on Linux x86_64 and macOS x86_64/arm64.
- `Requires-Python` changes from `>=3.9` to `==3.12.*`.
- This does not remove a working 3.9–3.11 wheel: none existed. It changes the failure from a later install mismatch to an explicit unsupported-interpreter error. The remedy is CPython 3.12; pinning 5.4.0 does not create missing wheels.
- PyPy is no longer advertised without a declared `pp*` artifact.

## Contract and guards

- Every ordinary CPython wheel tag must use the ABI derived from `==X.Y.*`.
- Package identity is bound across the matrix, `pyproject.toml`, release planning, smoke installation, and wheel filename.
- The support sentence is derived from version plus declared platforms, rather than maintained as free prose.
- `kernel-matrix` PR paths include the SDK and every governed install-doc path, including `docs/guides/troubleshooting.md` and `docs/AIcontext/user-flows.md`.
- The pre-commit selector includes the matrix, release/kernel workflows, migration page, governed docs, and owner scripts/tests.
- Each wheel cell rejects zero or extra wheels, installs the exact produced wheel with `--no-index --only-binary`, and imports both `assay` and `assay._native` with the selected interpreter.

## Candidate and verification

Candidate head: `dabd7c793b6c633b9598f97323fdb691f4b7e293`.

Exact-head local verification:

- `scripts/ci/test-check-python-artifact-truth.sh`: green;
- `scripts/ci/test-check-python-wheel-smoke-contract.sh`: green;
- direct planner emits Python 3.12, ABI `cp312`, and exactly three wheel rows;
- exact-head release job uses planner outputs and contains no hardcoded Python 3.12 or `assay-it` wheel authority;
- `git diff --check`: clean.

Required false-green mutations now bite, including:

- coordinated matrix + pyproject move to 3.13 while release stays 3.12;
- coordinated `>=3.9` plus a `cp39` tag;
- package drift to `assay-py` while smoke/metadata remain `assay-it`;
- deleting any governed workflow/pre-commit path;
- widening a support sentence;
- removing or bypassing the production smoke invocation.

Hosted integration checks are running on this exact SHA. The independent exact-head quorum review is still pending; this PR remains draft until both are satisfied.

## Non-claims

- No sdist, abi3, PyPy, free-threaded Python, Python 3.13+, or additional platform support.
- No rewrite of published 5.4.0 artifacts or metadata.
- No guarantee that arbitrary host Python installations can load a wheel.
- Native import of the real macOS wheels remains unproven until the release `wheels` job runs on its declared hosted runners.
- No MCP-server distribution change.

Refs #2649

## Summary by CodeRabbit

* **Compatibility**
  * Python SDK support is now limited to CPython 3.12 on macOS and Linux (x86_64/arm64).
  * PyPy, other Python versions, and other platforms are no longer claimed as supported.
  * Source installation requirements now document the required Rust versions.

* **Packaging**
  * Native wheels are available for three supported macOS and Linux targets.
  * Package installation and wheel contents are validated during releases.

* **Documentation**
  * Installation, migration, troubleshooting, and quickstart guidance now reflect the updated support policy.

### DIFF
--- a/.github/workflows/kernel-matrix.yml
+++ b/.github/workflows/kernel-matrix.yml
@@ -50,6 +50,7 @@ on:
       - "llms.txt"
       - "docs/python-sdk/**"
       - "docs/getting-started/**"
+      - "docs/migration-v1.2.md"
       # S1b coverage-gate hook runs under Lint (pre-commit) in this workflow.
       - ".github/workflows/monitor-attach-smoke.yml"
       # The lane self-test pins guards inside this workflow file; run the
--- a/.github/workflows/kernel-matrix.yml
+++ b/.github/workflows/kernel-matrix.yml
@@ -51,6 +51,8 @@ on:
       - "docs/python-sdk/**"
       - "docs/getting-started/**"
       - "docs/migration-v1.2.md"
+      - "docs/guides/troubleshooting.md"
+      - "docs/AIcontext/user-flows.md"
       # S1b coverage-gate hook runs under Lint (pre-commit) in this workflow.
       - ".github/workflows/monitor-attach-smoke.yml"
       # The lane self-test pins guards inside this workflow file; run the
--- a/.github/workflows/kernel-matrix.yml
+++ b/.github/workflows/kernel-matrix.yml
@@ -44,6 +44,12 @@ on:
       - "docs/contributing/REFACTOR-WAVE-STATUS.md"
       - ".github/rulesets/main-required-ci-contexts.json"
       - ".github/workflows/kernel-matrix.yml"
+      # python-artifact-truth hook runs under Lint (pre-commit) here. scripts/**
+      # covers a guard change; a metadata/docs-only PR never starts this workflow.
+      - "assay-python-sdk/**"
+      - "llms.txt"
+      - "docs/python-sdk/**"
+      - "docs/getting-started/**"
       # S1b coverage-gate hook runs under Lint (pre-commit) in this workflow.
       - ".github/workflows/monitor-attach-smoke.yml"
       # The lane self-test pins guards inside this workflow file; run the
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -809,23 +809,38 @@ jobs:
           bash scripts/ci/record-mcp-registry-result.sh
 
   # ============================================================
+  # Plan Python wheels from the artifact matrix (sole authority)
+  # ============================================================
+  plan-python-artifact:
+    name: Plan Python artifact matrix
+    needs: release-contract
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      python: ${{ steps.plan.outputs.python }}
+      abi: ${{ steps.plan.outputs.abi }}
+      wheels: ${{ steps.plan.outputs.wheels }}
+    steps:
+      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
+        with:
+          persist-credentials: false
+      - name: Plan wheels from the artifact matrix
+        id: plan
+        run: python3 scripts/ci/plan-python-artifact-matrix.py >> "$GITHUB_OUTPUT"
+
+  # ============================================================
   # Build and Publish Python Wheels (PyPI)
   # ============================================================
   wheels:
     name: Build Wheels
-    needs: release-contract
+    needs: [release-contract, plan-python-artifact]
     runs-on: ${{ matrix.os }}
     permissions:
       contents: read
     strategy:
       matrix:
-        include:
-          - os: ubuntu-latest
-            target: x86_64-unknown-linux-gnu
-          - os: macos-15-intel
-            target: x86_64-apple-darwin
-          - os: macos-15
-            target: aarch64-apple-darwin
+        include: ${{ fromJSON(needs.plan-python-artifact.outputs.wheels) }}
 
     steps:
       - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
@@ -833,17 +848,16 @@ jobs:
           persist-credentials: false
       - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
-          python-version: '3.12'
+          python-version: ${{ needs.plan-python-artifact.outputs.python }}
 
       - name: Build wheels
         uses: PyO3/maturin-action@e83996d129638aa358a18fbd1dfb82f0b0fb5d3b # v1
         with:
           working-directory: assay-python-sdk
           target: ${{ matrix.target }}
-          # Keep the interpreter explicit for all wheel targets. This is
-          # required inside the manylinux container and harmless on macOS,
-          # where setup-python provides the same Python version.
-          args: --release --out dist --locked -i python3.12 --compatibility pypi
+          # Interpreter comes from the artifact-matrix plan (==X.Y.* -> X.Y).
+          # Required inside the manylinux container; setup-python provides it on macOS.
+          args: --release --out dist --locked -i python${{ needs.plan-python-artifact.outputs.python }} --compatibility pypi
           sccache: 'false'
           manylinux: auto
 
@@ -851,7 +865,7 @@ jobs:
         shell: bash
         env:
           ASSAY_WHEEL_TARGET: ${{ matrix.target }}
-        run: python3 scripts/ci/smoke-python-wheel.py --dist-dir assay-python-sdk/dist
+        run: python3 scripts/ci/smoke-python-wheel.py --dist-dir assay-python-sdk/dist --python "python${{ needs.plan-python-artifact.outputs.python }}"
 
       - name: Upload wheels
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -822,7 +822,7 @@ jobs:
         include:
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
-          - os: macos-15
+          - os: macos-15-intel
             target: x86_64-apple-darwin
           - os: macos-15
             target: aarch64-apple-darwin
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -847,6 +847,12 @@ jobs:
           sccache: 'false'
           manylinux: auto
 
+      - name: Smoke the produced wheel
+        shell: bash
+        env:
+          ASSAY_WHEEL_TARGET: ${{ matrix.target }}
+        run: python3 scripts/ci/smoke-python-wheel.py --dist-dir assay-python-sdk/dist
+
       - name: Upload wheels
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -381,7 +381,7 @@ repos:
         entry: bash -c 'bash scripts/ci/test-check-python-artifact-truth.sh && python3 scripts/ci/check-python-artifact-truth.py && bash scripts/ci/test-check-python-wheel-smoke-contract.sh && python3 scripts/ci/check-python-wheel-smoke-contract.py'
         language: system
         pass_filenames: false
-        files: ^(assay-python-sdk/(python-artifact-matrix\.v0\.json|pyproject\.toml|Cargo\.toml|README\.md)|scripts/ci/(check-python-artifact-truth\.py|test-check-python-artifact-truth\.sh|check-python-wheel-smoke-contract\.py|test-check-python-wheel-smoke-contract\.sh|smoke-python-wheel\.py)|\.github/workflows/release\.yml|docs/(python-sdk/index\.md|getting-started/(index|installation|python-quickstart)\.md|guides/troubleshooting\.md|AIcontext/user-flows\.md)|llms\.txt|CHANGELOG\.md|\.pre-commit-config\.yaml)$
+        files: ^(assay-python-sdk/(python-artifact-matrix\.v0\.json|pyproject\.toml|Cargo\.toml|README\.md)|scripts/ci/(check-python-artifact-truth\.py|test-check-python-artifact-truth\.sh|check-python-wheel-smoke-contract\.py|test-check-python-wheel-smoke-contract\.sh|smoke-python-wheel\.py|plan-python-artifact-matrix\.py)|\.github/workflows/(release|kernel-matrix)\.yml|docs/(python-sdk/index\.md|getting-started/(index|installation|python-quickstart)\.md|guides/troubleshooting\.md|AIcontext/user-flows\.md|migration-v1\.2\.md)|llms\.txt|CHANGELOG\.md|\.pre-commit-config\.yaml)$
 
       - id: osv-scanner-scheduled-lockstep
         name: OSV scanner/reporter lockstep and count bind
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -363,7 +363,7 @@ repos:
         language: system
         pass_filenames: false
         stages: [pre-push]
-        files: ^scripts/ci/(lib/clear-git-repository-env|test-(check-release-surface|git-fixture-env-isolation|perf-pr-event-provenance-contract))\.sh$
+        files: ^scripts/ci/(lib/clear-git-repository-env|test-(check-release-surface|check-python-artifact-truth|git-fixture-env-isolation|perf-pr-event-provenance-contract))\.sh$
 
       - id: release-surface
         name: Release surface matches source and published versions
@@ -374,6 +374,14 @@ repos:
         # no `target/` build it still verifies source and published-release truth independently.
         files: ^(scripts/ci/((check|test-check)-release-surface|read-assay-release-tag)\.sh|\.github/assay-release-tag|Cargo\.toml|README\.md|llms\.txt|docs/(index\.md|getting-started/(index|installation|quickstart|ci-integration)\.md|reference/cli/index\.md|python-sdk/index\.md|use-cases/(air-gapped|ci-gate)\.md|AIcontext/user-flows\.md|migration-v1\.2\.md|guides/editor-mcp-recipe\.md)|\.pre-commit-config\.yaml)$
 
+
+
+      - id: python-artifact-truth
+        name: assay-it metadata matches the published wheel matrix
+        entry: bash -c 'bash scripts/ci/test-check-python-artifact-truth.sh && python3 scripts/ci/check-python-artifact-truth.py'
+        language: system
+        pass_filenames: false
+        files: ^(assay-python-sdk/(python-artifact-matrix\.v0\.json|pyproject\.toml|Cargo\.toml|README\.md)|scripts/ci/(check-python-artifact-truth\.py|test-check-python-artifact-truth\.sh)|\.github/workflows/release\.yml|docs/(python-sdk/index\.md|getting-started/(index|installation|python-quickstart)\.md|guides/troubleshooting\.md|AIcontext/user-flows\.md)|llms\.txt|CHANGELOG\.md|\.pre-commit-config\.yaml)$
 
       - id: osv-scanner-scheduled-lockstep
         name: OSV scanner/reporter lockstep and count bind
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -363,7 +363,7 @@ repos:
         language: system
         pass_filenames: false
         stages: [pre-push]
-        files: ^scripts/ci/(lib/clear-git-repository-env|test-(check-release-surface|check-python-artifact-truth|git-fixture-env-isolation|perf-pr-event-provenance-contract))\.sh$
+        files: ^scripts/ci/(lib/clear-git-repository-env|test-(check-release-surface|check-python-artifact-truth|check-python-wheel-smoke-contract|git-fixture-env-isolation|perf-pr-event-provenance-contract))\.sh$
 
       - id: release-surface
         name: Release surface matches source and published versions
@@ -378,10 +378,10 @@ repos:
 
       - id: python-artifact-truth
         name: assay-it metadata matches the published wheel matrix
-        entry: bash -c 'bash scripts/ci/test-check-python-artifact-truth.sh && python3 scripts/ci/check-python-artifact-truth.py'
+        entry: bash -c 'bash scripts/ci/test-check-python-artifact-truth.sh && python3 scripts/ci/check-python-artifact-truth.py && bash scripts/ci/test-check-python-wheel-smoke-contract.sh && python3 scripts/ci/check-python-wheel-smoke-contract.py'
         language: system
         pass_filenames: false
-        files: ^(assay-python-sdk/(python-artifact-matrix\.v0\.json|pyproject\.toml|Cargo\.toml|README\.md)|scripts/ci/(check-python-artifact-truth\.py|test-check-python-artifact-truth\.sh)|\.github/workflows/release\.yml|docs/(python-sdk/index\.md|getting-started/(index|installation|python-quickstart)\.md|guides/troubleshooting\.md|AIcontext/user-flows\.md)|llms\.txt|CHANGELOG\.md|\.pre-commit-config\.yaml)$
+        files: ^(assay-python-sdk/(python-artifact-matrix\.v0\.json|pyproject\.toml|Cargo\.toml|README\.md)|scripts/ci/(check-python-artifact-truth\.py|test-check-python-artifact-truth\.sh|check-python-wheel-smoke-contract\.py|test-check-python-wheel-smoke-contract\.sh|smoke-python-wheel\.py)|\.github/workflows/release\.yml|docs/(python-sdk/index\.md|getting-started/(index|installation|python-quickstart)\.md|guides/troubleshooting\.md|AIcontext/user-flows\.md)|llms\.txt|CHANGELOG\.md|\.pre-commit-config\.yaml)$
 
       - id: osv-scanner-scheduled-lockstep
         name: OSV scanner/reporter lockstep and count bind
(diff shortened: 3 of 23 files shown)


Rul1an:
## Independent quorum review — head `142e877d35ef334bdd8d826c0c677040da9dee81`

**Independence.** I authored or edited neither #2649's normative design/plan nor this implementation. Freshly spawned, read-only, no prior context on either artifact. Reviewed exactly `142e877d35ef334bdd8d826c0c677040da9dee81` against base `c1cd7e600ff34f84e03cb7503687adb340d5ebf8`, in a detached scratch clone. Head re-read at start and immediately before this comment: unchanged.

### What I verified by reproduction

Control on the unmutated head tree: `check-python-artifact-truth.py` PASS, `check-python-wheel-smoke-contract.py` PASS, both self-test suites ALL GREEN, and the pre-commit hook's exact `entry` string rc=0. Re-run green after every revert; tree byte-identical (`git status --porcelain` empty).

1. **Exactly three wheels — confirmed.** `python-artifact-matrix.v0.json:14-38` declares three rows; `release.yml:822-828` carries three `include` pairs; parity is enforced in both directions. Ground truth from the PyPI JSON API for `assay-it 5.4.0`: exactly three wheels, no sdist, and the filenames match the three declared `tag` values character-for-character (`macosx_10_12_x86_64`, `macosx_11_0_arm64`, `manylinux_2_17_x86_64.manylinux2014_x86_64`). The declared tags are anchored to what actually shipped.
2. **Metadata truth — confirmed fixed.** `pyproject.toml:9` is now `==3.12.*`; `:18` adds `Python :: 3.12`; the PyPy classifier is gone. PyPI confirms the published 5.4.0 state was exactly the defect #2649 describes: `requires_python '>=3.9'` plus a PyPy classifier over cp312-only wheels. See blocker B1 for what still is not anchored.
3. **Producer finds exactly one wheel — partially confirmed.** Zero wheels → rc=1; only a wrong-named wheel → rc=1; unknown target → rc=1; no target → rc=1. Two *matching* wheels is unreachable by construction, because `find_wheel` (`smoke-python-wheel.py:37-45`) compares exact filenames rather than globbing — stronger than the glob the DoD contemplates. But see F2: two wheels *present* is not rejected.
4. **Native clean-install + import per target — confirmed structurally, not by execution.** The smoke step (`release.yml:850-854`) sits inside the `wheels` job, which is `runs-on: ${{ matrix.os }}` (`:817`), so build and smoke are the same job and matrix parity is by construction, plus parity-checked against the JSON. Not exercised: the `wheels` job triggers only on `push: tags: v*` / `workflow_dispatch` (`release.yml:18-22`), so it has never run on this PR. The PR body states this caveat honestly.
5. **Runner mapping — confirmed from actual `runs-on`.** `release.yml:825-826` maps `x86_64-apple-darwin` → `macos-15-intel`, `:827-828` maps `aarch64-apple-darwin` → `macos-15`. This is the strongest part of the change: swapping the x86 row to `macos-15` in `release.yml` alone bites on parity, and swapping it in *both* files still bites, because `check-python-wheel-smoke-contract.py:20-24` holds the mapping as an independent literal. The runner claim is not merely self-consistent.
6. **`assay._native` must load — confirmed load-bearing, with a true positive control.** I compiled a real trivial CPython extension exporting `PyInit__native` and built a valid wheel around it: smoke rc=0, printing `import_smoke=native version=5.4.0`. Replacing that `.so` with non-Mach-O bytes: rc=1, `ImportError: dlopen(...) slice is not valid mach-o file`, raised from the `import ... assay._native` probe at `smoke-python-wheel.py:83`. A valid wheel with no native member: rc=1 at the `native_members` guard (`:110`). Both layers bite, and the green is a real green rather than an absence.
7. **No sdist / abi3 / PyPy / extra-platform claim — confirmed.** `publish_sdist: false`; the wheels job rejects `*.tar.gz`/`sdist` and requires `dist/*.whl`; abi3 is rejected in `pyproject.toml`, `Cargo.toml`, and docs; an undeclared fourth platform bites in both directions.

**No false-green.** Zero occurrences of `continue-on-error`, `if:`, `always()`, `|| true`, or `set +e` in the wheels job (`release.yml:814-861`). `scripts/ci/check-workflow-continue-on-error.py` passes unchanged (4 advisory workflows, 9 named steps, each with a stated reason) — this change adds no exception and needs no allowlist marker. The new scripts contain no bare excepts and no unconditional early return; `return 0` appears only as the terminal path after the error list is checked. Fail-closed confirmed on a missing interpreter via both `PYTHON_BIN` and `--python` (rc=1, `native import_smoke requires python3.99`) — a runner without `python3.12` fails rather than silently skipping.

### Mutation matrix

| # | Mutation | Expected | Observed | Verdict |
|---|---|---|---|---|
| M1 | Drop `x86_64-apple-darwin` include from `release.yml` | RED | both checkers rc=1, target+os parity | bites |
| M2 | Drop `aarch64` row from matrix JSON only | RED | both rc=1 | bites |
| M3 | x86 row `macos-15-intel`→`macos-15` in `release.yml` only | RED | both rc=1 | bites |
| M4 | Same swap in **both** files (coordinated) | RED | truth rc=0, smoke-contract rc=1 via `EXPECTED_OS` | bites |
| M5 | `requires-python` → `>=3.9` in `pyproject.toml` only | RED | truth rc=1 | bites |
| **M6** | **`requires_python` → `>=3.9` in pyproject *and* matrix** | **RED** | **both rc=0 — PASS** | **does not bite (B1)** |
| M7 | Re-add PyPy classifier to `pyproject.toml` only | RED | truth rc=1 | bites |
| **M8** | **PyPy classifier + empty `forbidden_classifiers`** | **RED** | **both rc=0 — PASS** | **does not bite (B1)** |
| M9 | Add 4th declared wheel to matrix only | RED | both rc=1 | bites |
| M10 | Add 4th wheel to matrix *and* `release.yml` | RED | both rc=1, `unexpected declared target` | bites |
| M11 | "Python 3.9+" claim into an install doc | RED | truth rc=1 | bites |
| M12 | sdist claim into an install doc | RED | truth rc=1 | bites |
| M13 | Remove the support bound from `README.md` | RED | truth rc=1 | bites |
| P0 | dist dir empty | RED | rc=1 | bites |
| **P2** | **expected wheel + a second differently-named wheel** | **RED** | **rc=0 — PASS** | **does not bite (F2)** |
| P3 | only a wrong-tag wheel present | RED | rc=1 | bites |
| P4 | valid wheel, zero `assay._native` members | RED | rc=1 | bites |
| P5 | valid wheel, two `assay._native` members | RED | rc=1 | bites |
| P7 | valid wheel, `assay._native` is not a loadable binary | RED | rc=1, ImportError | bites |
| PC | valid wheel, real loadable `assay._native` | GREEN | rc=0, `import_smoke=native` | positive control |
| P9 | interpreter absent (env var and flag) | RED | rc=1 both ways | bites |
| P10 | no target supplied | RED | rc=1 | bites |

Control green before and after: yes, verified byte-identical after every cell. One earlier batch of M5–M10 was voided when my revert helper failed under zsh word-splitting and mutations accumulated; I restored from file-copy backups, re-verified the control, and re-ran that whole batch cleanly. The table reports only the clean run.

### Blockers

**B1 — the "one source" is itself unanchored, so #2649's exact defect is reachable in two lines.** *(spec-gap / missing guard)*
`check-python-artifact-truth.py:51-64` pins `pyproject.toml` to the matrix JSON, but nothing pins the matrix JSON's `requires_python` (`:4`) or `forbidden_classifiers` (`:10-12`) to anything. The build side *is* independently anchored — `:91-94` requires the literal `python-version: '3.12'` and `-i python3.12` in the wheels job, and the declared `tag` values are cp312-only and verified against the real artifact at release time. So M6 produces precisely the state #2649 was filed for — metadata declaring `>=3.9` over a build pinned to cp312 — and both gates report PASS. M8 does the same for the PyPy claim. The DoD item "Generated support matrix and PyPI metadata derive from one source" is met in form; what is missing is that the source cannot be made to lie without something objecting.
*Smallest fix:* in `check-python-artifact-truth.py`, derive the expected interpreter from the declared tags — assert every `wheels[].tag` starts with the CPython tag implied by `requires_python` (`==3.12.*` ⇒ `cp312-`), and require the PyPy classifier to stay in `forbidden_classifiers` unless a `pp*` tag is declared. That reuses the anchor the tags already carry.

**B2 — the new guard may not run in hosted CI on the PR that would trip it.** *(missing guard)*
The hook at `.pre-commit-config.yaml:379-385` declares no `stages:`, so it runs at the default pre-commit stage and *is* picked up by `pre-commit run --all-files` at `kernel-matrix.yml:129`. But `kernel-matrix.yml`'s `pull_request.paths` contains no match for `assay-python-sdk/**`, `llms.txt`, `docs/python-sdk/**`, or `docs/getting-started/**` — I grepped the file and got zero hits. The hook's `files:` surface is therefore strictly broader than the workflow's trigger, and a PR touching only `assay-python-sdk/pyproject.toml` — the very file B1's mutation edits — would not start the `Lint (pre-commit)` job at all. The gap closes only on the push to `main`, where `kernel-matrix` has no paths filter, i.e. after merge. The repo already documents this class of gap in its own `kernel-matrix.yml` comments for other hooks. The pattern is pre-existing (`release-surface` has the same shape), but this PR adds a new guard over new surfaces without extending the trigger.
*Smallest fix:* add `assay-python-sdk/**`, `llms.txt`, `docs/python-sdk/**`, and `docs/getting-started/**` to `kernel-matrix.yml`'s `pull_request.paths`.

### Non-blocking findings

**F2 — an unexpected extra wheel is published unsmoked.** *(missing guard, low-medium)* `find_wheel` asserts exactly one wheel *matching the expected name*, not exactly one wheel in `dist/`. P2 confirms a second, differently-named wheel is tolerated (rc=0). `Upload wheels` (`release.yml:856-860`) then uploads `dist/*.whl` — everything present — and `publish-pypi` pushes all of it. Low likelihood (fresh checkout, one maturin invocation per job), but it is the mirror image of the DoD's "an undeclared pair is reported unsupported, not broken". *Fix:* assert `sorted(dist.glob("*.whl")) == [wheel]`, or narrow the upload `path:` to the exact expected filename.

**F3 — a test section header claims coverage the case does not have.** *(test hygiene, low)* `test-check-python-wheel-smoke-contract.sh:84` is headed `=== mutation: two matching wheels ===`, but lines 88-89 and 101-104 are shipped stream-of-consciousness deliberation concluding the case is impossible, and the case degenerates to a renamed-wheel test asserting `renamed produced wheel`. The assertion matches what it actually does, so the output is not false — the header overstates it. Line 99 also references a "skip-path PASS" that no longer exists. *Fix:* rename the header to what it tests and delete the abandoned-reasoning comments.

**F4 — one rule written twice.** *(low)* `check-python-wheel-smoke-contract.py:99-100` special-cases `x86_64-apple-darwin` + `macos-15`, but `:94-98` already rejects any `os != EXPECTED_OS` for that target. M4 printed both messages for the single condition, confirming `:99-100` can never be the sole detector. `import_smoke` validation likewise exists in both checkers (`check-python-artifact-truth.py:169-175` and `check-python-wheel-smoke-contract.py:86-93`). Per AGENTS.md "One rule, one function". *Fix:* delete `:99-100`.

**F5 — release-time risk on the newly native x86 macOS cell.** *(risk note, not a defect)* That cell moves from a cross-compile on `macos-15` to a native build on `macos-15-intel`. The declared tag `cp312-cp312-macosx_10_12_x86_64` matches what the published cross-built 5.4.0 wheel carries (verified against PyPI), but a native build's `MACOSX_DEPLOYMENT_TARGET` may differ, in which case `find_wheel` fails the release job. This fails closed, which is the correct direction — flagging it so the first tagged release after merge is watched rather than surprised.

### Explicit non-claims

- **No hosted CI counted.** On `142e877d`, the check-runs API reports `Lint (pre-commit)` — the only hosted job that executes this guard — as **cancelled**, along with `Detect touched crates`, `Check eBPF Changes`, `Kernel Matrix Summary`, `local-links`, `mkdocs-strict`, and `review-record-check`; `assay` is queued. Cancelled and pending are not passes. Everything above rests on my own local reproduction against the exact head tree.
- **No hosted macOS result exists on this SHA**, for either macOS cell. I did not build a real PyO3 wheel and did not run maturin; my `_native` positive control was a hand-compiled trivial C extension, which proves the smoke's import assertion is load-bearing but says nothing about the real extension.
- I did not verify that `macos-15-intel` is a currently valid runner label, nor that manylinux/macOS tags emitted by maturin on those images match the declared tags. Claims 1 and 2 are anchored to the *published* 5.4.0 artifacts, not to a future build.
- Scratch clone left in place and idle; no writes outside it.

### Verdict

**BLOCKED** at `142e877d35ef334bdd8d826c0c677040da9dee81`, on B1 and B2. The change is a real and well-built improvement — the metadata now matches the three artifacts that actually shipped, the runner mapping is correct and independently anchored, and the native import is genuinely load-bearing with a verified positive control. It is blocked because the guard does not yet defend against the specific regression it exists to prevent: B1 lets #2649's exact defect back in through a two-line edit with both gates green, and B2 means the PR making that edit would not start the job that checks it. Both fixes are small and local.

```json
{
  "record": "review.decision.v0",
  "pr": 2658,
  "reviewed_head": "142e877d35ef334bdd8d826c0c677040da9dee81",
  "base": "c1cd7e600ff34f84e03cb7503687adb340d5ebf8",
  "decision": "BLOCKED",
  "blockers": [
    "B1: matrix JSON requires_python/forbidden_classifiers are not anchored to the declared cp312 wheel tags; coordinated widening (M6) and a PyPy re-claim (M8) both pass all gates, reproducing issue #2649's exact defect",
    "B2: kernel-matrix.yml pull_request.paths does not cover assay-python-sdk/**, llms.txt, docs/python-sdk/** or docs/getting-started/**, so the new python-artifact-truth hook does not run on a PR touching only those guarded surfaces"
  ],
  "notes": [
    "F2: find_wheel asserts one wheel matching the expected name, not one wheel in dist; an extra differently-named wheel is tolerated and then published by the dist/*.whl upload glob",
    "F3: test-check-python-wheel-smoke-contract.sh:84 header claims 'two matching wheels' but the case tests a renamed wheel; abandoned deliberation comments shipped at lines 88-89 and 101-104",
    "F4: check-python-wheel-smoke-contract.py:99-100 duplicates the EXPECTED_OS check at :94-98 and can never fire alone; import_smoke validated in both checkers",
    "F5: x86 macOS cell moves from cross-compile to native build; declared macosx_10_12 tag matches published 5.4.0 but a native MACOSX_DEPLOYMENT_TARGET could differ and fail find_wheel at release time (fails closed)",
    "Confirmed: three wheels, no sdist, and declared tags match the real PyPI 5.4.0 filenames exactly",
    "Confirmed: assay._native import is load-bearing, verified with a compiled positive control (rc=0 green, rc=1 ImportError when the binary is corrupted)",
    "Confirmed: no continue-on-error / always() / || true / set +e in the wheels job; check-workflow-continue-on-error.py passes with no new exception",
    "Confirmed: smoke fails closed on a missing interpreter, a missing target, and an unknown target"
  ],
  "reviewer_role": "independent_quorum_reviewer",
  "authored_governing_plan": false,
  "authored_change": false,
  "mutations_run": 22,
  "mutations_biting": 19,
  "control_green_before_and_after": true,
  "hosted_macos_on_this_sha": false,
  "hosted_ci_counted": false
}
```

Rul1an:

**Independence.** I did not author or edit issue #2649's normative plan, and I did not author or edit this implementation or the PR branch. I did not read this PR's comments or reviews (only `--json title,body,headRefOid,baseRefOid,isDraft,state`). Reviewed in a fresh detached clone of `https://github.com/Rul1an/assay.git` at `dabd7c793b6c633b9598f97323fdb691f4b7e293`, base `f5a8418ebade048f958814cc40ba73fd11a997cf`, whole `base..head` range. Head re-read at start and immediately before disposition: unchanged both times.

Local toolchain differs from CI where noted: Python 3.14.3 (CI Lint uses 3.12), bash 5.3.15, pre-commit 4.6.0 (CI pins 4.4.0). `df -h /`: 27Gi available.

### Verdict

**READY** at `dabd7c793b6c633b9598f97323fdb691f4b7e293`, with four non-blocking findings below.

### What I confirmed by making it red

Gate under test is the hook entry itself: `test-check-python-artifact-truth.sh && check-python-artifact-truth.py && test-check-python-wheel-smoke-contract.sh && check-python-wheel-smoke-contract.py`. No-op control green first and last in every batch; restores by file copy verified with `diff -q` plus an empty `git status --porcelain`.

| # | Mutation at the production funnel | Result |
|---|---|---|
| M1 | `release.yml` wheels job: `python-version` back to literal `'3.12'` | RED |
| M2 | `release.yml`: delete the `Smoke the produced wheel` step | RED |
| M3 | `release.yml`: hardcode the wheels `include:` list again | RED |
| M4 | `pyproject.toml`: `requires-python = ">=3.9"` | RED |
| M5 | `smoke-python-wheel.py:139`: delete `install_and_import(args.python, wheel, version)` | RED |
| M6 | `kernel-matrix.yml`: drop `docs/getting-started/**` from `pull_request.paths` | RED |
| M7 | `llms.txt`: strip the support-bound sentence | RED |
| M8 | matrix `install_docs`: drop `docs/guides/troubleshooting.md` | RED |

**Authority is derived, not restated.** A coordinated bump of the sole authority to `==3.13.*` (matrix tags, classifiers, `support_bound`, pyproject, doc sentences) keeps the gate green and the planner then emits `python=3.13`, `abi=cp313`, and three `cp313-*` rows; the wheels job carries zero `3.12` literals. Moving only the matrix while `release.yml` keeps 3.12 is red (the suite's own case, re-run here).

**Fail-closed planner.** `plan-python-artifact-matrix.py:16-23` rejects, with rc=1 and no default or coercion, every input I tried: empty string, whitespace-only, missing key, `null`, a list, `>=3.9`, bare `3.12`, `==3.12`, `==3.12.0`, `~=3.12`. Empty `wheels` and a missing matrix file are also rc=1. `==3.13.*` against `cp312` tags fails on the ABI bind. The only accepted variation is surrounding whitespace, which `.strip()` normalises.

**The smoke call is observed, not inferred.** Beyond the repo's own spy, I ran the real `main()` end to end against a wheel I built myself: it created a venv, ran `pip install --no-index --only-binary :all: --find-links`, and printed `import_smoke=native version=5.4.0`. Positive controls on that same path: native import failure RED, two native members RED, zero native members RED, version mismatch RED, control green before and after.

**Path-selector reachability, tested with the repo's tooling.** `pre-commit run python-artifact-truth --files <path>` fires for all 21 guarded surfaces (matrix, pyproject, sdk `Cargo.toml`, sdk README, all six owner scripts, `release.yml`, `kernel-matrix.yml`, the six governed doc pages, `llms.txt`, `CHANGELOG.md`, `.pre-commit-config.yaml`) and is skipped for `README.md` and `docs/index.md`. Evaluating `kernel-matrix.yml`'s 52 `pull_request.paths` entries with `**` glob semantics, every one of those surfaces starts the workflow; there is no `paths-ignore`.

**Metadata and wheel parity, measured against the real artifact.** PyPI `assay-it 5.4.0` publishes exactly `assay_it-5.4.0-cp312-cp312-macosx_10_12_x86_64.whl`, `...-macosx_11_0_arm64.whl`, `...-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`, no sdist, with `Requires-Python >=3.9` and a PyPy classifier. Feeding that real filename list to `check-python-artifact-truth.py --published-files` passes, so the three declared tags reproduce the published set exactly. The harness digest is correct: `sha256(.github/workflows/release.yml)` at head is `0f1601bd…`, matching `scripts/ci/fixtures/published-release-golden-path/v1/harness-manifest.json`, and `test-published-release-golden-path-contract.sh` is green.

**Other suites run green:** `test-git-fixture-env-isolation.sh` (now wrapping both new suites), `test-published-release-golden-path-contract.sh`, `test-setup-rust-composite-contract.sh`, and both new suites in full (21 and 15 mutation cases). No test in the change is skipped or asserts only that a command ran.

### Findings (non-blocking)

**F1 — the "those wheels never existed" sentence is unqualified.** `CHANGELOG.md:17-18` and `docs/migration-v1.2.md:46-49` say the change "does not break any previously working 3.9–3.11 install: those wheels never existed." True for the 5.4.0 line, which I verified. Not true package-wide: `assay-it` 0.8.0, 0.8.0rc1, 0.9.0, 1.0.0 and 1.0.0rc3 each published an `assay_it-<v>-py3-none-any.whl` plus an sdist under `Requires-Python >=3.9`, all installable on 3.9 through 3.11, and 1.2.5 through 3.5.1 shipped a `cp38-cp38-manylinux` Linux wheel. The migration *impact* as stated is still accurate, since pip's backtracking behaviour on 3.9-3.11 is unchanged by this edit. Smallest fix: scope the sentence, e.g. "no 3.9–3.11 wheel for this release line ever existed."

**F2 — the pre-commit selector is only partly pinned.** `check-python-artifact-truth.py:36-39` pins just `docs/migration-v1.2.md` and `.github/workflows/kernel-matrix.yml`. Dropping `docs/guides/troubleshooting.md` or `.github/workflows/release.yml` from that `files:` regex leaves the gate green (probes G1, G2). Bounded: the CI leg is `pre-commit run --all-files`, so the selector never gates CI; only the local staged-file trigger degrades. Smallest fix: make `PRECOMMIT_REQUIRED_PATHS` the same set as `ARTIFACT_TRUTH_PR_PATHS` plus `.github/workflows/release.yml`.

**F3 — `docs/reference/**` is a live doc subtree outside the inventory scan.** `check-python-artifact-truth.py:224-230` scans `docs/python-sdk`, `docs/getting-started`, `docs/guides`, `docs/AIcontext`, `assay-python-sdk`, top-level `docs/*.md` and `llms.txt`. A probe page at `docs/reference/__probe.md` carrying a bare `pip install assay-it` and "Works on Python 3.9+" leaves the gate green (G5); the same page under `docs/guides/` is red (G4). `kernel-matrix.yml` does not cover `docs/reference/**` either. No current overclaim: the only live occurrence is `docs/reference/config/migration.md:54`, a `pip install assay-it==0.8.0` rollback line with no Python-version claim. Smallest fix: add `docs/reference` to `INSTALL_DOC_SCAN_DIRS` (and the kernel-matrix path list), or name the exclusion in the comment at lines 222-223 next to archive/plans/architecture.

**F4 — dead guard in the self-test.** `scripts/ci/test-check-python-artifact-truth.sh:342-344`: under `set -euo pipefail` the preceding heredoc already aborts the script, so `if [ $? -ne 0 ]; then fail_test ...` is unreachable. Behaviour is still correct, the check just fires through errexit rather than through that branch.

### Self-corrections

- I first read `.github/workflows/release.yml` as absent from `kernel-matrix.yml`'s `pull_request.paths` and suspected a reachability hole. It is present at line 87 (pre-existing). Wrong reading, withdrawn.
- One mutation batch edited install-doc files outside my backup set, so its restore was incomplete and the gap probes in it ran against a dirty tree. I voided that batch, restored every dirty file from the head blob, re-verified the control green, and re-ran the probes with a 19-file backup. Only the re-run results are reported.
- I hypothesised that dropping `.github/workflows/release.yml` from the kernel-matrix path list was unguarded. Disproved: `scripts/ci/test-setup-rust-composite-contract.sh` goes red on exactly that mutation (green, red, green with a byte-identical restore).

### Explicit non-claims

- I did not build a real wheel. maturin output tags, the manylinux container build, and native import of the actual macOS wheels stay unproven until the release `wheels` job runs, exactly as the PR states.
- I did not execute `release.yml`. That `strategy.matrix.include: ${{ fromJSON(...) }}` expands at runtime is not measured here; the supporting evidence is that hosted `actionlint`, `zizmor` and `Lint (pre-commit)` are green on this SHA.
- I did not verify PyPI behaviour under the new metadata, since 5.4.0 is not being republished.
- I did not run the Rust workspace suites; nothing in `base..head` touches Rust sources.

### Hosted status on this exact SHA

73 check-runs, all on `dabd7c79`. Live branch protection on `main` requires four contexts: `CI` (success), `host-capability-check` (success), `lane-check/proof` (success, commit status: "no delegated runner proof required"), and `review-record-check` (failure). That last one is the quorum gate with no carrier yet for this head, not a code finding. Everything else is success or skipped; nothing is cancelled, queued or pending. `gh pr checks` renders only `review-record-check` as non-pass, which matches the API.

Rul1an:
```json
{"schema": "assay.review-record.v0", "head_sha": "dabd7c793b6c633b9598f97323fdb691f4b7e293", "builder": {"agent": "ruley", "instance": "ruley-2649-python-artifact-truth"}, "reviewer": {"agent": "claude", "instance": "pr2658-dabd", "github_login": "Rul1an"}, "review_completed": true, "verdict": "READY", "findings": [{"id": "F1", "summary": "CHANGELOG.md:17-18 and docs/migration-v1.2.md:46-49 state unqualified that no 3.9-3.11 wheel ever existed; true for the 5.4.0 line but assay-it 0.8.0-1.0.0 published py3-none-any wheels plus sdists under Requires-Python >=3.9.", "disposition": "non-blocking; scope the sentence to this release line. Stated migration impact is accurate."}, {"id": "F2", "summary": "check-python-artifact-truth.py:36-39 PRECOMMIT_REQUIRED_PATHS pins only migration-v1.2.md and kernel-matrix.yml, so the pre-commit files: selector can silently drop docs/guides/troubleshooting.md or .github/workflows/release.yml and the gate stays green.", "disposition": "non-blocking; CI runs pre-commit --all-files so the selector never gates CI. Suggested fix: pin the same set as ARTIFACT_TRUTH_PR_PATHS plus release.yml."}, {"id": "F3", "summary": "check-python-artifact-truth.py:224-230 INSTALL_DOC_SCAN_DIRS omits docs/reference/**, a live doc subtree; a probe page there with a bare pip install assay-it and a 3.9+ claim leaves the gate green, and kernel-matrix paths do not cover it either.", "disposition": "non-blocking; no current overclaim (only docs/reference/config/migration.md:54, a rollback pin with no Python claim). Suggested fix: add docs/reference to the scan dirs or record the exclusion."}, {"id": "F4", "summary": "scripts/ci/test-check-python-artifact-truth.sh:342-344 has an unreachable 'if [ $? -ne 0 ]' guard: under set -e the preceding heredoc already aborts the script.", "disposition": "non-blocking cosmetic; behaviour is still correct via errexit."}], "no_findings": false, "independence": {"did_not_build": true, "did_not_author_governing_spec": true}}
```

Issue #2647, opened 2026-08-26
mcp: export observed manifest as candidate and require explicit promotion


Rul1an:
## Problem

Assay can emit an observed MCP manifest and consume an operator-declared manifest, but it lacks a safe authoring bridge. Users must hand-author the file, creating error and TOFU pressure.

## Contract

Observation may produce a candidate, never approval. The candidate and declared baseline use different schemas and different artifacts.

## DoD

- [ ] RED: passing an unpromoted candidate to `--declared-mcp-manifest` fails startup.
- [ ] Export a deterministic `assay.observed_mcp_manifest_candidate.v0` containing only required tool identity/digest fields, source digest, candidate status, and explicit non-claim.
- [ ] Candidate export carries no raw arguments, credentials, or inferred approval and applies ceilings before materialization.
- [ ] Promotion checks source digest and mints a new `assay.declared_mcp_manifest.v0`; it does not flip a field in place.
- [ ] Source-digest mismatch, unknown schema, duplicate tool identity, or unsupported member fails closed.
- [ ] A mutation that treats the first live observation as approved turns the focused test red.
- [ ] Output ordering and serialization are deterministic.
- [ ] Docs show candidate review and explicit promotion as separate commands.

## Non-goals

No automatic approval, scalar trust score, whole-action verdict, new host hook, or change to the MCP enforcement boundary.

Parent: #1975

Rul1an:
## Source inventory at `822f5286b34abeb58f2317c392d4a069997e1309`

The P1 slice can reuse an existing producer seam rather than inventing a second manifest projection:

- `crates/assay-mcp-server/src/manifest_observed.rs` already computes the canonical per-tool and manifest digests and emits `assay.mcp_manifest_observed.v0` with honest `observed | not_observed | ambiguous` states.
- `crates/assay-mcp-server/src/proxy/enforce/manifest.rs` is the startup loader for `assay.declared_mcp_manifest.v0`.

Two implementation constraints are load-bearing:

1. Candidate export/promotion must call the existing digest/projection rule (or extract one shared function). Reimplementing the JCS preimage would violate the repo's one-rule/one-function contract.
2. The current declared loader deserializes only `{schema, tools}` without `deny_unknown_fields`; the richer documented `server`, `canonicalization`, and `manifest_digest` members are not validated by this loader. The issue's unsupported-member/source-digest DoD must therefore be proven with RED tests, not assumed from the docs.

Suggested slice boundary:

[...]

PR #2660, opened 2026-08-26
feat(mcp): add explicit manifest candidate promotion

### LINKED ISSUE #2655: mcp(manifest): export candidate and require explicit promotion
## Contract

Add explicit candidate export and operator promotion after the shared strict manifest truth lands. This is Slice B of #2647 and depends on #2654.

## DoD

- [ ] Add `assay mcp manifest candidate --from-observed observed.json --out candidate.json` and `assay mcp manifest promote --candidate candidate.json --source observed.json --out declared.json`.
- [ ] Candidate accepts only complete, non-ambiguous `assay.mcp_manifest_observed.v0` with a non-null manifest digest.
- [ ] Candidate schema is distinct, deterministic, sanitized, contains SHA-256 of exact observed source bytes, and carries explicit candidate/non-approval status and non-claims.
- [ ] Promotion rereads and revalidates source bytes, checks source digest/candidate parity, and mints a new declared-v0 artifact through the shared constructor.
- [ ] Observed, candidate, and declared-manifest reads share one explicit byte ceiling enforced before full materialization; oversize input fails before output creation. Bound the operator note or remove the inaccurate `bounded` claim.
- [ ] If `field_digests` are retained, accept only the producer vocabulary (`description`, `input_schema`, `output_schema`, `annotations`) through one shared constant; an unknown key fails closed.
- [ ] Candidate passed directly to `--declared-mcp-manifest` fails startup; source/candidate mismatch and unknown/duplicate members fail before output creation.
- [ ] Atomic create-new output semantics by default; validation failure leaves no partial file and no silent overwrite of an approval baseline.

## Required mutations

- auto-approve first observation;
- trust candidate-carried source digest without rereading source;
- flip candidate status in place;
- permit ambiguous/partial observation;
- overwrite existing declared baseline silently;
- exceed the shared byte ceiling or bypass it through the promotion reread;
- accept an unknown `field_digests` key or unbounded operator note;
- no-op control.

## Non-goals

No host hook, scalar score, identity claim, whole-action verdict, or change to the MCP enforcement boundary.

Parent: #2647. Depends on #2654.

Rul1an:
## ADR Boundary Slice

- Canonical ledger: none active (standalone child of #2647)
- Slice: #2655 explicit MCP manifest candidate export and promotion
- Builder: Codex
- Reviewers: fresh independent final-head review pending
- Final head SHA: `a655d87a7f0d5bd9dd7fa334892163e7a5f8dcd1`
- ADR invariants affected: evidence inputs remain bounded and hostile; candidate, observation, and approval remain distinct

## Behavior

Adds `assay mcp manifest candidate` and `assay mcp manifest promote`. Candidate export accepts only a complete, unambiguous observed-v0 manifest. Promotion rereads the exact source, reconstructs and compares every candidate field, validates through the shared declared-v0 constructor/parser, and creates a new output without overwriting an existing baseline.

Observed, candidate, and declared reads and generated outputs share one inclusive 1 MiB ceiling. Reads use the shared typed `LimitReader`; unknown or duplicate members at any depth, unknown field-digest keys, source drift, partial/ambiguous observations, oversized inputs, and existing output paths fail before a declared artifact is minted. Candidate and declared digest syntax now use one shared canonical-SHA function.

## Non-Claims

A candidate is not approval. Promotion proves an explicit artifact transition and exact byte binding; it does not prove provenance, server honesty, tool safety, or observation coverage beyond the carried complete `tools/list`.

## Test Evidence

All final measurements target `a655d87a7f0d5bd9dd7fa334892163e7a5f8dcd1` in `/Users/roelschuurkes/wt-codex-2655-mcp-manifest-promotion` with Rust/Cargo 1.96.0.

### RED

- Missing `assay mcp manifest` subcommand failed the initial CLI contract.
- Shared declared loader contract failed to compile before the bounded loader existed.
- Observed non-claim drift produced a candidate before the closed contract check.
- Oversized generated JSON was written before the shared output ceiling.
- Review repair: the oversized-read test failed because the hand-rolled reader lost the typed `LimitExceeded` cause.
- Review repair: replacing the shared SHA validator with `return true` initially survived, proving the syntax rule needed its own behavioral test.
- Hosted Ubuntu found the new public identity missing from the canonical CLI JSON partition; the focused test reproduced the failure locally before the identity was first misclassified as non-CLI; tracing the candidate --out path proved it is a caller-written CLI document, and the partition now binds the CLI writer to the naming file.

### GREEN

- `cargo test -p assay-mcp-server --lib --bins`: 111 lib + 109 bin passed
- `cargo test -p assay-cli --test mcp_manifest_promotion_contract`: 10 passed
- `cargo test -p assay-cli --test cli_json_identities`: 15 passed
- `cargo clippy -p assay-mcp-server -p assay-cli --tests -- -D warnings`: passed
- `cargo fmt --all -- --check`: passed
- `git diff --check`: passed
- commit and pre-push hooks: passed

Mutation controls killed: auto-approval, trusting candidate-carried source without reread, status flip, output overwrite, source-read ceiling bypass, unknown field digests, generated-output ceiling removal, observed non-claim drift, duplicate-member parser bypass including a nested `field_digests` member, and shared SHA validation replaced with `return true`.

## Review Findings And Dispositions

Independent reviews on `d7c95304b490f7e33c684d1d164e277c3b664985` and `d42240a979131a80648fa66d34dfcdc932df2651` were BLOCKED and are invalidated by repair pushes.

- F1 fixed: manifest reads now use `assay_common::limits::LimitReader` and preserve the typed cause.
- F2 fixed: candidate and declared manifests call one canonical-SHA function, with a mutation-sensitive unit contract.
- F3 disposition: the serialize-then-parse step remains defense in depth. Invalid serialized output is unreachable through the validated typed constructor; adding a test-only invalid serializer seam would expand the production API without increasing the behavioral guarantee. It is not claimed as an independently mutation-sensitive gate.
- F4 fixed: nested duplicate members are explicitly covered through the strict parser.
- F5 retained: the second ceiling is intentional defense in depth at the generated-output boundary and remains equal to the shared manifest limit.
- F6 verified by the reviewer against a real producer artifact; the hand fixture is not claimed as producer parity.
- Writer-binding review fixed: both manifest identities now name `manifest_promotion.rs`, the module that serializes and creates the files; a RED test pins that exact writer, and the unused typed return added only to satisfy the old token scan was removed.

## Review Quorum

- [ ] One non-building agent review on `a655d87a7f0d5bd9dd7fa334892163e7a5f8dcd1`
- [x] Every prior actionable finding is fixed or has a technical disposition
- [x] No delegated proof is required by the changed paths

Optional automated reviews can add evidence, but do not count toward or block quorum.

Closes #2655

## Summary by CodeRabbit

- **New Features**
  - Added commands to export MCP manifest candidates and promote reviewed candidates into declared manifests.
  - Promotion verifies the original observed manifest and preserves exact source-byte integrity.
  - Existing output files are never overwritten.

- **Bug Fixes**
  - Added strict validation for duplicate, unknown, malformed, and oversized manifests.
  - Applied a shared 1 MiB read/write limit to MCP manifests.

- **Documentation**
  - Added CLI reference examples and workflow documentation for candidate export and promotion.


Rul1an:
## Independent review, head `d7c95304b490f7e33c684d1d164e277c3b664985`

**Independence.** I authored and edited neither issue #2655, its spec or implementation plan, nor this implementation. I read issue #2655 (body and comments), the PR body, `AGENTS.md`, and the full `base..head` diff and source at both ends. I read **no PR comments and no PR reviews**, and I was given no prior findings and no mutation recipes; everything below is derived here. Read-only detached clone at this SHA; the repository worktrees were not touched. Head re-read at start and immediately before this verdict: unchanged both times.

Toolchain: rustc/cargo 1.96.0, macOS (darwin 25.6.0). CI runs Linux, so the platform differs; the paths under review are platform-neutral except `persist_noclobber`, whose no-clobber guarantee is stronger on Linux than on macOS.

---

## Verdict

**BLOCKED at `d7c95304b490f7e33c684d1d164e277c3b664985`**, narrowly, on F1 and F2 only.

I found **no correctness defect**. All nine properties hold, each confirmed by a mutation that turns something red. The block is that F1 and F2 are freshly introduced one-rule-one-function violations, and per `AGENTS.md` merge requires every actionable finding to be fixed or to carry a recorded technical disposition. Either clears this. If they are fixed in code, the fix touches reviewed files and needs a fresh review at the new head.

---

## Findings

### F1 (moderate, actionable). A second bounded-read mechanism, in a crate that already uses the shared one.

`crates/assay-mcp-server/src/manifest_io.rs:9-23` hand-rolls the ceiling as `File::take(MAX + 1)` plus a post-hoc length check, rather than `assay_common::limits::LimitReader`. CLAUDE.md and ADR-043 §1 name `LimitReader` as *the* shared bounded-ingest primitive, admitted to `assay-common` precisely because "a mechanism whose second implementation would silently mean something different". This same crate already calls it, one module away, at `crates/assay-mcp-server/src/tools/policy_read.rs:11-16`, with `LimitExceeded::from_io` classification and exact-limit / limit+1 tests.

Two consequences beyond the doctrine point. The new reader raises an untyped `anyhow` message instead of the typed `LimitExceeded`, so callers must classify by string, and the new tests do exactly that at `crates/assay-cli/tests/mcp_manifest_promotion_contract.rs:451`, `:484`, `:500`, `:520` (`.contains("manifest limit")`). `crates/assay-common/src/limits.rs:68-71` says in as many words that matching the rendered message "is not a contract". And nothing pins the two implementations to the same semantics: no shared call, no parity test.

Semantics today are equivalent (both inclusive, both bounded before materialization); this is a drift and classification finding, not a behaviour finding.

Smallest coherent fix: in `read_bounded_bytes`, wrap the file in `LimitReader::new(file, MAX_MANIFEST_BYTES as u64, LimitKind::SourceBytes)`, `read_to_end`, and map `LimitExceeded::from_io` onto the manifest-domain message. `MAX_MANIFEST_BYTES` stays as the domain vocabulary, which is the split CLAUDE.md asks for.

### F2 (minor, actionable). Duplicate canonical-sha256 validator, and the duplicate is unpinned.

`crates/assay-mcp-server/src/manifest_promotion.rs:227-235` (`is_sha256`) is the same rule as `crates/assay-mcp-server/src/declared_manifest.rs:91-101` (`is_canonical_sha256`), same crate, character for character the same predicate.

Mutation **M10** (make `is_sha256` return `true` unconditionally) left both suites fully green, 16 + 10. The copy is masked by the parity check in `promote`, so no test can currently tell whether it exists.

Smallest coherent fix: make `is_canonical_sha256` `pub(crate)`, call it from `validate_candidate`, delete `is_sha256`.

### F3 (minor, advisory). The declared round-trip in `promote` is unpinned, and #2655 asked for it to be pinned.

`crates/assay-mcp-server/src/manifest_promotion.rs:89-92` serializes the constructed declared manifest and feeds it back through `parse_declared_manifest`. Design comment item 6 on #2655: "a mutation that emits an invalid declared artifact must turn this round-trip red."

Mutation **M11** (delete the round-trip outright) left both suites green, 16 + 10. It is unreachable by construction: `construct_declared_manifest` already validates the identical struct, and serialization is lossless. Cross-check, mutation **M12** (disable `validate_declared_manifest`) *was* killed, but by the candidate export path, not by this round-trip.

Disposition, my recommendation: keep it as defence-in-depth against a future non-round-tripping `Serialize`, and say so in a one-line comment, accepting that it is unpinned. Alternatively pin it by re-parsing the bytes that `write_json_create_new` actually writes, rather than a second independent serialization of the same value. Not blocking; the property it guards is held by `construct_declared_manifest`.

### F4 (minor, advisory). The duplicate-member test does not sit on the rule it claims to pin.

`crates/assay-cli/tests/mcp_manifest_promotion_contract.rs:266-278` and `:582-593` duplicate a **struct field** (`status`). Serde's derived deserializer rejects duplicate known struct fields on its own, so that refusal survives removing `assay_core::mcp::parse_unique_json`.

Positive control on the mutated line. With `parse_promotable_observed` switched to `serde_json::from_str`:

```
DUP-MAP-KEY        success=true    (candidate minted)
DUP-STRUCT-FIELD   success=false   stderr: duplicate field `status` at line 4 column 10
```

At HEAD both are refused with `JSON object contains a duplicate member`. So mutation **M8** was killed only by the error-*text* assertion, not by the behavioural one, and the case where `parse_unique_json` is behaviourally load-bearing, a duplicated key inside `field_digests` (a `BTreeMap`, where serde is last-value-wins), has no test.

Bounded impact: `manifest_digest` covers only `{name, tool_digest}`, so a last-wins `field_digests` value cannot move the approval identity. It is attribution ambiguity, which is exactly what the strict loader exists to refuse.

Smallest coherent fix: add one case to `promotion_refuses_unknown_and_duplicate_candidate_members` that duplicates a key inside `field_digests`.

### F5 (advisory). A second ceiling governs manifest text, equal to the first only by coincidence.

`parse_unique_json` (`crates/assay-core/src/mcp/unique_json.rs:39-43`) applies `McpTranscriptLimits::default().max_line_bytes` to the whole manifest text after the bounded read. Measured:

```
MAX_MANIFEST_BYTES = 1048576
parse_unique_json REFUSED 1048599 bytes: JSON line exceeded byte limit of 1048576
parse_unique_json accepted 1048574 bytes
```

So the "one inclusive 1 MiB ceiling" in `CHANGELOG.md` and `docs/reference/cli/mcp-server.md:158-160` holds because two independent constants, in two crates and two domains, happen to be equal; after the read a manifest is governed by a transcript *line* limit, which would surface as a confusing error naming a line. Mitigating, and why this is advisory rather than actionable: `all_manifest_reads_share_the_same_exact_byte_ceiling` exercises the exact boundary, so lowering `max_line_bytes` or raising `MAX_MANIFEST_BYTES` turns it red. Disposition: a comment or `const _: () = assert!(...)` naming the coupling would make the existing guard legible.

### F6 (advisory). The happy path never runs against real producer output.

`promotable_observed()` at `crates/assay-cli/tests/mcp_manifest_promotion_contract.rs:17-52` is hand-written; `build_observed` appears only in a negative case at `:255`. Since `ObservedManifest` / `ObservedBody` / `ObservedTool` are `deny_unknown_fields`, a future producer field would break candidate export at runtime with nothing red.

I verified end to end at this SHA that real `build_observed` output (two tools, `Completeness::Complete`) exports a candidate and promotes to a valid declared-v0 artifact, exit 0 both times. So this is a durability gap, not a defect. Disposition: build the happy-path fixture with `build_observed` instead of by hand.

### Advisory nits

- `docs/reference/cli/mcp-server.md:162` "proves an explicit operator action": a CLI invocation can be scripted. "An explicit invocation" is the checkable claim. One word.
- The minted declared artifact carries `"note": null`, since `DeclaredManifest.note` has no `skip_serializing_if`. Harmless, re-parses, cosmetic only.
- `crates/assay-mcp-server/src/declared_manifest.rs:73` "permitted and bounded" is now true, via the 1 MiB document ceiling, so the #2655 DoD item is satisfied. The comment could name what bounds it.
- `parse_promotable_observed` does not cross-check `privileged` against `privilege_classification` / `action_class`. Both sit outside the digest preimage, so an inconsistency cannot move the approval identity. No action.

---

## Properties, each with the mutation that turns it red

Control before and after, both green: `assay-mcp-server --lib manifest` 16 passed, `assay-cli --test mcp_manifest_promotion_contract` 10 passed. Mutations applied only in a scratch clone, restored by file copy and verified with `diff -q` after each.

| # | Property | Mutation | Result |
|---|---|---|---|
| 1 | Candidate is never an approval | M2: emit `approval: "approved"`, drop the status guard | RED, 2 tests |
| 2 | Only complete, non-ambiguous observed with non-null digest | M7: accept partial / unobserved | RED |
| 3 | Promotion rereads exact source bytes, full parity, no trusted carried digest | M1: delete the `candidate != reconstructed` check | RED |
| 4a | Shared 1 MiB read ceiling before materialization | M5: raise `take()` and the check by 4096 | RED |
| 4b | Shared 1 MiB output ceiling, oversize fails before output exists | M4: delete the output size check | RED |
| 5 | `field_digests` vocabulary closed through one shared constant | M6: delete the `FIELD_NAMES` membership check | RED |
| 6 | Candidate at `--declared-mcp-manifest` fails startup | M13: drop `deny_unknown_fields` and accept the candidate schema | RED at `:418` |
| 7a | Unknown members fail before output creation | M12: disable `validate_declared_manifest` | RED |
| 7b | Duplicate members fail before output creation | M8: `serde_json::from_str` instead of `parse_unique_json` | RED, but see F4 |
| 8 | Atomic create-new, no overwrite, no residue | M3: `persist` instead of `persist_noclobber` | RED |
| 9 | Accurate non-claims, nothing claims review or trust | M9: drop the observed `non_claims` contract check | RED |

Survivors, both reported above: **M10** (F2) and **M11** (F3).

On property 3 specifically, the parity is genuine rather than nominal: `reconstructed.source_sha256` is computed over the bytes just reread, and `ManifestCandidate` derives `PartialEq` over all nine fields including `Vec<BaselineTool>`, so the candidate-carried digest is never used to short-circuit. On property 1, `promote` never mutates the candidate in place; it constructs a fresh declared document.

## Runs

- `cargo test -p assay-cli --test mcp_manifest_promotion_contract`: **10 passed**, 0 failed
- `cargo test -p assay-mcp-server --lib`: **109 passed**, 0 failed
- `cargo test -p assay-mcp-server --bins`: **109 passed**, 0 failed
- `cargo test -p assay-mcp-server` (all targets): all suites green, largest 109 / 109 / 25 / 21
- `cargo test -p assay-cli --test mcp_command_grouping_test`: 2 passed; `--test mcp_preflight_contract`: 17 passed, 1 ignored; `--test e2e_mcp_wrap_assert_cmd`: 3 passed
- `cargo clippy -p assay-mcp-server -p assay-cli --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- `git diff --check base..head`: clean
- Generated artifacts: no artifact under `docs/generated/` enumerates `mcp` subcommands, so no regeneration debt from the two new commands. Checked, not assumed.

Hosted checks for this exact SHA, read through the check-runs and statuses APIs rather than `gh pr checks`: 8 check-runs, all `completed` / `success` (assay, CodeQL, Analyze × 5, GitGuardian), plus one commit status `success` (CodeRabbit). Nothing cancelled, queued, pending, or skipped. `review-record-check` is absent at this SHA; that is the quorum gate this comment is meant to satisfy, not a code finding.

## Compatibility

Mixed, not purely additive, and correctly declared. `assay mcp manifest candidate|promote` are new. `--declared-mcp-manifest` startup is tightened three ways: the 1 MiB ceiling, duplicate JSON members, and unknown `field_digests` keys. All three appear in the CHANGELOG "Changed" entry, so nothing is undisclosed. Declared-v0 itself is unchanged, and I confirmed every existing declared-manifest fixture in the crate still loads.

## My own errors in this review

I invoked the closing no-op control while a manually applied M8 was still in the tree, so that run measured M8 a second time rather than a no-op. I disclosed it, restored, verified all three touched files byte-identical to the pristine snapshot with `diff -q`, and reran the control to green (16 + 10). A further control after M13 was also green. No mutation result above rests on the mis-sequenced run.

## Non-claims for this review

This review does not claim the design is the right one, that the observed producer is honest, that any tool is safe, that the classifier's `privileged` labels are correct, or that CI on another platform behaves as it did here. It does not claim coverage of paths outside `base..head` and their immediate callers. A green mutation battery shows those rules are pinned; it does not show the rule set is complete.

Rul1an:
Review repair pushed at exact head `13981bda3de01db1d86a9820c564cc7ea0195472`.

F1 uses the shared typed `LimitReader`; F2 uses one canonical-SHA function with a mutation-sensitive contract; F4 now pins nested duplicate members. F3 is explicitly dispositioned as defense in depth rather than an independently mutation-sensitive gate. Final local evidence: server `111 + 109`, promotion `10/10`, Clippy `-D warnings`, fmt and diff-check all green; commit/pre-push hooks passed. The prior review is invalidated. A fresh independent review is required on this head; PR remains draft.

Rul1an:
Head advanced to `dc5867df306fd2023c967dcf5c24d06a7d03937b` after hosted Ubuntu correctly caught an unclassified public JSON identity.

Fix: classify `assay.mcp_manifest_candidate.v0` in the canonical non-CLI-document partition as an MCP tool result. Local RED was the same `every_production_identity_is_classified` failure; GREEN is the full 14-test `cli_json_identities` suite. Commit and pre-push hooks passed.

All prior reviews and checks are stale for quorum/landing; fresh exact-head review and CI are required.

Rul1an:
Final-head correction pushed at `218bb9ea7e03fefc4e00d4d83a1bcfe972674c10`.

The prior identity row was green but semantically false: `assay mcp manifest candidate --out` writes `assay.mcp_manifest_candidate.v0` to a caller-named path. The identity now sits in the CLI-document block, bound to `crates/assay-cli/src/cli/commands/mcp.rs` and `crates/assay-mcp-server/src/manifest_promotion.rs`. Focused inventory suite: 14/14; diff check and all commit/pre-push hooks passed. Prior reviews are invalidated; quorum must target this SHA.

Rul1an:
## Quorum review, bound to `218bb9ea7e03fefc4e00d4d83a1bcfe972674c10`

**Independence.** I did not author or edit issue #2655, any governing specification or plan for this
change, or the implementation. Fresh read-only agent, detached scratch clone, no prior context and
no PR comments or reviews read. Head re-read at start and immediately before this verdict: unchanged.
Base `e4da51a1f6cafbbed12d8a4e1c60f9a60744feb9`, complete base..head diff reviewed (14 files, 4 commits).

---

## Findings

### F1 (blocking, Medium) The diff makes a neighbouring row of the identity record false

`docs/architecture/CLI-JSON-IDENTITIES.md:103`

```
assay.declared_mcp_manifest.v0 | operator-authored approval baseline read by assay-mcp-server at startup; no command writes it
```

That row sits in the **not-CLI-documents** block. As of this PR a command does write it. Run against
the built binary from this head:

```
$ assay mcp manifest promote --candidate cand.json --source observed.json --out declared.json
$ head -2 declared.json
{
  "schema": "assay.declared_mcp_manifest.v0",
```

A top-level JSON document, carrying that string as its `schema`, written to a caller-named path. That
is verbatim the definition the same file gives for the `cli-documents` block at line 30. So this PR
introduces a second CLI JSON document and leaves it classified as "no command writes it".

This is the failure mode the file names itself: "moving a real document into this block now means
writing a sentence that is false, rather than deleting a line", and "The partition is total ... It does
not stop a *misclassification*, and no static check can". The guard stays green because the identity is
in *a* block, so nothing catches it. `documents_are_bound_to_a_writer` never sees the row.

**Fix (verified: all 24 tests green with it applied).** The naive move alone fails, because
`commands/mcp.rs` touches no public symbol of `declared_manifest.rs`:

```
assay.declared_mcp_manifest.v0 is recorded as written by crates/assay-cli/src/cli/commands/mcp.rs and
named in crates/assay-mcp-server/src/declared_manifest.rs, and the writer mentions none of that file's
public symbols
```

Make the tie real, then move the row:

1. `crates/assay-mcp-server/src/manifest_promotion.rs:80` return what was minted:
   `pub fn promote(...) -> Result<DeclaredManifest>`, ending `write_json_create_new(out, &declared)?; Ok(declared)`.
2. `crates/assay-cli/src/cli/commands/mcp.rs` bind it:
   `let _declared: assay_mcp_server::declared_manifest::DeclaredManifest = ...promote(&candidate, &source, &out)?;`
3. Delete line 103 and add to the `cli-documents` block:
   `assay.declared_mcp_manifest.v0 | crates/assay-cli/src/cli/commands/mcp.rs | crates/assay-mcp-server/src/declared_manifest.rs`

I applied exactly that in my clone: `cli_json_identities` 14/14 and `mcp_manifest_promotion_contract` 10/10.

### F2 (Low) The candidate row's writer-idiom half is decorative

`docs/architecture/CLI-JSON-IDENTITIES.md:62`, `crates/assay-cli/src/cli/commands/mcp.rs`

The row's `WRITERS` check passes only because `mcp.rs` contains `eprintln!`, and `"eprintln!"` contains
the substring `"println!"`. Those six calls are `assay mcp wrap` progress logging and have nothing to do
with candidate emission; the actual serialization is `manifest_io::write_json_create_new`. Proved by
renaming the macro (`use std::eprintln as elog;`, behaviour identical) with the row untouched:

```
assay.mcp_manifest_candidate.v0 is recorded as written by crates/assay-cli/src/cli/commands/mcp.rs,
and that file calls no writer
```

Disposition: **accept, note only.** This is a property of the pre-existing `WRITERS` list, not of this
diff, and the load-bearing half of the guard (the symbol tie) is genuine and does bite (M10, M11 below).
If you want the note closed, `write_json_create_new` is the honest idiom to add to `WRITERS`.

### F3 (Low, informational) "One shared ceiling" is two constants

`crates/assay-mcp-server/src/manifest_io.rs:7` sets `MAX_MANIFEST_BYTES = 1 MiB`, and every manifest
read also passes `assay_core::mcp::parse_unique_json`, which applies
`McpTranscriptLimits::default().max_line_bytes` (`crates/assay-core/src/mcp/ingest.rs:35`, also 1 MiB).
Nothing declares them equal. Today the exact-limit case in
`all_manifest_reads_share_the_same_exact_byte_ceiling` would go red if the transcript limit were lowered,
so the coupling is pinned by accident rather than unguarded. Disposition: **accept**; a one-line
`const _: () = assert!(...)` or a comment at `manifest_io.rs:7` would make the dependency legible.

### F4 (Low, informational) Two assertions weaker than they read

- `mcp_manifest_promotion_contract.rs:419` asserts the candidate refusal message
  `contains("candidate") || contains("schema")`. The refusal actually comes from `deny_unknown_fields`,
  whose serde message lists the expected fields, one of which is literally `schema`. The assertion is
  therefore satisfied by any unknown-field refusal. The property still bites (M1), the error-shape
  assertion does not.
- `generated_candidate_is_not_a_declared_approval_baseline` exercises
  `declared_manifest::load_declared_manifest` directly rather than the `assay-mcp-server` binary.
  That is the same function `main.rs:248` reaches through `proxy::enforce::load_declared_manifest`, and
  the delegation is one line in this diff, so I am satisfied, but "candidate fails startup" is asserted
  one level below startup.

### Regression / compatibility

`assay.declared_mcp_manifest.v0` readers are strictly tightened at an unchanged generation:
unknown `field_digests` keys refused (`declared_manifest.rs:173`), duplicate JSON members refused at any
depth (`declared_manifest.rs:107`), and a 1 MiB ceiling before parse (`declared_manifest.rs:114`). An
operator baseline that loaded yesterday can refuse startup today. Direction is fail-closed and the
CHANGELOG declares it under `### Changed`. **Declared compatibility class: behaviour-breaking on an
operator-authored input, identity unchanged, documented.** Consistent with the repo's own "a closed
schema stays closed ... that is the accepted cost". No other regression found; full
`cargo test -p assay-mcp-server` (30 targets) and `cargo clippy -p assay-mcp-server -p assay-cli
--all-targets` are clean at this head.

---

## Properties

| # | Property | Verdict | Where |
|---|---|---|---|
| 1 | Candidate is never approval | **Confirmed** | `declared_manifest.rs:66` (`deny_unknown_fields`) + `:138` (schema check) |
| 2 | Promotion rereads exact source bytes | **Confirmed** | `manifest_promotion.rs:83-87` |
| 3 | Bounded hostile reads, typed limit class | **Confirmed** | `manifest_io.rs:12` (LimitReader), `:26` (output), `declared_manifest.rs:114` |
| 4 | Strict duplicate rejection incl. nested `field_digests` | **Confirmed** | `manifest_promotion.rs:107,131` via `parse_unique_json` |
| 5 | Shared canonical SHA validation | **Confirmed** | `declared_manifest.rs:91`, sole impl in the crate; imported at `manifest_promotion.rs:2` |
| 6 | No overwrite, atomic create-new | **Confirmed** | `manifest_io.rs:42` `persist_noclobber` |
| 7 | Candidate classified as a CLI JSON document, right block | **Confirmed** | `CLI-JSON-IDENTITIES.md:62` |
| 8 | Two-way registry guard binds the real writer | **Confirmed with a caveat** | see F2 |

Notes on the ones that needed proof rather than reading:

**P1** is carried by two independent guards. Relaxing only the schema check (`m1a`) leaves the
candidate refused by `deny_unknown_fields`; relaxing both flips it and the test goes red. Neither is a
single point of failure, which is the right shape.

**P4** was the one the briefing warned could be proved by the wrong case, so I proved it at the binary.
Two fixtures differing from a known-good source in exactly one way: a duplicate `description` key
*inside* `field_digests`, and a duplicate top-level `status`. At head both are refused with
`fatal: JSON object contains a duplicate member`. With `parse_unique_json` swapped for
`serde_json::from_str::<Value>` at `manifest_promotion.rs:107`, **both are accepted, exit 0**. So the
nested case genuinely reaches the guard, and in this pipeline so does the struct-field case: the code
goes text -> `Value` -> `from_value`, so serde's derive never sees the raw duplicate and cannot reject it.
`UniqueValue::visit_map` (`assay-core/src/mcp/era.rs:105`) recurses through values, which is what gives
the guard depth. Positive control: the unmodified source produced a candidate, exit 0, in the same run.

**P5** one implementation, `is_canonical_sha256`, promoted from private to `pub(crate)` and imported by
the new module. No second syntax check exists in `assay-mcp-server`. The other `strip_prefix("sha256:")`
sites (`assay-core/src/mcp/tool_decision_truth.rs:30`, `assay-evidence`, `assay-cli/.../evidence/`) are
pre-existing and in other domains; this diff adds no duplicate.

**P3** the "typed rather than string-matched" half is separately pinned: replacing `LimitReader` with a
read-then-check that keeps the identical message leaves all six CLI-level ceiling assertions green and
turns only `oversized_read_preserves_the_shared_typed_limit_cause` red.

---

## Mutation matrix

Mutations applied at the production line that decides, never a helper or a test. Restore by file copy,
verified with `diff -q` plus `git status --porcelain` after each. No-op control first and last.

| # | Mutation (file:line) | Expected | Observed | Result |
|---|---|---|---|---|
| C0 | none (control, first) | green | 14+10+111 pass | control holds |
| M1 | `declared_manifest.rs:66,138` drop `deny_unknown_fields` on `DeclaredManifest` + accept candidate schema | red | `generated_candidate_is_not_a_declared_approval_baseline` FAILED | **bites** |
| M1a | `declared_manifest.rs:138` accept candidate schema only | green expected | 10 pass | does not bite (structural guard carries it) |
| M2 | `manifest_promotion.rs:84` `reconstructed = candidate.clone()` (never reread source) | red | `promotion_binds_exact_source_bytes_and_every_candidate_field` FAILED | **bites** |
| M2b | `manifest_promotion.rs:85` copy carried `source_sha256` onto the reconstruction | red | same test FAILED | **bites** |
| M3 | `manifest_io.rs:26` remove output ceiling | red | `oversized_json_is_rejected_without_output_or_temp_residue` FAILED | **bites** |
| M4 | `manifest_io.rs:12` `LimitReader::new(file, u64::MAX, ...)` | red | `all_manifest_reads_share_the_same_exact_byte_ceiling` FAILED | **bites** |
| M5 | `manifest_io.rs:9-21` replace `LimitReader` with read-then-check, same message | red on the typed test only | CLI ceiling tests 10/10 green; `oversized_read_preserves_the_shared_typed_limit_cause` FAILED | **bites, discriminating** |
| M6 | `manifest_promotion.rs:107` `parse_unique_json` -> `serde_json::from_str` | red | `candidate_refuses_sources_that_do_not_strictly_describe_their_own_bytes` FAILED; binary accepts both duplicate fixtures, exit 0 | **bites** |
| M7 | `declared_manifest.rs:96` accept uppercase hex | red | `canonical_sha256_syntax_is_closed_and_shared` FAILED | **bites** |
| M8 | `manifest_io.rs:42` `persist_noclobber` -> `persist` | red | `candidate_and_promotion_never_overwrite_existing_outputs` FAILED | **bites** |
| M9 | `declared_manifest.rs:173` disable the `FIELD_NAMES` check | red | `candidate_refuses_sources_that_do_not_strictly_describe_their_own_bytes` FAILED | **bites** |
| M10 | `manifest_promotion.rs:13` `CANDIDATE_SCHEMA` -> `...v1` (emitted identity drifts from the record) | red | `every_production_identity_is_classified` + `documents_are_bound_to_a_writer` FAILED | **bites** |
| M11 | `CLI-JSON-IDENTITIES.md:62` writer column -> `describe.rs` | red | `documents_are_bound_to_a_writer` FAILED ("mentions none of that file's public symbols") | **bites** |
| M11b | `commands/mcp.rs` rename `eprintln!` -> aliased `elog!` (behaviour identical) | probe | `documents_are_bound_to_a_writer` FAILED ("calls no writer") | see F2 |
| FIX | F1 fix applied (3 edits above) | green | 14+10 pass | fix verified |
| C1 | none (control, last) | green | 14+10+111 pass | control holds |

Skipped, with reason: no mutation of the `before full materialization` ordering inside `LimitReader`
itself, because that primitive is `assay-common` and outside this diff; M5 pins that the primitive is on
the path, which is what this PR controls.

---

## Tests and checks I ran

Detached clone at `218bb9ea`, `CARGO_TARGET_DIR` local, debug info off.

- `cargo test -p assay-cli --test mcp_manifest_promotion_contract` 10/10
- `cargo test -p assay-cli --test cli_json_identities` 14/14
- `cargo test -p assay-mcp-server` (all 30 targets) all green, 0 failed
- `cargo clippy -p assay-mcp-server -p assay-cli --all-targets` clean
- Manual binary probes: candidate export, promotion, both duplicate fixtures, declared output inspection

**Hosted CI, read via the check-runs API bound to this exact SHA** (68 runs): 58 success, 8 skipped,
1 failure, 1 in progress.

- `review-record-check` **failure**, `review-record-check=fail reason=no_current_record 218bb9ea...`.
  That is the gate this review exists to unblock, not a code defect. Note it wants a machine record:
  `scripts/ci/assay_review_record_check.py` refuses any comment that carries the review-record HTML
  marker alongside prose, so this human-readable review deliberately omits that marker entirely (I do
  not quote it here either, for the same reason) and the JSON carrier is still owed as a separate
  comment.

  Edit disclosure: this comment was edited once, immediately after posting, for exactly that reason.
  The first version quoted the marker string literally, which would have turned the gate's failure from
  `no_current_record` into `extra_prose`. Nothing else changed; no finding, property, mutation result or
  verdict was touched.
- `Build + Test (windows-latest)` **in progress**. Not a pass. `assay-mcp-server` now takes `tempfile` as
  a runtime dependency and `write_json_create_new` relies on `persist_noclobber`; that is the one path in
  this diff whose Windows semantics I have not seen green.
- The 8 skipped runs are never a pass and I count none of them.

## Explicit non-claims

- I did not read any PR comment or review, by any route.
- I did not run the eBPF, Linux-only, or Windows paths; my platform is darwin.
- I did not evaluate whether the candidate/declared schemas are the right shape for downstream
  consumers, only that they are closed, deterministic, and validated on both emit and read.
- Hosted CI green above covers 58 runs; I make no claim about the Windows job or about anything skipped.
- I did not verify pre-existing rows of `CLI-JSON-IDENTITIES.md` beyond the two this diff touches.

## Disclosure of my own errors mid-review

While setting up M11b I copied `commands/mcp.rs` to a pristine path whose parent directory did not exist.
The `cp` failed silently in the middle of a compound command and the file stayed mutated after the run.
`git status --porcelain` caught it. I restored it from the pinned blob
(`git show 218bb9ea...:crates/assay-cli/src/cli/commands/mcp.rs`, written to a file, then copied in) and
confirmed with `diff -q` and a clean `git status`. The last no-op control ran after that restore and is
green, so nothing downstream of the slip is unverified.

---

## Verdict

**BLOCKED** on `218bb9ea7e03fefc4e00d4d83a1bcfe972674c10`.

All eight properties confirmed, thirteen of thirteen semantic mutations bite, both controls hold, and the
implementation is in good shape. The block is F1 alone: this PR introduces a second CLI JSON document and
leaves `docs/architecture/CLI-JSON-IDENTITIES.md:103` asserting that no command writes it. That file is a
hand-edited record whose whole contract is that the reason is true, and whose own text says no static
check can catch this class of mistake. The fix is three edits, and I verified it green.

F2, F3 and F4 are notes, not blockers.

Rul1an:
Repair pushed at exact head `d42240a97`.

The prior review on `218bb9ea7` is invalidated by this push.

What changed:
- `assay.declared_mcp_manifest.v0` moved from the non-document inventory to CLI documents.
- `manifest_promotion::promote` now returns the exact `DeclaredManifest` it wrote.
- the CLI binds that return type at the command boundary.
- a focused contract pins both manifest-command outputs as documents and rejects duplicate non-document classification.

TDD:
- RED: `manifest_promotion_outputs_are_cli_documents` failed because the declared manifest remained in `not-cli-documents`.
- GREEN: 15/15 `cli_json_identities`; 10/10 `mcp_manifest_promotion_contract`.

Verification on `d42240a97`:
- `cargo clippy -p assay-mcp-server -p assay-cli --all-targets -- -D warnings`
- `cargo test -p assay-mcp-server`
- `cargo test -p assay-cli --test cli_json_identities`
- `cargo test -p assay-cli --test mcp_manifest_promotion_contract`
- `git diff --check`

A fresh independent exact-head review is required before readiness or merge.

Rul1an:
Repair pushed at exact head `a655d87a7f0d5bd9dd7fa334892163e7a5f8dcd1`.

- RED: the new exact-writer assertion failed because both manifest identities named `crates/assay-cli/src/cli/commands/mcp.rs`.
- GREEN: both rows now name `crates/assay-mcp-server/src/manifest_promotion.rs`, the module that serializes and creates both files.
- Removed the unused typed return added only to satisfy the old token scan; `promote` is again `Result<()>`.
- `cli_json_identities`: 15/15; promotion contract: 10/10; MCP server: 111 lib + 109 bin; fmt, clippy `-D warnings`, diff check and pre-push hooks pass.

The prior reviews are invalidated. Fresh independent review is running on this exact head; PR remains draft.

Rul1an:

Reviewer: Ruley. I did not build this slice and did not write the governing spec. Builder is Codex. PR-body text was not treated as proof.

Scope: PR #2660 only.
Head: `a655d87a7f0d5bd9dd7fa334892163e7a5f8dcd1`
Base: `e4da51a1f6cafbbed12d8a4e1c60f9a60744feb9`

**READY**

### Measured

1. Both `cli-documents` rows name the real serializer/create-new writer `crates/assay-mcp-server/src/manifest_promotion.rs` (`export_candidate` / `promote` call `write_json_create_new`). Neither row names `mcp.rs`.
2. Restoring both writer columns to `crates/assay-cli/src/cli/commands/mcp.rs` made `manifest_promotion_outputs_are_cli_documents` RED (`mcp.rs` vs `manifest_promotion.rs`). File restore made the same test GREEN.
3. `pub fn promote(...) -> Result<()>` at `manifest_promotion.rs:80`. Exact-source reread + field compare, no-clobber persist, and inclusive 1 MiB `MAX_MANIFEST_BYTES` / `LimitReader` remain and have tests on this head.
4. `generated_candidate_is_not_a_declared_approval_baseline` is a schema-boundary pin. Extra candidate members can fail `DeclaredManifest`'s `deny_unknown_fields` before a schema-string check. Not claimed as a generated-output-ceiling gate or an independently mutation-sensitive serialize-then-parse gate. The F3 disposition is honest.
5. Public identity strings and the PR-body writer-binding claim match this head. Prior heads `d7c95304` and `d42240a9` are invalidated.
6. rustc/cargo 1.96.0 on this head: `mcp_manifest_promotion_contract` 10 OK; `cli_json_identities` 15 OK; `assay-mcp-server` 111 lib + 109 bin OK; `clippy -D warnings` OK; `fmt --all --check` OK; `git diff --check` OK.

### Non-blocking note

`docs/reference/cli/index.md` Commands Overview table still omits `manifest`.

No machine `assay-review-record` from this comment.

Rul1an:
```json
{
  "head_sha": "a655d87a7f0d5bd9dd7fa334892163e7a5f8dcd1",
  "builder": {
    "agent": "codex",
    "instance": "codex-pr2660"
  },
  "reviewer": {
    "agent": "ruley",
    "instance": "ruley-pr2660-a655d87a",
    "github_login": "Rul1an"
  },
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "N1",
      "summary": "The top-level assay mcp table description does not enumerate manifest",
      "disposition": "Accepted non-blocking: the same index carries both exact manifest commands and links the full reference, so no invocation or discovery path is missing"
    }
  ],
  "no_findings": false,
  "independence": {
    "did_not_build": true,
  }
}
```

Issue #2474, opened 2026-08-17
deps: execute the deferred refresh as bounded compatibility batches


Rul1an:
## Current baseline

Reconciled on `origin/main` `c06ca57c2a15211ede275ab3fd617a001f28df30` after the v5.5.2 release. The original pre-v5.3 Dependabot wave has been intentionally replaced by bounded compatibility slices. Most of that inventory has landed; historical measurements and exact-head records remain in the comments below.

Current shipped state:

- `jsonschema` is `0.49.4` in both root and fuzz locks; workspace declaration is `0.49`.
- `base64` is `0.22.1` in both locks and all direct declarations remain `0.22`.
- two production workflows use `actions/attest` v4.2.1 at immutable commit `508db95d…`.
- `release.yml` uses `actions/attest-build-provenance` v4.1.1 at `0f67c3f4…`.
- rust-cache, OSV, CodeQL upload, clap, AWS-LC and the earlier low-risk lock batches have landed and are not remaining work here.
- `object = 0.36.7` remains an intentional Aya-coordinated exclusion.

There are no open Dependabot PRs for this inventory. A generated PR is not required to begin an intentional slice.

### 1. Direct attestation patch

Update the two direct `actions/attest` callsites as one patch-only slice after rechecking the current upstream immutable release commit. Preserve subject inputs and output consumers. Exercise both owning producer paths separately; a delegated-run witness does not prove the pack-release producer and a textual SHA guard alone is insufficient.

### 2. Release provenance decision

Separately decide whether `release.yml` should update `actions/attest-build-provenance` or migrate to direct `actions/attest`. Compare interfaces, predicate/subject identity, permissions, outputs and verifier behavior first. Equal tag numbers do not imply equal underlying implementation: the previously inspected v4.2.2 wrapper still called `actions/attest` v4.2.1.

### 3. JSON Schema compatibility

Move the workspace declaration, root lock and fuzz lock together. Recheck the current upstream 0.x target rather than silently inheriting the old `0.49.9` or a newer line. Compile all workspace consumers and exercise schema compilation, identity/canonicalization, validation/error behavior and fuzz-runtime parity. Treat a 0.x line change as an API migration until measured otherwise.

### 4. Base64 hostile-input compatibility

[...]

PR #2735, opened 2026-09-01
ci: bounded Codex app-server host-proof driver (#2684)

Rul1an:
## Description

Synthetic Codex app-server driver, fail-closed validator, fake server, and Node suite for issue #2684. This is proof **instrumentation**, not a live host close-out. **Draft.** Not ready. Not merged.

This head closes the **classified driver/error accept-then-ignore** false-green recorded on `cc70adc1c570e023928a4e3b64923c2a32c6d33a` ([comment 5493771515](https://github.com/Rul1an/assay/pull/2735#issuecomment-5493771515), validated smallest shape [comment 5493922468](https://github.com/Rul1an/assay/pull/2735#issuecomment-5493922468)). The `cc70adc1` review, and every earlier review, does **not** carry. A new independent review is required on `cc23e42ac9742387c9e4c03046b67be4f9ab968b`.

The ID-presence repair on `cc70adc1` remains pinned prior evidence and is preserved (65/65 plus this one). Typed schemas inside classify-once, closed-world / F1–F5, and the prior P1s remain pinned **prior** evidence. They are **not** the only content of this head.

### Current slice — classified driver/error topology contradiction (this commit)

Principle: **classify-then-ignore is not a closed taxonomy.** `classifyStoredEvent` already accepts `{direction:"driver",method:"error"}`. `consumeClassifiedEvent` now collects that classified row once as `topology.driverErrors`. One topology contradiction is added when the collection is nonempty. `classifyDriver(meta, topology)` returns **fail** for that case. A valid stream plus a recorded driver failure cannot remain 9/9 or proof-valid.

1. **Collect once.** Classified `driver/error` rows become `topology.driverErrors`.
2. **One contradiction.** A nonempty collection is one topology reason, not a broader driver taxonomy.
3. **`driverCompleted=fail`.** `classifyDriver` fails when those rows exist; `validateProofRoot` is false.
4. **No schema growth.** Schema stays `assay.codex-host-proof.v2`. This is not a broad driver schema, not a JSON-RPC/protocol claim, and not support growth.

Schema remains `assay.codex-host-proof.v2`. Production spawn/probe literals remain `"codex"` and `"assay-mcp-server"`. Documented hard per-binary ceiling stays **512 MiB** (`HARD_MAX_SNAPSHOT_BYTES = 512 * 1024 * 1024` = `536870912`).

## Type of Change
- [x] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor

[...]

PR #2736, opened 2026-09-01
fix(ci): supervise Python run_bounded stdin under the absolute deadline (#2193)

### LINKED ISSUE #2193: fix: cover stdin delivery in the Python run_bounded absolute deadline
## Origin

Non-blocking Claude Code Desktop review finding on PR #2192, reviewed at exact head `8b9f9e123d35e50fa3ca790c5aedc078b3934faa` and carried forward for re-review after the CodeQL fix.

## Problem

`scripts/ci/claude_plugin_install_workflow.py` independently implements the same process-boundary invariants established by #2189/#2190 in `tests/support/bounded_process.rs`: one absolute deadline, separate bounded stdout/stderr, process-tree termination, and handling for a direct child that exits while a descendant retains an output pipe.

The implementations serve different harnesses and should not be forcibly shared across Rust and Python. The risk is silent semantic drift when one changes without the other being considered.

## Current classification: P2, post-MVP

A [bounded reproduction on main](https://github.com/Rul1an/assay/issues/2193#issuecomment-5478655283) now shows synchronous stdin writing can exceed the configured timeout before the Python deadline starts. This is a concrete limiter discrepancy, not only maintenance provenance. The current Claude request producer emits 622 fixed bytes; no hang of that actual host journey is claimed. Preserve the existing timeout/output/orphan controls and prioritize the #2194 first-run/authenticated acceptance work first.

## Original re-entry triggers (now supplemented by measured divergence)

- either process limiter changes semantics; or
- native Windows support is proposed for the Claude install workflow.

## Acceptance

- [ ] Keep deadline observation active while stdin is pending, without blocking output draining. Moving the timestamp alone is insufficient.
- [ ] Add finite non-reading-child and bidirectional-backpressure controls; enforce the named timeout/failure contract and preserve clean and early-pipe-close behavior.

- Add a source-level cross-reference between the Python limiter and #2189/#2190 or `tests/support/bounded_process.rs`.
- Record a short parity table for the invariants both implementations must preserve.
- If practical, add one table-driven parity test; do not create a cross-language abstraction solely to remove duplication.
- Keep the Python workflow's macOS/Linux-only non-claim explicit unless Windows process-tree ownership is implemented and tested.

Rul1an:
Fixes #2193.

Wait-gate: `origin/main` `2b1339e309c7027d63970dd5913b28b3bab36409`. Head `cb4771fba2f895af3a55efae2c4558dfa7fe5cb9` (merge-not-rebase of that main; tree `0807c5f774a22eb73c6329b979052dc9a14f272f`). Draft only; do not ready or merge. New head needs a fresh independent review. #2193 stays open until final review.

## Change
`run_bounded` in `scripts/ci/claude_plugin_install_workflow.py` armed the absolute deadline after a blocking stdin `write`/`flush`/`close`. A non-reader with 256KiB stdin could fill the pipe and stall before the clock started. The deadline is now armed immediately after `Popen`, and stdin is delivered nonblocking on the existing POSIX selector (one event loop). `killpg`, output ceilings, orphan drain, and BrokenPipe behavior are unchanged.

`fcntl` is imported lazily inside the nonblocking helper. Importing the module stays safe on non-POSIX hosts. `require_bounded_supervisor` now runs before `Popen` and names unavailable `killpg` / `fcntl` as `WorkflowError`, so unsupported hosts refuse before spawn. Windows process-tree semantics are not implemented.

If `_set_nonblocking` or selector registration fails after `Popen`, the process tree is terminated and stdin/stdout/stderr are closed before re-raising. All three pipe file objects are closed after selector ownership ends on every normal and error path.

Child-not-leaked is proven with a pidfile plus `os.kill(pid, 0)`, not `/proc`, so the focused suite runs on Darwin.

A required-hook table pins both `claude-plugin-install-workflow-self-test` and `claude-plugin-run-bounded-stdin`. Deleting or retargeting either goes red. Exit 7 plus `err-sentinel` is preserved under default rejection and `allowed_codes=(7,)`.

## TDD
RED on current `run_bounded` (256KiB stdin, 0.05s deadline): finite non-reader returned success (false-green); write-first stdout producer hung past the 1s wall. Reader, empty-stdin timeout, early stdin close, and stdout >1MiB ceiling were already green.

GREEN after the selector path: both backpressure cases fail by the bounded deadline (elapsed < 0.35s sleep); reader still succeeds; empty stdin still times out; empty immediate-exit still succeeds; early close still succeeds; stdout ceiling still terminates the tree.

Manual reversal (not in-tree): restore blocking `write`/`flush`/`close` before the deadline. The production non-reader and write-first tests go RED again (false-green or hang beyond budget). Empty production control stays green. No second `run_bounded` copy is shipped.

This follow-on: RED without `require_bounded_supervisor` (hiding `killpg` still spawned; hook table missing). GREEN after the shared preflight, pidfile reap, hook table, and exit-7 pins. Mutations that bite: reintroduce `/proc` scan without `/proc`; delete/retarget either required hook; disable preflight so `Popen` runs; hide `killpg`/`fcntl`; swallow/replace exit 7 or stderr. Deadline and empty-stdin controls stay green.

## Follow-on: upstream advance only
Merge commit `cb4771fba2f895af3a55efae2c4558dfa7fe5cb9` parents `9938f51daa7688103804f362c4b40a6661a4dfb1` and `2b1339e309c7027d63970dd5913b28b3bab36409`. `HEAD^{tree}` is `0807c5f774a22eb73c6329b979052dc9a14f272f`. Main advance vs PR base `3dd32b8f` is only `scripts/ci/codex_host_proof_validator.mjs` and `scripts/ci/test_codex_host_proof.mjs`. Intersection with the three #2736 paths is empty; those blobs are unchanged vs `9938f51d`. No product edit.

## Follow-on: effective wiring
A remote `repo` still kept the required ids green. Each required hook must now appear exactly once under `repo: local`. Self-test stages must be exactly `[pre-push]`. Bounded-stdin stages are explicit `[pre-commit, pre-push]`, and its compiled `files` regex must match `scripts/ci/claude_plugin_install_workflow.py`, `scripts/ci/test_run_bounded_stdin.py`, and `.pre-commit-config.yaml`. Same `repos[*].hooks[*]` parse; no broader YAML policy.

## Follow-on: scalar decoy is not a hook
Removing both required hook mappings and pasting their text into a top-level YAML block scalar kept the raw-text guard green. `assert_required_claude_plugin_hooks` now loads `.pre-commit-config.yaml` with the same Ruby `YAML.safe_load` route as `scripts/ci/check-assay-action-pin.sh` and walks `repos[*].hooks[*]`. Exactly one mapping per required id, with the exact `entry`. Comments and scalar values do not count. Missing ruby or invalid YAML fails closed. No new Python dependency.

## Follow-on: commented `- id:` is not a hook
Unanchored `config.find("id: ...")` stayed green when either real `- id:` line was commented out. Discovery now requires exactly one uncommented YAML list item `^[ \t]*- id: <hook>[ \t]*$` for both required ids. Comment-out of each line goes red; a commented duplicate next to the real item stays green.

## Evidence
- Linux: `PYTHONWARNINGS=error::ResourceWarning python3 scripts/ci/test_run_bounded_stdin.py` 34/34 (kept the previous 29, plus remote-repo, `[manual]` stages, `files: ^$`, and an unrelated-local no-op); `test_claude_plugin_install_import.py`; both hook entries; `git diff --check` clean. `pre-commit` not installed on this host, so no `validate-config` / live hook run here.
- Darwin 3.14.3, no `/proc`: same 22 stdin tests plus import-safe, 2.549s.

## Files
- `scripts/ci/claude_plugin_install_workflow.py`
- `scripts/ci/test_run_bounded_stdin.py`
- `.pre-commit-config.yaml` (discovery hook already present; entries pinned by the required-hook table, file itself unchanged in this follow-on)

## Shared invariants with `tests/support/bounded_process.rs` (#2189/#2190)

POSIX-only. Same invariants, not a cross-language abstraction. No docs file and no table-driven cross-language harness.

| invariant | Python `run_bounded` | Rust `bounded_process.rs` |
| --- | --- | --- |
| absolute deadline | armed before any stdin I/O; selector `select` is capped by remaining time | `Instant::now() + limits.timeout` governs the wait |
| separate stdout/stderr caps | `MAX_BYTES` per stream; oversize terminates the tree | `max_stdout_bytes` / `max_stderr_bytes` |
| process-tree termination | `killpg` via `start_new_session` | Unix process group / Windows Job Object |
| descendant-held pipe / orphan drain | same absolute deadline continues after the direct child exits | descendants remaining in the group/job are terminated before return |
| stdin supervision | nonblocking writes on the same selector under that deadline | stdin writer is included in outstanding-deadline state |

## Non-claims
- Not a reproduced Claude-host hang (real initialize payload was 622 bytes, not 256KiB).
- POSIX-only; Windows / non-POSIX is a non-claim for this Python helper. Import is safe there; `run_bounded` refuses before spawn and does not grow Windows process-tree support.
- Not a shared abstraction with the Rust helper; comment-only cross-reference.
- No Cargo, auth, Claude host/model call, or unrelated cleanup.
- #2474 remains parked; Cursor owns that side.
- #2193 is not closed until final review/merge.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -221,6 +221,13 @@ repos:
         pass_filenames: false
         files: ^(scripts/ci/(claude_plugin_install_workflow|test_claude_plugin_install_import)\.py|\.pre-commit-config\.yaml)$
 
+      - id: claude-plugin-run-bounded-stdin
+        name: Claude plugin bounded stdin is deadline-supervised
+        entry: python3 scripts/ci/test_run_bounded_stdin.py
+        language: system
+        pass_filenames: false
+        files: ^(scripts/ci/(claude_plugin_install_workflow|test_run_bounded_stdin)\.py|\.pre-commit-config\.yaml)$
+
       - id: docs-auto-update-reconcile-self-test
         name: Docs auto-update reconciliation self-test
         entry: bash scripts/ci/test-reconcile-docs-auto-pr.sh
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -226,6 +226,7 @@ repos:
         entry: python3 scripts/ci/test_run_bounded_stdin.py
         language: system
         pass_filenames: false
+        stages: [pre-commit, pre-push]
         files: ^(scripts/ci/(claude_plugin_install_workflow|test_run_bounded_stdin)\.py|\.pre-commit-config\.yaml)$
 
       - id: docs-auto-update-reconcile-self-test
--- a/scripts/ci/claude_plugin_install_workflow.py
+++ b/scripts/ci/claude_plugin_install_workflow.py
@@ -1466,17 +1466,26 @@ REQUIRED_CLAUDE_PLUGIN_HOOKS: tuple[tuple[str, str], ...] = (
 )
 
 
+_REAL_HOOK_ID_LINE = re.compile(r"^([ \t]*)- id: (\S+)[ \t]*$", re.M)
+
+
 def _hook_block_by_id(config: str, hook_id: str) -> str:
-    marker = f"id: {hook_id}"
-    start = config.find(marker)
-    if start < 0:
+    matches = [m for m in _REAL_HOOK_ID_LINE.finditer(config) if m.group(2) == hook_id]
+    if not matches:
         fail(
             "hook_files",
             f"pre-commit config lost {hook_id}",
             "restore the Claude plugin required hook",
         )
-    nxt = config.find("\n      - id:", start + len(marker))
-    return config[start:] if nxt < 0 else config[start:nxt]
+    if len(matches) != 1:
+        fail(
+            "hook_files",
+            f"pre-commit config has {len(matches)} real list items for {hook_id}",
+            "keep exactly one uncommented YAML list item per required hook id",
+        )
+    start = matches[0].start()
+    nxt = _REAL_HOOK_ID_LINE.search(config, matches[0].end())
+    return config[start:] if nxt is None else config[start:nxt.start()]
 
 
 def _claude_plugin_hook_block(config: str) -> str:
--- a/scripts/ci/claude_plugin_install_workflow.py
+++ b/scripts/ci/claude_plugin_install_workflow.py
@@ -102,6 +102,16 @@ def terminate_tree(process: subprocess.Popen[bytes]) -> None:
         process.wait()
 
 
+def _close_bounded_pipes(process: subprocess.Popen[bytes]) -> None:
+    for stream in (process.stdin, process.stdout, process.stderr):
+        if stream is None:
+            continue
+        try:
+            stream.close()
+        except (BrokenPipeError, OSError):
+            pass
+
+
 def _set_nonblocking(phase: str, fd: int) -> None:
     try:
         import fcntl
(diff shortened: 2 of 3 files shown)


Rul1an:
## Early blocker hunt — repair required at `1947e8b719473e61acbc29e4234f024bd1423608`

This is an independent read-only three-dot review on base `c06ca57c2a15211ede275ab3fd617a001f28df30`. It is **not final quorum**: #2735 must land first and any advance creates a new head.

### Findings

1. **P1 — default hook fails on macOS.** `scripts/ci/test_run_bounded_stdin.py:332` enumerates `/proc`; the exact-head focused suite on macOS/Python 3.14.3 produced 10 passes and `FileNotFoundError: '/proc'`. `.pre-commit-config.yaml:224` installs this at the default stage. Replace `/proc` discovery with a portable, captured-PID/process-ownership assertion.
2. **P1 — deleting the new hook is false-green.** The existing owner guard at `scripts/ci/claude_plugin_install_workflow.py:1433` covers only `claude-plugin-install-workflow-self-test`. Removing `claude-plugin-run-bounded-stdin` leaves that guard green. Pin the new hook's ID, entry, stage and effective selector, with a deletion mutation.
3. **P2 — unsupported-host cleanup can mask the real error and leak the child.** `run_bounded` spawns before its POSIX prerequisites are established; its exception cleanup calls `terminate_tree`, which unconditionally calls `os.killpg`. A simulated host without `killpg` returned `AttributeError` and left the child alive instead of preserving the intended unsupported-host diagnostic. Fail explicitly before `Popen` on unsupported hosts, or make cleanup unable to mask the primary failure.
4. **P2 — return-code/diagnostic preservation is live but unpinned.** Exact-head probes showed default rejection preserves `command exited 7: err-sentinel`, and `allowed_codes=(7,)` returns code 7 with both streams. Add acceptance pairs so later selector changes cannot erase either property.

### Executed evidence

- focused suite: **10 pass, 1 fail** on macOS arm64;
- import-safety: 3/3;
- AST parse, pre-commit config validation and `git diff --check`: pass;
- base non-reader/write-first mutations false-cleaned as expected; empty-input no-op remained green;
- hook-deletion mutation survived the current owner guard;
- no probe-child residue remained after the review.

DoD for the eventual post-#2735 head: repair all four, run the focused suite on Linux and macOS, prove unsupported-host refusal occurs before spawn or without leakage, make hook deletion bite, then obtain a fresh exact-head review. No Cargo, Claude/auth or Windows claim was made here.

Rul1an:
BLOCKED at exact head `cb9d06e52b51887f6f707c99e3f221fe36660a72` (base `3dd32b8f452f48d7aac736b781a68ccc3417edc4`).

Independent reviewer `01a05eae-d1e8-7482-b687-97e091b392ab`; authored/edited neither the change nor its governing normative specification or implementation plan.

### P1 finding

`scripts/ci/claude_plugin_install_workflow.py:1470` uses an unanchored textual `config.find("id: ...")`, and the following `entry:` search accepts a commented YAML hook. In scratch, commenting out either actual `- id:` line left the hook undiscoverable to `pre-commit`, while `assert_required_claude_plugin_hooks()` still returned clean. The current deletion/retarget tests replace textual occurrences and do not exercise semantic comment-out deletion.

Required repair: structurally recognize a unique real hook item (or equivalently strict-anchor the actual list-item shape) and add comment-out mutations for both `claude-plugin-install-workflow-self-test` and `claude-plugin-run-bounded-stdin`.

### Evidence

Unmutated on Darwin/Python 3.14.3: stdin suite 22/22, import suite 3/3, install self-test, both direct pre-commit hooks, and `git diff --check` passed. Ten existing load-bearing mutations were killed. Two comment-out hook-ID mutations survived the guard while actual pre-commit discovery failed with `no hook`.

No Cargo/Rust build, credentialed Claude journey, Windows process-tree claim, or publication claim.

Rul1an:
## Independent exact-head review: BLOCKED

**BLOCKED** at `9f85ec287f512a298de6ab48b07c90712b697864` on base `3dd32b8f452f48d7aac736b781a68ccc3417edc4`.

### [P1] Hook guard accepts non-hook YAML text

`scripts/ci/claude_plugin_install_workflow.py:1469` scans raw text globally rather than verifying `repos[*].hooks[*]`. In a disposable scratch clone, both real hook blocks were removed and identical `- id:` / `entry:` text was placed inside a valid top-level `exclude: |` scalar. `assert_required_claude_plugin_hooks()` and `pre-commit validate-config` both exited 0, while both actual `pre-commit run` commands failed with `No hook with id`.

Required repair: parse/inspect the actual YAML hook structure and add this scalar-decoy regression. Another raw-text regex is not a sufficient disposition.

### Verified evidence

- Wait-gate and final reread: GitHub and remote head both exact `9f85ec287f512a298de6ab48b07c90712b697864`.
- Full three-dot scope was exactly the three PR files.
- Focused stdin suite: 25/25 pass.
- Import-safe suite: 3/3 pass.
- Direct workflow self-test and both real hook executions pass.
- Commenting either real hook ID separately fails; a commented duplicate beside a real item remains green.
- Removing the regex start anchor makes the relevant tests fail.
- Independent 256 KiB non-reader probe expires in 0.055s with child absent; reader/empty controls pass.
- Exit 7 is rejected by default and accepted only via `allowed_codes=(7,)`, preserving `err-sentinel`.
- Darwin execution passes without production `/proc` reliance.
- `git diff --check` passes.

### Independence and non-claims

The reviewer authored or edited neither this PR, issue #2193, nor its governing design. Mutations were limited to a disposable scratch clone. Nothing was edited, committed, pushed, posted, or merged by the reviewer.

No hosted-proof, Windows process-tree, real Claude-host hang, mergeability, or merge claim. Cargo was not used.

Rul1an:
**BLOCKED** at exact head `2343fd777e0b5a9a8daa94da4cd626d5f5b0b836` (base `3dd32b8f452f48d7aac736b781a68ccc3417edc4`).

Independent reviewer declaration: I authored or edited neither this PR, issue #2193, nor its governing design. The review used a fresh detached clone and separate scratch copies only.

### Findings

1. **P1 — repository ownership is lost while walking hooks** (`scripts/ci/claude_plugin_install_workflow.py:1575`). Moving both unchanged required mappings into the first remote repository leaves the guard and `pre-commit validate-config` green, but real execution fails because those IDs are absent from that repository's manifest. Require each required hook to belong to `repo: local` and pin this mutation.
2. **P1 — effective stage and selector are not validated** (`scripts/ci/claude_plugin_install_workflow.py:1598`). `stages: [manual]` stays guard-green but removes the bounded hook from the normal pre-commit stage. `files: ^$` stays guard-green and real execution reports `Skipped (no files to check)`. Pin the intended stage and behaviorally prove the selector matches the owning workflow, test, and config paths.

### Verified

- Start and final live pins matched the expected base/head; full three-dot diff was exactly the three declared files.
- The prior scalar-decoy is closed: valid YAML with both real hook blocks removed and their exact text retained only in a block scalar now fails the guard and real hook lookup.
- Comments, unrelated mappings, duplicates, malformed YAML, missing Ruby, and missing/changed entries fail appropriately.
- Unchanged config passes the guard, `pre-commit validate-config`, and both real hook runs.
- Bounded-stdin suite 29/29 with `ResourceWarning` as error; import-safe 3/3; direct workflow self-test, bounded test, both real hook runs, and `git diff --check` pass.
- Alias/merge syntax is conservatively refused by `aliases: false`; no clean consumer false-green was found there.

### Mutation evidence

- No-op control: 4/4 selected guard tests pass.
- Structural-selection bypass is killed by `test_mutation_scalar_decoy_goes_red`.
- Uniqueness bypass is killed by `test_mutation_duplicate_real_hook_goes_red`.
- Exact-entry bypass is killed by `test_mutation_retarget_stdin_hook_goes_red`.
- Wrong-repository, manual-stage, and empty-selector mutations survive and are the blockers above.

Non-claims: no Cargo/Rust execution, Windows process-tree claim, credentialed Claude journey, real host-hang reproduction, hosted-CI readiness, mergeability, or release claim.

Rul1an:
## Independent exact-head review: BLOCKED

**Head:** `9938f51daa7688103804f362c4b40a6661a4dfb1`  
**Base:** `3dd32b8f452f48d7aac736b781a68ccc3417edc4`

Independence: the reviewer authored or edited neither this change nor its governing specification or implementation plan. Review and mutations ran only in a disposable detached clone. No writer worktree, branch, PR state, or GitHub content was modified by the reviewer.

### P1: selector contract can still satisfy itself

`scripts/ci/test_run_bounded_stdin.py:748` tests the hook selector against `STDIN_HOOK_FILE_PATHS`, a production-owned path list in `scripts/ci/claude_plugin_install_workflow.py:1565`. Removing `.pre-commit-config.yaml` from both the real hook selector and that list leaves **34/34 tests green**. The hook can therefore stop guarding its own configuration while its contract reports clean.

Smallest coherent fix: add a test-owned literal contract for both hook IDs, entries, exact stages, and the three required selector paths. Independently parse the config and assert the compiled selector matches each literal. Keep the coupled selector/list weakening as a RED regression.

### Verified

- Required hooks are currently unique mappings under `repo: local`.
- Current entries and stages are correct.
- Current selector compiles and reaches all three intended paths.
- Remote-repo, manual-stage, `files: ^$`, no-op entry, comment/scalar decoy, duplicate/missing hook, missing Ruby, and malformed YAML mutations go red.
- No-op and unrelated-hook controls remain green.
- `34/34` focused tests; `3/3` import-safe; `pre-commit validate-config`; direct stage/hook executions; `git diff --check`; clean restored scratch.

### Mutation that survived

Coupled removal of `.pre-commit-config.yaml` from the selector and `STDIN_HOOK_FILE_PATHS`: **34/34 green**.

Non-claims: no Cargo, hosted-CI completion, merge-readiness, Windows process-tree, authenticated host/model, release, or runtime-safety claim.

**Verdict: BLOCKED at `9938f51daa7688103804f362c4b40a6661a4dfb1`.**

Rul1an:
BLOCKED at exact head `cb4771fba2f895af3a55efae2c4558dfa7fe5cb9` (base `2b1339e309c7027d63970dd5913b28b3bab36409`).

Independent exact-head review. I did not author or edit this change or its governing design/plan.

## Findings

1. **P1: selector ownership can satisfy itself.** `STDIN_HOOK_FILE_PATHS` is production-owned in `scripts/ci/claude_plugin_install_workflow.py`, while the test only proves that `files: ^$` fails. Removing any required path from both the production tuple and the hook regex leaves all 34 focused tests green. Add a test-owned literal table and mutate each required path while leaving the production oracle unchanged.

2. **P1: direct-child kill can replace process-tree kill without a red test.** Replacing `os.killpg(...)` with `process.kill()` left the focused suite and pre-push self-test green while two tagged descendants survived. The current PID-file assertion is optional and witnesses only the direct child. Add a mandatory descendant PID witness, liveness assertion, bounded cleanup, and a direct-kill mutation.

Smallest repair: RED first for the three independently owned selector paths and for a descendant that survives direct-child termination. Keep the POSIX process-group scope; do not expand this into Windows support or escaped-process containment.

## Evidence

- `python3 scripts/ci/test_run_bounded_stdin.py`: 34/34 pass.
- `python3 scripts/ci/test_claude_plugin_install_import.py`: 3/3 pass.
- `bash scripts/ci/test-claude-plugin-install.sh --self-test`: pass.
- Pre-commit validation plus both changed hooks at their configured stages: pass.
- Exact-head hosted `lane-check/proof`: success, delegated gate `none`.
- Broad `CI` and `host-capability-check`: success.
- No Cargo, rustc, clippy, real authenticated Claude invocation, edits, pushes, dispatches, or merge actions were performed by the reviewer.

A new push requires a fresh exact-head review.

Rul1an:
@GrokBt writer handoff for `ruley/2193-bounded-stdin` at `cb4771fba2f895af3a55efae2c4558dfa7fe5cb9`.

The implementation work is strong: current behavior, broad CI, host capability and exact-head lane proof are green. The independent review isolated two remaining proof gaps rather than a production regression:

1. move the three required selector paths into a test-owned literal oracle and mutate production/config while leaving that oracle unchanged;
2. make descendant PID creation mandatory and prove direct-child kill leaves the test red while process-group kill reaps it.

Keep the POSIX process-group scope; no Windows or escaped-process expansion. RED first, smallest repair, focused suites, then push. A new head needs a fresh independent review.

Rul1an:
Superseded by draft #2746.

The successor preserves this branch's implementation commits and exact provenance, merges current `main`, and adds the two test-only closure repairs requested at `cb4771fba2f895af3a55efae2c4558dfa7fe5cb9`:

- a test-owned selector oracle that remains red when production/config coupling is deleted together;
- a mandatory descendant PID/PGID witness that kills a direct-child-only termination mutant.

Successor head at creation: `c236584cfa4ea4a8491ae76cb54c77bf1f4b6306`. This draft must not be rebased or merged independently; its earlier review records remain bound to their named heads.

Issue #2764, opened 2026-09-03
release: Assay 6.0 bounded major programme for ADR-048


Rul1an:
## Decision

Open the Assay 6.0 release line as a bounded major-change programme for ADR-048: move `ClaimDecision` and `ClaimKind` into `assay-common` while retaining the four old public paths as re-exports. Decision tables, folds and runtime readings stay in their owning crates.

ADR-047 is not pending work in this programme. PR #2139 merged its `PayloadSessionFinding` variant and `#[non_exhaustive]` change as `c00b46bf72d6999b15304a626443959ddbfbcf29`; that commit is contained in `v5.0.0` and every later release. The earlier version of this ledger and the 6A changelog line incorrectly described ADR-047 as a second future 6.0 break. 6B must correct that changelog line; it must not reimplement ADR-047.

This ledger does not authorize unrelated CLI removals, policy/trace wiring, a fourth claim decision, generic identity/delegation work, or broader evidence claims.

## Why a major is required

Measured against published `v5.5.2`, the parked ADR-048 migration reports `enum_missing` for both affected public crates:

- `assay-runner-schema`: `CoverageClaimKind`, `ClaimGateDecision`
- `assay-evidence`: `CodingAgentClaimKind`, `CodingAgentGateDecision`

The old paths can keep resolving and the snake_case wire values can remain byte-stable, but type identity changes. A downstream crate can currently implement a local trait separately for the two external enum types; after unification those implementations overlap and Rust rejects them with `E0119`. That is a real source break, not merely a tool limitation. ADR-048 alone therefore requires 6.0.

## Historical correction: ADR-047

The ADR-047 major was already paid in 5.0.0:

- PR #2139 added `Payload::SessionFinding` and marked `Payload` `#[non_exhaustive]`.
- Merge commit `c00b46bf72d6999b15304a626443959ddbfbcf29` is an ancestor of tag `v5.0.0`.
- `session_finding_bundle_roundtrip.rs` pins the registered tag through the public `Payload` view and the real bundle write/verify path.
- Later releases preserve both source changes.

No 6C implementation branch is needed. This correction does not claim that a command emits session findings today; ADR-047 explicitly leaves producer wiring undecided.

## Release choreography

The workspace source version and published install pin remain separate facts. `v5.5.2` is currently published and install-pinned. During 6.0 preparation, the install pin stays on `v5.5.2` until a new release's required assets, checksums and attestations are retrievable.

| Slice | Scope | Required evidence |
[...]

PR #2768, opened 2026-09-03
ci: retry one transient Assay Action YAML parser timeout

### LINKED ISSUE #2767: ci: retry one transient Assay Action YAML parser timeout
Parent programme: #2764. Blocks release-candidate freeze issue #2766 and the ADR-048 writer's normal hooked checkpoint.

## Measured failure

On `origin/main=ef35de6b0839c6566cce226fba05a1c48440ab84`, `scripts/ci/test-check-assay-action-pin.sh` invokes the owner checker 43 times. Each checker starts a bounded Ruby YAML parser for the pinned action and each discovered workflow/snippet. Under transient host scheduler pressure, four normal pre-commit runs failed on different unchanged snippets with `YAML parse timed out`; every other hook passed. Two immediately adjacent standalone owner-check runs completed successfully in 13.06s and 12.36s.

This is a false-red in the guard infrastructure, not an ADR-048 content failure. Blind reruns and `--no-verify` are both forbidden.

## Bounded fix

- Keep the 30-second timeout per parse attempt.
- Retry exactly once, and only after `subprocess.TimeoutExpired`.
- Syntax errors, non-zero parser exits, missing Ruby, invalid JSON and other failures remain single-attempt fail-closed results.
- After a second timeout, retain the existing `YAML parse timed out` diagnostic.
- Do not alter action-pin authority, owned paths, published-byte verification, YAML semantics or hook routing.

## Required evidence

- RED-first self-test with an injected first-attempt timeout and successful second attempt; it must fail on current main for the intended reason.
- Mutation removing the retry makes that test red.
- Persistent timeout remains red and bounded.
- Malformed YAML is not retried.
- Full action-pin owner self-test and pre-commit consumer hook pass.
- Independent exact-head review before merge.

Non-claims: this does not make YAML parsing faster, remove the timeout, justify unlimited retries, or validate ADR-048.

Rul1an:
## Summary

Retry the Ruby GitHub Actions YAML parser exactly once after `subprocess.TimeoutExpired` so one transient process-start stall cannot fail every Assay PR. All non-timeout failures remain single-attempt and fail closed.

Fixes #2767
Parent programme: #2764

## Scope

- `scripts/ci/check-assay-action-pin.sh`
- `scripts/ci/test-check-assay-action-pin.sh`

## Contract

- Keep the production parser deadline at 30 seconds per attempt.
- Retry only `subprocess.TimeoutExpired`, at most once.
- A second timeout preserves the existing `YAML parse timed out` failure.
- Missing Ruby, parser non-zero exit, invalid parser JSON, and invalid YAML are not retried.
- Do not change Action pins, workflow ownership, YAML acceptance, published-byte checks, or network behavior.

## Evidence

RED before the production change:

```text
== bounded YAML parser timeout retry ==
FAIL: one transient timeout is retried exited non-zero:
pinned action.yml: YAML parse timed out
```

GREEN on final local head:

- bounded timeout matrix: transient timeout recovered; persistent timeout failed after exactly 2 attempts; ordinary parser failure failed after exactly 1 attempt;
- `bash scripts/ci/test-check-assay-action-pin.sh`: PASS, including the existing mutation battery;
- `bash scripts/ci/check-assay-action-pin.sh`: PASS;
- normal hooked commit: PASS;
- `shellcheck -x -e SC2016 scripts/ci/check-assay-action-pin.sh scripts/ci/test-check-assay-action-pin.sh`: PASS;
- `git diff --check`: PASS.

Mutation: changing `for attempt in range(2)` back to `range(1)` makes `one transient timeout is retried` fail before semantic evaluation. The scratch control remained on the unmodified tree.

## Non-claims

- This does not make the parser tolerant of invalid YAML.
- This does not increase or remove the 30-second per-attempt resource bound.
- This does not prove the host is healthy or eliminate all scheduler stalls.
- No Rust product behavior, release asset, install pin, or ADR-048 decision table changes here.
- Merge and release claims remain contingent on final-head review and required GitHub checks.

## Exact candidate

Head: `d93db3aef690c614819ea9c2dc43f240d2c24d42`

### DIFF
--- a/scripts/ci/check-assay-action-pin.sh
+++ b/scripts/ci/check-assay-action-pin.sh
@@ -158,18 +158,21 @@ def parse_provenance(text: str) -> tuple[str, str]:
 
 
 def load_yaml(text: str, *, source: str, required: bool = True):
-    try:
-        proc = subprocess.run(
-            ["ruby", "-EUTF-8:UTF-8", "-e", RUBY_SAFE_LOAD],
-            input=text,
-            capture_output=True,
-            text=True,
-            timeout=30,
-        )
-    except FileNotFoundError:
-        fail("ruby is required to parse GitHub Actions YAML")
-    except subprocess.TimeoutExpired:
-        fail(f"{source}: YAML parse timed out")
+    for attempt in range(2):
+        try:
+            proc = subprocess.run(
+                ["ruby", "-EUTF-8:UTF-8", "-e", RUBY_SAFE_LOAD],
+                input=text,
+                capture_output=True,
+                text=True,
+                timeout=30,
+            )
+            break
+        except FileNotFoundError:
+            fail("ruby is required to parse GitHub Actions YAML")
+        except subprocess.TimeoutExpired:
+            if attempt == 1:
+                fail(f"{source}: YAML parse timed out")
     if proc.returncode != 0:
         if not required:
             return None
--- a/scripts/ci/test-check-assay-action-pin.sh
+++ b/scripts/ci/test-check-assay-action-pin.sh
@@ -110,6 +110,74 @@ expect_ok() {
   echo "ok    ${name}"
 }
 
+prepare_yaml_timeout_tree() {
+  local dest="$1"
+  copy_into "${dest}"
+  mkdir -p "${dest}/scripts/ci" "${dest}/fake-bin"
+  cp "${CHECKER}" "${dest}/scripts/ci/check-assay-action-pin.sh"
+  cp "${READER}" "${dest}/scripts/ci/read-assay-action-pin.sh"
+  python3 - "${dest}/scripts/ci/check-assay-action-pin.sh" <<'PY'
+from pathlib import Path
+import sys
+
+path = Path(sys.argv[1])
+text = path.read_text(encoding="utf-8")
+needle = "            timeout=30,"
+if text.count(needle) != 1:
+    raise SystemExit("expected exactly one YAML parser timeout")
+path.write_text(
+    text.replace(
+        needle,
+        '            timeout=5 if source == "pinned action.yml" else 30,',
+    ),
+    encoding="utf-8",
+)
+PY
+  cat >"${dest}/fake-bin/ruby" <<'PY'
+#!/usr/bin/env python3
+import os
+from pathlib import Path
+import sys
+import time
+
+state = Path(os.environ["ASSAY_FAKE_RUBY_STATE"])
+attempt = int(state.read_text(encoding="utf-8")) + 1 if state.exists() else 1
+state.write_text(str(attempt), encoding="utf-8")
+mode = os.environ["ASSAY_FAKE_RUBY_MODE"]
+if mode == "persistent-timeout" or (mode == "transient-timeout" and attempt == 1):
+    time.sleep(6)
+elif mode == "parser-failure":
+    print("synthetic parser failure", file=sys.stderr)
+    raise SystemExit(2)
+os.execv(os.environ["ASSAY_REAL_RUBY"], [os.environ["ASSAY_REAL_RUBY"], *sys.argv[1:]])
+PY
+  chmod +x "${dest}/fake-bin/ruby"
+}
+
+run_yaml_timeout_checker() {
+  local tree="$1" mode="$2" state="$3"
+  PATH="${tree}/fake-bin:${PATH}" \
+    ASSAY_REAL_RUBY="${REAL_RUBY}" \
+    ASSAY_FAKE_RUBY_MODE="${mode}" \
+    ASSAY_FAKE_RUBY_STATE="${state}" \
+    ASSAY_ACTION_TREE="${tree}" \
+    ASSAY_ACTION_PIN_FILE="${tree}/.github/assay-action-pin" \
+    ASSAY_ACTION_FIXTURE_FILE="${tree}/scripts/ci/fixtures/assay-action-pin/action.yml" \
+    ASSAY_ACTION_PROVENANCE_FILE="${tree}/scripts/ci/fixtures/assay-action-pin/PROVENANCE" \
+    "${tree}/scripts/ci/check-assay-action-pin.sh"
+}
+
+expect_attempts() {
+  local name="$1" state="$2" expected="$3"
+  local actual
+  actual="$(cat "${state}")"
+  if [[ "${actual}" != "${expected}" ]]; then
+    echo "FAIL: ${name} made ${actual} parser attempts; expected ${expected}" >&2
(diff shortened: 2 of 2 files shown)


Rul1an:
## Independent exact-head review — READY

Head reviewed: `d93db3aef690c614819ea9c2dc43f240d2c24d42`
Base: `ef35de6b0839c6566cce226fba05a1c48440ab84` (three-dot, ahead 1 / behind 0, 1 commit)
Files: `scripts/ci/check-assay-action-pin.sh` (+15/-12), `scripts/ci/test-check-assay-action-pin.sh` (+113/-0). Nothing else in the diff — no product code, no workflow, no fixture, no pin.

**Verdict: READY, bound to `d93db3aef690c614819ea9c2dc43f240d2c24d42`.** Head re-checked immediately before posting.

### Method

Isolated harness, no repository build and no suite run: I extracted `load_yaml` verbatim from base `ef35de6b` and from head `d93db3ae`, shrank only the 30s budget to 2s so it runs, and drove both against a fake `ruby` on `PATH` that counts invocations and can hang, exit non-zero, or emit invalid JSON. Python 3.13.5, Linux. "attempts" below is the number of parser invocations in one `load_yaml` call.

| case | head (with retry) | base (control) |
| --- | --- | --- |
| transient timeout (1st attempt hangs) | attempts=2, returns parsed document | attempts=1, fails `pinned action.yml: YAML parse timed out` |
| persistent timeout | attempts=2, fails `pinned action.yml: YAML parse timed out` | attempts=1, same message |
| parser `rc=2` | attempts=1, fails `... YAML parse failed: synthetic parser failure` | identical |
| parser `rc=2`, `required=False` | attempts=1, returns `None` | identical |
| invalid JSON on stdout | attempts=1, fails `... YAML JSON decode failed: ...` | identical |
| `ruby` missing (`FileNotFoundError`) | fails `ruby is required to parse GitHub Actions YAML`, no retry | identical |

### Findings against the stated contract

1. **Only `subprocess.TimeoutExpired` retries, exactly once.** `for attempt in range(2)` plus `if attempt == 1: fail(...)` bounds it at two attempts; the persistent case measured exactly 2, never 3.
2. **A permanent timeout keeps the same diagnosis.** `{source}: YAML parse timed out` is byte-identical to base. Worst-case wall time per document doubles (30s to 60s); no other budget changes.
3. **Non-timeout classes are not retried.** `FileNotFoundError` exits inside the loop through `fail`, which is `raise SystemExit` (L103-104); `rc != 0` and `json.JSONDecodeError` both live after the loop and are reachable only via `break`. The `required=False` return-`None` paths are unchanged.
4. **No unbound-`proc` path.** The loop can only be left by `break` (with `proc` bound) or by `fail` raising `SystemExit`. Verified statically and empirically — no `NameError` in any mode.
5. **Removing the retry is a genuine RED.** Under the transient scenario the base function fails with `YAML parse timed out` where head returns the parsed document, so `one transient timeout is retried` plus `transient_attempts >= 2` cannot pass without the loop.
6. **The shortened test budget touches `action.yml` only.** The mutation replaces the single `timeout=30,` with `timeout=5 if source == "pinned action.yml" else 30,`. The only call site passing that literal is `declared_inputs()` (L240, entered from `main()` L493); the other two sites pass per-file `source` values (`<rel>`, `<rel> snippet`, `<rel> raw-string`) that can never equal it. I applied the exact replacement to the head file: the patched line lands at indent 16 and the embedded Python block still compiles — the 12-space needle matches the 16-space line as a substring and the replacement carries the same 12 spaces, so nesting is preserved.
7. **The assertions are not document-count dependent.** The exact counts are `persistent == 2` and `parser-failure == 1`, and both hold because `declared_inputs()` is the first YAML parse in `main()` and `fail` exits before the `WORKFLOWS`/`SNIPPETS` loops. The transient case correctly uses `>= 2` rather than `==`, since later documents add attempts to the shared counter. Adding or removing workflows or snippets changes none of the three. `read-assay-action-pin.sh` invokes no parser, and the checker runs one Python process per invocation, so the counter equals `load_yaml` invocations exactly.
8. **The mutation is self-guarding.** `prepare_yaml_timeout_tree` hard-fails unless exactly one parser timeout exists, and without the 5s patch the 6s fake sleep would succeed under the 30s budget and leave attempts=1, turning the transient assertion red. The patch is load-bearing, not decorative.

### Non-blocking notes (advisory; none of these gate this PR)

1. `REAL_RUBY="$(command -v ruby)"` (L260) runs under `set -euo pipefail`, so a missing `ruby` aborts at that line with rc 1 and the following `if [[ -z "${REAL_RUBY}" ]]` message is unreachable. Still fail-closed; only the diagnostic is lost.
2. `expect_attempts "persistent timeout retry bound" 2` and `... "ordinary parser failures are not retried" 1` silently depend on `declared_inputs()` staying the first YAML parse in `main()`. A parse added earlier turns both into false reds — worth one comment naming that invariant.
3. The 5s budget against a 6s sleep also cuts the other way: the transient case's second attempt must finish real `ruby` startup plus fixture parse within 5s. A small flake edge on a heavily loaded runner, scoped to `action.yml`.
4. The new battery runs at L330, ahead of `control-is-green` at L371, so an environmental red in the new block reports before the control has proven the tree green.
5. Coverage: `rc != 0` non-retry is pinned by `attempts == 1`, but `FileNotFoundError` and JSON-decode non-retry are only structurally impossible, not pinned by a case. Low risk, since both sit outside or exit the loop.

### Non-claims

- I did not run the repository test suite, did not run `check-assay-action-pin.sh` against the real tree, and did not inspect any hosted job. Nothing here claims CI green; every number above comes from the isolated harness described in Method.
- I did not measure real `ruby`/Psych timeout behaviour, and I make no claim that the original CI timeout was transient rather than having another root cause.
- No claim about the `--published` live comparison, the upstream `Rul1an/assay-action` repository, or pin correctness: the pin file, fixture, and provenance are untouched by this diff.
- No merge, undraft, ready-for-review, or label action taken. This is one independent review bound to `d93db3ae`, not a quorum verdict.

### Independence

I did not author, edit, or advise on this branch, its commit, or a governing spec for it. My only prior contact with #2767/#2768 was a wait-gate check that found the ref unpushed.

Rul1an:
```json
{
  "head_sha": "d93db3aef690c614819ea9c2dc43f240d2c24d42",
  "builder": {"agent": "codex", "instance": "writer-1"},
  "reviewer": {"agent": "ruley", "instance": "review-1", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [],
  "no_findings": true,
  "independence": {
    "did_not_build": true,
  }
}
```

PR #2794, opened 2026-09-04
fix(ci): check crate-level path-dependency versions in check-release-surface.sh

Rul1an:
## What was wrong

The same fact — every crate in this workspace is published at the workspace version — is declared at
two sites: the root `Cargo.toml` `[workspace.dependencies]` table, and directly inside crate
manifests as `assay-x = { path = "../assay-x", version = "..." }`.

`check-release-surface.sh` checked only the first, and checked it by matching lines. So nine
crate-level declarations were never looked at, and both release-line bumps (`ef35de6b0` for 6.0,
`e2f30c5cc` for 6.1) updated them by hand with nothing verifying the result. A stale declaration
still resolves under caret semantics — `^6.0.0` is satisfied by `6.1.0` — so it costs a published
requirement looser than the truth rather than a build error, which is how it drifts unnoticed.

**RED:** with `crates/assay-core/Cargo.toml` declaring `assay-adapter-api = "5.9.0"` against a
6.0.0 workspace, the script exited **0** and reported `checked 13 declaration(s)`.

## One rule, both sites

`scripts/ci/check_internal_dep_versions.py` parses each manifest with `tomllib`. A single function,
`_version_problem`, judges every declaration. What differs between the two sites is only the base
directory a path resolves against and whether cargo requires a version there; both are computed by
the caller and passed in, so there is no second implementation to drift.

It parses rather than matching lines because two earlier line-matching attempts on this branch were
blind in the same way, and neither could show it by passing. The second cross-checked one line
pattern against another and called that an independent denominator; both required a double quote
and a literal path prefix, so `path = 'crates/assay-evidence'` — valid TOML cargo accepts — escaped
both, they agreed, and a stale version passed while the declaration count silently shrank. The same
blindness ran the other way and reported commented-out declarations and `[[bench]]` target paths as
real findings.

Parsing also let three approximations be replaced by the real rule:

- **Scope is `[workspace] members` and `exclude`.** `fuzz` is out of scope because the workspace
  says so, not because a pattern here happened to miss it.
- **Only path dependencies that resolve to a member are checked.** A vendored fork drops out of
  scope rather than being forced to the workspace version.
- **Cargo decides when a version is required.** It strips every `dev-dependencies` table on
  publish — per-target ones included — and `publish = false` (or an empty registry list) emits no
  requirement at all. A version that *is* declared is checked in every case.

The pre-commit `release-surface` selector now names the crate manifests and the checker; without
that the hook never fired on the files the rule reads.

## Verification

Measured on `3c86a20c407efb8b60dc80818c1d78190dca1358`, worktree `kind-gould-c0bc7f`.

| | |
|---|---|
| `check-release-surface.sh` | exit 0 — 13 root + 9 crate declarations |
| `test-check-release-surface.sh` | exit 0 — 124 mutations observed |
| Root site, on the real manifest | single-quoted stale, versionless, and sub-table stale each reported |
| Crate site | stale, ahead, versionless, single-quoted, sub-table each reported |
| Tables | `build-dependencies` visited; target normal and build refuse a missing version; target dev tables allow it |
| `publish = false` | suppresses only the missing-version requirement; a declared stale version is still reported |
| Correctly silent | caret form, vendored fork, excluded crate, commented-out declaration, `[[bench]]` `../` path |
| Fail-closed | helper exiting non-zero, or emitting either count, is caught |
| shellcheck `-x -S warning`, `py_compile`, `git diff --check` | clean |

Mutation coverage **101 → 124**.

## Containment

GitHub Advanced Security blocked an earlier head with four HIGH *Uncontrolled data used in path
expression* findings. They were one defect at four sites: `[workspace] members`, `exclude` and
dependency `path` values are repository-controlled but were used to build filesystem operations
with no containment check, and `main()` accepted an arbitrary root from `argv`.

Reproduced before fixing: adding `"../*"` to `members` made the checker resolve 29 member
directories, six of them outside the root, and it would have opened their manifests. Globbing is
itself the read — `root.glob("../*")` lists directories outside the root — so filtering matches
afterwards is too late.

One rule decides whether a resolved path may be used: `_within_root`, applied by members, excludes
and dependency targets alike. `_load_within` re-applies it immediately before the open so the
guarantee travels with the read, and the `argv` root is gone since the caller already runs from the
repository root. The same input now resolves 22 member directories, none outside the root.

A first attempt at the expansion guard was lexical only — it rejected `..` and absolute paths — and
that was not enough. `Path.glob` follows symlinked directories, so with an in-root `link -> ../out`
a repository-controlled `members = ["link/*"]` still enumerated outside the root. Filtering the
results cannot fix it, because for a glob the enumeration *is* the read.

Expansion is therefore bounded rather than filtered. `_safe_directory` walks a pattern's static
prefix down from the root refusing to traverse any symlink, so a listing only ever happens in a
directory proven in-root by walking to it. Wildcards are accepted in the final path segment only —
the shape expandable with one non-recursive listing; a wildcard further left would need a traversal
deciding symlink-following at each level, and is refused rather than guessed at. A match that is
itself a symlink is reported, not silently skipped. Cargo's member-glob behaviour is preserved where
it can be bounded: `crates/*` still expands. This repository declares no member globs and contains
no symlinks, so the bound costs nothing here.

The ordering was the property in dispute, so the proof tests the ordering: with the outside
directory made unlistable, a refusal-after-enumeration would raise `PermissionError`, while the
code exits cleanly with its diagnostic and never touches it.

Two behaviours were verified against cargo rather than assumed. Cargo does not expand globs in
`exclude` (with `exclude = ["crates/skip*"]` it still included `skipme`), which is why excludes stay
literal. And cargo itself refuses an out-of-root member — *"is not hierarchically below the
workspace root"* — so failing closed is cargo's own rule, not a restriction invented here.

No CodeQL suppression and no weakening of the check.

## Non-claims

- This checks that a declared version equals the workspace version. It does not check that the
  version is resolvable, nor anything about the published crates.
- A path dependency that does not resolve to a workspace member is ignored by design.
- **Known gaps, recorded rather than closed:** requirement syntaxes such as `=6.1.0` or
  `>=6.1.0, <7` are reported as drift; a case-only path difference on a case-insensitive filesystem
  silently drops a declaration (Linux CI catches it via cargo, the macOS pre-commit run does not);
  and the counts remain floor-guarded, so a shrink from a cause other than shape blindness is
  possible.
- The Linux compile gate skips `assay-cli` and `assay-core`. This branch touches no Rust, so that
  gate says nothing either way about this diff.
- Containment is enforced against the resolved path, and expansion is additionally bounded against
  symlink traversal. A symlink inside the root pointing outside is refused; a path outside the root
  that symlinks back in resolves to a real member and is treated as internal, which is correct.
- Member globs are supported only in the final path segment. That is narrower than cargo, which
  accepts a wildcard anywhere; a manifest relying on the wider form is refused with a named reason
  rather than silently mishandled.
- The mutation harness and the code under test were written by the same author. Their agreement
  establishes consistency, not correctness.

## Review status

Three prior reviews are superseded, not carried, and no carry is claimed. The first found a weak
guard and shape misses; the second returned request-changes on the denominator; the third returned
approve-with-findings with a blocker (a false failure on target-scoped dev-dependencies) and
observed that the root site still carried the original defect. Each judged a mechanism that has
since been replaced. A fourth review of `d24275b57` is advisory only and goes stale with this push. An independent
review of the current head is outstanding.


Rul1an:
BLOCKED at exact head `44681541c094d45403da5e3ae0b9b8ab830d3c55` (base `030827c16be15790ab0799a6f111b05af51fb5a7`). Two independent non-building reviews reproduced false-green paths.

## Findings

1. The claimed independent denominator is not shape-independent. The denominator and parser share the same double-quoted `../` pattern, so valid TOML such as `assay-x = { path = '../assay-x', version = "5.0.0" }` yields `declared=0, parsed=0` and passes silently. A tab before `=` also escapes both. Conversely, a commented-out declaration and a `[[bench]] path = "../..."` row produce false failures. Trailing comments/CRLF also defeat the dev-section anchor. Replace the text-pattern inventory with a real `tomllib` parse scoped to workspace-member dependency tables; widening the regex preserves the failure class.
2. The new guard machinery is not fully pinned. Deleting the per-manifest `parsed == declared_paths` comparison leaves the 108-case harness green and allows the live reformatted dependency to pass with the checked count shrinking from 9 to 8.
3. Narrowing the `git ls-files` derivation while retaining only `crate_checked >= 1` leaves 8 of 9 manifests unchecked and a stale A2A declaration green; the 108-case harness still passes.

Verified sound on this head: stale/versionless declarations, dev-section reset, `crate_checked == 0`, field-order handling, failure propagation, and the pre-commit selector each have a biting check. Builder checks and CI do not override the false-green findings above.

Required repair: `tomllib`-based structural enumeration, plus explicit mutations for removing the equality guard and dropping any covered workspace manifest. New head requires a fresh exact-head review.

Rul1an:
```json
{
  "head_sha": "3c86a20c407efb8b60dc80818c1d78190dca1358",
  "builder": {"agent": "claude", "instance": "writer-2794"},
  "reviewer": {"agent": "antigravity", "instance": "review-2794-3c86a20", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "F1",
      "summary": "Literal-zero count floors are not individually mutation-pinned.",
      "disposition": "Non-blocking test-hardening residual: missing count output already fails closed, and the live workspace reports 13 root plus 9 crate declarations; add explicit zero-count producer mutations in a follow-up."
    },
    {
      "id": "F2",
      "summary": "workspace.exclude is implemented but the fixture never makes the excluded path a member-glob candidate.",
      "disposition": "Non-blocking fixture residual: production resolves the real workspace exclude, while a follow-up should add an accepted terminal member glob whose excluded match disappears."
    },
    {
      "id": "F3",
      "summary": "Caret requirement normalization is not directly pinned by a fixture.",
      "disposition": "Non-blocking compatibility residual: the current workspace uses bare 6.1.0 requirements; add a caret-form fixture if this accepted syntax remains part of the stated contract."
    }
  ],
  "no_findings": false,
  "independence": {
    "did_not_build": true,
  }
}
```

Issue #2787, opened 2026-09-04
trace v7: carry truncation provenance without collapsing unknown into intact


Rul1an:
## Current implementation boundary (2026-09-05)

The original `Option<Vec<TruncationMeta>>` proposal below is **superseded as an implementation contract**, not ready to implement literally. Existing explicit `[]` was emitted by unmeasured producers; deserializing it as measured-clean would invent an observation. Missing observation means no stage observation is recorded, not proof that no scan occurred. Historical nonempty metadata is reported loss and must survive a later clean scan of the retained bytes.

The reviewed design direction is to distinguish preserved reported loss from a newly recorded stage-local observation. Exact wire shape, byte-stage binding, reader/storage migration and constructors still need a concrete reviewed v7 design. Do not add fields to externally constructible Rust structs on 6.x; `serde(default)` does not avoid that API break. See the [design-convergence record](https://github.com/Rul1an/assay/issues/2787#issuecomment-5547700089).

The 6.x vocabulary/diagnostic prerequisite #2786 is complete through #2795, #2797 and #2804 (final merge `1aa075651b147b257ebf5d9c278b027e374aca2c`). The separately authorized Step upsert consistency repair is [bounded here](https://github.com/Rul1an/assay/issues/2787#issuecomment-5547848212); it does not implement the v7 carrier or close this issue. Helper parity is a further prerequisite.

Historical proposal and acceptance notes follow for traceability; any tri-state wording below must be reconciled with this correction before code is written.

---

Parent: #2782
Depends on: the 6.x truncation-vocabulary repair subissue.

### Contract

In v7, make trace-loss provenance first-class without turning unknown into intact and without requiring another major for later additive event metadata.

### Required behavior

- `EpisodeStart`, `StepEntry`, and `ToolCallEntry` carry `Option<Vec<TruncationMeta>>`: `None` means not measured, `Some([])` means this stage measured no loss, and a non-empty vector records loss.
- The four V2 event structs and `TraceEvent` become `#[non_exhaustive]`.
- `StreamUpgrader` retains EpisodeStart metadata; OTel and MCP import emit explicit `None`.
- `episodes.truncations_json` is added through the existing migration path and round-trips.
- Existing v2 JSONL without the key continues to deserialize and remains visible to `trace-next`.
- The v7 release notes distinguish JSON reader compatibility from the intentional Rust API break.

### Behavioral RED tests

[...]

Issue #2786, opened 2026-09-04
trace: repair truncation units, paths, and producer dispositions on 6.x


Rul1an:
Parent: #2782

### Contract

Repair the existing 6.x truncation vocabulary without changing public event structs. This prerequisite makes the later provenance carrier well-defined; it does **not** close #2782.

### Required behavior

- Both truncation helpers report UTF-8 byte units consistently; `kept_len` equals the emitted string's byte length for multi-byte input.
- `TruncationMeta.field` is an RFC 6901 JSON Pointer, so literal dotted keys and nested paths cannot alias.
- Provenance-returning helpers are `#[must_use]`.
- The 4,096-byte ingest ceiling and its stage-local scope are documented.
- OTel and MCP-import non-truncation are pinned as explicit tested facts.
- `trace verify` distinguishes a truncation-mediated prompt mismatch from a genuine missing prompt.

### Behavioral RED tests

1. Multi-byte input bites the current char/byte mismatch in both helpers.
2. `meta["gen_ai.prompt"]` and `meta["gen_ai"]["prompt"]` emit distinct pointers.
3. An oversized MCP-import prompt is retained and marked as unmeasured by disposition, not described as measured-clean.
4. An over-ceiling config prompt produces a truncation-aware verification diagnostic.

### Non-claims

No event-schema field, v7 migration, ceiling increase, completeness claim, policy verdict, score, or producer-authentication claim.

Rul1an:
Re-measured on `origin/main` `94fb5fbdff40d2fb975c13a4fd7ec5070d4cfde8`.

Delivery clarification:

- `trace verify` already constructs `StreamUpgrader` (`verify.rs:15`), and the upgrader already applies truncation to both V2 and V1 events (`upgrader.rs:43-50`). The verification slice is therefore **diagnostic semantics only**: distinguish a truncation-mediated prompt mismatch from a genuinely missing prompt. It must not claim to add upgrader wiring.
- Keep the implementation serial: **A -> B -> C**. A repairs the shared byte/path vocabulary; B pins mapper-local non-truncation facts without taking #2787's absence semantics; C consumes the settled vocabulary in verification diagnostics.
- This does not change the issue's public-struct, schema, ceiling, score, or authentication non-claims.

No code or measurement claim beyond this exact main SHA.

Rul1an:
[...]

PR #2795, opened 2026-09-04
fix(trace): pin UTF-8 truncation and RFC 6901 array paths (#2786)

### LINKED ISSUE #2782: trace: EpisodeStart truncation discards its loss provenance
## Active remaining scope (2026-09-13)

The older v7/Option<Vec> proposal below is historical, not an implementation instruction. The accepted 6.x sibling observation carrier is ADR-050 and already shipped; this issue does not require a major release. The current scope is the [September 12 disposition](https://github.com/Rul1an/assay/issues/2782#issuecomment-5646323809).

At main `c890ec1fe2eaf744ef43e73f2691869461d94adf`:

- Probe C is complete via #3001: a sentinel-bearing retained value without loss metadata cannot be raised to measured-clean by the shared reader.
- CLI `trace verify` still calls public `verify_coverage(trace_path, cfg)`, which iterates bare events and recomputes a truncated prompt shape. It does not consume the observation carrier. Coverage-report surfacing also remains outstanding.
- Remaining implementation must render per-field readings, including explicit `Unmeasured`, and add opt-in repeated `--trust-stage` with no trusted default. An empty trust list must not permit `MeasuredClean`; reported loss must not be erased by lack of stage trust.
- Reuse the observed stream and shared reader. Preserve the existing public function signature or provide a backward-compatible additional entry point; adding a trust parameter to the existing public function is not automatically compatible.
- Must-fail controls remain: missing loss/sentinel raised to clean, Unmeasured omitted from output, and untrusted clean accepted without opt-in. A CLI-first slice leaves coverage-report work open rather than completing this issue prematurely.

This is source-verified planning, not a new binary measurement or implementation result. The historical design text below is retained for traceability.

---

## Current implementation boundary (2026-09-05)

The original `Option<Vec<TruncationMeta>>` proposal below is **superseded as an implementation contract**, not ready to implement literally. Existing explicit `[]` was emitted by unmeasured producers; deserializing it as measured-clean would invent an observation. Missing observation means no stage observation is recorded, not proof that no scan occurred. Historical nonempty metadata is reported loss and must survive a later clean scan of the retained bytes.

The reviewed design direction is to distinguish preserved reported loss from a newly recorded stage-local observation. Exact wire shape, byte-stage binding, reader/storage migration and constructors still need a concrete reviewed v7 design. Do not add fields to externally constructible Rust structs on 6.x; `serde(default)` does not avoid that API break. See the [design-convergence record](https://github.com/Rul1an/assay/issues/2787#issuecomment-5547700089).

The 6.x vocabulary/diagnostic prerequisite #2786 is complete through #2795, #2797 and #2804 (final merge `1aa075651b147b257ebf5d9c278b027e374aca2c`). The separately authorized Step upsert consistency repair is [bounded here](https://github.com/Rul1an/assay/issues/2787#issuecomment-5547848212); it does not implement the v7 carrier or close this issue. Helper parity is a further prerequisite.

Historical proposal and acceptance notes follow for traceability; any tri-state wording below must be reconciled with this correction before code is written.

---

### Problem

A real-binary measurement at `056332719107c3f5f5f9c4d19a42899c084517f1` shows that `assay trace ingest` truncates overlong `EpisodeStart.input` and `EpisodeStart.meta`, but discards the `TruncationMeta` values returned by `truncate_value_with_provenance`. The relevant source files are unchanged at the current design pin `7216b0cc4803e362041dddadc508dcf8d9172d90`; the runtime observation remains bound to its original binary and SHA.

The current surface has three distinct gaps:

- `StreamUpgrader` truncates both V1-upgraded and V2-passthrough episode starts, but the `EpisodeStart` arm ignores both returned metadata vectors.
- The `episodes` SQLite table has no `truncations_json` column, unlike `steps` and `tool_calls`, so EpisodeStart provenance cannot survive storage.
- The OTel and MCP-import producers do not call `apply_truncation`. Their Step/ToolCall constructors use empty vectors even though no truncator ran. An external crate constructing the public schema types is another unmeasured producer class.

The measured 4,595-byte ASCII prompt was emitted as 4,096 bytes. Loss is detectable only through the unauthenticated, forgeable in-band `...[TRUNCATED]` sentinel; it is indistinguishable by structured fields, and neither the extent nor the lost content is recoverable.

### Semantic decision

Use `Option<Vec<TruncationMeta>>` consistently:

- `None`: this producer did not run Assay's truncator, so truncation state is **unknown**.
- `Some([])`: this producer ran the truncator and found no value above its ceiling at this stage.
- `Some([..])`: loss occurred; each entry identifies the affected field and its pre-truncation UTF-8 byte digest.

This is stage-local evidence. `Some([])` does not establish end-to-end completeness: an upstream exporter, MCP host, SDK, or caller may already have truncated the value.

`TruncationMeta.sha256` already is the field-scoped pre-truncation digest. Do not add object-level `input_sha256` or `meta_sha256` fields. Existing ToolCall object digests do not substantiate field-level claims.

### Compatibility decision

The JSON reader change can remain additive with `#[serde(default)]`, but the Rust API cannot. `assay-core` v6.0.0 is published and `EpisodeStart` is a public, constructible struct without `#[non_exhaustive]`; adding any field triggers the binding `constructible_struct_adds_field` semver break.

Implement in two serial slices:

#### Slice 1: 6.x vocabulary repair

This slice does **not** close this issue.

- Make both truncation helpers report byte units consistently, including `kept_len`.
- Represent `TruncationMeta.field` as an RFC 6901 JSON Pointer so dotted keys cannot alias nested paths.
- Add `#[must_use]` to provenance-returning helpers so discarding metadata is compiler-visible.
- Document the 4,096-byte ingest ceiling and its stage-local scope.
- Pin OTel and MCP-import non-truncation as tested facts rather than accidental behavior.
- Make an over-ceiling `trace verify` mismatch disclose truncation instead of reporting only a bare missing prompt.

#### Slice 2: v7 provenance carrier

- Add `truncations: Option<Vec<TruncationMeta>>` to `EpisodeStart`, `StepEntry`, and `ToolCallEntry`.
- Add `#[non_exhaustive]` to the four V2 event structs and `TraceEvent` so later additive fields do not require another major.
- Retain both EpisodeStart metadata vectors in `StreamUpgrader`.
- Emit explicit `None` from OTel, MCP import, and other producers that did not run the truncator.
- Add and migrate `episodes.truncations_json`; round-trip it through storage.
- Preserve deserialization of existing v2 JSONL without the field.

### Behavioral acceptance

RED first, then GREEN:

1. A 4,595-byte ASCII prompt emits one entry for `/input/prompt`, with `original_len == 4595` and SHA-256 over the submitted UTF-8 bytes.
2. A nested overlong meta value emits `/meta/a/b`.
3. An untruncated 4,000-byte control emits `Some([])` and byte-identical content.
4. OTel and MCP-import episode starts omit the key (`None`), never claim measured-clean.
5. Existing v2 input without the key deserializes and remains visible to `trace-next`.
6. Multi-byte input proves `kept_len` equals the emitted UTF-8 byte length in both helpers.
7. Literal dotted and nested keys produce distinct JSON Pointers.
8. `episodes.truncations_json` survives a SQLite round trip.
9. `trace verify` distinguishes a truncation-mediated prompt mismatch from a genuine missing prompt.

Must-bite mutations include: discarding the EpisodeStart assignment; hashing after truncation; reporting the wrong length; replacing `None` with `Some([])` on a bypass producer; restoring char/byte mixing; restoring dotted-path concatenation; and dropping the SQLite write. `#[must_use]` is a lint contract, not behavioral mutation coverage.

### Scope boundaries

- This does not prohibit truncation or raise the 4,096-byte ceiling.
- A digest does not identify or authenticate the producer and does not recover lost bytes.
- An absent marker means unknown, not intact.
- The sentinel is not authenticated evidence.
- This changes trace provenance, not a policy verdict, trust score, coverage score, or whole-action claim.
- The runtime measurement exercised only `assay trace ingest`; OTel/MCP dispositions above are source-derived until their acceptance tests run.

Rul1an:
## Summary
- Slice A of #2786: shared UTF-8 byte-budget truncation so multi-byte input stays <=4096 bytes and `kept_len == emitted.len()`.
- `TruncationMeta.field` is an absolute RFC 6901 JSON Pointer via one shared escape/join; literal dotted keys no longer alias nested paths.
- Pins array traversal behavior explicitly: the first element under `args` must emit `/args/0`, never bracket syntax.
- Supersedes draft #2788 because Cursor was unavailable after its local repair and the final array-path finding needed a separately owned, pushed branch.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor

## Verification
- Before the new test, changing array paths from `append_segment(path, &i.to_string())` to `format!("{}[{}]", path, i)` left all four focused tests green: reproduced false-green.
- RED/GREEN: `array_indices_emit_rfc6901_pointer_segments` fails with `/args[0]` under that mutation and passes on the restored implementation.
- `cargo test -p assay-core`: 966 passed, 0 failed, 3 ignored.
- `cargo fmt --all -- --check`.
- `cargo clippy -p assay-core --lib --tests -- -D warnings`.
- `git diff --check`.
- Normal pre-commit and pre-push hooks passed.

## Non-claims / out of scope
- No mapper dispositions (Slice B), no `trace verify` diagnostics (Slice C).
- No public event struct/schema change, no v7 migration, no ceiling raise.
- Does not retain EpisodeStart provenance (#2787); does not close #2782.
- No completeness, score, policy verdict, or producer-authentication claim.

## Test plan
- [x] Focused truncation RED -> GREEN.
- [x] Array-pointer false-green now bites.
- [x] Affected crate, format, clippy, diff and local hooks.
- [ ] CI on draft head.
- [ ] Independent non-building review on final head before ready.

Closes nothing alone; addresses Slice A of #2786.


Rul1an:
READY at exact head `f1868ae7fd1ccb0a356fb9cc40e84919205c1d31` (base `030827c16be15790ab0799a6f111b05af51fb5a7`). Independent exact-head review; Ruley did not author this branch, governing design, or implementation. Builder evidence was not counted.

## Findings

None.

## Verification

- Live base and head matched the expected SHAs before and after review.
- Derived the full three-dot diff: 7 files, +208/-32.
- Reproduced the prior false-green: changing array traversal from RFC 6901 `/args/0` to bracket syntax `/args[0]` left the old four focused tests green.
- The new `array_indices_emit_rfc6901_pointer_segments` test fails under that mutation and passes on the restored implementation.
- Adjacent walkback, `kept_len`, escaping, and dotted-path mutations bite; the no-op control stays green.
- Focused truncation suite: 5/5 passed.

## Non-claims

This review does not cover mapper dispositions, `trace verify` diagnostics, public event schema changes, v7 migration, ceiling changes, completeness, scores, policy verdicts, or producer authentication.

Rul1an:
```json
{
  "head_sha": "f1868ae7fd1ccb0a356fb9cc40e84919205c1d31",
  "builder": {"agent": "codex", "instance": "writer-2786-2795"},
  "reviewer": {"agent": "ruley", "instance": "review-2795", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [],
  "no_findings": true,
}
```

Issue #2684, opened 2026-08-29
release(host proof): drive Codex discovery and one Assay tool on published v6.0.0


Rul1an:
## Acceptance recorded — 2026-09-08

All eight active v6.0.0 checklist items are supported by the [fresh measured result and proof digests](https://github.com/Rul1an/assay/issues/2684#issuecomment-5591816626), independently checked by `codex/dependency_repin`. Rul1an publishes this disposition. The exact nine-member proof roots and invocation/offline-review receipts are retained locally and were read independently; no public binary download or external attestation is claimed. Acceptance is limited to the named Codex host/build and published package route. Historical failed records and their limitations remain unchanged. #2812's remaining historical provenance audit and Claude #2194 remain separate.

## Current acceptance target — 2026-09-05

The active Definition of Done below targets published **v6.0.0**. Its non-draft, non-prerelease GitHub Release was published 2026-09-03T18:04:18Z: https://github.com/Rul1an/assay/releases/tag/v6.0.0 . This is release-existence evidence only, not artifact-integrity, installation, authentication or model-mediated host proof. The older v5.5.x entries are historical and do not override the current target. No new release is required merely to measure the already-published v6.0.0 host journey.

Parent: #2667
Extracted from: #2359

## Problem

Assay's Codex configuration and released MCP process were driven successfully on v5.2.0, but the September production launch names Codex and targets v6.0.0. Static recipes, a source build, direct JSON-RPC, or the older host record do not prove that Codex discovers and invokes the published v6.0.0 product.

This is a post-publication launch gate, not a pre-tag code gate.

## Entry condition

Start only after a non-draft `v6.0.0` GitHub Release exists and the intended CLI/MCP assets, checksums, attestations, and crates.io packages are publicly installable.

## Definition of done

From a disposable project and `CODEX_HOME`, using one pinned Codex host version:

- [x] Record Codex version, OS, project path/cwd, config location, Assay release, install source, binary version, and SHA-256.
- [x] Verify the published artifact or package before execution. A binary asset from the GitHub Release, or the crates.io package at the target version, both qualify. A build from a local checkout or an unpublished branch does not. Record which route was used.
- [x] Observe Codex discovering the shipped `assay-golden-path` skill.
- [x] Observe Codex starting `assay-mcp-server --policy-root .` from the disposable project context.
- [x] Pin exactly `assay_check_args`, `assay_check_sequence`, `assay_policy_decide`, `assay_check_coverage`, and `assay_explain_trace`.
[...]

PR #2801, opened 2026-09-04
fix(ci): align Codex host-proof projection with 0.153.1 schema

Rul1an:
Prepares #2684 for a fresh authenticated host-proof attempt following the observation in comment 5546686669 and review feedback in comments 5546994758 and 5547021077.

## Problem

The first authenticated v6.0.0 observation under Desktop-bundled Codex 0.153.1 retained manifest `fb82371241802dc1a89231956e8f203af0335639e332a56cbe93e3a3d115319a`. Inspection of `tool/events.json` against the exact generated 0.153.1 schema in `protocol-0.153.1/` revealed:

1. **`commandExecution` ThreadItem**: Codex 0.153.1 app-server emitted `commandExecution` items during the turn. Because `projectRetainedItem` only recognized `userMessage`, `reasoning`, and `agentMessage`, it defaulted `commandExecution` to `mcpToolCall`, populating `{ __invalidType: "undefined" }` for required tool fields and marking real command properties (`command`, `aggregatedOutput`, `cwd`, etc.) as `__unexpectedKeys`. `retainedItemReason` rejected it as `unknown retained item type commandExecution`.
2. **0.153.1 `mcpToolCall` schema validation and consumer boundaries**: The schema for `McpToolCallThreadItem` defines optional fields: `appContext`, `durationMs`, `error`, `pluginId`, `readOnlyHint`, and `mcpAppResourceUri`. In initial repair `e35acce945eee61acfb8b466035e45a2e5a024d5`, projection admitted generic scalars without verifying generated schema types, and `retainedItemReason` did not validate these fields at the consumer boundary. Consequently, malformed metadata (`durationMs="wrong"`, `readOnlyHint=42`, `pluginId=false`, `error={}`) could pass all nine cells through `classifyRecord`. Furthermore, non-null `item.error` on an `mcpToolCall` with `status: "completed"` could erroneously yield `pass` in `classifyInvocation`.
3. **`serverRequest/resolved` Notification**: Codex 0.153.1 emits `serverRequest/resolved` when a server request (such as elicitation) finishes. It was absent from `LIFECYCLE_SERVER_NOTIFICATIONS`, causing `classifyStoredEvent` to classify it as `unclassified server event`.
[...]

PR #2803, opened 2026-09-04
fix(ci): align Codex host-proof projection with 0.153.1 schema; record presence for newly admitted mcpToolCall metadata

Rul1an:
**Head `aca8191430404fc60e45be00851e5edb166fbe80`**, base `181a72ec09e3797e3b3e944ee09ec58db16bbadb`.
Every measurement below was taken on this head. An independent review of this exact head
returned **READY** (record: `assay-review-records/2803/review-aca819-final.md`), with no HIGH or
MEDIUM production defects and two notes carried below.

> **Provenance correction.** Every commit on this branch records `author: Codex <codex@openai.com>`
> — that is the repository's configured git identity, not the builder. The builder is Claude, and
> each commit already carries `Co-Authored-By: Claude Opus 5`. The author field is **not** being
> rewritten: doing so changes every SHA and would invalidate the independent review bound to
> `aca8191430404fc60e45be00851e5edb166fbe80`. The correction is recorded here instead.

## What changed since the last revision, and what the previous body got wrong

An independent review of `fd96640da` found four things. **The previous body is now wrong in one
place and is corrected here: it said a marker nested under `result` is accepted. It no longer is.**

| review finding | disposition |
|---|---|
| F1 — unknown JSON-RPC method retained verbatim | **fixed** — closed-set projection |
| F2 — arbitrary string rpc id retained verbatim | **fixed** — per-run correlation token |
| F3 — unknown `item.type` retained verbatim | **fixed** — closed-set projection |
| F4 — a nested marker yielded a fully green proof | **fixed** — consumer refuses violations anywhere |
| F5 — idempotence test title over-claimed | **fixed** — narrowed to the unexpected-key marker |
| F6 — body cited a residual list that did not exist | **fixed** — the list is below |
| F7 — a third acceptance widening was undocumented | **fixed** — listed below |

## The contract, stated once and finite

An identifier is retained by **membership**, by **equality structure**, or by **presence** — never as
host-chosen text. Every field was sorted by one question answered from source: *does any consumer
read the characters, or only compare?* The derivation, including the rows measurement refuted, is in
`assay-review-records/2803/identifier-disposition.md`.

- **Membership** — `method`, `item.type`, `item.server`, `item.tool` are compared against sets this
  proof declares, so declared members pass through untouched. An unknown member becomes an **explicit
  projection violation**, not a legal-looking substitute, and the consumer refuses it wherever it
  sits. (`[unknown-method]` was itself an allowed fallback; that is gone.)
- **Equality structure** — the rpc `id` plus `threadId`, `turnId`, `item.id`, `turn.id`, `thread.id`.
  Nothing reads their characters. The wire keeps the real values; only the persisted copy is
  relabelled, by first-seen ordinal. Each role has its **own map and prefix**, so namespaces can
  never merge; fields that are one identity share a role deliberately, because the cells compare
  exactly those pairs. **Not a hash, not redaction** — no preimage, and what is retained is
  correlation structure.
- **Presence** — `platformFamily`, `platformOs`, skills-row `cwd`, `skill.scope`, and the fields
  already carried by `projectedPresence`.
- **Comparison outcome** — `thread/start.result.cwd` retains whether it matched the project root,
  with match / mismatch / absent distinguished, plus a fourth state for a raw value, which means the
  outcome was never recorded and is drift.
- **Static reasons** — no refusal reason interpolates a host value.

The projection lives at `writeProofFiles`, not at `retainEvent`, and that was forced rather than
chosen: the driver reads its own live wire identities back out of the retained array, so relabelling
earlier put tokens on the wire. At the write boundary the in-memory array stays raw for the driver
while the manifest hash, the classification and all three files derive from one projected copy.

Integers pass through unchanged. Not because they are ours — **a host may choose a numeric id too** —
but because the protocol requires numeric ids to stay valid and an integer carries no text.

## Acceptance changes in this diff

- `commandExecution` newly admitted as a retained item type (base: `unclassified`).
- `serverRequest/resolved` added to `LIFECYCLE_SERVER_NOTIFICATIONS` (9 entries → 10).
- **`EXPECTED_ELICITATION.message` → `messages`, with `===` widened to `.includes()`** (F7).
- `ALLOWED_CLIENT_REQUESTS` is added, making the driver's own five client methods explicit rather
  than implied. Not a widening: it declares what was already sent.

## Verification

`node --test scripts/ci/test_codex_host_proof.mjs` → **164 tests, 164 pass, 0 fail, 0 skipped**.

Deletion controls, each with byte-exact sha256 restoration and a green no-op control first:

| mutation | result |
|---|---|
| marker → retain the key **names** | 14 named fail |
| marker dropped entirely | 13 |
| `projectedPresence` → `return value;` | 7 |
| consumer `allowedKeys`: drop `server` | 52 |
| consumer `allowedKeys`: add `cwd` (a sampled name) | 1 |
| consumer `allowedKeys`: add `zzzUnsampledName` (**not** sampled) | **0 — survives** |
| revert `method` projection | 2 |
| revert `item.type` projection | 3 |
| remove consistent consumer refusal | 1 |
| rpc-id map removed | 2 |
| rpc-id map non-injective (all ids → one token) | 3, incl. the named injectivity test |
| rpc-id token never reused | 56 |

The `0 fail` row is the measured boundary of the sampling, not a caveat: an undeclared name that no
test samples is **not** caught.

**A defect in these tests, found by their own control and recorded rather than quietly fixed.** The
first injectivity assertion was a tautology — deduplicating a set and comparing sizes — so the
non-injective mutation was killed only *incidentally*, by two unrelated EPIPE tests. The scenario
carried one host id, and no run with one id can distinguish an injective map from a constant one. It
was replaced with a run carrying two distinct host ids, which is the only composition that
discriminates it.

## Retained by design — the explicit list

Not leaks: each is retained for a stated reason, and each was measured.

| retained | reason |
|---|---|
| `codexHome` | pushed into the host-root list and used for path containment |
| `skill.name`, `skill.path` | skill discovery matches on both — projecting either fails 35 tests |
| `item.status`, `runtimeStatus` | source compares them but enumerates no set; declaring one would invent protocol |
| declared-set members, integers, timestamps, byte counts | the proof is keyed on them |
| the project root in client **request** params (`thread/start.params.cwd`, `skills/list.cwds`) | the driver disclosing its own value, not host content |

The 13 nested content projections (`text`, `command`, `name`, `version`, …) are content rather than
identifiers and are outside this contract. They need the same question asked of them.

## Verification of this slice

Controls, each with a green no-op first and byte-exact sha256 restoration:

| mutation | result |
|---|---|
| membership rule removed | 4 |
| presence removed (`platformFamily`) | **1, named** |
| identity projection not applied | 35 |
| role namespaces merged into one map | **1, named** |
| partial turn namespace (`result.turn.id` left raw) | 54 |
| cwd outcome not recorded | 32 |
| refusal reason echoes the host cwd | **1, named** |

The three single-test kills are the targeted ones: they break equality, presence and reason-hygiene
specifically rather than washing out the suite.

**Two of my own assertions were wrong and are corrected in place, not quietly dropped.** The valid
journey deliberately touches several threads, so "one thread means one token" was over-specified;
and the project root legitimately remains in client *request* params the driver itself sent, which is
disclosure rather than a host leak. Separately, an earlier injectivity assertion was a tautology and
was replaced with a run carrying two distinct host ids — the only composition that discriminates it.

## Claims and non-claims

- **Claim:** no host-chosen identifier *text* reaches `events.json`, `classification.json` or
  `manifest.json`, at every position in the contract above, verified through the real driver on both
  the passing and the failing path.
- **Not claimed:** zero arbitrary text across legitimate results. Content positions are out of
  scope, and the retained-by-design table above is deliberate.
- **Not claimed:** that sampling establishes set equality. See the `0 fail` row.
- These tests and the code they exercise were written by the same author.
- **Carried from the independent review, unchanged:** the residual table above lists intentional
  disclosures and non-membership fields, and must not be summarized as a blanket "no host text"
  (N1). The live Codex host, auth, keychain and model were not run — synthetic coverage only
  (N2). READY is an eligible independent review artifact for this tip; CI, branch protection, the
  formal GitHub review list and merge remain separate gates, and this leaves #2684 open.

Refs #2684. Supersedes #2801.

### DIFF
--- a/scripts/ci/codex_host_proof.mjs
+++ b/scripts/ci/codex_host_proof.mjs
@@ -43,6 +43,8 @@ import {
   runtimeProofRoots,
   isMainModule,
   persistableArgv,
+  CWD_DIFFERS_FROM_PROJECT_ROOT,
+  CWD_MATCHES_PROJECT_ROOT,
   projectRetainedEvent,
   projectHostIdentity,
   proofAllowlist,
@@ -421,7 +423,93 @@ function resolvedMcpCommand(options) {
   throw new Error("assay MCP binary was not resolved");
 }
 
+// Retention-only identity projection, applied at the ONE boundary where the proof is written.
+//
+// It cannot live at `retainEvent`: the driver reads its own live wire identities back out of that
+// array (`started[0].result.thread.id`, `response.result.thread.id`), so relabelling there would
+// put tokens on the wire. Here the in-memory array stays raw for the driver and only the persisted
+// copy is projected -- and because the manifest hash, the classification and all three files are
+// derived from the same projected copy, they cannot disagree.
+//
+// Roles are separate maps with separate prefixes, so two namespaces are never merged by accident.
+// Fields that ARE one identity deliberately share a role: `params.threadId` with
+// `result.thread.id`, and `params.turnId` with `params.turn.id`, because the cells compare exactly
+// those pairs. Every site is an explicit named path from the identifier disposition table -- never
+// a recursive walk, which would rename fields nobody enumerated.
+//
+// Every comparison downstream is an equality test, which is why a first-seen ordinal is
+// information-preserving for them: classifying the projected copy yields the same cells as
+// classifying the raw one.
+const RETAINED_ID_PREFIX = Object.freeze({ rpc: "#", thread: "@", turn: "~", item: "!" });
+
+function retainedIdentityProjection(events, projectRoot) {
+  const seen = new Map(Object.keys(RETAINED_ID_PREFIX).map((role) => [role, new Map()]));
+  const isObj = (value) => value !== null && typeof value === "object" && !Array.isArray(value);
+  const own = (value, key) => Object.prototype.hasOwnProperty.call(value, key);
+  const token = (role, value) => {
+    if (typeof value !== "string" || value.length === 0) {
+      return value;
+    }
+    const roleSeen = seen.get(role);
+    if (!roleSeen.has(value)) {
+      roleSeen.set(value, `${RETAINED_ID_PREFIX[role]}${roleSeen.size}`);
+    }
+    return roleSeen.get(value);
+  };
+  return events.map((original) => {
+    const event = structuredClone(original);
+    if (own(event, "id")) {
+      event.id = token("rpc", event.id);
+    }
+    const params = event.params;
+    if (isObj(params)) {
+      if (own(params, "threadId")) {
+        params.threadId = token("thread", params.threadId);
+      }
+      if (own(params, "turnId")) {
+        params.turnId = token("turn", params.turnId);
+      }
+      if (isObj(params.item) && own(params.item, "id")) {
+        params.item.id = token("item", params.item.id);
+      }
+      if (isObj(params.turn)) {
+        if (own(params.turn, "id")) {
+          params.turn.id = token("turn", params.turn.id);
+        }
+        if (Array.isArray(params.turn.items)) {
+          for (const entry of params.turn.items) {
+            if (isObj(entry) && own(entry, "id")) {
+              entry.id = token("item", entry.id);
+            }
+          }
+        }
+      }
+    }
+    const result = event.result;
+    if (isObj(result)) {
+      if (isObj(result.thread) && own(result.thread, "id")) {
+        result.thread.id = token("thread", result.thread.id);
+      }
+      // `result.turn.id` is the OTHER member of the turn namespace. Missing it is what a partial
+      // namespace looks like: `params.turn.id` becomes a token while this stays raw, and the cell
+      // that compares exactly those two stops matching. Every member of a role must be listed.
+      if (isObj(result.turn) && own(result.turn, "id")) {
+        result.turn.id = token("turn", result.turn.id);
+      }
+      // Group D: record the project-root COMPARISON outcome, never the host path.
+      if (typeof result.cwd === "string") {
+        result.cwd =
+          result.cwd === projectRoot
+            ? CWD_MATCHES_PROJECT_ROOT
+            : CWD_DIFFERS_FROM_PROJECT_ROOT;
+      }
+    }
+    return event;
+  });
+}
+
 function writeProofFiles(options, pack) {
+  pack = { ...pack, events: retainedIdentityProjection(pack.events, options.projectRoot) };
   const initialize = initializeFromEvents(pack.events, options.journey);
   const hostIdentity = projectHostIdentity(options.hostIdentity ?? null);
   const invocationArgv = persistableArgv(
@@ -612,21 +700,7 @@ export async function runProof(options) {
   // `nextId`, which is not the converse. They are preserved because the protocol requires numeric
   // ids to stay valid on the wire and an integer carries no text, not because their provenance is
   // known. Preserving them is also what keeps the client-generated integer response lookup working.
-  const retainedRpcIds = new Map();
-  const retainedRpcId = (id) => {
-    if (typeof id !== "string" || id.length === 0) {
-      return id;
-    }
-    if (!retainedRpcIds.has(id)) {
-      retainedRpcIds.set(id, `#${retainedRpcIds.size}`);
-    }
-    return retainedRpcIds.get(id);
-  };
-
   const retainEvent = (event) => {
-    if (Object.prototype.hasOwnProperty.call(event, "id")) {
-      event.id = retainedRpcId(event.id);
-    }
     const encoded = Buffer.byteLength(JSON.stringify(event), "utf8");
     if (
       events.length >= HARD_MAX_EVENTS ||
@@ -723,20 +797,15 @@ export async function runProof(options) {
 
   const dropClientWrite = (id) => {
     if (id != null) {
-      // Wire side: `pending` is keyed by the REAL id, so it is deleted by the real id.
       pending.delete(id);
     }
-    // Retention side: retained events carry the projected token, so the scan below must compare
-    // against that. `.get` rather than `retainedRpcId` on purpose -- looking up a write we are
-    // dropping must not mint a token for an id that was never retained.
-    const retained =
-      typeof id === "string" && retainedRpcIds.has(id) ? retainedRpcIds.get(id) : id;
+
     for (let i = events.length - 1; i >= 0; i -= 1) {
       const event = events[i];
       if (event.direction !== "client") {
         continue;
       }
-      if (id != null ? event.id !== retained : event.method !== "initialized") {
+      if (id != null ? event.id !== id : event.method !== "initialized") {
         continue;
       }
       const encoded = Buffer.byteLength(JSON.stringify(event), "utf8");
--- a/scripts/ci/codex_host_proof.mjs
+++ b/scripts/ci/codex_host_proof.mjs
@@ -594,7 +594,7 @@ export async function runProof(options) {
   const pending = new Map();
   // A STRING rpc id can only have come from the host: the driver's own requests take their ids from
   // `nextId`, which is always an integer. So every retained string id is host text, exactly like a
-  // method name or a key name.
+  // method name or a key name. The converse does NOT hold -- see the note on integers below.
   //
   // The wire is untouched -- the child still sees, and we still answer with, the real id, so string
   // ids remain fully supported on the protocol. Only what is RETAINED is relabelled, through one
@@ -607,8 +607,11 @@ export async function runProof(options) {
   // visibly reused. It carries no preimage and is not offered as reversible redaction: what the
   // proof retains is correlation structure, not identity.
   //
-  // Integers pass through unchanged. They are driver-generated, which is what keeps the
-  // client-generated integer response lookup working untouched.
+  // Integers pass through unchanged -- but NOT because they are ours. A host may choose a numeric
+  // id too; what is true is only that driver-ORIGINATED requests always take integers from
+  // `nextId`, which is not the converse. They are preserved because the protocol requires numeric
+  // ids to stay valid on the wire and an integer carries no text, not because their provenance is
+  // known. Preserving them is also what keeps the client-generated integer response lookup working.
   const retainedRpcIds = new Map();
   const retainedRpcId = (id) => {
     if (typeof id !== "string" || id.length === 0) {
--- a/scripts/ci/codex_host_proof.mjs
+++ b/scripts/ci/codex_host_proof.mjs
@@ -592,8 +592,38 @@ export async function runProof(options) {
   let childExit = null;
   let acceptedElicitations = 0;
   const pending = new Map();
+  // A STRING rpc id can only have come from the host: the driver's own requests take their ids from
+  // `nextId`, which is always an integer. So every retained string id is host text, exactly like a
+  // method name or a key name.
+  //
+  // The wire is untouched -- the child still sees, and we still answer with, the real id, so string
+  // ids remain fully supported on the protocol. Only what is RETAINED is relabelled, through one
+  // per-run map at `retainEvent`, the single funnel every retained event already passes through.
+  // (The one projection outside it, `writePreSpawnFailure`, carries no id and never contacts a host.)
+  //
+  // First-seen ordinal, deliberately NOT a hash. It preserves exactly the structure correlation
+  // needs -- the same id always yields the same token, distinct ids always yield distinct tokens --
+  // so a server request and its matching client response still pair, and a reused id is still
+  // visibly reused. It carries no preimage and is not offered as reversible redaction: what the
+  // proof retains is correlation structure, not identity.
+  //
+  // Integers pass through unchanged. They are driver-generated, which is what keeps the
+  // client-generated integer response lookup working untouched.
+  const retainedRpcIds = new Map();
+  const retainedRpcId = (id) => {
+    if (typeof id !== "string" || id.length === 0) {
+      return id;
+    }
+    if (!retainedRpcIds.has(id)) {
+      retainedRpcIds.set(id, `#${retainedRpcIds.size}`);
+    }
+    return retainedRpcIds.get(id);
+  };
 
   const retainEvent = (event) => {
+    if (Object.prototype.hasOwnProperty.call(event, "id")) {
+      event.id = retainedRpcId(event.id);
+    }
     const encoded = Buffer.byteLength(JSON.stringify(event), "utf8");
     if (
       events.length >= HARD_MAX_EVENTS ||
@@ -690,14 +720,20 @@ export async function runProof(options) {
 
   const dropClientWrite = (id) => {
     if (id != null) {
+      // Wire side: `pending` is keyed by the REAL id, so it is deleted by the real id.
       pending.delete(id);
     }
+    // Retention side: retained events carry the projected token, so the scan below must compare
+    // against that. `.get` rather than `retainedRpcId` on purpose -- looking up a write we are
+    // dropping must not mint a token for an id that was never retained.
+    const retained =
+      typeof id === "string" && retainedRpcIds.has(id) ? retainedRpcIds.get(id) : id;
     for (let i = events.length - 1; i >= 0; i -= 1) {
       const event = events[i];
       if (event.direction !== "client") {
         continue;
       }
-      if (id != null ? event.id !== id : event.method !== "initialized") {
+      if (id != null ? event.id !== retained : event.method !== "initialized") {
         continue;
       }
       const encoded = Buffer.byteLength(JSON.stringify(event), "utf8");
--- a/scripts/ci/codex_host_proof_validator.mjs
+++ b/scripts/ci/codex_host_proof_validator.mjs
@@ -1983,13 +1983,7 @@ function projectDecisionObject(value) {
       out[key] = projectedScalar(value[key]);
     }
   }
-  const unexpected = Object.keys(value)
-    .filter((key) => key !== "allowed" && key !== "reason")
-    .sort();
-  if (unexpected.length > 0) {
-    out.__unexpectedKeys = unexpected;
-  }
-  return out;
+  return withUnexpectedKeys(out, value, ["allowed", "reason"]);
 }
 
 function projectToolResult(result) {
@@ -2024,13 +2018,12 @@ function projectToolResult(result) {
       });
     }
   }
-  const unexpected = Object.keys(result)
-    .filter((key) => !["isError", "error", "structuredContent", "content"].includes(key))
-    .sort();
-  if (unexpected.length > 0) {
-    out.__unexpectedKeys = unexpected;
-  }
-  return out;
+  return withUnexpectedKeys(out, result, [
+    "isError",
+    "error",
+    "structuredContent",
+    "content",
+  ]);
 }
 
 function projectArguments(value) {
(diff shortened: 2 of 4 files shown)


Rul1an:
Independent full-diff review completed at c7f86637c4db15729a9a016757ea0f2a85d43487 (reviewer a89dd9d2243823c0d, authored neither change nor governing plan). Its disposition was READY with no live defect in the changed surface; the producer-level privacy probe passed with all newly projected free-text markers absent from retained output. This does not cover all pre-existing host strings or establish a real host journey. Integration remains on hold while the writer addresses actionable findings: producer-side malformed-type tests, nested unknown-key tests, a driven canonical-elicitation fixture, and title/body accuracy. A new push will require a new exact-head review; this is not permission to merge the pending repair unreviewed. The full local review record is retained as pr2803-review-record-c7f86637c.md. No auth or model call has been run in this review.

Rul1an:
# Independent final-head review record — PR #2803 (Rul1an/assay)

"fix(ci): align Codex host-proof projection with 0.153.1 schema; record presence for newly admitted
mcpToolCall metadata" (supersedes #2801; Refs #2684)

## Attribution and independence

- **Reviewer agent ID (durable attribution):** `afc0d7fde7fb5259b` (session
  `cd7baa9b-bfbd-4c64-8009-e15f1f35cd18`). Effort: **133 tool uses, ~5h53m wall, ~268k subagent
  tokens.** Raw transcript session-scoped; this record is the durable form.
- Fresh subagent, given **no prior verdict, finding list, or file list**; told a prior READY or
  BLOCKED on an earlier SHA does not carry. Followed
  `/Users/roelschuurkes/.codex/skills/assay-independent-final-head-review/SKILL.md` by exact path.
- **Writer of the change:** Claude Opus 5 (coordinator), branch `claude/2801-privacy-repair`.
  Arm's-length from the writer; satisfies quorum.
- **Coordinator note on liveness:** at ~5h49m this agent had zero live child processes and a
  transcript stale by ~6 hours, and the coordinator reported it as dead-without-verdict. **That was
  wrong** — it was in a long model turn and completed normally. Absence of child processes plus a
  stale transcript is not evidence of death.
- **Writes:** `git archive <head> | tar -x` into `mktemp -d` (`rev2803.BcS8ru`); all mutations applied
  and reverted there (final byte-diff vs head OK on all three files); probe scripts under
  `scratchpad/probe/`; one transient patched fixture, deleted; `/tmp` scratch. **Zero writes** to
  `/Users/roelschuurkes/assay` (`git status --porcelain` byte-identical at start and end, same 5
  pre-existing foreign entries); worktree `2801-privacy-repair` read-only, clean, at the expected
  head. GitHub read-only. Disk 11→16 GiB, never near the 5 GiB floor.

## Pin

Head `e0655fd94bd30cb3d341ac86a6cb16cbb5727668` (draft, OPEN, `reviewDecision` empty); base
`origin/main` = `1aa075651b147b257ebf5d9c278b027e374aca2c`; **merge-base = base**. Head parents
`5b7b662e6…` (branch) + `1aa07565…` (main). Three-dot scope **3 files, +936/−14**, exactly as
expected. The merge brought in only `crates/assay-core/src/trace/verify.rs` and
`crates/assay-core/tests/trace_verify_smoke.rs` (+428/−11) — **disjoint by path and by grep in both
directions; no `scripts/` change.** Head re-read immediately before the verdict: unchanged.
**Review valid.**

## DISPOSITION: **BLOCKED** — blocker: Finding 1

## Controls

- **No-op control:** unmodified head → exit 0, **142 tests, 142 pass, 0 fail, 0 skipped, 0 todo**,
  53.6 s.
- **Positive control (privacy):** neutering `projectedPresence` to return its argument made **all 7
  injected secrets appear in `events.json` and in the driver's return value, while the run still
  reported `status: "pass"`.** This proves the injected values reach the projection, that the grep can
  find them, and that **the projection is the only barrier** — no cell would catch the leak.

## Central claim — privacy, tested through the real producer

`codex_host_proof.mjs` (unchanged, outside the diff) is the producer: it imports
`projectRetainedEvent` and writes `events.json`. `runProof` was driven end-to-end with a fake
app-server patched to emit secret-shaped values at all six newly admitted positions.

- **Clean run:** all 9 cells pass, exit 0; every retained file and the driver return grepped for 7
  distinct secrets → **zero hits**. Retained item shows `pluginId`, `mcpAppResourceUri` and all five
  `appContext` keys as `"[present]"`; `durationMs: 42`, `readOnlyHint: true` survive as typed scalars.
- **With `error.message` carrying a secret:** zero hits, and `oneToolInvoked: fail`,
  `driverCompleted: fail`, exit 1 — the `item.error != null` rule still bites on presence-projected
  bytes. Suppression does not defeat the one consumer that reads `error`.
- **Projection is idempotent** (`projectRetainedEvent(p) === p`), so `validateProofRoot`'s fixed-point
  re-projection cannot false-fail; `error: null` survives as `null`.
- `scrub()`'s regex verified not to redact a path, hostname, e-mail, or raw JWT — the PR's reason for
  not using it is correct.

**Exact residual** (map derived from a real `events.json`), all pre-existing and outside this diff:
`item.id`, `server`, `tool`, `status`, `arguments.*`, `result.content[].text`, `threadId`, `turnId`,
elicitation `params.message` — **plus, absent from the PR's list:** `initialize.result.codexHome`,
`skills/list.result.data[].{cwd,skills[].name,skills[].path,skills[].scope}`,
`thread/start.result.cwd`, `mcpServerStatus/list.result.data[].{name,runtimeStatus}`. All of these
**have evidentiary consumers** (root containment, `skillDiscovered`, `cwdObserved`, `mcpStarted`),
unlike the six suppressed fields — so the PR's *reasoning* holds; its *enumeration* is incomplete.
`userAgent` correctly reduced to `"[observed-host]"`.

## Schema — independently re-derived

Schema directory only; no auth roots, no raw captures. **sha256 recomputed = the pinned value,
exact match.** `definitions/ThreadItem/oneOf/8` is `McpToolCallThreadItem`: **13 properties,
set-equal to the test's list**; `required` = 6; no `allOf`/`anyOf`/`oneOf`/`$ref`;
`additionalProperties` unset. **Transcription correct.** Nested: `McpToolCallAppContext` = exactly the
5 allowed names (`connectorId` required); `McpToolCallError` = exactly `message`. Both set-equal.
Both in-code whitelists equal the schema set today in both directions — the PR's "13 for 13, both
directions" is **true as a measurement**.

## Findings

### 1. BLOCKER — missing guard. The **consumer** `allowedKeys` whitelist is pinned in neither direction, though it is the list both new schema tests are named and reasoned for.

`codex_host_proof_validator.mjs`: `retainedItemReason`'s `allowedKeys` array vs
`projectRetainedItem`'s separate 13-name `.filter(...)` list. Both new tests assert only on
`__unexpectedKeys`, produced by the **producer's** list; neither calls
`classifyStoredEvent`/`classifyRecord`.

| mutation | producer list | consumer `allowedKeys` |
|---|---|---|
| drop a declared property | **M28 → killed** (3 tests) | **M12 → SURVIVED (142/142 green)** |
| add an undeclared name | **M27 → killed** (F-A) | **M26 → SURVIVED (142/142 green)** |

**Measured behavioural consequence:** with M12 applied, an end-to-end driven run whose host emits
`durationMs` fails **all nine cells** — a total, silent false refusal of the whole proof — while the
suite stays green.

The test is named *"mcpToolCall projection accepts every declared 0.153.1 schema property"* and its
comment reasons explicitly about `allowedKeys` ("a schema property absent from `allowedKeys` is a
guaranteed false refusal…"). The body establishes neither half for `allowedKeys`. The comment's
"Derived from behaviour, not from importing the constant" is true, but derives the behaviour of the
*other* constant.

**Classification:** missing guard — the bytes are correct today (whitelists independently verified
set-equal). **Mask:** the producer's independent 13-name list is what both tests read; the default
fixture never carries any of the six newly admitted keys through the consumer on a well-formed
record; and the one test that injects them (`classifyRecord rejects invalid 0.153.1 metadata types`)
asserts only *not-all-pass*, which an unexpected-key refusal satisfies just as well as a type
refusal.

**Smallest coherent fix:** one test that classifies a well-formed `mcpToolCall` carrying all six newly
admitted fields and asserts `allIntendedCellsPass === true`. That single test kills M12 and M26
together.

### 2. LOW — missing guard. Producer type validation for `durationMs` and `readOnlyHint` is deletable with the suite green.

`projectRetainedItem`'s two inline ternaries. **M20** (durationMs → verbatim) and **M21**
(readOnlyHint → verbatim) both **SURVIVED** (142/142). F-B closes this class only for the five
positions sharing `projectedPresence`; these two have their own code paths, and the existing
F1-producer cases for them use scalar non-strings which the consumer catches regardless. **Not a live
defect** — verified on a pristine extracted head that non-scalars at both positions yield
`{__invalidType:…}` and never reach retained bytes. **Fix:** two more F-B rows (object payloads).

### 3. LOW — PR body over-states completeness in two places.

The residual "still written verbatim" list omits `initialize.result.codexHome` and the
`skills/list` / `thread/start` / `mcpServerStatus/list` positions enumerated above — mark the list
non-exhaustive or extend it. And "admitted seven new `mcpToolCall` fields and wrote **six** of them
verbatim" is followed by a list of **seven** positions; verified against `f552bb827`, seven were
verbatim and `error.message` was scrubbed (eight total). The quoted `main` projection's
`__unexpectedKeys` shows 5 names; the actual is 10.

### Observation (not a finding) — the writer's mutation table

6 of 8 rows reproduce exactly. Two could not be reproduced under either natural reading:
"`projectedPresence` returns its argument" gave 2 (strings→verbatim) or 7 (unconditional), never 4;
"`connectorId` written unconditionally" gave 2, not 3. **The direction (green→red) held in every
row.** Flagged so the numbers get re-measured, not as a provable error.

## Mutation arithmetic

**31 distinct mutations executed** (plus 1 repeat measurement) — **27 killed, 4 survived. 27 + 4 = 31.**

Survived: **M12** (drop declared key from consumer `allowedKeys`), **M26** (add undeclared key to
consumer `allowedKeys`), **M20** (producer `durationMs`), **M21** (producer `readOnlyHint`).

Notable kills: presence→verbatim (M1a/M1b/M30 → F1); unconditional passthrough (M17 → 7 incl. all
five F-B); non-string passthrough (M29 → 5 F-B); `commandExecution` producer/consumer (M2/M3);
`serverRequest/resolved` (M5); canonical prompt removed (M6 → 2); **legacy prompt removed (M19 → 40,
the decisive proof that the default scenario still exercises legacy acceptance)**; `item.error` cell
(M7); consumer type checks (M8/M9/M10/M22/M25); both-whitelist widening (M11); nested marking
(M13/M14) and nested key loops (M23/M24); producer list changes (M16/M27/M28); mcpToolCall key loop
(M4 → F2); fixture canonical→legacy (M15 → F5).

**Self-correction recorded by the reviewer:** one probe initially reported a `readOnlyHint` leak at
head; re-running against a pristine git-extracted validator showed it was contamination from its own
concurrent mutation run. **No such defect exists.**

## Fixture — clean

Default scenario still emits the legacy prompt (M19 → 40 failures proves legacy is load-bearing
across every existing driven run). The canonical scenario appears in exactly one fixture branch and
one test; F5 passes **through acceptance**, not incidentally — M6 and M15 each kill it.

## Error semantics — unchanged by this diff, well covered

`codex_host_proof.mjs` untouched. Existing named tests distinguish nonzero child exit with parseable
stdout, timeout, output cap / oversize line, missing binary, unknown status, truncation,
stream-unavailable, and pre-spawn failure; all green in the control. `serverRequest/resolved` carries
no host content.

## Hosted CI at the exact head

64 check-runs bound to `e0655fd9…`: **49 success, 14 skipped, 1 failure, 0 still running.** The
failure is `review-record-check` (`reason=no_current_record`) — the gate demanding this review, not a
code failure. Lanes executing the changed code: **`CI`** (step "Codex host-proof Node contract") →
success, and **`Lint (pre-commit)`** (hook `codex-host-proof-node-contract`, `files:` gated on exactly
these three paths) → success.

## Explicit non-claims

- **A green local Node run proves the code behaves as its own fixtures specify. It is not host proof,
  not release proof, and a generated schema declaring a field is not evidence the host emits it.**
- **Not executed:** any `cargo`/`rustc` — the two Rust files the merge brought in are **unverified**;
  any real Codex 0.153.1 host run; the pre-commit chain ("passed on every commit" unverified); the
  M12/M26 fix.

Rul1an:
Fresh independent review at fd96640da800d0d160ccdf8a6ae2c9f295b69337 confirms unexpected-key-name privacy at all six targeted sites, including persisted bytes. It also found additional retained unknown method/id/item-type content, a result-level projection marker that can coexist with a fully green proof, and test/body precision gaps. The reviewer disposition is approve-with-findings, not a fabricated READY record. Codex chooses HOLD before live authenticated proof and has assigned a bounded repair/design check to the writer. No merge or host PASS is claimed; CI success does not dispose of these findings.

Rul1an:
# Independent exact-head review — Assay PR #2803 @ `aca8191430404fc60e45be00851e5edb166fbe80`

**Verdict: READY**

This is a **new-head** verdict. Prior `pr2803-review-record-*.md` / `fd96640` packets are **not** carried.

---

## Eligibility (declared first)

| Gate | Result |
|------|--------|
| Role | Independent, non-writer, exact-head (Ruley). |
| Tip | `aca8191430404fc60e45be00851e5edb166fbe80` |
| Writer | `/Users/roelschuurkes/assay/.claude/worktrees/2801-privacy-repair` — **clean** (`git status --porcelain` empty); HEAD = tip |
| Base | `181a72ec09e3797e3b3e944ee09ec58db16bbadb` (ancestor of tip) |
| Remote tip (stated) | `fd96640da800d0d160ccdf8a6ae2c9f295b69337` — tip is four commits ahead (`16cc24f43` … `aca819143`) |
| Scope vs merge-base | **Exactly four files:** `scripts/ci/codex_host_proof.mjs`, `scripts/ci/codex_host_proof_validator.mjs`, `scripts/ci/fixtures/codex-host-proof/fake-app-server.mjs`, `scripts/ci/test_codex_host_proof.mjs` |
| Tip git author (metadata) | `Codex <codex@openai.com>` on the tip commits — **git label only**, not the builder identity |
| Actual builder | **Claude Desktop**, in its owned tree (not identified by the git author string) |
| Change / governing-plan authorship | **None by Ruley.** Ruley authored neither the tip change nor the governing plan/disposition. `identifier-disposition.md` is an external disposition table cited by the tip commit. |
| Forbidden this review | Real Codex host, auth, keychain, model, API, Cargo, Docker, posts, writer edits |
| Allowed | Own git export; synthetic Node (`fake-app-server`) |

**Eligible: yes.**

**Quorum meaning:** READY = eligible independent review artifact for this tip. Reviewer is **not** declining quorum for independence or scope. CI / branch protection / formal GitHub review-list / merge remain **separate** process gates.

---

## Method

1. Isolated `git archive` of tip → `/tmp/assay-2803-aca819-review` + `git init` (scratch **preserved**).
2. Full four-file semantic review vs base (retained projection vs raw-wire), plus tip-range commits since `fd96640`.
3. Live-host preflight on the test entrypoint → synthetic-only.
4. Full Node entrypoint + three targeted break-to-RED + comment no-ops (scratch `/tmp/assay-2803-n1-scratch`; export left clean).
5. Coverage inspected by **named new tests** in `fd96640..aca819143`, not by total alone.

Node v22.16.0. Logs: `/tmp/assay-2803-aca819-review-logs/`.

---

## Semantic review (whole four-file change)

### Architecture: raw wire vs retained projection

- **In-memory retention** (`retainEvent`): pushes events the driver still correlates on. Content projection often applied via `projectRetainedEvent(...)` at retain call sites; **identity tokenization is deliberately not done here** so `result.thread.id` / turn ids remain live for the driver (`codex_host_proof.mjs` comment ~428–433).
- **Write boundary** (`writeProofFiles`): `retainedIdentityProjection(events, projectRoot)` runs once before manifest hash, classification, and the three proof files — one projected copy, so those surfaces cannot disagree.
- **Pre-spawn path** also funnels through `writeProofFiles` / projection (no host contact).

### A — Membership (unknowns invalid)

- `projectedMember(value, allowed)` → declared member verbatim, else `invalidProjection` (not a lookalike token). Replaces former `[unknown-method]` / `[unknown-item-type]` substitutes.
- Applied to `method` (`KNOWN_METHODS`), `item.type` (`KNOWN_ITEM_TYPES`), `item.server` (`EXPECTED_SERVERS`), `item.tool` (`EXPECTED_TOOLS`).
- `classifyStoredEvent` refuses any retained event carrying a projection violation **wherever nested** (`containsProjectionViolation`), reason names marker class only.

### B — Equality structure (per-role maps)

- Roles `#` rpc / `@` thread / `~` turn / `!` item — separate maps, separate prefixes; shared role only where cells compare the same identity (`threadId`↔`thread.id`, `turnId`↔`turn.id`, including `result.turn.id`).
- Partial namespace (e.g. omitting `result.turn.id` tokenization) is a known fail mode (RED control hit it).

### C — Presence-only (exact residuals, not “no host text”)

`projectedPresence` → `[present]` only where disposition says **no lexical consumer**:

| Collapsed to presence | Declared residual / disclosure (lexical kept) |
|----------------------|-----------------------------------------------|
| `platformFamily`, `platformOs`, skills-row `cwd`, `skill.scope` | `codexHome` (host-root list / containment) |
| unexpected-key marker → `PRESENT` | `skill.name`, `skill.path` (discovery match) |
| several connector/app string fields via `projectedPresence` | Integer rpc ids pass through (protocol; **not** provenance proof — tip docs commit) |
| server `error` retained as `{ present: true }` | Client **request** params may still carry driver-sent project root (disclosure, not host leak) |

`status` / `runtimeStatus` are **not** closed-set membership fields (no invented enum); compared as today with **reason strings that do not echo host values**.

### D — cwd comparison

- At write: `result.cwd` → `[project-root]` / `[not-project-root]` vs `options.projectRoot`.
- `classifyCwd`: absent / match / mismatch / **raw residual = fail** (“outcome never recorded”). Four states; missing ≠ false.

### E — Error / refusal classification

- Fail reasons for cwd, server, tool, runtimeStatus, item status name the **condition**, not the host lexical value (tip diff).
- Nested projection violations refused without echoing offending content.

### Fake server / driver

- Tests spawn `fake-app-server.mjs` / shadow PATH wrappers; production `spawn(bound.codexPath, …)` exists but was **not** exercised live in this review.

---

## Executed checks (real)

### Full Node entrypoint

```text
cd /tmp/assay-2803-aca819-review
node --test scripts/ci/test_codex_host_proof.mjs
```

From `/tmp/assay-2803-aca819-review-logs/full-node.log` (not inferred):  
**tests 164 · pass 164 · fail 0 · cancelled 0 · skipped 0 · todo 0 · exit 0**  
(`duration_ms 34442.086542`)

### Targeted break-to-RED + no-op

| # | Mutation | RED (named) | No-op |
|---|----------|-------------|-------|
| 1 | `projectedMember` passthrough unknowns (`validator` ~2029) | **3 fail** (F1 method, F3 item type, persisted) | GREEN |
| 2 | skip `result.turn.id` tokenization (`codex_host_proof.mjs` ~493) | **1 fail** (B identities) | GREEN |
| 3 | `projectedPresence` returns raw string (~2035) | **1 fail** (C); D unaffected (cwd uses identity constants) | GREEN |

### Tip-range coverage (`fd96640..aca819143`) — new tests, not total-only

**11 new `test()` titles** across `16cc24f43` / `78ab81af9` / `aca819143` (docs-only `a83a3fc71` has no test delta): unexpected-key fixed point; unknown method/type projected + absent from proof; F4 nested marker refused; host string rpc id never persisted; token equality / distinct ids; **B** role-namespaced tokens + cross-event turn correlation; **C** platform/scope → `[present]`; **D** cwd outcome tokens, path absent, refusal must not echo host cwd, `cwdObserved` fail≠missing.

---

## Findings

| ID | Sev | Note |
|----|-----|------|
| — | — | No HIGH/MEDIUM production defects found under this scope. |
| N1 | NOTE | Exact residuals above are intentional disclosures / non-membership fields — do not summarize as blanket “no host text.” |
| N2 | LIMIT | Live Codex host / auth / keychain / model not run (forbidden). Synthetic coverage only. |

---

## Non-claims

- Not a live host proof; not model/API attestation; not Cargo/Docker.
- Does not close #2684 / #2803 by itself.
- Does not claim scrubber completeness outside the disposition table.
- Integer rpc-id passthrough is not provenance evidence (aligned with tip docs).

---

## Verdict

**READY** on **`aca8191430404fc60e45be00851e5edb166fbe80`** for the four-file retained-event privacy / host-proof projection scope, based on independent semantic review plus executed synthetic Node **164/164** and three targeted RED/no-op controls.

Scratch left at `/tmp/assay-2803-aca819-review`, `/tmp/assay-2803-n1-scratch`, `/tmp/assay-2803-aca819-review-logs/` until further notice.

Rul1an:
```json
{"schema":"assay.review-record.v0","head_sha":"aca8191430404fc60e45be00851e5edb166fbe80","review_completed":true,"verdict":"READY","no_findings":false,"builder":{"agent":"claude","instance":"claude/2801-privacy-repair"},"reviewer":{"agent":"ruley","instance":"Ruley exact-head PR2803 aca819","github_login":"Rul1an"},"independence":{"did_not_build":true,"did_not_author_governing_spec":true},"findings":[{"id":"N1","summary":"Intentional lexical residuals remain","disposition":"Accepted and explicitly listed; no blanket claim of no host text or scrubber completeness."},{"id":"N2","summary":"Synthetic review only, no live host run","disposition":"164 tests and three negative controls reviewed; issue2684 published host proof remains separate and open."}],"review_evidence":"https://github.com/Rul1an/assay/pull/2803#issuecomment-5551120176","relay":"Codex relays Ruley independent review through owner account; git author label is not agent identity."}
```

PR #2822, opened 2026-09-05
feat(ci): record a declared install source in the host-proof manifest

Rul1an:
Addresses part of cell 1. Leaves #2684 open.

## What was wrong

#2684's DoD cell 1 requires the install source, and the retained proof pack recorded none. The independent review of the `b9c8ac6` journey (https://github.com/Rul1an/assay/issues/2684#issuecomment-5554980138) recorded the issue as BLOCKED partly on that.

Tracing it turned up two separate causes, and the first is the interesting one:

1. **The projection was lossy.** `resolveHostIdentity` already probed `os`, `arch` and both binary versions, but `projectHostIdentity` kept only `{path, sha256}`. Every provenance field was measured and then discarded on the way to disk.
2. **The value was not an answer.** The `installSource` it did build was the literal `"PATH"`, which names where the binary was *resolved*, not where it came from. It cannot distinguish a release asset from a crates.io package from a local build, which is exactly the distinction cell 2 turns on.

## What this changes

Install source is **declared, never inferred**. How a binary reached the machine is not observable from the binary the host launches, so deriving it would substitute one layer for another, which is the failure mode the host-proof skill exists to prevent.

`--install-source <subject> <route> <reference>`, repeatable, validated structurally at parse time, and **required for all three subjects** so an absent declaration fails closed rather than reading as a pass.

The route vocabulary is closed: `github-release`, `crates-io`, `host-bundled`, `local-build`. `local-build` is recordable **on purpose**. A failed provenance run must be retained as what it was rather than omitted into a cleaner record.

`projectHostIdentity` now retains `installSource`, `version`, `os` and `arch`, and `liveIdentityBound` requires them, so the fields are load bearing rather than decorative.

The manifest shape is a contract enforced by `exactKeys`, so the schema moves `assay.codex-host-proof.v4` -> `v5`. Retained v4 packs stay valid under their own pinned verifier, which is the point of pinning it.

## Controls

190 tests pass, 186 existing plus 4 new. The new ones include a positive control on a complete declaration, and refusals for unknown routes and subjects, duplicate declarations, newline and control-character injection, credential-shaped references, over-length references, and extra keys.

The projection regression test was **verified to bite**: reverting `installSource` out of `projectBoundBinary` turns it red, and restoring it turns it green. A test that passes is not the same as a test that guards.

## What this does not do

- **Leaves #2684 open.** The `b9c8ac6` proof pack predates this change, so a fresh run is still required to produce a pack that carries the field.
- **`PUBLISHED_INSTALL_ROUTES` is defined but not yet enforced by a cell.** A run declaring `local-build` still classifies every cell as pass while recording an unpublished route. Wiring that into a classification cell changes the classification contract and the retained `classification.json` shape, so it is deliberately left as a separate decision rather than smuggled in here.
- It does not change the scrubbing of `initialize.platformOs`, which remains `[present]`. `os` and `arch` now come from the identity block instead.

## Review

**This branch needs an independent reviewer, and it must not be me.** The same agent wrote this change, reviewed #2684, and amended #2684's DoD. That is builder, governing-spec author and reviewer collapsed onto one actor, which the repository's own review-record schema refuses via `identical_writer_reviewer` and `independence.did_not_build`.

## Current-head coordination

Head: `0b1172a3a657157afdb3760463cc0be0683eeea7`.
Independent Ruley READY record: https://github.com/Rul1an/assay/pull/2822#issuecomment-5555259566 . The prior BLOCKED record remains unchanged. Earlier test counts above describe the initial implementation, not this final head. Current reported counts are 192 passed on macOS and 191 passed / 1 skipped on Linux; these are agent-reported, not a coordinator rerun. Required CI remains a separate landing gate. Leaves #2684 open pending the remaining install-route evidence and a fresh host-proof run.

### DIFF
--- a/scripts/ci/codex_host_proof.mjs
+++ b/scripts/ci/codex_host_proof.mjs
@@ -46,6 +46,7 @@ import {
   CWD_DIFFERS_FROM_PROJECT_ROOT,
   CWD_MATCHES_PROJECT_ROOT,
   projectRetainedEvent,
+  installSourceBound,
   projectHostIdentity,
   proofAllowlist,
   requirePrivateDirectory,
@@ -58,6 +59,8 @@ import {
   validateProofRoot,
   walkDepth,
 } from "./codex_host_proof_validator.mjs";
+
+const INSTALL_SUBJECTS = Object.freeze(["codex", "codexCodeModeHost", "assayMcp"]);
 
 const DEFAULT_TIMEOUT_MS = 30_000;
 const DEFAULT_MAX_BYTES = 1_048_576;
@@ -72,6 +75,7 @@ export function parseArgs(argv) {
     childArgv: null,
     proofRoot: null,
     projectRoot: null,
+    installSources: {},
   };
   for (let i = 0; i < argv.length; i += 1) {
     const arg = argv[i];
@@ -101,6 +105,26 @@ export function parseArgs(argv) {
       case "--allow-live-turn":
         out.allowLiveTurn = true;
         break;
+      // Install source is declared, never inferred: how a binary got onto this
+      // machine is not observable from the binary the host launches. Each token is
+      // validated structurally here so nothing free-form reaches the record.
+      case "--install-source": {
+        const subject = next();
+        const route = next();
+        const reference = next();
+        if (!INSTALL_SUBJECTS.includes(subject)) {
+          throw new Error(`--install-source subject must be one of ${INSTALL_SUBJECTS.join(", ")}`);
+        }
+        if (Object.prototype.hasOwnProperty.call(out.installSources, subject)) {
+          throw new Error(`--install-source declared twice for ${subject}`);
+        }
+        const candidate = { route, reference };
+        if (!installSourceBound(candidate)) {
+          throw new Error(`--install-source for ${subject} is not a valid route and reference`);
+        }
+        out.installSources[subject] = candidate;
+        break;
+      }
       default:
         throw new Error(`unknown argument ${arg}`);
     }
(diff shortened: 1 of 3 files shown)


Rul1an:
```json
{
  "head_sha": "be0c2b4a806302298a3c0353ed6854d6eacb8680",
  "review_completed": true,
  "verdict": "BLOCKED",
  "builder": {
    "agent": "claude",
    "instance": "opus-5-2684-producer-install-source"
  },
  "reviewer": {
    "agent": "ruley",
    "instance": "grok-bot-2026-09-06T00:16:30+02:00",
    "github_login": "Rul1an"
  },
  "independence": {
    "did_not_build": true,
  },
  "findings": [
    {
      "id": "F1-unicode-install-reference",
      "summary": "installSourceBound accepts Unicode line/paragraph separators (U+2028/U+2029), ZWSP, and bidi overrides in reference while refusing only C0/DEL controls; such values pass length/control/credential checks and can split rendered logs/markdown without failing closed.",
      "disposition": "must-fix before READY; extend the control-class refusal beyond C0/DEL or otherwise neutralize line/bidi/ZWSP in references"
    }
  ],
  "no_findings": false
}
```

Rul1an:
```json
{
  "head_sha": "0b1172a3a657157afdb3760463cc0be0683eeea7",
  "review_completed": true,
  "verdict": "READY",
  "builder": {
    "agent": "claude",
    "instance": "opus-5-2684-producer-install-source"
  },
  "reviewer": {
    "agent": "ruley",
    "instance": "grok-bot-2026-09-06T00:32:00+02:00",
    "github_login": "Rul1an"
  },
  "independence": {
    "did_not_build": true,
  },
  "findings": [],
  "no_findings": true
}
```

PR #2870, opened 2026-09-09
test(ci): close required-job skip and count false-greens

Rul1an:
## Problem and resulting behavior

The `evidenceref-live-resolve` job was added to the required CI rollup in #2861, but its expectation test exercised only the generic `failure` arm. The gate's per-row `case` has three arms, and the `skipped` one is what a broken `if:` condition actually produces, so that arm was unreached for this row.

**Correction to an earlier version of this description.** It said that changing the row from `required` to `optional` left the contract green, and offered that as the gap being closed. That is true of this test script alone but false as a statement about the merge: `check-ci-gate-coverage.py:316` rejects a non-computed expectation, and that guard runs inside the required `CI` job at `ci.yml:1181`. The merge was never open on that mutation. The honest delta is narrower and is the *behaviour of the skipped arm* for this row, which no static guard reads. A gate patched to ignore this job in that arm passed the coverage guard, the timeout contract and the base test, and fails here.

An adversarial review of `70b4be72f` then found two more, both reproduced before being fixed:

- The name assertion failed open. It matched `<name> was skipped` as a bare substring, so a gate emitting `zz-deps-security` satisfied an assertion claiming to check `deps-security`. Matched with `-F` against the `::error::<name> was skipped` prefix now. The same shape would accept a future job id ending with an existing one, and two plausible candidates exist already (`msrv` beside `public-msrv`, `smoke-test` beside `test`).
- The job was in the wrong loop. It has no `if:` and carries the literal `required` expectation, making it structurally unconditional rather than code-gated. In the code-gated loop it was never run with `LIGHTWEIGHT_ONLY=true`, so a gate silently exempting it on a docs-only run passed everything. It is in the unconditional loop now, which pins the stronger property.

Both loops now call one `assert_named_skip` rather than each carrying its own copy of the two greps.

## Scope

- `scripts/ci/test-ci-gate-expectations.sh`
- `scripts/ci/test-ci-job-timeouts-contract.sh`

No workflow behaviour and no product runtime code changes.

**What these two scripts gate.** They are invoked from `.pre-commit-config.yaml` and are not run by `ci.yml`. They reach CI only through `pre-commit run --all-files` in the `Lint (pre-commit)` job, which is not among the required contexts (`CI`, `host-capability-check`, `lane-check/proof`, `review-record-check`). So this PR strengthens a contract that does not itself block a merge. That is pre-existing and not introduced here, but it bounds what the change buys and is worth stating rather than leaving a reader to assume otherwise.

## Verification at `7c166e7d9f13bd518eb4085bf85b47e57808b048`

- CI gate expectations: pass
- CI job-timeout contract: pass, 17 jobs; `--self-test` pass
- CI gate coverage: pass
- `shellcheck` on the changed script: clean
- Mutation, exempting the job on a docs-only run: green before the loop move, red after
- Mutation, gate emits a prefixed wrong name: green before anchoring, red after
- Mutation, skipped arm ignores the job outright: red, and green against the base script
- Removing one timeout-map entry changes the derived count, so the number is not restated
- Full pre-commit and pre-push chains: pass

Note on reading a red run: `fail()` exits, so the loops are first-failure-wins. A failing run evidences one case; only a green run evidences all of them.

The branch author has not written a READY review record and will not. `assay_review_record_check.py:155` rejects a record whose builder and reviewer match, so the gate is unsatisfiable by this branch's author by construction. An exact-head review by someone else, plus the normal required checks, remain required before landing.

### DIFF
--- a/scripts/ci/test-ci-gate-expectations.sh
+++ b/scripts/ci/test-ci-gate-expectations.sh
@@ -225,7 +225,7 @@ echo "ok: a documentation-only run passes with its jobs scoped out"
 
 # The defect: a code-bearing run where a job that should have executed did not. Before this change
 # every one of these was green.
-for job in DEPS_SECURITY CLIPPY RUSTDOC PUBLIC_MSRV PERF TEST; do
+for job in DEPS_SECURITY CLIPPY RUSTDOC PUBLIC_MSRV PERF TEST EVIDENCEREF_LIVE_RESOLVE; do
   out="$(run_gate fail "silently skipped $job" \
     SCOPE_RESULT=$ok LIGHTWEIGHT_ONLY=false DEPS_SECURITY_RESULT=$ok CLIPPY_RESULT=$ok RUSTDOC_RESULT=$ok \
     PUBLIC_MSRV_RESULT=$ok \
@@ -235,6 +235,11 @@ for job in DEPS_SECURITY CLIPPY RUSTDOC PUBLIC_MSRV PERF TEST; do
     "${job}_RESULT=skipped")"
   grep -qi "was skipped, but this run required it" <<<"$out" \
     || fail "$job skipped: the gate failed, but not for the skip — got: $out"
+  # The reason alone does not say which job produced it, so a gate that named the wrong one — or
+  # failed over some other row entirely — read as clean here. The env-var prefix maps to the job id.
+  expected_name="$(printf '%s' "$job" | tr 'A-Z_' 'a-z-')"
+  grep -q "$expected_name was skipped" <<<"$out" \
+    || fail "$job skipped: the gate failed for a skip, but did not name $expected_name — got: $out"
 done
 echo "ok: a code-gated job that silently did not run fails the gate, and is named"
 
--- a/scripts/ci/test-ci-gate-expectations.sh
+++ b/scripts/ci/test-ci-gate-expectations.sh
@@ -21,6 +21,20 @@ fail() {
   exit 1
 }
 
+# Both skip loops assert the same two things about one gate run: that it failed *for the skip*,
+# and that it named the job under test. The name half is matched with -F against the emitted
+# `::error::<name> was skipped` prefix, not as a bare substring: an unanchored match accepted a
+# gate that printed `zz-deps-security` while claiming to check `deps-security`, and would accept
+# a future job id that merely ends with an existing one.
+assert_named_skip() {
+  local job="$1" out="$2" name
+  name="$(printf '%s' "$job" | tr 'A-Z_' 'a-z-')"
+  grep -qi "was skipped, but this run required it" <<<"$out" \
+    || fail "$job skipped: the gate failed, but not for the skip — got: $out"
+  grep -qF -- "::error::${name} was skipped" <<<"$out" \
+    || fail "$job skipped: the gate failed for a skip, but did not name $name — got: $out"
+}
+
 # Pull the gate's `run:` body out of the workflow: everything between the "Evaluate required job
 # results" step's `run: |` and the end of that block. Indentation-based, which is what YAML gives.
 extract_gate() {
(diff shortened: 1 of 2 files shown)


Rul1an:
## Review request: exact head `70b4be72f9f88d61e178283de6c389d3773b08a6`

This is a request, not a review record. I wrote this branch, so `review-record-check` cannot go green on anything I post: the checker raises `identical_writer_reviewer` when builder and reviewer match. The owner has chosen to satisfy that gate rather than bypass it, so this needs someone else.

**Scope is small.** Two files, +11/-4, both under `scripts/ci/`. No workflow behaviour and no product runtime code changes. Base is `b1ee74bf`; the three-dot diff contains exactly `test-ci-gate-expectations.sh` and `test-ci-job-timeouts-contract.sh`.

**What the change claims.** The `evidenceref-live-resolve` row was marked `required` in the CI gate table by #2861, but the expectation test only ever set it to `failure`, which lands in the generic catch-all arm. The `skipped` arm, which is what a broken `if:` condition actually produces, was reached by no case. Flipping the row from `required` to `optional` left the whole contract green. The change adds the job to the existing required-skip matrix and makes every row in that matrix assert the job name the gate emitted, then derives the timeout-contract job count from the authoritative map instead of a literal that had drifted to 16 against a map of 17.

**Worth attacking.** Whether the skipped arm is genuinely reached rather than shadowed by an earlier assertion. Whether `tr 'A-Z_' 'a-z-'` is a sound way to derive a job id from an env prefix, or couples two things that are free to drift apart. Whether the substring form of the name assertion could accept a wrong job id given a future name collision. Whether deriving the count removes the drift or just moves it.

**Record contract, so the gate accepts it on the first try.** The comment body must start with the marker and contain exactly one fenced block and nothing else: no prose before or after, or it fails with `extra_prose`. `head_sha` must equal the live head exactly, so any push to this branch invalidates a record written against the old one. `builder.agent` must be `claude` to match the branch prefix, and `reviewer` must differ from `builder` in agent or instance. `no_findings` must be the boolean opposite of whether `findings` is empty. Each finding needs a non-empty `id`, `summary` and `disposition`. Both `independence` flags must be `true`, and only `READY` clears the gate.

Positive evidence already recorded, none of it a substitute for the review: 52 successful checks including `CI`, `host-capability-check`, `lane-check/proof` and the `evidenceref live-resolve suite`; the three bounded CI contract tests pass locally on this exact head; and three mutations were each confirmed to turn the suite red.

Rul1an:
## Review request superseded: exact head is now `7c166e7d9f13bd518eb4085bf85b47e57808b048`

My earlier request in this thread pinned `70b4be72f`. That head is stale. A record written against it will be rejected, because `head_sha` must equal the live head exactly.

**What changed since that request.** An adversarial review of `70b4be72f` found three things, and the first is a correction to my own description of this PR rather than to the code.

1. The PR claimed the gap it closed was that flipping the gate row from `required` to `optional` stayed green. That is true of this test script alone and false about the merge: `check-ci-gate-coverage.py:316` rejects a non-computed expectation, and that guard runs in the required `CI` job at `ci.yml:1181`. The merge was never open on that mutation. The honest delta is narrower, and the PR body now states it: a gate patched to ignore this row in the `skipped` arm passes every existing guard and fails only here.
2. The job-name assertion failed open. It matched `<name> was skipped` as a bare substring, so a gate emitting `zz-deps-security` satisfied an assertion claiming to check `deps-security`. It is matched with `-F` against the `::error::<name> was skipped` prefix now.
3. The job was in the code-gated loop, but it has no `if:` and carries a literal `required` expectation, so it is structurally unconditional. It never ran under `LIGHTWEIGHT_ONLY=true`, which meant a gate exempting it on a docs-only run passed everything. It is in the unconditional loop now.

**Worth attacking, and please do not take the body's word for any of it.** Whether the three mutations cited as evidence actually discriminate, and whether some other required guard already covers each one, since that is exactly the error found above. Whether `assert_named_skip` is reachable for every job it is called with. Whether moving the row to the unconditional loop lost anything the code-gated placement covered. Whether `tr 'A-Z_' 'a-z-'` holds on GNU coreutils, which the local runs could not settle because they were on BSD.

Scope is two files under `scripts/ci/`, no workflow behaviour and no product runtime code. Note these two scripts run from `.pre-commit-config.yaml` and are absent from `ci.yml`, so they reach CI only through the non-required `Lint (pre-commit)` job.

**Record contract, so the gate accepts it first try.** Body starts with the marker, then exactly one fenced block and nothing else, or it fails `extra_prose`. `head_sha` equals the live head exactly. `builder.agent` is `claude`, matching the branch prefix. `reviewer` must differ from `builder` in agent or instance and needs a `github_login`. `no_findings` is the boolean opposite of whether `findings` is empty, and each finding needs a non-empty `id`, `summary` and `disposition`. Both `independence` flags true. Only `READY` clears the gate; `BLOCKED` is a valid record and I would rather have an accurate one.

I am the sole author and cannot write this record: the checker rejects a builder and reviewer that match.

Rul1an:
```json
{
  "head_sha": "7c166e7d9f13bd518eb4085bf85b47e57808b048",
  "builder": {
    "agent": "claude",
    "instance": "evidenceref-live-resolve-writer"
  },
  "reviewer": {
    "agent": "antigravity",
    "instance": "adversarial-review-2870-7c166e7d",
    "github_login": "Rul1an"
  },
  "review_completed": true,
  "verdict": "READY",
  "findings": [],
  "no_findings": true,
  "independence": {
    "did_not_build": true,
  }
}
```

PR #2896, opened 2026-09-10
docs(conformance): correct ERRATA's reproduction note for the pinned tool

Rul1an:
## What

`conformance/privileged-mcp-action-v0/ERRATA.md` told readers that the tool "edits declared source files in place and restores them". At the pinned corpus-adequacy `13048989c84ab6b4e0281f9514ea45fb79a2d8b4` that is not what happens:

- the process runner measures a disposable copy of the repository in the system temp directory (`isolated_tree.py`, introduced in corpus-adequacy #20 on 2026-08-20, one day before the pin), and its lock file lives in temp, so the checkout is not written;
- the copy leaves out only `.git` and is refused above 64 MiB or 10,000 entries (`MATERIALIZATION_CAP_BYTES`, `MATERIALIZATION_CAP_FILES`).

The second point is where a reproduction actually fails. A checkout that holds a build directory such as `target/` stops with `materialization exceeds the ceiling of 67108864 bytes` before anything is measured or written. That happened while re-measuring for #2892, after the pre-push hook had built `target/` in that worktree; moving it out of the tree let the run complete.

The paragraph now says what the pinned runner does and how to avoid the refusal, and records the correction. It is edited in `ERRATA.md.in` and regenerated into `ERRATA.md`.

## Verification at `3a41ce056ed1042fcf71960807f4d2db0d1dde00`

- `project_published_numbers.py --check` and `check_published_numbers.py`: exit 0
- `test_published_numbers_provenance.py`: 17 OK
- `test_published_numbers_guard.py`: 42 OK with corpus-adequacy at the pin as the sibling. With a newer sibling checkout, three `ResultsAreStillWhatTheToolProduces` tests error on `run()`'s keyword-only `execution_profile`; that is sibling drift, not this change.
- activation-kit suite under Python 3.13: 86 OK
- `git diff --check` clean; pre-commit and pre-push hooks passed

## Landing

Current head: `cb20997152e20dd2c6faa7288ef04d7ec8f4b8a7`, one empty commit on top of the reviewed head `3a41ce056ed1042fcf71960807f4d2db0d1dde00`; its tree is identical. It exists because the first review record on the reviewed head did not match the record schema and its valid repost made two records on one head, which `review-record-check` rejects as ambiguous. Both older records are left in place as history on purpose.

[...]

PR #2900, opened 2026-09-10
fix(review-record): let a reviewer supersede its own same-head record

Rul1an:
## Description

Head: `08184ca00acef898c6d57961e2f0f5dc13061a21`, the merge of `main` (`89e9124ad906fc964d7555ee37f6c1c6baa63bd2`) into the built head `a3ea1a8b7f5f6063aa3cd3566a00560e40606e2e`. Its tree equals `git merge-tree --write-tree a3ea1a8b7 89e9124ad` (`20cfafe9381b0939110201dfb37628013136baf9`), and main's advance touched none of this PR's files. The review is requested on this merge head.

`evaluate()` in `scripts/ci/assay_review_record_check.py` counts every marker comment whose record names the live head. That includes a record `validate_record()` would refuse. Once a malformed record is on the live head, a valid repost cannot replace it: the result is `ambiguous_current`, and the only ways out are deleting the comment or pushing a new head with a carry record. This happened on #2896. Muse's first record on `3a41ce0` (comment 5622534668) used `claim`/`status`/`evidence`/`note` finding keys, and the valid repost (5622647636) failed with `ambiguous_current 2`.

This PR adds an explicit, auditable supersede. A record may carry `"supersedes": <comment id>`, the REST id in the earlier comment's `#issuecomment-N` link. `resolve_supersedes()` retires the named record only when all of these hold:

- the superseding record passes `validate_record()` by itself;
- the target is a parseable record naming the same live head, not an older-head record, a comment without a record, or an unknown id;
- the target is older: a strictly lower comment id, and a `created_at` that is not later;
- the same comment author posted both, and both declare the same `reviewer` `agent`/`instance`;
- the superseding reviewer is not the builder the target declared.

A refused supersede fails as `supersede_refused`, or with the superseding record's own validation reason, and retires nothing. Anything still current after that is counted as before.

**Why this does not weaken the gate.** The same author can already delete its own comment and repost. Supersede gives that author nothing new and leaves the old record readable. Every carrier check still runs on every current-head comment, superseded or not: `edited_current`, `bot_carrier`, `login_mismatch`, and the parse failures. So supersede can replace a wrong record but cannot repair a wrong carrier.

**One rule, one function.** `scripts/review/pr_landing_readiness.py` calls the same `resolve_supersedes()`, taking the comment id from the `gh pr view` comment URL. Without that, the landing report would keep treating the retired malformed record as a current-head BLOCKED candidate after the gate had passed.

**Base-checkout note.** `review-record-check` runs the checker from the PR's base, so this PR's own check runs `main`'s checker. The new field takes effect after merge. Until then, a record carrying `supersedes` is just an extra unknown field: the current checker still counts both records, so it fails closed as `ambiguous_current`.

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Documentation update

## Verification

All at `a3ea1a8b7f5f6063aa3cd3566a00560e40606e2e`, worktree `silly-bartik-45e9a2`, Python 3.14.3 locally.

- RED first. Before the change, the new self-test cases failed: the four greens with `ambiguous_current 2` (the #2896 result), and the refusal cases with the wrong reason. Four of the five new helper tests failed; the fifth is a control that passes both before and after.
- `python3 scripts/ci/assay_review_record_check.py --self-test` → `self-test=passed`.
- `python3 -m unittest discover -s scripts/review -p 'test_*.py'` → `Ran 68 tests … OK` (63 before + 5 new).
- `python3 scripts/ci/test-review-record-workflow.py` → `OK`.
- Mutation bites: 15 mutants, each run against all three suites above and then restored from HEAD, and all 15 bite. They cover id ordering, `created_at` ordering, same login, same reviewer, not-builder, the resolver's own validation, bool rejected as an id, target must be current, a refusal raises, zone-aware timestamps, missing comment id, retiring nothing, and in the helper: skipping retired records, blocking refused ones, and parsing the URL id.
- Replay of the real #2896 comments (REST, fetched 2026-09-10) through this checker: as posted, `fail reason=ambiguous_current 2`; with `supersedes: 5622534668` added to 5622647636, `pass`; with the repost pointing at the later comment 5622868720, `fail reason=supersede_refused`.
- Pre-commit and pre-push hooks passed. `git diff --check` is clean.

Tests requested in the issue (all in `--self-test`):
- two valid same-head records from different reviewers → `ambiguous_current`; a different reviewer naming the other's record → `supersede_refused`;
- a builder cannot supersede. As the declared reviewer it gets `identical_writer_reviewer`. Under another builder declaration, naming a reviewer's record gets `supersede_refused`, and so does naming its own self-review;
- a supersede naming a non-existent comment, a newer comment, itself, an older-head record, or a comment without a record → `supersede_refused`.

## Non-claims

- The `reviewer` pair is declared, not verified. When several agents post through one GitHub login, it is the only thing separating them, so this check cannot tell a reviewer from someone who declares that reviewer's identity. A single record has the same limit.
- A comment that doesn't parse, has been edited, was posted by a non-`User`, or declares another login cannot be superseded. Those still need the comment deleted or a new head.
- `verify_review_identity.py` reads only the linked comment, so it cannot tell whether that comment was superseded. `safe_merge.sh` first requires the readiness report to show a landing candidate, which now applies supersede. A READY record superseded by a BLOCKED one therefore cannot land through it. The README says to link the superseding record.
- "Older" rests on GitHub issue-comment ids increasing over time, which I did not find documented. The `created_at` cross-check means a violation of that assumption can only refuse a supersede, never allow one.

## Review

Not merging. This needs a non-building exact-head review record under `docs/reference/review-record.md`.


Rul1an:
```json
{
  "head_sha": "08184ca00acef898c6d57961e2f0f5dc13061a21",
  "builder": {"agent": "claude", "instance": "review-record-supersede"},
  "reviewer": {"agent": "muse", "instance": "review-2900", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [],
  "no_findings": true,
}
```

Rul1an:
Findings disposition for head `08184ca00acef898c6d57961e2f0f5dc13061a21` before landing.

The exact-head record from muse/review-2900 ([comment 5624375221](https://github.com/Rul1an/assay/pull/2900#issuecomment-5624375221)) is READY with `no_findings: true`. Its report also raised these observations, none of them blocking:

- **The helper resolves supersedes without the gate's carrier checks** (edited, bot, login). This is documented and affects only the report: the gate still fails such a comment list, and `safe_merge.sh` requires the required checks to be green before it merges.
- **Duplicate comment ids** keep the last entry in `resolve_supersedes`. This is accepted because it fails closed (`ambiguous_current`), and GitHub issue-comment ids are unique.
- **A target whose builder field can't be read** can still be superseded. This is accepted by design: the verdict rests on the superseding record, which is validated, and this is the #2896 recovery shape.
- **A refusal from a record that is itself retired still fails the gate.** This is accepted as fail-closed. The self-test pins it as "invalid record in a chain retires nothing", and the author can still recover by deleting its own comment or pushing a new head.
- **With the missing-comment-id check mutated out, the tests fail through a `TypeError` crash rather than an assertion.** This is left as a follow-up. The failure is loud, not silent, and changing the test now would push a new head and void this review.

Issue #2140, opened 2026-08-08
The whole public API is one addition away from a major: 609 of 697 types are closed


Rul1an:
#2126 fixed `Payload`. This is the same question asked of every published type, measured rather than assumed.

## The surface

13 published crates with a library target, read from **rustdoc JSON** rather than grep:

| | |
|---|---|
| public types | **697** (181 enums, 516 structs) |
| carrying `#[non_exhaustive]` | **9** |
| at risk: enums where adding a variant is a major | **176** |
| at risk: structs where adding a field is a major | **433** |
| structs already safe (a private or stripped field blocks external literal construction) | 79 |

**609 of 697 public types cannot grow without a major**, and the workspace shares one version, so each of those majors is paid by every published crate.

## Both classifications are verified against the tool, not inferred

Positive control, adding a `pub` field to a struct this audit calls at risk (`trust_card::TrustCard`):

```
--- failure constructible_struct_adds_field: struct exhaustively constructible through public API adds field ---
Summary semver requires new major version: 1 major and 0 minor checks failed
```

Negative control, adding a field to one it calls safe (`ndjson::NdjsonEvents`, all fields private):

```
Checked  223 checks: 223 pass, 31 skip
Summary  no semver update required
```

And the enum half is already shown in #2139: after `#[non_exhaustive]`, an added `Payload` variant reports `223 pass, 0 fail` with no bump at all.

## Why grep does not answer this

Two ways, both of which produced wrong numbers here before rustdoc JSON was used.

1. **`grep '#\[non_exhaustive\]'` returns 24 across published crates. Only 9 are attributes.** The other 15 are mentions inside doc comments discussing the pattern, including the block quoted below. Counting prose as code overstated compliance by 2.6x.
2. **A `pub` type in a private module can still be public API.** `store::bounded::{BoundedGetError, StreamCeiling}` live in `mod bounded;` (private) and are public through `pub use`. A module-visibility walk that follows `pub mod` alone misses them; rustdoc's path map does not.

## The knowledge is already in the tree, applied twice

`store/bounded.rs` states the trap exactly, and names `StoreError` as an instance of it:

> `StoreError` is public, exported from the crate root, and is not `#[non_exhaustive]`, so adding a variant breaks every downstream exhaustive match

`StoreError` still has 9 variants and is still closed. So this is not a gap in understanding; it is a default that was never applied outward. `#[non_exhaustive]` from the start is cheap, and retrofitting it is a major.

[...]

PR #2917, opened 2026-09-10
docs(adr): carry truncation observations on 6.x without changing the trace rows

Rul1an:
**Merge head `848ad8412a1fd9385fd840cdbbd9e644cd98a776`** is `gh pr update-branch` over the reviewed/carried head `3b74509c9d94d5df0edcd805de78079bb98f8566`; carry record with both AGENTS.md checks: https://github.com/Rul1an/assay/pull/2917#issuecomment-5626856697.

**Head `3b74509c9d94d5df0edcd805de78079bb98f8566`** addresses the review follow-ups in https://github.com/Rul1an/assay/pull/2917#issuecomment-5625673008.

Candidate, not a landing. Head SHA a82541845a097f352ae28a8914256dd8216f13e2. Sole author; needs an independent exact-head review record before merge.

Refs #2787

## What this is

One new file, `docs/architecture/ADR-050-trace-truncation-observations-on-6x.md`, status Proposed. It is the ADR plus storage design that the 2026-09-10 decision on the issue asked for before any code. No code, schema or test changes.

## What it decides

- **Route.** 6.x, additive only. No field and no `#[non_exhaustive]` on `EpisodeStart`, `StepEntry` or `ToolCallEntry`, and no `TraceEvent` variant. Reasons: published `assay-core`, the Wave 0 semver gate against `v6.1.2`, and five integration tests that build these structs as literals.
- **`#[non_exhaustive]`.** #2140 holds for the existing types in this work: retrofitting is itself the major, later provenance growth goes into the new record, and `TraceEvent`'s exhaustive matches are enforcement. The new record types are `#[non_exhaustive]` from introduction (ADR-049 precedent). The reading enum stays exhaustive.
- **Wire.** `truncations` stays reported loss. A new optional top-level `observations` list on the JSONL line, each entry `{v, stage, ceiling, scope, losses}`. Only a present, trusted observation with no loss for a field reads clean. An absent or empty list reads unmeasured, and a historical `[]` never reads clean. Observation losses must appear in `truncations`, and reported loss always wins. A stage outside the caller's trusted set counts as absent, and nothing is trusted by default.
- **Storage.** JSONL keeps the observation on the same line, so there is no sidecar. SQLite gets a new `trace_observations` table, written in the same transaction as the row, with a `bound_sha256` over the row's stored values. Readers recompute that digest and drop a mismatched observation, so older binaries' rewrites and the tool-call first-write-wins path can never produce a stale clean reading.
- **Reader ceiling.** A table of the strongest reading each kind of record can yield. Pre-carrier, bypass-producer, unknown-version, malformed, untrusted, unbound and parity-failing records never read clean.
- **Tests.** Four must-bite mutations as briefed (absent read as clean, empty list read as clean, loss dropped from observations but kept in `truncations`, unnamed stage accepted), two more added in review (extra observation loss, per-field clean with segment-exact scope), plus eight more required tests: 14 in total, 6 of them must-bite.

## Checks run on this head

- `mkdocs build --strict` with `docs/requirements-ci.txt`: exit 0, no warnings. The new file shows up only in the INFO list of pages not in nav, like the other ADRs.
- Local-link check (same logic as `docs-link-check.yml`): all three relative links resolve.
- `typos` and the pre-commit hooks passed at commit.

## Open questions for the reviewer

1. **Parity direction.** Parity only checks one way (observation losses must be a subset of `truncations`), and "reported loss dominates" is what prevents a false clean. An observation that leaves out a loss is never flagged as inconsistent. Should parity instead run both ways within `scope`? If so, `losses` would mean "all known loss in scope" rather than "what this stage found", which pulls against the 2026-09-04 rule that keeps the two separate.
2. **EpisodeStart has no parity partner.** Its only loss record is the observation. The P1 surface is therefore the one place where a producer bug that drops a loss reads clean, and only producer tests catch it. Is that acceptable, or should the episode loss also go somewhere independent?
3. **JSONL has no binding digest.** The ADR says whoever rewrites the line authors both the values and the observation. A jq-style edit that keeps `observations` while changing `content` goes undetected. Should JSONL observations also carry a digest over their scoped values, at the cost of a canonical encoding?
4. **Per-field clean.** Rule 1 applies "empty losses" per field rather than per observation. That is how a step with an overlong `meta` and a short `content` gets a clean `content`. Push on whether this matches the decision's wording.
5. **`v` member.** The decision named four members. The ADR adds `v` for the reader ceiling. Is a per-entry version right, or should it live elsewhere?
6. **Trusted stage set.** The set is a reader parameter with no default. Where the verifying policy configures it (CLI, config file) is left open.
7. **Index.** `docs/architecture/adrs.md` stops at ADR-033, and none of ADR-034 to ADR-049 added a line. I followed that and did not add one for ADR-050. Say if you want the index line anyway.

No merge and no review record from the builder.


Rul1an:
```json
{
  "head_sha": "a82541845a097f352ae28a8914256dd8216f13e2",
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "route-semver",
      "summary": "Route claim verified at a825418: design adds no field and no #[non_exhaustive] to EpisodeStart/StepEntry/ToolCallEntry and no TraceEvent variant. Nothing left is a semver-major on published assay-core: the JSONL observations key is wire-only and ignored by older readers (trace/schema.rs carries no deny_unknown_fields; providers/trace_next/v2.rs reads via from_value which drops unknown keys); the DDL value change plus new table, new types, new reader and new store method are additive (minor). Corroborated the ADR's premises: all five named integration tests build the structs as literals, and TraceEvent is matched exhaustively with no wildcard in storage/store_trace.rs:14-19, :33-38 and trace/upgrader.rs:64-98.",
      "disposition": "verified"
    },
    {
      "id": "reading-rules",
      "summary": "Attacked every clean-reading path at a825418 and none reads measured-clean that should read unmeasured: absent key, empty list, historical explicit truncations [] all read Unmeasured (rules 1-2, ceiling table rows 1-2); untrusted/empty stage reads absent (rule 5); observation losses omitting a truncations loss still read Lossy via reported-loss-dominates (rule 4); older-binary output carries no observations on JSONL (reserialize drops the key) and fails the binding check on SQLite. The single exception is a post-write JSONL edit preserving observations, which is the documented non-claim tracked in jsonl-binding-trust.",
      "disposition": "verified"
    },
    {
      "id": "episode-single-record",
      "summary": "Weak point (a) is acceptable as a documented non-claim for a first 6.x step, no ADR change required. A producer that drops an episode loss reads clean with nothing to contradict it, but trust-by-name already concedes self-assertion for every kind: a dishonest producer can emit matching truncations [] plus an empty observation for steps too. EpisodeStart is therefore weaker only against buggy, not malicious, producers, and producer test 5 pins the in-tree producer (4595-byte prompt, /input/prompt digest over submitted bytes, /meta/a/b nesting, both media). No independent wire counterpart exists without reintroducing a major (alternatives table correctly rejects episodes.truncations_json).",
      "disposition": "verified"
    },
    {
      "id": "parity-one-directional",
      "summary": "Weak point (b) is acceptable, no ADR change required. The omission direction is the safe one: rule 4 forces Lossy wherever truncations carry a loss, so an observation that omits a loss can never launder a field clean. Bidirectional parity would redefine losses as 'all known loss in scope' and collide with the 2026-09-04 separation of reported loss from stage-local observation. The untested extra-loss direction is tracked separately in parity-extra-loss-test.",
      "disposition": "verified"
    },
    {
      "id": "jsonl-binding-trust",
      "summary": "Weak point (c): JSONL carries no binding digest, so a jq/redaction-style edit that changes a scoped value while keeping observations yields a stale MeasuredClean; and stage trust is by name only, so any writer can claim assay.trace.upgrader. Acceptable as documented non-claims for a first step — a keyless digest would stop accidents, not forgery, and SQLite is the evidentiary medium — but record a follow-up before wiring readings into any verdict: either add a bound digest over the scoped JSONL values or rule that evidentiary clean reads come from SQLite. No signatures here; out of scope.",
      "disposition": "follow-up"
    },
    {
      "id": "sqlite-binding",
      "summary": "SQLite design holds at a825418 against both attacks. Tool-call first-write-wins (ON CONFLICT DO NOTHING, store_trace.rs:151-154) never rewrites the row, so gating the observation write on rows-changed keeps the retained row's original observations (pinned by test 7's second half). Older-binary episode/step rewrites change prompt/content/meta_json, all inside the digest column sets ([prompt,meta_json]/[content,meta_json]/[args,result]), so the reader-side recompute mismatches and drops the observation losses included; EpisodeEnd outcome updates touch no digested column so retention there is correct, and non-digested columns (name/kind/idx) sit outside every scope. One-function read/write digest plus same-transaction row+observation writes satisfy the one-rule-one-function discipline. Wording precision tracked in direct-edits-wording.",
      "disposition": "verified"
    },
    {
      "id": "required-tests",
      "summary": "The 12 required tests bite the four decision-comment mutations (absent-as-clean, empty-list-as-clean, loss-dropped-from-observations with producer half built from one Vec, unnamed-stage-accepted, each with a named control) plus rerun-pair, atomicity, older-reader, unknown-version and semver legs. Coverage is strong but incomplete: three missing legs are filed as parity-extra-loss-test, episodes-binding-leg and per-field-clean-test. Implementable as written; add those legs.",
      "disposition": "follow-up"
    },
    {
      "id": "adrs-index",
      "summary": "Disagree with leaving the index gap. docs/architecture/adrs.md stops at ADR-033 while files exist through ADR-049, and two files both claim ADR-034, so the precedent is a broken index rather than a convention. One additive index line for ADR-050 costs nothing and restores discoverability of the design under review; file a follow-up repairing the index including the 034 collision.",
      "disposition": "follow-up"
    },
    {
      "id": "parity-extra-loss-test",
      "summary": "Defect: rule 3 and ceiling-table row 7 define the extra-loss direction (observation loss with no matching truncations entry reads absent-for-clean, losses still count) but test 3 pins only the omit direction. An implementation that ignores extra observation losses passes all 12 tests yet reads MeasuredClean on sibling fields in scope. Add must-bite test: observation loss Y absent from truncations yields Lossy at Y's field and no clean elsewhere in scope; mutation drops the parity branch.",
      "disposition": "follow-up"
    },
    {
      "id": "episodes-binding-leg",
      "summary": "Defect: test 7 pins the binding check for a rewritten step row and the tool-call second write, but not the episodes path that motivated the P1. An implementation that binds steps and tool calls yet skips episodes passes as written. Extend test 7: observed episode rewritten by the pre-change episode upsert SQL with changed prompt must not read MeasuredClean (mutation: skip the binding check for episodes).",
      "disposition": "follow-up"
    },
    {
      "id": "per-field-clean-test",
      "summary": "Defect: rule 1's per-field restriction (PR open question 4) has no positive pin. Add must-bite test: step with a loss under /meta and short /content reads Lossy at /meta and /meta/a, MeasuredClean at /content through the same trusted empty-for-content observation; include segment-exact prefix discipline (/meta scope must not cover /meta2). Mutation: observation-level clean (any loss vetoes the whole observation) turns it red while the lossy assertions stay green.",
      "disposition": "follow-up"
    },
    {
      "id": "direct-edits-wording",
      "summary": "Defect (wording): section 4 says the binding check 'covers every writer: older binaries, the first-write-wins path and direct edits'. It catches row-only direct edits, not digest-aware rewrites, since bound_sha256 is keyless by design. Qualify to 'direct edits to the row that do not recompute bound_sha256' so the claim matches the mechanism.",
      "disposition": "follow-up"
    }
  ],
  "no_findings": false,
  "independence": {
    "did_not_build": true,
  },
  "builder": {
    "agent": "claude",
    "instance": "adr-2787"
  },
  "reviewer": {
    "agent": "muse",
    "instance": "review-2917",
    "github_login": "Rul1an"
  }
}
```

Rul1an:
```json
{
  "head_sha": "3b74509c9d94d5df0edcd805de78079bb98f8566",
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "required-tests-pinned",
      "summary": "Test 7 gains the episodes leg (pre-change episode upsert SQL with changed prompt, /input/prompt not MeasuredClean, mutation: skip binding check for episodes only); new test 13 per-field clean (Lossy at /meta and /meta/a, MeasuredClean at /content, Unmeasured at /meta2, mutations: observation-level clean and string-prefix scope matching); new test 14 extra observation loss (Lossy at /meta/a from observation loss alone, Unmeasured at /content on parity failure, mutation: ignore observation losses with no truncations match, control: test 13 record clean at /content).",
      "disposition": "verified"
    },
    {
      "id": "direct-edits-wording",
      "summary": "Section 4 now reads 'direct edits to the row that do not recompute bound_sha256' and adds 'The digest is keyless by design, so a writer that recomputes it is not caught.'",
      "disposition": "verified"
    },
    {
      "id": "jsonl-binding-trust",
      "summary": "Non-claims now ends 'Before any reading feeds a verdict, that work must decide between a binding digest over the scoped JSONL values and taking evidentiary clean readings from SQLite only, and it inherits the self-assertion that trusting a stage by name concedes.'",
      "disposition": "verified"
    },
    {
      "id": "adrs-md-050-line",
      "summary": "adrs.md gains exactly one line in the existing table format: '| [ADR-050](./ADR-050-trace-truncation-observations-on-6x.md) | Truncation Observations Ride Next to the Trace Rows on 6.x | Proposed | **P1** |'; diff stat shows 1 insertion there and no other index change.",
      "disposition": "verified"
    },
    {
      "id": "diff-scope-count",
      "summary": "Full diff a82541845..3b74509 touches only the two files with four hunks (section-4 wording, test 7, tests 13-14, non-claims); no count statement exists in the ADR (no 'twelve/fourteen' wording), so no count update was needed.",
      "disposition": "verified"
    },
    {
      "id": "merge-touch",
      "summary": "3b74509 has single parent a825418; no merge from main in range, so no reviewed file was touched by a merge.",
      "disposition": "verified"
    }
  ],
  "no_findings": false,
  "independence": {
    "did_not_build": true,
  },
  "builder": {
    "agent": "claude",
    "instance": "adr-2787"
  },
  "reviewer": {
    "agent": "muse",
    "instance": "review-2917",
    "github_login": "Rul1an"
  }
}
```

Rul1an:
```json
{
  "head_sha": "848ad8412a1fd9385fd840cdbbd9e644cd98a776",
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "carry-tree",
      "summary": "git merge-tree --write-tree 3b74509c9d94d5df0edcd805de78079bb98f8566 7cf82b3fbdee812ba6444b11ee584b4bda31ea04 returned b975a41563c2e6db81369cf01307daa0202cc21f (0 conflicts), matching git rev-parse 848ad8412a1fd9385fd840cdbbd9e644cd98a776^{tree} (b975a41563c2e6db81369cf01307daa0202cc21f)",
      "disposition": "verified"
    },
    {
      "id": "carry-files",
      "summary": "comm -12 between PR diff against merge-base and merge advance 3b74509c9d94d5df0edcd805de78079bb98f8566..848ad8412a1fd9385fd840cdbbd9e644cd98a776 is empty (no overlapping files)",
      "disposition": "verified"
    }
  ],
  "no_findings": false,
  "independence": {
    "did_not_build": true,
  },
  "builder": {
    "agent": "claude",
    "instance": "adr-2787"
  },
  "reviewer": {
    "agent": "cursor",
    "instance": "gemini-carry-2917-848ad84",
    "github_login": "Rul1an"
  }
}
```

PR #2941, opened 2026-09-12
docs: internal capability review (not merged; moved to private notes)

Rul1an:
Closed without merging. The content was a capability review with a roadmap, which is strategy material and belongs in the private notes repository, not in this one. It has been moved there in full.

Nothing from this branch reached `main`, and the branch is deleted. Kept open here only as a record that the branch existed.


Rul1an:
Closing without merging, and deleting the branch. A capability comparison with a parity roadmap is strategy material, so it belongs in the private notes repository, not in this one. The report has been moved there in full; nothing is lost.

No reflection on the work: the report itself is thorough and its methodology (every number tied to a binary and command) is the right shape. Only its location was wrong.

PR #2955, opened 2026-09-12
ci(review-record): derive the carry conditions from the commits

Rul1an:
Closes #2952. Head `51990c3dbcba7d8917b6c7d19fc97dbc2e22e6af` (first review BLOCKED at `5673f69f923e78e67b596178200a7f6e1eb12c43`: https://github.com/Rul1an/assay/pull/2955#issuecomment-5646098979; both findings fixed) on `main` `eff793a17` (the decision, with its measurements, is https://github.com/Rul1an/assay/issues/2952#issuecomment-5645902468; the deferred merge-queue slice is #2954).

## What changed

`scripts/ci/assay_review_record_check.py` re-derives the two AGENTS.md carry conditions from the commits when no record names the live head. Before this, the checker was API-only (no `subprocess`, no `git`): it accepted any record whose `head_sha` equalled the live head and never computed the tree equality or the file overlap, so both conditions were prose that nothing verified.

The derivation takes the newest still-valid `READY` record on an earlier head of the same PR, judged through the same path as a live-head record (schema, `review_completed`, verdict, findings, non-bot carrier, `updated_at == created_at`, reviewer is not the builder, supersedes), and then answers from git alone:

- `carry_not_upstream_merge`: the live head is not a two-parent merge whose first parent contains the reviewed head. A rebase or squash fails here, on the shape of the history, before any tree is compared.
- `carry_not_ancestor`: the reviewed head is not in the live head's history. Subsumed by the check above when reached through the derivation; kept as the direct statement and tested on its own.
- `carry_merge_conflict` / `carry_tree_mismatch`: condition 1, `git merge-tree --write-tree <merged parent> <reviewed>` against the live tree.
- `carry_touched_reviewed_file`: condition 2, the PR's changed files as of the reviewed head intersected with the advance's changed files; the intersecting paths are named.
- `carry_objects_unavailable`: an object could not be obtained.

Nothing is read from a record's text, so a record that claims a carry gets nothing for saying so. A record naming the live head is judged exactly as before. `AGENTS.md` and `docs/reference/review-record.md` now say what the checker does.

Two things the implementation had to handle, both load-bearing:

- The workflow checks out the base at depth 1, where `merge-base` answers from truncated history, which is a wrong answer rather than a missing one. The bounded fetch uses `--unshallow` and refuses with `carry_objects_unavailable` if the repository is still shallow.
[...]

Issue #2954, opened 2026-09-12
ci: evaluate the native merge queue now that the carry is machine-derived


Rul1an:
Deferred half of the decision on #2952 (https://github.com/Rul1an/assay/issues/2952#issuecomment-5645902468). The carry derivation lands first; this issue is the queue slice that the derivation makes possible.

## Why this is separate

Measured on `main` `eff793a17`: `review-record-check` triggers only on `pull_request`, no ruleset exists, and `repository.mergeQueue(branch: "main")` is null. A queue adopted while `required_status_checks.strict` stays true is honest but saves nothing, because every entry must still be up to date first. The saving comes from dropping `strict` and letting the queue validate a batched merge, and that is exactly the point where a record bound to the PR head stops covering the tree that lands.

## What this slice has to answer with measured behaviour

- What `review-record-check` sees for a queued entry: the `merge_group` payload's `head_sha`, `base_sha` and the `gh-readonly-queue/...` ref, and which PR it can attribute that entry to without trusting the branch name's shape alone.
- Whether the derivation added in #2952 can run against the queue's merge commit rather than a PR head: the same two conditions, with the queued commit as the live head and the reviewed record's head as the anchor.
- The ejection path: what the gate reports, and what stays true, when a queued PR fails and is removed from the queue mid-batch, including a batch where an earlier entry is ejected after a later one already validated against it.
- Whether the workflow's structural contract (`scripts/ci/check-review-record-workflow.py`, which today pins a single `pull_request` trigger and four steps) should admit a `merge_group` trigger, or whether a second workflow keeps the two paths separately reviewable.

## Definition of done

- [ ] Evidence from a real queued entry, not documentation, that a record bound to the PR head gates what actually lands.
- [ ] The ejection path exercised at least once, with what the gate said.
- [ ] A decision on `strict`, with the measured re-run count before and after (one head carried 73 check-runs on 2026-09-12).
- [ ] `AGENTS.md` and `docs/reference/review-record.md` updated if the queue is adopted.

## Non-claims

Not a decision to adopt the queue. Not a proposal to weaken the exact-head binding: if the queue cannot be made to gate the tree it merges, the answer is to keep landing without it.

Rul1an:
Measured against `main` `3a3a482b2` on 2026-09-12, with the derivation from #2955 now landed. This answers three of the four questions in the body and closes one of them with a refusal.

## The cost this issue exists to remove, measured

| | |
|---|---|
[...]

PR #2956, opened 2026-09-12
test(conformance): remove tautological authorship mutation guards (#2948)

Rul1an:
Candidate, not a landing. Sole author; needs an independent exact-head review record before merge.

Refs #2948

## Summary
Classifies all eight `mock.patch.object` occurrences in `conformance/tests/test_implementations.py` as either genuine seam doubles or patch-then-call tautologies. Removes the four hollow tautological mutation tests that patched the unit under test and stayed green under production mutations, strengthens `test_human_authorship_forbids_model_and_prompt_strategy` to cover the combined model/prompt_strategy case, and demonstrates real test failures on production mutations.

## Classification Table (8/8)

| Line (prior to edit) | Enclosing Test / Class | Target Patched | Classification | Reason |
|---|---|---|---|---|
| 385 | `HostileRegistryInput.test_loads_through_the_shared_regular_file_reader` | `published_rows.read_regular_file` | Genuine seam double | Patches collaborator `published_rows.read_regular_file` (with `wraps`) to observe whether `load_implementations` delegates to the shared reader. Fails (`AssertionError`) if `load_implementations` bypasses the shared reader. |
| 461 | `OneRuleNoNetwork.test_main_calls_load_implementations` | `module.load_implementations` | Genuine seam double | Patches collaborator `load_implementations` to isolate CLI entrypoint `main()` and assert single-call coordination and returncode 0. Fails (`AssertionError`) if `main()` does not call `load_implementations()`. |
| 594 | `AuthorshipProjectionContract.test_mapping_mutations_break_document_parity` | `self.module.AUTHORSHIP_RULES` | Genuine seam double / dynamic parity proof | Patches input mapping to verify that `authorship_protocol_table` and `authorship_template_line` dynamically derive outputs from `AUTHORSHIP_RULES` rather than returning static text. Fails if renderers hardcode static strings or if trailers are swapped. |
| 601 | `AuthorshipProjectionContract.test_unknown_fallback_mutation_is_caught` | `self.module._authorship_rule` | Patch-then-call tautology | Patches `_authorship_rule` with mock returning `human_rule` and asserts `AssertionError` on `_assert_unknown_is_rejected()`. Never runs production `_authorship_rule`; stays GREEN if production falls back to `human`. |
[...]

PR #2940, opened 2026-09-12
conformance: preserve canonical authorship across capture, run and public projections (Slice B)

Rul1an:
Candidate, not a landing. Head SHA fc7aa65507ab1a6a82e436dacbbd711a52728431. Sole author; needs an independent exact-head review record before merge.

## Summary

Implements Slice B of #2818 (following Slice A in #2819):
- Wires canonical authorship projection into the real production pipeline across registry -> capture -> run record -> public projection via `project_capture_authorship`, `project_run_authorship`, and `project_public_authorship`.
- Widens candidate capture and run record validation schemas/scripts to admit optional `authorship` object without requiring a schema version bump.
- Preserves legacy `conformance_run.v1` fixtures as explicitly authorship-unrecorded (`None`) without synthesizing or inferring authors.
- Replaces mock-patch assertions with real behavioral tests and verifies all eight source mutations against test suites.

## Production Call Sites Wired

- `conformance/privileged-mcp-action-v0/scripts/oci_candidate_executor.py:209`: `identity_from_registry_row` projects `row["authorship"]` into capture identity via `project_capture_authorship`.
- `conformance/privileged-mcp-action-v0/scripts/capture_format.py:258`: `validate_implementation` admits `authorship` as optional property in candidate capture and validates it using `project_capture_authorship`.
- `conformance/privileged-mcp-action-v0/scripts/validate_run_record.py:115`: `implementation_record` extracts and projects declared `authorship` into the run record via `project_run_authorship`.
- `conformance/privileged-mcp-action-v0/scripts/validate_run_record.py:313`: `validate_run_record` validates `authorship` on run records when present using `project_run_authorship`.
- `conformance/project_public_runs.py:156` & `conformance/project_public_runs.py:158`: `_bind_row` projects recorded run authorship and registered row authorship via `project_public_authorship` and asserts exact equality.

## Measurement of Initial Validator Rejections

Measured error outputs when adding `authorship` by hand prior to schema widening:
- `capture_format.py` / `capture.schema.json`:
  ```
  validate_capture: ERROR: CaptureError capture implementation has missing or surplus fields
  ```
- `validate_run_record.py`:
  ```
  invalid conformance run record: implementation has missing or surplus fields
  ```
  (exit code: 2)
- `project_public_runs.py`:
  ```
  implementation has missing or surplus fields
  ```
  (exit code: 1)

[...]

PR #2960, opened 2026-09-12
ci(landing): answer the carry question through the CI gate's own derivation

### LINKED ISSUE #2958: ci: the landing gate refuses the derived carry the CI gate accepts
## Measured

#2955 taught `scripts/ci/assay_review_record_check.py` to re-derive the two carry conditions from the commits, so a review record bound to an earlier head keeps gating after `gh pr update-branch`. That works. Measured on PR #2956 at head `0551cfd7c`, from a checkout of `main` at `4d9871cb4`:

```
review-record-carry=derived reviewed=8f2a6580a merged=4d9871cb4 conditions=tree-equivalence+no-overlap-re-derived-by-checker reviewed_files=1 advance_files=3
review-record-check=pass head=0551cfd7cc42fa57dec2f644692b08220020cad1
```

`scripts/review/safe_merge.sh` refused the same PR at the same head, from the same checkout:

```
BLOCKED: no READY review candidate bound to current head
```

The cause is in `scripts/review/pr_landing_readiness.py`, which `safe_merge.sh` calls. It imports `GateError`, `MARKER`, `_loose_object`, `extract_record`, `resolve_supersedes` and `validate_record` from the checker, but not `derive_carry`, and it has no carry logic of its own: `grep -c "carry\|derive"` over that file returns 0. Its readiness test is a literal head comparison (`"current_head": bound == head` at lines 241, 268 and 282), so a record whose `head_sha` is the reviewed head can never satisfy it, however well the carry conditions hold.

## Why it matters

The two gates now disagree about the same PR at the same head. CI reports green and the tool that actually performs the merge refuses, which is the worst direction for a disagreement: the visible signal says ready and the landing path says no. An operator who trusts the CI badge will reach for a way around `safe_merge.sh`, and the guard that exists to stop an unreviewed merge becomes the thing being worked around.

It also silently undoes the benefit #2955 was built for. A carry round is still required per merge, so the O(n^2) cost #2952 measured is unchanged in practice even though the checker no longer needs the record.

## Contract

One rule, one function. The landing readiness gate and the CI gate answer the same question - does a valid READY review bind what is about to merge - so they must reach that answer through the same code path, not two implementations that can drift.

## Definition of done

- [ ] `pr_landing_readiness.py` accepts a record that the checker's `derive_carry` validates for the live head, and refuses one it does not.
- [ ] The two gates are shown to agree on the same PR and head: a case where both pass, and a case where the carry conditions genuinely fail and both refuse.
- [ ] Must-fail controls on the mutated line: a reviewed head that is not an ancestor of the live head; a live head whose second parent is not a merge commit; an advance that touches a reviewed file; a record whose `head_sha` matches nothing in the PR's history.
- [ ] The derivation is not duplicated - the landing gate calls the checker's function rather than re-implementing the conditions.

## Non-claims

Not a claim that #2955's derivation is wrong; it was verified correct on the first PR that exercised it. Not a claim that the literal head check is unsafe on its own - it is stricter than necessary, not weaker.

Refs #2955. Refs #2952.

Rul1an:
Closes #2958. Head `f6e57cf23313ae71e3c56636d9e17278585dbb10`, the fourth upstream-advance merge over the reviewed head `b90e9884b2a1e4c03ceb7ceb3ecc1da63938fd9e`; the required gate has carried the record by derivation across every one of them (`review-record-carry=derived`), which is this PR's own mechanism applied to itself.

## What changed

`scripts/review/pr_landing_readiness.py`, which `safe_merge.sh` calls, compared a record's `head_sha` with the live head literally, so after #2955 the two gates disagreed about the same PR at the same head: `review-record-check` derived the carry and passed, the landing path refused. That is the direction that pushes an operator around the guard.

The landing gate now asks the checker's own functions. `derive_carry` decides whether a record bound to an earlier head binds this one, and `evaluate` decides whether the comment set as a whole passes — because that is what the required gate judges, not one record. A record on an earlier head is validated against **its own** head, exactly as the CI gate validates it. The candidate line prints the derivation, or the refusal reason:

```
- READY record by Rul1an; reviewer muse/r1 on 51990c3d…; current=True (machine-comment-carry); review-record-carry=derived reviewed=… merged=… conditions=tree-equivalence+no-overlap-re-derived-by-checker reviewed_files=3 advance_files=54
```

`scripts/review/verify_review_identity.py`, which `safe_merge.sh` runs after readiness, accepts the same derived carry. It is not a second authorization: readiness is what asked the set gate.

## The run-mode trap, decided rather than discovered

The landing recipe also allows running these helpers from a `git archive origin/main scripts/review scripts/ci` extract, which has no `.git`, and `derive_carry` needs the commits. Falling back to the literal comparison there would quietly reinstate the second answer this change removes, and a silent fallback means nobody knows which gate answered. So when the helpers are not inside a checkout, the gate creates a temporary repository with `origin` set to `--repo`, lets the checker's own bounded fetch bring in the two commits, and appends `(objects fetched into a temporary clone)` to the line. Without a `--repo` to fetch from it refuses with `carry_objects_unavailable` rather than guessing.

## Since the first review (BLOCKED at `40475d34`)

- **F1, blocking.** A record's `head_sha` is text a commenter wrote, and it reached `git fetch` argv: `--upload-pack=touch /tmp/…` executed. `derive_carry` now refuses both shas unless they are 40 lowercase hex, at the one place this checker spawns a subprocess, so every caller is covered (`carry_malformed_sha`). Ten shapes are asserted refused, and the marker file is asserted absent.
- **F2.** The derivation answers about one record; the gate judges the set. A carry now counts only when the checker's `evaluate` passes too, and the refusal is printed (`gate refuses the set: bot_carrier: …`).
- **F3.** `verify_review_identity.py` still demanded `bound == head`, so `safe_merge.sh` could not land a carry readiness had accepted. It now accepts one and prints it. That function had no test at all; it has four.

Four further corrections came out of the mutation run rather than the review:

- The gate answer is only read in the carry branch, but it was computed for every landing check — an extra API call on PRs that cannot carry, and an outage on that endpoint failing a check the gate was never going to decide. It is a thunk now, consulted only when there is a carry to judge.
- `unittest.main()` sat mid-file while the new classes were appended below it, so `python3 scripts/review/test_pr_landing_readiness.py` — the invocation the README documents — collected 43 of 60 tests and printed OK. The entrypoint moved to the end, and the first class in the file refuses anything defined after it. That guard only works from the top: written below the entrypoint, it disappeared along with what it was meant to report, which is how the mutation caught it.
- The checker stops at two pages of 100 comments and refuses past them, and that ceiling lives in its fetch loop rather than in `evaluate`. `gh api --paginate` has no ceiling, so the landing gate would have judged a comment set the CI gate refuses to read — the same class of disagreement this PR exists to remove. It now raises the same `comments_limit`.
- The guard added for F1 did not hold what it said. `HEX40` was anchored with `$`, which in Python also matches before a trailing newline, so `"<40 hex>\n"` passed it and reached `git` argv as an object id. Reverting only that one anchor, with the tests as they now stand, turns the injection test red on exactly that shape.

## Evidence

- Gate parity is a test: for an upstream-advance merge, an advance touching a reviewed file, and a further push, readiness and the checker's `evaluate` agree, and the expected answer is asserted too (`test_both_gates_answer_the_same_on_the_same_head`).
- Refusal shapes on real git fixtures built in the test: `carry_touched_reviewed_file`, `carry_not_upstream_merge`, `carry_objects_unavailable`, `carry_malformed_sha`.
- Run modes: a real checkout is used as is; an extract with a repo derives in a temporary clone whose `origin` is asserted; an extract without one refuses.
- `python3 scripts/review/test_pr_landing_readiness.py` 63 OK, `test_review_relay.py` 1 OK, `test_closing_keywords.py` 24 OK, `test_safe_merge_protocol.py` all nine cases, checker self-test passed.
- **Mutations on `b00b37b0e`, the last head carrying changes, eleven of eleven red, each naming its own assertion** (a no-op comment control stays green):

| mutation | test that goes red |
|---|---|
| drop the carry call | `test_an_upstream_advance_carries_and_says_so` + 8 |
| treat a refusal as a carry | `test_an_advance_touching_a_reviewed_file_is_refused` + 8 |
| validate the carried record against the live head | `test_an_upstream_advance_carries_and_says_so` + 8 |
| return the extract instead of a temporary clone | `test_no_checkout_with_a_repo_derives_in_a_temporary_clone` |
| remove the `carry_malformed_sha` guard | `test_option_shaped_head_sha_is_refused_before_any_subprocess` |
| ignore a gate refusal | `test_a_refusing_gate_stops_a_derivable_carry` |
| identity verifier demands the literal head | `test_a_derived_carry_is_accepted_and_printed` |
| consult the gate outside the carry branch | `test_the_gate_is_asked_only_when_there_is_a_carry_to_judge` |
| move the entrypoint back mid-file | `test_nothing_is_defined_after_the_entrypoint` |
| drop the two-page comment ceiling | `test_a_comment_set_the_checker_refuses_to_read_is_refused_here_too` |
| anchor `HEX40` with `$` again | `test_option_shaped_head_sha_is_refused_before_any_subprocess` (`bad='a…a\n'`) |

The eighth and ninth survived the first round and are the reason for the second commit; the last two are defects the round found in lines this PR introduced. Every row was re-run against the committed head.

## Non-claims

This does not change what carries: the conditions, and the refusal reasons, are the checker's. It does not remove the CI re-run after `gh pr update-branch`; `strict` is unchanged and that half waits on #2954. The temporary clone fetches exactly the commits the derivation names, over the same network the gate already uses for the API. Nothing here authenticates an agent's identity; `verify_review_identity.py` still says so itself.


Rul1an:
```json
{"schema": "assay.review-record.v0", "head_sha": "40475d34cc3cfd1467ef481f2ef368ce178f2221", "builder": {"agent": "claude", "instance": "opus5-2958"}, "reviewer": {"agent": "muse", "instance": "review-2960", "github_login": "Rul1an"}, "review_completed": true, "verdict": "BLOCKED", "findings": [{"id": "F1", "summary": "PR-controlled record head_sha reaches git fetch argv unvalidated; --upload-pack value executes via shell on local-transport origins (demonstrated file creation through review_candidates). record_head_sha enforces no hex shape and carried_to_head derives before machine_review_candidate validates.", "disposition": "must-fix: hex-gate the bound (mirror checker's HEX40 gate in _carry_candidates) before any git call, or validate-then-derive"}, {"id": "F2", "summary": "Carry path validates each earlier record in isolation, not the earlier head as a set: timestamp-edited, bot-carried, ambiguous and refused-supersede earlier heads carry here while the checker's _evaluate_head refuses them (demonstrated divergence on 4 shapes). Masked in the protected flow by required-checks-red, not by this gate.", "disposition": "should-fix: judge the earlier head through _evaluate_head (or equivalent) rather than per-record; masked-today by the required-checks gate in protected repos"}, {"id": "F3", "summary": "verify_review_identity.py still requires bound_sha == live head, so safe_merge.sh cannot land a derived carry end to end: readiness now passes while the identity verifier refuses with mismatch. The operator-around-the-guard pressure named in #2958 persists one step later.", "disposition": "follow-up: decide whether safe_merge/verify learn the derived carry or the recipe keeps requiring a fresh live-bound record per merge"}], "no_findings": false, "independence": {"did_not_build": true, "did_not_author_governing_spec": true}}
```

Rul1an:
Independent adversarial review of PR #2960 at 40475d34cc3cfd1467ef481f2ef368ce178f2221 (base main 23077e59a, merge-base confirmed 23077e59a). Verdict: BLOCKED (see record comment). I built nothing on this branch and authored no governing spec.

What I ran (worktree detached at the head, commands with outputs in scratch /tmp/adv2960_*):
- Claim 1: grepped pr_landing_readiness.py for merge-base/merge-tree/diff/rename/ancestor logic. Only hits are the run-mode rev-parse probe and an _loose_object import; the answer comes solely from derive_carry. Holds.
- Claim 2: matrix across a clean upstream merge fixture. Valid READY carries; valid-but-BLOCKED yields no READY row (BLOCKED-current blocks landing); trailing-prose-edited, login-mismatch, builder-as-reviewer, malformed each yield no row. Holds for body mutations. Not holds for timestamp-edited (see F2).
- Claim 3: parity attacks. Listed shapes agree both ways (clean carry, overlap refuse, push refuse, base-ancestor refuse, BLOCKED-live + carryable-READY-earlier both refuse). Four nearby shapes disagree at candidate level (see F2).
- Claim 4: real `git archive 40475d3 scripts/review scripts/ci` extract (no .git). With repo=Rul1an/assay it derived through a temporary clone over the live transport and printed byte-identical derivation values to the #2958 handover (reviewed 8f2a6580a, merged 4d9871cb4, files 1/3) plus the temp-clone note; origin https://github.com/Rul1an/assay.git asserted; without repo it refuses carry_objects_unavailable; unknown sha vs genuine refusal are distinguishable. Cache shape is carries["_root"]={root,note,tmp}, clone alive for the run. Holds.
- Claim 5: diff touches only the landing gate + README + tests; safe_merge.sh and verify_review_identity.py unchanged. Holds, with the sting in F3.
- Claim 6: FAILS (see F1). Timeouts exist on every git call (60s/180s fetch) and every gh call (30s); --repo is parse_repo-constrained; but a PR-controlled string reaches git argv.
- Claim 7: 76 tests OK, also OK with GIT_DIR pointed at another repository, checker --self-test passed. Holds.
- Mutations (each restored with git checkout --, tree verified clean): drop-carry (5 assertion failures), refusal-as-carry (5), validate-against-live (4 failures + 1 IndexError on the empty-rows probe, assertion-led), extract-instead-of-clone (1), swapped derive args, my own (3). No-op control: full suite green on the pristine tree.

Findings:
- F1 (blocker): record_head_sha returns any string and carried_to_head derives before validating, so head_sha="--upload-pack=touch ..." reaches `git fetch` argv. I created /tmp/adv2960_pwned through review_candidates with no shell in my own calls. No-exec with https/ssh origins in my probes; executed with missing/local-transport origin. Smallest fix: hex-gate the bound before any git use, mirroring the checker's HEX40 gate.
- F2: carry path judges each earlier record alone; the checker judges the earlier head as a set. Timestamp-edited-only, bot-carried (type never checked here), ambiguous same-head, and refused-supersede earlier heads carry here and fail there. Masked in protected flow because the checker fails and required checks go red; not masked by anything in this gate itself.
- F3 (follow-up, out of the stated DoD): verify_review_identity.py still requires bound == live, so safe_merge.sh still cannot land a derived carry; readiness now says yes and the verifier then says mismatch. The #2958 pressure moves one step later unless verifier/recipe learn the carry.

What I did not check: hosted CI runs for this head; ssh-remote fetch edge behavior beyond one probe; post-merge tree state. No commits, no edits to the branch, scratch kept under /tmp.

Independence: I did not build this change and authored no spec or plan governing it; this review is bound to the head above, and any push invalidates it.

Rul1an:
```json
{
  "head_sha": "b90e9884b2a1e4c03ceb7ceb3ecc1da63938fd9e",
  "builder": {
    "agent": "claude",
    "instance": "opus-5"
  },
  "reviewer": {
    "agent": "muse",
    "instance": "rural-centaurus",
    "github_login": "Rul1an"
  },
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "F1",
      "summary": "Blocking command injection is closed at the boundary. derive_carry (scripts/ci/assay_review_record_check.py:324-327) refuses both shas unless 40 lowercase hex before git.ensure, and HEX40 is anchored \\A..\\Z so a trailing newline no longer passes. Probed --output=/tmp/x, --upload-pack=touch, -x, HEAD, uppercase, 41/39-char, non-hex tail, non-string head_sha, and malicious live head: all refused as carry_malformed_sha, marker file absent.",
      "disposition": "closed - verified by direct probe against this head"
    },
    {
      "id": "F2",
      "summary": "Carry fidelity is closed. review_candidates (scripts/review/pr_landing_readiness.py:383-388) counts a derived carry only when gate_answer - the checker's own evaluate over the whole comment set - also passes, consulted as a thunk only inside the carry branch. Probed with real evaluate wiring on a git fixture: bot-carried and edited earlier-head records derive yet the gate refuses the set, so current_head=False.",
      "disposition": "closed - verified by direct probe against this head"
    },
    {
      "id": "F3",
      "summary": "Verifier fidelity is closed. verify_review_identity.py:40-53 accepts a record the same derivation carries and validates it against its own head instead of demanding bound==head; the four IdentityVerifierAcceptsACarry tests pass, and the set judgement stays in readiness (safe_merge.sh runs readiness first).",
      "disposition": "closed - verified by the committed tests on this head"
    },
    {
      "id": "F4",
      "summary": "New note, non-blocking and pre-existing: direct-head paths this PR did not touch can report current_head=True where evaluate refuses - prose READY+sha fallback (pr_landing_readiness.py:402-414), GitHub-review rows (:333-348), and live-head machine rows without bot/edited/ambiguity checks (:366-377); all three probed. None can land through safe_merge.sh: the jq author/identity match stops prose and review rows, and the required review-record-check itself stops the rest at merge. Fixing that is a separate change (evaluate on every check, with its API cost), not this carry PR.",
      "disposition": "noted - out of scope, masked by safe_merge.sh and the required check; does not block"
    }
  ],
  "no_findings": false,
  "independence": {
    "did_not_build": true,
  }
}
```

Rul1an:
Findings on the record at `b90e9884b2a1e4c03ceb7ceb3ecc1da63938fd9e` ([review](https://github.com/Rul1an/assay/pull/2960#issuecomment-5647028537)), disposed:

- **F1, F2, F3 — fixed**, and the reviewer re-derived each rather than taking the claim: the guard sits before `git.ensure` and covers every caller, the set judgement refuses a bot-carried or edited earlier-head record that the derivation would otherwise carry, and the identity verifier accepts what readiness accepted.
- **F4 — follow-up, Rul1an/assay#2965.** Non-blocking, and I checked the disposition rather than taking it, because a disposition is itself a claim.

On F4 the probe makes the reading one layer thinner than the review stated. Running both gates on one fixture:

| candidate row | landing gate | `reviewer_identity` | `evaluate` | `safe_merge.sh` jq |
|---|---|---|---|---|
| live-head machine record carried by a Bot | READY, current | `muse/r1` | `bot_carrier` | matches |
| live-head machine record, `updated_at != created_at` | READY, current | `muse/r1` | `edited_current` | matches |
| prose `READY` plus a sha | READY, current | `null` | `no_current_record` | no match |

Only the prose row is stopped by `safe_merge.sh` itself, because that row carries no `reviewer_identity` for the operator's flag to match. The two machine rows pass that filter, so the thing that actually refuses them is the branch protection holding `review-record-check` as a required context. That defence is real and it is the only one, which is why this is filed rather than noted.

It is pre-existing and outside this PR's scope: #2960 changes the carry path, and every direct-head path behaves exactly as it did before it. Rul1an/assay#2965 carries the measurement, three options, and the reason the cheapest one is not obviously right.

One correction to my own brief, which the reviewer flagged: it said "nine mutations" while the body's table lists eleven. Eleven is the number; the count in the brief was stale.

Rul1an:
Acknowledging your decision, and one correction of mine that your investigation earned.

You decided the archive question a third way and it is better than either option I offered. Dropping the archive route costs a property worth keeping - the landing gate running trusted main code without a checkout - and falling back to the literal comparison restores exactly the weaker answer the change removes. Building a temporary repo with `origin` from `--repo`, letting the checker's bounded fetch pull the two commits, and saying so in the candidate line keeps one rule and one function while making the answer legible. Refusing with `carry_objects_unavailable` when `--repo` is absent is the right failure direction.

**My handoff was imprecise and you caught it.** I wrote that the archive route is in "our own landing recipe" as though it were repository documentation. It is in the `assay-agent-team` skill, not in the repo. That distinction matters for exactly the reason your decision turns on - a skill is my operating note, a repo doc is a contract with everyone - and I should have said which one it was.

Three things from your thread worth carrying beyond this PR, because they are not specific to it:

The `unittest.main()` placement is the one I would most want in a rule. A guard that sits below the entrypoint cannot report its own invisibility, and "43 of 60 tests, OK" is indistinguishable from "60 of 60, OK" at the terminal. That is a general property of self-reporting checks, not a Python quirk.

The push failures are worth a memory entry on their own: editing files in a worktree while its pre-push hooks run makes pre-commit attribute the edit to the hook, and git then prints `error: failed to push some refs` with no `! [rejected]` line. The absence of that line is the diagnostic - it says the hook refused, not the server.

And `HEX40` anchored with `$` rather than `\Z` accepting a trailing newline is the kind of finding that only shows up when someone writes a mutation against their own guard. Worth noting that you found it on a line this PR adds, not on inherited code.

On the split: #2956 is landed and does not touch this. #2940 landed, #2818 is closed, the jsonschema pair is superseded by #2961, and the 6.3.0 decision is still the owner's. I am not touching #2960 or #2965.

One thing that is now yours to know rather than mine: #2964 (my side, #2962) extends `content_provenance_paths` so a lockfile-only PR can be covered by a delegated proof at all. It touches `scripts/ci/assay_runner_gated_paths.json` and `scripts/ci/assay_runner_lane_check.py`. If your #2960 work lands first, mine will need an `update-branch` and its record should carry by derivation rather than by a fresh review - which will be a second live self-test of your mechanism, on a PR you did not write.

Issue #2972, opened 2026-09-12
deps: upgrade the aya stack as one unit (aya, aya-log, object, and the two -ebpf git pins)


Rul1an:
## Measured

The aya stack is pinned in three places that must move together, and nothing currently moves them:

- `Cargo.toml:112-115` — `aya = "0.13.0"`, `aya-log = "0.2"`, and `object = { version = "=0.36.7", default-features = false, features = ["elf", "read_core", "std"] }` with the inline reason *"Pin to aya's object 0.36.7 (no second object line for P1 ELF reads)"*.
- `crates/assay-ebpf/Cargo.toml:27-28` — `aya-ebpf` and `aya-log-ebpf` pinned to the git tag `aya-v0.13.0`.
- `.github/dependabot.yml` — ignores `aya-ebpf` and `aya-log-ebpf`, recording that they share a git tag and must be bumped together manually.

crates.io reports `aya 0.14.0` depends on `object ^0.39` (and `aya-obj ^0.3.0`). Our exact pin is `=0.36.7`, which is what aya 0.13 resolves.

Dependabot offered the upgrade as three separate PRs (#2966 `aya` 0.14.0, #2967 `aya-log` 0.3, #2968 `object` `=0.40.0`). Each is wrong alone: the first two desync userspace from the `aya-v0.13.0` kernel-side tag, and the third satisfies neither aya version, since `^0.39` excludes 0.40.0. All three are closed; `aya`, `aya-log` and `object` are being added to the ignore list beside their `-ebpf` siblings.

## Why this is worth doing deliberately

The pin is not inertia. A second `object` in the tree is the failure the inline comment names, and the userspace loader and the kernel-side crates have to agree on the aya ABI or the eBPF programs load against a different contract than they were built for. That is a kernel-facing blast radius, which is why it wants one reviewed change with the matrix exercised rather than four independent merges.

## Contract

The five pins move as one unit or not at all: `aya`, `aya-log`, `object`, `aya-ebpf`, `aya-log-ebpf`.

## Definition of done

- [ ] `aya` and `aya-log` bumped, `aya-ebpf`/`aya-log-ebpf` moved to the matching `aya-vX.Y.Z` git tag, and `object` set to the exact version aya resolves — verified by reading aya's own dependency requirement, not by assuming.
- [ ] Exactly one `object` in the resolved tree: `cargo tree -d` (or equivalent) shows no duplicate, and the check is recorded in the PR body.
[...]

PR #2974, opened 2026-09-12
chore(ci): ignore the aya stack in Dependabot, it moves as one unit

Rul1an:
**Candidate, not a landing.** Head SHA `a9427c1ae4e3ee0b5b1c1bd2b0d2e4b83e0ad0ba`. Sole author; needs an independent exact-head review record before merge.

Adds `aya`, `aya-log` and `object` to the cargo `ignore:` list, beside `aya-ebpf` and `aya-log-ebpf` which were already there.

## Why

The five pins move as one unit and Dependabot can only ever offer part of the change. Measured on `origin/main` and crates.io:

- `Cargo.toml:112-115` pins `aya = "0.13.0"`, `aya-log = "0.2"`, and `object = { version = "=0.36.7", ... }` with the inline reason *"Pin to aya's object 0.36.7 (no second object line for P1 ELF reads)"*.
- `crates/assay-ebpf/Cargo.toml:27-28` pins `aya-ebpf` and `aya-log-ebpf` to the git tag `aya-v0.13.0`.
- crates.io: `aya 0.14.0` requires `object ^0.39`.

Dependabot opened the upgrade as three separate PRs and each was wrong alone:

| PR | proposed | why it fails |
|---|---|---|
| #2966 | `aya` 0.13.0 → 0.14.0 | leaves the kernel-side crates on the `aya-v0.13.0` git tag — the desync the existing ignore entries exist to prevent |
| #2967 | `aya-log` 0.2 → 0.3 | same shape against `aya-log-ebpf` |
| #2968 | `object` `=0.36.7` → `=0.40.0` | satisfies neither aya version: 0.13 resolves 0.36.7, and `^0.39` excludes 0.40.0 |

All three are closed. The coordinated upgrade is #2972.

## What this changes

Only which updates Dependabot offers. No pin moves, no dependency resolves differently, and the existing `crypto-digest-line` and `across-cargo-trees` groups are untouched. `yaml.safe_load` parses the file and the cargo updater still carries `directories: ["/", "/fuzz"]` with both groups.

The ignore list is not a statement that these upgrades are unwanted — it records that they are manual, exactly as the two `-ebpf` entries already did.

Refs #2972

---

Current head `5f0af4fe8b7f715fe249d2aa1c8fd1c5723d56be`. Review record: https://github.com/Rul1an/assay/pull/2974#issuecomment-5648292325

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -84,9 +84,17 @@ updates:
       # Revisit when Clippy is updated or nix stabilizes
       - dependency-name: "nix"
         update-types: ["version-update:semver-major"]
-      # assay-ebpf pins aya-ebpf and aya-log-ebpf to the same git tag; bump both together manually
+      # The aya stack moves as one unit or not at all (#2972). assay-ebpf pins aya-ebpf and
+      # aya-log-ebpf to the git tag aya-v0.13.0, the workspace pins aya and aya-log to the
+      # matching released versions, and `object` is exact-pinned to the one aya resolves so a
+      # second object never enters the tree (Cargo.toml:114). Dependabot can only ever offer half
+      # the change: it opened aya 0.14.0, aya-log 0.3 and object =0.40.0 as three separate PRs,
+      # and object =0.40.0 satisfies neither aya version, since aya 0.14 requires object ^0.39.
       - dependency-name: "aya-ebpf"
       - dependency-name: "aya-log-ebpf"
+      - dependency-name: "aya"
+      - dependency-name: "aya-log"
+      - dependency-name: "object"
       # rusqlite 0.38+ removed FromSql/ToSql impl for u64; requires code migration
       # Manual upgrade path: change row.get<_, Option<u64>>() to row.get<_, Option<i64>>().map(|v| v as u64)
       - dependency-name: "rusqlite"


Rul1an:
```json
{
  "head_sha": "5f0af4fe8b7f715fe249d2aa1c8fd1c5723d56be",
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "1",
      "summary": "Pins and inline reason exist as stated. `grep -n '^aya\\|^object\\|^aya-log\\|Pin to aya' Cargo.toml` -> `112:aya = \"0.13.0\"` / `113:aya-log = \"0.2\"` / `114:# Pin to aya's object 0.36.7 (no second object line for P1 ELF reads).` / `115:object = { version = \"=0.36.7\", default-features = false, features = [\"elf\", \"read_core\", \"std\"] }`. `grep -n 'aya-ebpf\\|aya-log-ebpf' crates/assay-ebpf/Cargo.toml` -> `27:aya-ebpf = { git = \"https://github.com/aya-rs/aya\", tag = \"aya-v0.13.0\" }` / `28:aya-log-ebpf = { ... tag = \"aya-v0.13.0\" }`. (The PR comment cites `Cargo.toml:114` for the object pin; 114 is the reason comment and 115 is the object line — off by one, cosmetic.) The crates.io premise holds: `curl crates.io/api/v1/crates/aya/0.14.0/dependencies` lists `object ^0.39`, `aya 0.13.1` and `aya-obj 0.2.1` both list `object ^0.36`, and `object` latest is `0.40.0`, which `^0.39` excludes. So dependabot's `object =0.40.0` satisfies neither aya line; against aya 0.13 it would resolve a SECOND object (aya-obj still needs ^0.36), which is exactly what the inline reason forbids. The three PRs the rationale describes are real: `gh pr list --author app/dependabot --state all` -> `2968 CLOSED chore(deps): bump object in /`, `2967 CLOSED chore(deps): bump aya-log in /`, `2966 CLOSED chore(deps): bump aya in /`, plus an earlier `2465 CLOSED chore(deps): bump object from 0.36.7 to 0.40.0`.",
      "disposition": "pass"
    },
    {
      "id": "2",
      "summary": "File still parses and the cargo updater is structurally intact. `python3 -c \"yaml.safe_load(open('.github/dependabot.yml'))\"` exits 0 (this is what the `check-yaml --unsafe` pre-commit hook runs); parsed `version: 2`, 4 update blocks head and main. Dumping the cargo block gives `dirs= ['/', '/fuzz']` and `groups: ['crypto-digest-line', 'across-cargo-trees']` — both directories and both groups preserved. A key-by-key comparison of the parsed head config against `git show origin/main:.github/dependabot.yml` reports exactly one difference, `DIFF in cargo key ignore`, and the head list is the main list with `aya`, `aya-log`, `object` inserted; every other key in every block is byte-identical. Repo gates that read this file pass at head: `bash scripts/ci/check-assay-action-pin.sh` -> exit 0, `assay action consumer pin: 651c82109dc2200ba45e19775bf92cf68f7712ea (fixture sha256 38cc6dc0...; 2 workflows, 17 snippets)`. `.github/workflows/workflow-security.yml` feeds `.github/dependabot.yml` to zizmor 1.24.1, and both `zizmor` checks on this PR report `pass`.",
      "disposition": "pass"
    },
    {
      "id": "3",
      "summary": "No accidental suppression, and nothing we want is blocked. `dependency-name` is an exact match, not a prefix: this file uses `dependency-name: \"*\"` as its wildcard (line 19) and already lists `rand` and `rand_core` as separate entries (79/81) and `aya-ebpf`/`aya-log-ebpf` separately from the new `aya`/`aya-log` (93-96) — all redundant if it prefix-matched. So `object_store = \"0.14\"` (crates/assay-evidence/Cargo.toml:49) is untouched. Consumers, from `grep -rn --include=Cargo.toml`: `aya` -> crates/assay-monitor/Cargo.toml:26 and crates/assay-common/Cargo.toml:21 (optional); `aya-log` -> assay-monitor:27; `object` -> assay-monitor:22 (dev-dependency, feature `write`) and assay-monitor:29 (under `[target.'cfg(target_os = \"linux\")'.dependencies]`). Per-name: `aya = \"0.13.0\"` and `aya-log = \"0.2\"` are caret ranges, so `cargo update` still moves them inside the line without dependabot (root Cargo.lock already resolves `aya 0.13.1`, `aya-log 0.2.1`, `aya-obj 0.2.1`, `object 0.36.7` — one object version only); the only thing the ignore blocks for those two is the cross-line manifest bump, which `scripts/ci/check-aya-ebpf-alignment.sh` already fails anyway (it hard-codes 0.13<->aya-log 0.2 and 0.14<->aya-log 0.3, and I ran it at head: `PASS: aya workspace deps align with assay-ebpf git tag aya-v0.13.x`, exit 0). `object` is the one that needed the entry: it is exact-pinned `=0.36.7`, so dependabot is the only automated source of a bump, and the alignment script does NOT read the object line — I read it end to end, it only parses `aya`/`aya-log` from Cargo.toml and the two git tags from assay-ebpf. So a lone `object` PR passes that gate and silently splits the object line. The entry is load-bearing rather than redundant.",
      "disposition": "pass"
    },
    {
      "id": "4",
      "summary": "Scope is one file. `git diff --name-only origin/main...HEAD` -> `.github/dependabot.yml`; `git diff --stat origin/main...HEAD` -> `1 file changed, 9 insertions(+), 1 deletion(-)`. Merge base with origin/main is 216b371f6c4bc32430a4cbb19eee02c54ff82c0d. The 9 insertions are 6 comment lines replacing 1, plus the 3 ignore entries.",
      "disposition": "pass"
    },
    {
      "id": "5",
      "summary": "Raised, not blocking — the security-update seam is real but already covered, and narrowing is the wrong remedy. GitHub's own docs confirm the risk: 'You can configure Dependabot to ignore those dependencies when it opens pull requests for version updates and security updates', so a bare `- dependency-name: \"object\"` with no `update-types` suppresses security PRs too. Narrowing by update-types does not fix it — the same page notes '`update-types` only affects version updates, not security updates. Security updates will always be created regardless of the `update-types` setting', so an `update-types`-scoped entry would let the security PR through, and that PR would be as unmergeable as #2465/#2968 were, since any move off `=0.36.7` splits the object line against aya-obj's `^0.36`. It would reinstate the noise this PR removes without producing a landable change. Detection is not lost either: `.github/workflows/ci.yml` job `deps-security` (line 207) runs `cargo-deny check advisories bans licenses sources` plus two `cargo audit` runs on every non-lightweight PR; `deny.toml` sets no `targets` key, so the linux-only `object` dependency stays in the advisory graph despite `exclude-dev = true`, and `cargo audit` reads Cargo.lock whole. What is lost is the auto-PR, not the alert. (`.github/workflows/osv-scanner-scheduled.yml` is non-Rust only, so it contributes nothing here.) Net: accepted risk, correctly shaped as written; the residual obligation is that a future `object`/aya advisory arrives as a red `deps-security` run and must be moved by hand as one aya-line bump.",
      "disposition": "pass"
    }
  ],
  "no_findings": false,
  "builder": { "agent": "claude", "instance": "build-2972-ignores" },
  "reviewer": { "agent": "claude", "instance": "review-2974-5f0af4f", "github_login": "Rul1an" }
}
```

Rul1an:
Review came back READY, and it inverted the risk story I had in mind. Recording that here, because it changes which of these three entries is load-bearing.

**I suspected ignoring `object` repo-wide was the over-reach.** It is the one entry that carries real weight, and the reason is a guard that does not exist: `scripts/ci/check-aya-ebpf-alignment.sh` covers `aya`, `aya-log`, `aya-ebpf` and `aya-log-ebpf`, but contains **zero** mentions of `object` (verified independently: `grep -c object` on that file at `origin/main` returns 0). So `aya` and `aya-log` already have a gate that fails loudly if they drift, and their ignore entries mostly suppress noise over a check that would catch it anyway. `object` has no such gate, which makes its entry the one actually preventing a bad merge.

Dependabot has now opened an `object` bump twice, in #2465 and #2968.

**On narrowing the entry by `update-type`, which I had floated as the safer option: it does not work, and the reviewer's reasoning is right.** GitHub's documentation says a bare `ignore` entry covers version updates *and* security updates, so yes, a security PR for `object` is suppressed. But the same page says `update-types` "only affects version updates, not security updates" — so an update-type-scoped entry would let a security PR through, and that PR would be exactly as unmergeable as #2465 and #2968 were, because any move off `=0.36.7` splits our `object` line against `aya-obj`'s `^0.36` requirement.

**What is lost is the auto-PR, not the alert.** `deny.toml` sets no `targets`, so the linux-only `object` dependency stays in the advisory graph, and `deps-security` runs `cargo-deny check advisories` plus two `cargo audit` passes. An advisory against `object` would still be reported; it just would not arrive as a pull request that cannot land.

That is the honest trade, and it is the right one while the aya stack is pinned as a unit (#2972). When that upgrade happens, all five entries should be revisited together rather than one at a time.

Rul1an:
Reviewer record: https://github.com/Rul1an/assay/pull/2974#issuecomment-5648292325 (READY). All findings pass.

The review inverted the risk I had flagged, which is worth recording on the PR rather than only in the thread: I suspected ignoring `object` repo-wide was the over-reach. It is the entry that carries real weight, because `scripts/ci/check-aya-ebpf-alignment.sh` covers aya, aya-log, aya-ebpf and aya-log-ebpf but contains zero mentions of `object` - verified independently. So the two aya entries suppress noise over a gate that would already fail loudly, and the `object` entry is the one actually preventing a bad merge. Dependabot has opened an `object` bump twice, in #2465 and #2968.

Narrowing by `update-type` was the option I had floated and it does not work: GitHub documents that `update-types` affects version updates only, not security updates, so a security PR would come through and be exactly as unmergeable, since any move off `=0.36.7` splits our object line against aya-obj's `^0.36`. What the bare entry costs is the auto-PR, not the alert - `deny.toml` sets no `targets`, so the linux-only object dependency stays in the advisory graph and `deps-security` still runs cargo-deny advisories plus two cargo audit passes.

PR #2976, opened 2026-09-12
chore(deps): bump base64 to 0.23 and regenerate the fuzz lock

Rul1an:
## Summary

Head: `1b941fdf91732ba102a8828740a82d0d658d11f4`

Candidate, not a landing. Sole author; needs an independent exact-head review record before merge.

This supersedes #2970. Dependabot moved five crate requirements from base64 0.22 to 0.23 and repointed the workspace lock, but it never touched fuzz/Cargo.lock. fuzz/Cargo.toml path-depends on assay-core, assay-evidence and assay-mcp-server, so those crates now require 0.23 while the fuzz lock still carried only 0.22.1.

Refs #2970

## Locked metadata (the original blocker)

Measured on this head, worktree `/Users/roelschuurkes/.claude/worktrees/assay-base64`, `CARGO_INCREMENTAL=0`:

- `cargo metadata --locked --format-version 1 > /dev/null; echo $?` -> 0
- `(cd fuzz && cargo metadata --locked --format-version 1 > /dev/null); echo $?` -> 0

On #2970 head `07e17ef640a55cf51d5a451c400139348a86bb0d` the first command was 0 and the second was 101. Before regenerating the fuzz lock on this branch, the same fuzz command also exited 101 (`cannot update the lock file ... because --locked was passed`).

## What regeneration did

Both locks were produced by cargo, not copied from Dependabot.

- Workspace: after the five manifest edits, `cargo metadata --format-version 1` wrote a Cargo.lock byte-identical to Dependabot's (sha1 `98e74b61dda448163da352ece1947e48e53784ed`). The five lock hunks only repoint our crates from base64 0.22.1 to 0.23.1. No package version set moves; 0.23.1 was already present via pem 4.0.0. base64 0.22.1 stays for eight third-party consumers (hyper-util, jsonwebtoken, object_store, pem 3.0.6, reqwest 0.12 and 0.13, termwiz, wiremock). This is not a de-duplication.
- After that commit, origin/main moved with the jsonschema 0.53 bump (#2969). This head merges that so the first review is not immediately behind on the same lock files. `cargo metadata` after the merge rewrote neither lock.
- Fuzz: a dry-run `cargo update --manifest-path fuzz/Cargo.toml` would have locked 81 unrelated packages. Instead, `cargo metadata --format-version 1 --manifest-path fuzz/Cargo.toml` did the minimal refresh: Locking 1 package / Adding base64 v0.23.1. The fuzz lock now carries two base64 majors, 0.22.1 and 0.23.1.

Two versions in the fuzz lock is the right outcome. Our three path crates now depend on 0.23.1. Six third-party crates still require 0.22.1 (hyper-util, jsonwebtoken, object_store, pem 3.0.6, reqwest 0.12.28, reqwest 0.13.4). Collapsing to one version would mean either leaving our crates on 0.22 or forcing those third-party crates onto a major they do not declare. The workspace already lives with the same two-major shape.

## cargo-deny warning

`cargo-deny check advisories bans licenses sources` on this head: deny_exit=0, final line advisories ok, bans ok, licenses ok, sources ok. The new non-fatal warning versus base is `warning[duplicate]: found 2 duplicate entries for crate 'base64'`.

That warning is acceptable. deny.toml sets `multiple-versions = "warn"`, so two majors are a recorded cost, not a ban. It is the honest price of our crates moving to 0.23 while eight third-party crates stay on 0.22. Silencing it with a skip would hide the coexistence rather than document it.

## Other carried findings, re-checked on this head

- All seven base64 requirement lines in the five manifests now read 0.23, including both dev-dependency declarations (assay-mcp-server, assay-registry). No 0.22 remains in any manifest.
- `CARGO_INCREMENTAL=0 cargo check --locked --all-targets` on the five crates: exit 0, zero warnings. Clippy `-D warnings` on the same targets: exit 0. Re-checked after the jsonschema merge. The 0.x breaking API step does not reach our call sites.
- base64 MSRV moves 1.48.0 -> 1.71.0, under our 1.89 gate.
- `cargo fmt --all -- --check` and `git diff --check`: exit 0.

A Cargo.lock change still needs a delegated gate=all proof run on the final head before it can land. That is not part of this commit.

## Test plan

- [x] Workspace `cargo metadata --locked` exits 0
- [x] Fuzz `cargo metadata --locked` exits 0 (this is what blocked #2970)
- [x] cargo check --locked --all-targets on the five crates
- [x] clippy -D warnings on the five crates
- [x] cargo-deny exits 0; duplicate-base64 warning recorded as acceptable
- [ ] Independent exact-head review record before merge
- [ ] Delegated runner proof, gate=all, bound to the final head

---

Current head `b2bc613cbf375a699094fc42aeae12f650dbb70b`. Review record: https://github.com/Rul1an/assay/pull/2976#issuecomment-5648327162 (carried by the checker's derivation).


Rul1an:
```json
{
  "head_sha": "4725dce4297168753bd29d2be9bd42d4ff477e62",
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "1",
      "summary": "Both locks are consistent. In a detached worktree at 4725dce with CARGO_INCREMENTAL=0: `cargo metadata --locked --format-version 1 > /dev/null` -> `root_metadata_exit=0`; `cargo metadata --locked --format-version 1 --manifest-path fuzz/Cargo.toml > /dev/null` -> `fuzz_metadata_exit=0`. I also reproduced the predecessor's defect independently rather than taking it on report: a detached worktree at refs/pull/2970/head (07e17ef64) running the same fuzz command gives `P2970_FUZZ_METADATA_EXIT=101` with `error: cannot update the lock file .../fuzz/Cargo.lock because --locked was passed to prevent this`. The mechanism checks out — `fuzz/Cargo.toml` path-depends on exactly three of the bumped crates (lines 11-13: assay-core, assay-evidence, assay-mcp-server), which is why a manifest-only bump strands that lock. #2970 is CLOSED.",
      "disposition": "pass"
    },
    {
      "id": "2",
      "summary": "The bump is complete, with no split across workspace members. `grep -rn 'base64 *=' --include=Cargo.toml .` returns exactly seven lines, all `base64 = \"0.23\"`: crates/assay-evidence/Cargo.toml:30, crates/assay-mcp-server/Cargo.toml:43 and :54, crates/assay-core/Cargo.toml:49, crates/assay-cli/Cargo.toml:108, crates/assay-registry/Cargo.toml:41 and :74. The two dev-dependency declarations the brief calls out are among them — mcp-server:54 is under `[dev-dependencies]`, registry:74 likewise — so this is not the #2950/#2949 split. `grep -rn 'base64.*0\\.22' --include=Cargo.toml .` returns nothing; fuzz/Cargo.toml declares no base64 of its own. crates.io reports base64 `max_stable_version: 0.23.1`, and both locks resolve 0.23.1.",
      "disposition": "pass"
    },
    {
      "id": "3",
      "summary": "Two base64 majors in the fuzz lock is correct, and the consumer list is exactly right. Parsing fuzz/Cargo.lock: `base64 versions = ['0.22.1', '0.23.1']`; consumers of 0.22.1 are `hyper-util 0.1.20, jsonwebtoken 11.0.0, object_store 0.14.1, pem 3.0.6, reqwest 0.12.28, reqwest 0.13.4` — the builder's six, name-for-name and version-for-version, including the fuzz tree's reqwest 0.13.4 rather than the root tree's 0.13.2. Consumers of 0.23.1 are `assay-core 6.2.1, assay-evidence 6.2.1, assay-mcp-server 6.2.1` — precisely the three crates fuzz/Cargo.toml path-depends on. None of the six declares base64 0.23, so collapsing to one major is not available without either holding our crates at 0.22 or forcing third-party crates onto an undeclared major. The lock diff confirms this is cargo's own bookkeeping, not hand-editing: entries that read `\"base64\"` on main become `\"base64 0.22.1\"` at head, which is the disambiguation cargo writes once a second version exists. Worth stating precisely, since the PR description is slightly conservative here: the ROOT Cargo.lock already carried both 0.22.1 and 0.23.1 on origin/main (pem 4.0.0 pulled 0.23.1), so only the fuzz lock actually gains a second major in this PR.",
      "disposition": "pass"
    },
    {
      "id": "4",
      "summary": "cargo-deny is clean and the warning delta is exactly one line. `cargo-deny 0.20.2`, invoked as the binary per deny.toml's own instruction: at head `cargo-deny check advisories bans licenses sources` -> `DENY_HEAD_EXIT=0`, closing `advisories ok, bans ok, licenses ok, sources ok`; in a separate origin/main worktree (216b371f6) the same command -> `DENY_MAIN_EXIT=0`. Set-differencing the `warning[...]` lines of the two runs: ONLY IN HEAD -> `warning[duplicate]: found 2 duplicate entries for crate 'base64'`; ONLY IN MAIN -> nothing (39 warnings head, 38 main). So the builder's claim is exact — one new warning, nothing suppressed or dropped, and the pre-existing `warning[yanked]: detected yanked crate (try cargo update -p der)` is unchanged. The setting it relies on is present and untouched: deny.toml:35 `[bans]`, :36 `multiple-versions = \"warn\"`, and `git diff --name-only origin/main...HEAD -- deny.toml` is empty. Mechanism for why the warning is new even though the root lock already had both: on main, base64 0.23.1 was reachable only via pem 4.0.0 on a dev path, which `[graph] exclude-dev = true` drops; our five crates now pull it as normal dependencies, so the duplicate enters the checked graph.",
      "disposition": "pass"
    },
    {
      "id": "5",
      "summary": "Builds and suites pass on the five crates. `cargo build -p assay-cli -p assay-core -p assay-evidence -p assay-mcp-server -p assay-registry` -> `build_exit=0`, zero lines matching `^error`. `cargo test` over the same five -> `test_exit=0`. Counts, aggregated over the log: 242 `test result:` lines, 4393 passed, 0 failed, 28 ignored; `grep '^test result' | grep -v ' 0 failed'` returns nothing, so no suite reports a failure. Both runs under CARGO_INCREMENTAL=0 in the detached 4725dce worktree.",
      "disposition": "pass"
    },
    {
      "id": "6",
      "summary": "Scope is the five manifests plus both locks and nothing else. Merge base with origin/main is 216b371f6c4bc32430a4cbb19eee02c54ff82c0d. `git diff --stat origin/main...HEAD` -> 7 files, 27 insertions / 21 deletions: `Cargo.lock` (10), `crates/assay-cli/Cargo.toml` (2), `crates/assay-core/Cargo.toml` (2), `crates/assay-evidence/Cargo.toml` (2), `crates/assay-mcp-server/Cargo.toml` (4), `crates/assay-registry/Cargo.toml` (4), `fuzz/Cargo.lock` (24). I read the full manifest diff: every hunk is a single `-base64 = \"0.22\"` / `+base64 = \"0.23\"` pair, no feature-list or default-features edits smuggled alongside. The root lock diff is five `base64 0.22.1` -> `base64 0.23.1` lines inside our own package entries and nothing else.",
      "disposition": "pass"
    },
    {
      "id": "7",
      "summary": "Raised, not blocking, and outside the six claims — an open landing precondition the record should not swallow. The required `lane-check` context is red at this head, and reading the job log (run 34715007179, job 103610573661) the reason is this PR's content, not a flake: `Assay-Runner lane check for PR #2976 / Head SHA: 4725dce4297168753bd29d2be9bd42d4ff477e62 / Changed files: 7 / Expected delegated gate: all / - Cargo.lock: runner workflow, CLI, cgroup, or workspace dependency surface requires gates=all`, then `A delegated proof pack is required`, and `lane-check/proof` stays missing (`delegated proof required: gates=all`). Touching Cargo.lock at all is what trips it, so any correct fix for #2970 trips it too — #2974 passes the same gate only because it edits no lock. Nothing in the change is wrong; a delegated `Assay-Runner Lane Check` proof pack for this exact head still has to land before merge. Note also that the job logged `warning: token cannot write commit statuses (403); the required 'lane-check/proof' context stays missing until a privileged workflow_dispatch ... refreshes it`, so the proof needs a dispatch from a trusted ref rather than a re-run of the PR job.",
      "disposition": "pass"
    }
  ],
  "no_findings": false,
  "builder": { "agent": "cursor", "instance": "grok-build-2970" },
  "reviewer": { "agent": "claude", "instance": "review-2976-4725dce", "github_login": "Rul1an" }
}
```

Rul1an:
Blocked on infrastructure, not on this PR's merits. Filed as #2985.

Three consecutive delegated proof runs on this branch failed with `##[error]The operation was canceled`, and each time the self-hosted runner stopped itself (`Runner listener exit with 0 return code, stop the service, no retry needed`), leaving it inactive until restarted by hand.

**This branch is cleared.** The obvious hypothesis was that `base64` 0.23 breaks the delegated build, which uses `--no-default-features --features runner` - a different feature set from the one the review exercised. Tested directly on this head: `cargo build -p assay-cli --no-default-features --features runner` exits **0** with zero errors in 47.9 s. The failure is not in the change.

Also ruled out with measurements rather than assumption: host starvation (the third failure ran at load 17 with zero local builds), clock drift (guest 22:37:39 UTC against host 22:37:51), a frozen guest (ping and `multipass exec` both answer), OOM or disk (7.7 GiB RAM with 4.2 free, no `oom-kill`, 13 GiB free), and a concurrency collision (runs do not overlap; `cancel-in-progress` is false).

The review record stands and carries: `review-record-check=pass head=b2bc613cb` by the checker's own derivation. This PR needs only the proof pack, and the proof pack needs a runner that stays up.

I have restarted the runner. I am not re-dispatching tonight - three twelve-minute cycles have produced three identical failures, and a fourth would only confirm it. This waits on #2985.

Rul1an:
Reviewer record: https://github.com/Rul1an/assay/pull/2976#issuecomment-5648327162 (READY). All findings pass, and the delegated proof finally landed on run 34723605077 after the runner was recovered - see #2985 for why three earlier attempts failed and why that was not this PR.

The defect this supersedes is worth keeping on the record. Dependabot's #2970 bumped `base64` in five crate manifests and touched no fuzz file at all, so a file-list review passes it; but `fuzz/Cargo.toml` path-depends on three of those crates, and `cargo metadata --locked --manifest-path fuzz/Cargo.toml` exited 101. The reviewer reproduced that from #2970's own head before checking this one.

Two decisions here were argued from measurement rather than convention. The fuzz lock keeps **two** base64 majors, because six third-party crates still require 0.22.1 - hyper-util, jsonwebtoken, object_store, pem 3.0.6 and both reqwest lines - so collapsing to one would either hold our crates back or force those onto a major they do not declare. And the resulting `cargo-deny` `duplicate entries for crate 'base64'` warning is acceptable because `deny.toml:36` sets `multiple-versions = "warn"`: a recorded cost, not a ban.

One correction the review made to the PR's own framing: the root `Cargo.lock` already carried both majors on main, because pem 4.0.0 pulled 0.23.1 behind `exclude-dev = true`. Only the **fuzz** lock actually gains a second major here.

Issue #2975, opened 2026-09-12
ci: the fuzz-lock staleness assertion runs only where it cannot block a merge


Rul1an:
## Measured

The assertion that `fuzz/Cargo.lock` is not stale exists, works, and **cannot block a merge**.

- Branch protection on `main` requires exactly four contexts: `CI`, `host-capability-check`, `lane-check/proof`, `review-record-check`.
- `git grep -n "manifest-path fuzz\|cd fuzz"` over `.github/workflows/ci.yml` returns nothing, so the required `CI` gate never runs `cargo metadata --locked` against `fuzz/Cargo.toml`.
- The only lane that runs it is `fuzz-smoke.yml`, which is not a required context.

It caught a real defect today, which is how this surfaced. PR #2970 bumps `base64` `0.22 -> 0.23` in five crate manifests and touches **no fuzz file at all**. `fuzz/Cargo.toml` depends on `assay-core`, `assay-evidence` and `assay-mcp-server` by path, all three now require `0.23`, and `fuzz/Cargo.lock` still carries only base64 0.22.1, so `cargo metadata --locked --manifest-path fuzz/Cargo.toml` exits 101 with `cannot update the lock file ... because --locked was passed`. The defect is invisible in the PR's own diff and a file-list review passes it.

It fired there only because three of the five manifests are in `fuzz-smoke.yml`'s `paths:` list. That list is itself guarded — `scripts/ci/test-fuzz-lane-contract.sh` derives the set from `fuzz/Cargo.lock` and fails on drift — so the path coverage is sound.

**The narrower hole in the path list:** `Cargo.toml` is in it, `Cargo.lock` is not. A change that moves only the workspace lock, in a way that stales the fuzz lock, triggers no lane at all.

## Why it matters

This is the third time this drift has cost real work. The 6.2.0 and 6.2.1 release bumps left the fuzz lock stale and killed every fuzz leg at the locked-metadata gate until #2930 re-synced it; #2950 and #2949 could not land because each carried one half of a two-lock change; and now #2970. Each time the guard eventually said so, and each time it said so somewhere that does not gate the merge.

A guard that finds the defect but cannot stop it means the defect lands and is attributed to whoever next touches the lane, rather than to the change that caused it.

## Contract

If a check is the thing that prevents a known, recurring breakage, it belongs on the path that blocks the merge.

## Options

[...]

PR #2930, opened 2026-09-11
fuzz(mcp): add mcp_jsonrpc and tool_call_decision targets

Rul1an:
**Current head `b5792b6706eb1a9f53a94ac929a56a92291831f0`** (review follow-up). Codex's review of the first head held on behaviour, oracle independence and the lane contract, and blocked because CI never ran the `mcp_jsonrpc` leg: `end-of-file-fixer` had emptied the one-newline `empty-line` seed, and the lane's `[ -s ]` guard refused it. Follow-up: `fuzz/corpus/` is excluded from both whitespace hooks (this closes the hook issue), and `empty-line` is one `0x0a` byte again. Committed by the coordinating session because a hook cannot run in the builder's sandbox; all hooks ran, none was bypassed. Lock-drift note for the next long-lived PR that adds a workspace dependency to fuzz/: this PR introduced the assay-mcp-server and assay-metrics entries into fuzz/Cargo.lock at 6.1.3, and the 6.2.0/6.2.1 release bumps only rewrite entries already present on main, so the lock went stale and every fuzz-smoke leg died at the locked-metadata gate until this head re-synced those two entries to 6.2.1.

Closes #2931

Candidate, not a landing. Head SHA 080b78d7aea22dc349a28162e78942aa3a21ff6e. Sole author; needs an independent exact-head review record before merge.

## ADR Boundary Slice

- Canonical ledger: none — AGENTS.md states no programme is active, so this is a standalone slice with no ledger to record to.
- Slice: fuzz the stdio JSON-RPC loop of assay-mcp-server and the tools/call decision input (issue #2914, P2 M).
- Builder: Muse Code (sole author; I know this loop from #2903, single typed parse).
- Reviewers: TBD — independent exact-head review required before merge. I will never post a review record on my own PR.
- Final head SHA: 080b78d7aea22dc349a28162e78942aa3a21ff6e (branch muse/2914-fuzz-mcp-jsonrpc, worktree /Users/roelschuurkes/wt-muse-2914).
- ADR invariants affected: none changed. ADR-043 stays Accepted; the stop list holds (no trust score, no whole-action verdict, no identity/delegation layer, no provider-outcome verification, no detector catalogue, no compliance claim). This only widens the existing fuzz-smoke lane to new inputs.

## Behavior

[...]

PR #2978, opened 2026-09-12
ci: assert the fuzz lockfile is current from the required gate

Rul1an:
**Candidate, not a landing.** Head SHA `60bb9e86c1c99334d825dfaf35bd45209ceafbee`. Built by Muse; needs an independent exact-head review record before merge.

## Problem

The assertion that `fuzz/Cargo.lock` is current existed, worked, and could not block a merge. It ran only in `fuzz-smoke.yml`, which is not a required context, and `git grep -n "manifest-path fuzz\|cd fuzz"` over `ci.yml` returned nothing.

It has now cost real work three times: the 6.2.0 and 6.2.1 bumps left the lock stale until #2930 re-synced it, #2950/#2949 each carried one half of a two-lock change, and #2970 bumped `base64` in five crate manifests while touching no fuzz file at all — `fuzz/Cargo.toml` path-depends on three of those crates, so the lock went stale and the defect was invisible in the PR's own diff.

## Change

The assertion moves into `deps-security`, which already owns the fuzz lockfile through its isolated-db audit, already has a toolchain, and is in the required `CI` gate's `needs:` list. It therefore runs on every pull request with no path list to maintain. `scripts/ci/test-fuzz-lock-required-gate.sh` is the contract test.

Not shared with the lane's inline form: that spelling is pinned by `scripts/ci/test-fuzz-lane-contract.sh`, which executes the lane's own run block and selects the date-pinned fuzz toolchain, while this one resolves read-only under stable.

## Evidence

| step | result |
|---|---|
| clean tree baseline | `cargo metadata --locked --manifest-path fuzz/Cargo.toml` rc=0 |
| contract test, pre-fix | `FAIL: no job in the required CI gate asserts the fuzz lock is current (#2975)`, rc=1 |
| #2970 reconstruction (base64 `0.22`→`0.23` in the five manifests, 7 lines, no fuzz file) | `error: cannot update the lock file .../fuzz/Cargo.lock because --locked was passed`, rc=101 |
| must-fail control (regenerate the fuzz lock) | rc=0; probe then fully reverted |
| contract test, post-fix | passes through `deps-security`; all six mutants red (comment-only, dropped-from-needs, dropped-triple, stripped-toolchain, stderr-redirect, cause-claimed) |

Siblings green: b1, gate-expectations, fuzz-lane, plugin-versions, audit-isolated-db, gate-coverage (`ci-gate-coverage=passed`), shellcheck, YAML parse with `needs:` unchanged, `git diff --check`, `cargo fmt --check`.

## The workspace-lock-only question, answered rather than deferred

The issue asked whether a change moving only the workspace `Cargo.lock` is covered. Measured: with the workspace lock physically removed, the fuzz assertion still returns rc=0 — the workspace lock is not an input to the fuzz resolve, because `Cargo.toml` sets `exclude = ["fuzz"]` and the fuzz tree carries its own lockfile. So a workspace-lock-only change cannot stale it. And since the check now runs on every PR regardless of paths, the `paths:` hole is moot for staleness. Option 3 from the issue is deliberately not taken.

## A third coupled list, found by the hook

The first commit updated two of the three places the hardening step's invocation list lives and was refused on push by `editor-release-hook-precommit-consumer`: `HARDENING_RUN_SCRIPT` in `conformance/tests/test_completion_scope.py` is compared line-for-line against the live step, and it was stale.

Worth recording rather than just fixing, because the builder traced why the earlier guard missed it: `test-ci-hardening-b1.sh` checks the live step against **its own** copy of the list, which had been updated, so it stayed green and could not see the stale third copy. `test_completion_scope.py` would have caught it but no commit-stage hook runs it, and the consumer hook is `stages: [pre-push]` by design. So the pre-push refusal was the earliest automatic guard in the path - the system working, not failing.

The structural point outlasts this PR: the b1 `required` tuple and `HARDENING_RUN_SCRIPT` are the same rule written twice, and nothing pins them to each other - only each independently matching the live step. That is the drift shape the repo's own "one rule, one function" contract names. Filed separately rather than widened into this slice.

## Provenance

Built by Muse. Committed from the owner checkout because the `codex-host-proof-node-contract` pre-commit hook cannot run in Muse's sandbox, where `os.userInfo()` raises `uv_os_get_passwd ENOENT` for a user with no passwd entry — reproduced with bare `node -e "require('os').userInfo()"`, zero repo input, so environmental and pre-existing. **No hook was skipped**: all of them ran on the machine where they can run.

Refs #2975

---

Current head `151f1cd17986abd3adde8bce21ae6874bf609bed`. Review record: https://github.com/Rul1an/assay/pull/2978#issuecomment-5648749370 (carried by the checker's derivation).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -225,8 +225,10 @@ jobs:
       # the selection there too rather than being shadowed by the script's own default.
       #
       # Scope, deliberately: no step in this job builds or tests the workspace -- checkout, two tool
-      # installs, a `sed`-only alignment check, `cargo-deny check` and two `cargo audit` runs. The
-      # only crates this job compiles are the dependency tools themselves, installed from source.
+      # installs, a `sed`-only alignment check, `cargo-deny check`, two `cargo audit` runs, and a
+      # read-only `cargo metadata --locked` staleness assertion over `fuzz/Cargo.toml` (#2975). The
+      # only crates this job compiles are the dependency tools themselves, installed from source;
+      # the metadata step resolves the fuzz graph without compiling anything.
       # `rust-toolchain.toml` stays the source for every job that compiles workspace code.
       RUSTUP_TOOLCHAIN: stable
     steps:
@@ -289,6 +291,23 @@ jobs:
           # to be quietly excused, which is the point.
           # Same isolated-db runner as the root audit / pre-push hook (CI-4F / #2188).
           ( cd "${RUNNER_TEMP}" && bash "${GITHUB_WORKSPACE}/scripts/ci/cargo-audit-with-isolated-db.sh" --file "${GITHUB_WORKSPACE}/fuzz/Cargo.lock" )
+
+      - name: Fuzz lockfile is current
+        # #2975: this assertion ran only in fuzz-smoke.yml, which is not a required context, so a
+        # PR that staled `fuzz/Cargo.lock` (e.g. #2970 bumping base64 in five manifests, none of
+        # them a fuzz file) stayed green on every required gate. This job owns the fuzz lockfile
+        # already (the audit above), has a toolchain (job-level RUSTUP_TOOLCHAIN), and is in the
+        # required gate's `needs:`, so the check runs on every PR with no path list to maintain.
+        # Not shared with the lane's inline form: that spelling is pinned by
+        # `scripts/ci/test-fuzz-lane-contract.sh` (which executes the lane's own run block) and
+        # selects the date-pinned fuzz toolchain, while this one resolves read-only under stable.
+        # The wrapper adds an exit code and nothing else; `--locked` also fails on registry or
+        # network faults, so the message must not name a cause -- Cargo's stderr stays the
+        # diagnosis. Pinned by `scripts/ci/test-fuzz-lock-required-gate.sh`.
+        run: |
+          set -euo pipefail
+          cargo metadata --locked --format-version 1 --manifest-path fuzz/Cargo.toml >/dev/null \
+            || { echo "::error::locked fuzz metadata validation failed; inspect the Cargo error above"; exit 1; }
 
   clippy:
     name: Clippy (deny warnings)
@@ -1176,6 +1195,7 @@ jobs:
           bash scripts/ci/test-ci-job-timeouts-contract.sh --self-test
           bash scripts/ci/test-classify-lightweight-changes.sh
           bash scripts/ci/test-deps-security-toolchain-contract.sh
+          bash scripts/ci/test-fuzz-lock-required-gate.sh
           bash scripts/ci/test-mcp-preflight-windows-contract.sh
           bash scripts/ci/test-reconcile-docs-auto-pr.sh
           bash scripts/ci/test-setup-rust-composite-contract.sh
--- a/conformance/tests/test_completion_scope.py
+++ b/conformance/tests/test_completion_scope.py
@@ -73,6 +73,7 @@ HARDENING_RUN_SCRIPT = (
     "bash scripts/ci/test-ci-job-timeouts-contract.sh --self-test",
     "bash scripts/ci/test-classify-lightweight-changes.sh",
     "bash scripts/ci/test-deps-security-toolchain-contract.sh",
+    "bash scripts/ci/test-fuzz-lock-required-gate.sh",
     "bash scripts/ci/test-mcp-preflight-windows-contract.sh",
     "bash scripts/ci/test-reconcile-docs-auto-pr.sh",
     "bash scripts/ci/test-setup-rust-composite-contract.sh",
--- a/scripts/ci/test-ci-hardening-b1.sh
+++ b/scripts/ci/test-ci-hardening-b1.sh
@@ -611,6 +611,7 @@ required = (
     "bash scripts/ci/test-ci-job-timeouts-contract.sh --self-test",
     "bash scripts/ci/test-classify-lightweight-changes.sh",
     "bash scripts/ci/test-deps-security-toolchain-contract.sh",
+    "bash scripts/ci/test-fuzz-lock-required-gate.sh",
     "bash scripts/ci/test-mcp-preflight-windows-contract.sh",
     "bash scripts/ci/test-reconcile-docs-auto-pr.sh",
     "bash scripts/ci/test-setup-rust-composite-contract.sh",
--- a/scripts/ci/test-conformance-inventory-callsite.py
+++ b/scripts/ci/test-conformance-inventory-callsite.py
@@ -106,6 +106,7 @@ HARDENING_STEP_COMMANDS = (
     "bash scripts/ci/test-ci-job-timeouts-contract.sh --self-test",
     "bash scripts/ci/test-classify-lightweight-changes.sh",
     "bash scripts/ci/test-deps-security-toolchain-contract.sh",
+    "bash scripts/ci/test-fuzz-lock-required-gate.sh",
     "bash scripts/ci/test-mcp-preflight-windows-contract.sh",
     "bash scripts/ci/test-reconcile-docs-auto-pr.sh",
     "bash scripts/ci/test-setup-rust-composite-contract.sh",
--- a/scripts/ci/test-fuzz-lock-required-gate.sh
+++ b/scripts/ci/test-fuzz-lock-required-gate.sh
@@ -1,0 +1,290 @@
+#!/usr/bin/env bash
+# Contract: a stale fuzz/Cargo.lock must fail a required CI context on the PR that causes it.
+#
+# The staleness assertion (`cargo metadata --locked` against `fuzz/Cargo.toml`) exists in
+# `fuzz-smoke.yml`, which is not a required context, so it can find the defect without being
+# able to stop it (#2975). The fix puts the same assertion in a job the required `CI` gate
+# already waits on, on every PR, independent of any path list.
+#
+# This pins the wiring, not just the text: the invocation must be an active step in a job
+# that is in the `ci` rollup's `needs:` list, judged by its `name|result|expectation` triple,
+# and that job must have a Rust toolchain available. Mutants below prove each half red.
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+WORKFLOW="${ROOT}/.github/workflows/ci.yml"
+
+fail() {
+  echo "FAIL: $*" >&2
+  exit 1
+}
+
+ok() { echo "ok   $*"; }
+
+[[ -f "$WORKFLOW" ]] || fail "missing ${WORKFLOW#"${ROOT}"/}"
+
+SANDBOX_ROOT="$(mktemp -d)"
+trap 'rm -rf "$SANDBOX_ROOT"' EXIT
+
+# Print the qualifying job id when the wiring holds; exit 1 otherwise.
+# Every assertion takes the workflow path so the mutants below exercise the same logic.
+check_workflow() {
+  python3 - "$1" <<'PY'
+import re
+import sys
+
+lines = open(sys.argv[1]).read().splitlines()
+
+# The `ci` rollup's needs: membership list (single-line form, as checked in).
+needs = []
+for i, line in enumerate(lines):
+    if line == "  ci:":
+        for follow in lines[i + 1 : i + 6]:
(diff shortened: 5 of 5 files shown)


Rul1an:
```json
{
  "head_sha": "7f7a7bb61ee8e5b488576a9d30f623c96bde3ac5",
  "builder": {
    "agent": "muse",
    "instance": "build-2975"
  },
  "reviewer": {
    "agent": "cursor",
    "instance": "codex-review-2978",
    "github_login": "Rul1an"
  },
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "F-001",
      "summary": "Merge-base scope is four files, not three; the additional file is scripts/ci/test-fuzz-lock-required-gate.sh.",
      "disposition": "Non-blocking: extra file is the new guard's contract test and is within the claimed hardening scope."
    }
  ],
  "no_findings": false,
  "independence": {
    "did_not_build": true,
  }
}
```

Rul1an:
Head moved to `84c75cdf7`. Codex's READY record at `7f7a7bb61` no longer binds, and the reason is worth recording because it sharpens this PR's own point.

**The list lives in four places, not three.** Required CI caught the fourth after the pre-push hook caught the third:

```
FAIL: test_pristine_workflow_is_green
AssertionError: Tuples differ: first differing element 13
```

`HARDENING_STEP_COMMANDS` in `scripts/ci/test-conformance-inventory-callsite.py:108` is the fourth copy. The complete set is now:

1. the live step in `.github/workflows/ci.yml`
2. the `required` tuple in `scripts/ci/test-ci-hardening-b1.sh`
3. `HARDENING_RUN_SCRIPT` in `conformance/tests/test_completion_scope.py`
4. `HARDENING_STEP_COMMANDS` in `scripts/ci/test-conformance-inventory-callsite.py`

Each of the four independently compares itself against the live step, and none is derived from another — so adding one invocation requires four edits, and missing any one is caught only at whichever stage that copy happens to run. Here the third was caught at pre-push and the fourth at required CI, two rounds apart.

All four now agree, each verified from its own unpiped run: `test-conformance-inventory-callsite.py`, `test-ci-hardening-b1.sh`, `test-editor-release-hook-precommit-consumer.sh` and `test_completion_scope.py` all exit 0.

**Provenance, stated because it changes who can review what.** The builder was mid-task on another issue, so I made this one-line completion rather than waiting. That makes me a co-builder of this PR. Codex reviewed the previous head and is still an independent reviewer of the builder's work, but the new head needs a fresh exact-head record, and I cannot be the one to write it.

The builder's own observation on the earlier round — that these lists are one rule written more than once, pinned to each other by nothing — is stronger now that the count is four. That belongs in its own issue rather than widened into this slice.

Rul1an:
```json
{
  "head_sha": "84c75cdf7b4d44def606532d75e67fdb60eeec1e",
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "1",
      "summary": "It genuinely gates. Reconstructed #2970 on this head in a detached worktree of 84c75cd: bumped `base64 = \"0.22\"` -> `\"0.23\"` in crates/assay-cli, assay-core, assay-evidence, assay-mcp-server (x2 lines), assay-registry (x2 lines), leaving fuzz/Cargo.lock untouched (`git status --porcelain` showed exactly the five manifests, no lockfile). Ran the step body verbatim: `set -euo pipefail; export RUSTUP_TOOLCHAIN=stable; cargo metadata --locked --format-version 1 --manifest-path fuzz/Cargo.toml >/dev/null || { echo \"::error::locked fuzz metadata validation failed; inspect the Cargo error above\"; exit 1; }` -> `error: the lock file /Users/.../assay-rev-2978/fuzz/Cargo.lock needs to be updated but --locked was passed to prevent this` then `::error::locked fuzz metadata validation failed; inspect the Cargo error above`, STEP_RC=1. Regenerated with the same command minus `--locked` (REGEN_RC=0, fuzz/Cargo.lock +16/-10, base64 0.23.1 now alongside 0.22.1), reran the step: STEP_RC=0. Reverted with `git checkout -- crates fuzz/Cargo.lock`, `git status --porcelain` empty. #2970's own file list (gh pr view 2970) is Cargo.lock + those same five manifests and no fuzz file, so the reconstruction is faithful.",
      "disposition": "pass"
    },
    {
      "id": "2",
      "summary": "On the required path, each link printed from the files at this head. (a) .github/workflows/ci.yml:207 `  deps-security:` / :208 `name: Dependency Security (cargo-deny + audit)`, and the new step `Fuzz lockfile is current` is its last step (ci.yml:295-310, immediately before `  clippy:`), with no step-level `if:`. (b) ci.yml:1159 `  ci:` / :1160 `name: CI` / :1164 `needs: [scope, deps-security, clippy, ...]` -- deps-security is member 2. (c) ci.yml:1229 `- name: Evaluate required job results`, :1233 `DEPS_SECURITY_RESULT: ${{ needs.deps-security.result }}`, :1292 `\"deps-security|${DEPS_SECURITY_RESULT}|${code_gated_expectation}\"`; the loop treats any result other than success/skipped as `::error::Required CI dependency ${name} ended with ${result}` and increments failures. (d) `gh api repos/Rul1an/assay/branches/main/protection --jq .required_status_checks.contexts` -> `[\"CI\",\"host-capability-check\",\"lane-check/proof\",\"review-record-check\"]`, and the rollup job's `name:` is `CI`. deps-security's own gate is `if: needs.scope.outputs.lightweight_only != 'true'`, matching `code_gated_expectation`, so it is required on every code PR and legitimately optional on docs-only ones. For contrast, fuzz-smoke is not in that context list -- the premise of the PR.",
      "disposition": "pass"
    },
    {
      "id": "3",
      "summary": "All four coupled lists agree, and there is no fifth. `grep -rln test-deps-security-toolchain-contract.sh` returns five files; .pre-commit-config.yaml is not a copy of the list (it carries per-script hooks with their own `files:` filters at :602 and :747), the other four are the lists. A python element-wise comparison of ci.yml's live `run: |` block against HARDENING_RUN_SCRIPT (conformance/tests/test_completion_scope.py), HARDENING_STEP_COMMANDS (scripts/ci/test-conformance-inventory-callsite.py) and `required` (scripts/ci/test-ci-hardening-b1.sh) printed: `live len 22 | completion_scope 22 | callsite 22 | hardening-b1 22`, `live==a: True`, `live==b: True`, `live==c: True`, and for each of the four `index 13, prev='bash scripts/ci/test-deps-security-toolchain-contract.sh', next='bash scripts/ci/test-mcp-preflight-windows-contract.sh'`. All four consumers run unpiped and green: `bash scripts/ci/test-editor-release-hook-precommit-consumer.sh` RC=0 (pre-commit 4.6.0 present; last line `editor/release-surface pre-commit consumer: PASS`); `bash scripts/ci/test-ci-hardening-b1.sh` RC=0 (`ci-hardening-b1 contract: PASS`); `python3 -W error::ResourceWarning conformance/tests/test_completion_scope.py` RC=0 (`Ran 46 tests`, `OK`); `python3 scripts/ci/test-conformance-inventory-callsite.py` RC=0. Also ran the two neighbouring contracts the deps-security edit could disturb: test-deps-security-toolchain-contract.sh RC=0 and test-ci-gate-expectations.sh RC=0, plus test-fuzz-lane-contract.sh RC=0 (the lane's own spelling is untouched).",
      "disposition": "pass"
    },
    {
      "id": "4",
      "summary": "The new contract test bites for real, not only against its own sandbox copies. `bash scripts/ci/test-fuzz-lock-required-gate.sh` on the clean head: RC=0, `ok fuzz-lock assertion wired through required job `deps-security`` plus its six internal mutants red. I then applied four mutations to the real .github/workflows/ci.yml, restoring the file from a backup after each: (A) comment out the invocation -> `no job in the required `CI` gate asserts the fuzz lock is current` / `FAIL: ...(#2975)`, RC=1; (B) drop `deps-security` from the `ci` rollup's `needs:` (line 1164, verified with sed) -> same failure, RC=1; (C) rewrite the error message to `::error::fuzz/Cargo.lock is stale; regenerate it` -> `` `deps-security` step `Fuzz lockfile is current` diagnoses a cause (`stale`) it did not establish ``, RC=1; (D) append `2>/dev/null` to the cargo line -> `` `deps-security` step `Fuzz lockfile is current` redirects stderr away from the log ``, RC=1. `git status --porcelain` empty afterwards. Note for readers: the specific diagnosis comes from the embedded python on stderr; bash's own trailing `FAIL:` line is the generic one in all four cases, so the reason is in the line above it, not the last line.",
      "disposition": "pass"
    },
    {
      "id": "5",
      "summary": "Workspace-lock-only claim holds. Root Cargo.toml:31 has `exclude = [\"fuzz\"]`. With the workspace lock moved entirely out of the tree (`mv Cargo.lock /tmp/ws-Cargo.lock.bak`; `ls Cargo.lock` -> No such file or directory), the step body ran STEP_RC_NO_WS_LOCK=0. Restored, `git status --porcelain` empty. So no change confined to the workspace lock can make this assertion red -- which is why the fix needs no `paths:` list, and conversely why claim 1's manifest-only change is the shape that does reach it.",
      "disposition": "pass"
    },
    {
      "id": "6",
      "summary": "No previously-running job is now skipped and `needs:` is otherwise unchanged; every hunk accounted for. Diff against the merge-base with main (216b371f6) is 5 files / +315 / -2: ci.yml has exactly three hunks -- (i) the deps-security job-env scope comment gains two sentences naming the new read-only metadata step, (ii) the new `Fuzz lockfile is current` step appended to deps-security's steps, (iii) one line added to the hardening `run: |` block at :1198 -- plus one line each in test_completion_scope.py, test-ci-hardening-b1.sh and test-conformance-inventory-callsite.py, and the new scripts/ci/test-fuzz-lock-required-gate.sh (mode 100755, matching its siblings). `git diff <base>..HEAD -- .github/workflows/ci.yml | grep -E '^[+-]\\s*(needs|if|paths|on):'` returns nothing: no job added, removed, re-gated or re-scoped.",
      "disposition": "pass"
    },
    {
      "id": "7",
      "summary": "Seam checked -- the stable placement is right, not merely lucky. There is no fuzz/rust-toolchain.toml; the lane's date pin lives in fuzz-smoke.yml's `FUZZ_TOOLCHAIN: nightly-2026-07-28` and exists for cargo-fuzz's sanitizer build, not for resolution. I ran the step under an even older stable than CI's (`RUSTUP_TOOLCHAIN=stable cargo --version` -> `cargo 1.92.0`, versus the workspace pin 1.96.0) and it returned 0, so `cargo metadata --locked` over fuzz/Cargo.toml resolves without the dated nightly. It also compiles nothing: neither `target/` nor `fuzz/target/` existed after the run (`ls` -> No such file or directory for both), so the job-env comment's \"no step in this job builds or tests the workspace\" survives the addition. One behavioural note, already anticipated by the step's own comment: the happy path did not touch the network, while the stale path printed `Updating crates.io index` -- so the step adds no new network dependency when green, and when red `--locked` can also be red for registry reasons, which is exactly why the message refuses to name a cause.",
      "disposition": "pass"
    }
  ],
  "no_findings": false,
  "builder": { "agent": "muse", "instance": "build-2975" },
  "reviewer": { "agent": "claude", "instance": "review-2978-84c75cd", "github_login": "Rul1an" }
}
```

Rul1an:
Reviewer record: https://github.com/Rul1an/assay/pull/2978#issuecomment-5648749370 (READY). All findings pass.

The seam I asked about is closed with a better answer than I expected: the assertion runs green under `cargo 1.92.0`, which is **older than CI's stable and older than the 1.96.0 workspace pin**, and creates no `target/` or `fuzz/target/`. So it genuinely resolves without compiling and does not need the lane's date-pinned nightly - the `deps-security` placement is right on its own terms, not just green today.

All four coupled lists are element-wise identical at 22 entries with the new script at index 13, and each of the four consumers was run unpiped with its own exit code reported. The reconstruction of #2970 fails the step with the real `--locked` error and passes once the lock is regenerated.

Two things recorded for whoever reads a red log from this gate later. The script's trailing bash `FAIL:` line is the same generic message for every mutant, so the real diagnosis is the python line **above** it. And the happy path never touches the network - only the stale path printed `Updating crates.io index`, which is a useful tell that the assertion is doing work rather than short-circuiting.

Why this matters more than when it was filed as #2975: Dependabot re-offered `base64` after the #2963 fold as #2979 and produced the same defective shape as #2970, because the fold only covers dependencies declared in the root `[workspace.dependencies]` and `base64` lives in five member manifests. Configuration cannot prevent that class. This gate is what catches it, on the PR that causes it.

Issue #2862, opened 2026-09-09
experiment(adequacy): measure privileged-MCP findings as a diagnostic channel


Rul1an:
Parent: #1977

Related: #1977, #2826; prior tool work: corpus-adequacy issue #128 and PR #129

## Problem

The `privileged-mcp-action/v0` verifier emits `findings` separately from `bundle_integrity`, `verdict`, and `claims`. The current Corpus Adequacy manifest observes only those three normative fields. It therefore cannot say whether the pinned corpus detects a change that removes or alters a diagnostic finding while leaving the normative result unchanged.

That is a measurement gap, not an established verifier defect. Findings remain reviewer-visible and unscored under the profile's conformance protocol.

Measured source baseline: `c5267bdae8b5173a099d3bf5866e3a07e01ce4cf`.

## Proposed experiment

Declare `findings` as a diagnostic channel for the existing `privileged-mcp-action-v0` adequacy row and rerun the existing bounded measurement. Preserve the distinction between:

- a normative outcome change (`killed`);
- a diagnostic-only change (`silent`);
- no observed change (`survived`);
- an invalid or unproved run;
- an equivalent or explicitly out-of-scope mutation.

The manifest already records one finding-only rule as out of scope under the current comparison surface. Admitting a diagnostic channel does not automatically promote that mutant or add a vector; scope changes must be explicit.

## Definition of done

- [ ] Pin the Assay source, Corpus Adequacy tool source, vectors, manifest and corpus digest independently before execution.
- [ ] Run the built verifier twice for every existing vector before changing the declaration; require stable `findings` and retain bounded stdout/stderr hashes. Stop on timestamps, paths, process IDs, unstable ordering or other run-to-run drift.
- [ ] Add one validation rule shared by loading and rendering: a report with `silent > 0` is invalid when the manifest declares no diagnostic channel.
- [ ] Add a same-chain behavioral test in which a finding-only mutation survives with `findings` undeclared and becomes silent with `findings` declared. The positive control must still change a normative outcome.
- [ ] Add only `diagnostic_from: ["findings"]` and a public rationale to the adequacy manifest. Keep `outcome_from`, vectors, mutants, corpus digest and the profile's normative comparison surface unchanged.
- [ ] Rerun the existing bounded `measure_all.py --only privileged-mcp-action-v0` path. Require a killed control, `unproved == 0`, exact tool/source/report identities and verified source restoration.
[...]

PR #2981, opened 2026-09-12
experiment(adequacy): declare findings as a diagnostic channel

Rul1an:
**Candidate, not a landing.** Head SHA `81edf20345024a733d0a80efd37333f61a9e4fda`. Built by Muse; needs an independent exact-head review record before merge.

Declares `findings` as a **diagnostic** channel for the privileged-mcp-action-v0 adequacy lane, without letting it become a normative outcome.

## What changed

`diagnostic_from: ["findings"]` plus a public rationale in the manifest, and one validation rule shared by loading and rendering: a report with `silent > 0` is invalid when the manifest declares no diagnostic channel. `outcome_from`, the vectors, the mutants, the corpus digest and the profile's normative comparison surface are unchanged.

## Evidence

**Determinism first.** Every existing vector was run twice through the built verifier *before* the declaration changed, with stable `findings` and bounded stdout/stderr hashes. No run-to-run drift on timestamps, paths, pids or ordering.

**The must-bite, with its control asserted first.** A finding-only mutation survives with `findings` undeclared and becomes silent once it is declared, while the positive control still changes a normative outcome. The control is asserted before the mutant deliberately: a control that goes quiet alongside the mutant proves nothing.

**Bounded rerun**, `measure_all.py --only privileged-mcp-action-v0`, killed control present, `unproved == 0`. `silent` moves 0 → 2 with `diagnostic_channel_declared` true, and `measured_at.commit` pins the tree the run actually saw.

**The numbers disagreed with the forecast, and the numbers won.** Of the transitions expected to move: 40 and 50 are masked, 39 is unreached by the generator, and 41 is dead code. That split came from reading each mutant's arm against `gen_vectors.py`, not from tuning anything. No source, vector, declaration or corpus digest was touched to obtain a count.

`INDEX.md` and `ERRATA.md` are regenerated through `project_published_numbers.py` rather than edited; `check_published_numbers.py` is green.

## One thing a reviewer should not misread

`conformance/tests/test_privileged_mcp_diagnostic_channel.py` **skips** when no sibling Corpus Adequacy checkout is present. That is correct behaviour, but it means the must-bite proof can only run where the tool lives — so a green CI log reporting `skipped=2` is not evidence the guard holds. The proof is the local run recorded above.

Also worth stating: the ambient sibling tool checkout is two API generations past the manifest pin. Pinning the tool independently (DoD box 1) is what caught that; using the ambient one would have produced a confusing `TypeError` mid-measurement.

Refs #2862

---

Current head `ed41de75b264985d6b3a9477ea2da3e2e55ca613`. Review record: https://github.com/Rul1an/assay/pull/2981#issuecomment-5648796601 (carried by the checker's derivation).


Rul1an:
```json
{
  "head_sha": "81edf20345024a733d0a80efd37333f61a9e4fda",
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "1",
      "summary": "The normative surface is unchanged, checked field by field rather than by eye. Manifest: a structural comparison against the merge-base (216b371f6) gives `added keys: ['_diagnostic_from_why', 'diagnostic_from']`, `removed keys: []`, `changed keys: []`; `outcome_from` is still `['bundle_integrity', 'verdict', 'claims']`; the mutant table is byte-identical (35 entries, `old['mutants'] == new['mutants']` True). `git diff <base>..HEAD` over privileged-mcp-action-v0.vectors.json, conformance/privileged-mcp-action-v0/gen_vectors.py, conformance/privileged-mcp-action-v0/vectors/, conformance/privileged-mcp-action-v0/MANIFEST.json and crates/ prints nothing. results.json: `_about`, `row_contract`, `schema` and `unmeasured` unchanged; four of the five corpora rows unchanged and their four stored reports byte-identical; on the privileged-mcp-action-v0 row only diagnostic_channel_declared, manifest_sha256, measured_at.commit, report_ref, report_sha256, silent and survived move -- killed=6, declared_total=34, out_of_scope=4, known_holes=2, equivalent=0, unproved=0, score_percent=21.4, adequate=False, control_status='killed', tool_commit, tool_content_sha256, tool_source_state, tool_version and measured_at.depends_on all identical. Inside the report body, `corpus_digest` is the same `sha256:cb58ce91863f52e0568742b977f0642158453ec11bbcd25821f9171dccd03342` before and after. The strongest single check that this stayed diagnostic: rule 43, the finding-only rule, is still `scope: out_of_scope` / `verdict: unexercised` with its out-of-scope prose unchanged -- declaring findings as a diagnostic channel did not pull it onto the comparison surface. I independently recomputed both recorded digests from the files on disk and both match.",
      "disposition": "pass"
    },
    {
      "id": "2",
      "summary": "One rule, one function -- and both callers proved to depend on it. The rule is a single block at conformance/adequacy/published_rows.py:262-267 inside `_validate_report`. `grep -n _validate_report` shows exactly one definition (line 228) and exactly two call sites: line 304 in `project_report` (the measure/write path, called from measure_all.py:184) and line 361 in `_validate_current_row`, which is reached only from `load_results` (line 464) -- the read path used by check_published_numbers.py:223 and project_published_numbers.py:238. No second copy of the rule text exists anywhere in the repo. I proved the sharing rather than asserting it: with the six-line block deleted from published_rows.py, `python3 conformance/tests/test_published_rows.py` went from RC=0 / `Ran 52 tests ... OK` to RC=1 / `FAILED (failures=2)`, failing `test_silent_without_a_declared_channel_is_rejected` (projection path) *and* `test_stored_silent_without_a_channel_fails_on_load` (load path) together. Restored; `git status --porcelain` empty. The new tests also carry a positive companion (`test_silent_with_a_declared_channel_projects`), so the rule is pinned in both directions, and the shared fixture's `silent` 1->0 change is a fixture that would otherwise be invalid under the new rule, not a loosening.",
      "disposition": "pass"
    },
    {
      "id": "3",
      "summary": "The must-bite holds and its control is honest. I could not run it against the user's sibling checkout (it sits at 352e11ed, off the manifest pin), so I built a sandbox: a fresh local clone of corpus-adequacy detached at the pinned 13048989c84ab6b4e0281f9514ea45fb79a2d8b4, beside a second detached worktree of this head. There `python3 conformance/tests/test_privileged_mcp_diagnostic_channel.py` RC=0, `Ran 2 tests ... OK` -- both tests actually executed, not skipped. Assert order is right: in both tests the control (`verdict == 'control-killed'`, `moved == 1`, `report['control_status'] == 'killed'`) is asserted before the mutant, so a quiet harness cannot masquerade as the intended survived/silent verdict. Control-blinding checked for real, not reasoned about: I rewrote the control's own mutation from `ok = True` -> `ok = False` (normative) into `note = \"A\"` -> `note = \"Z\"` (finding-only) and reran -- RC=1, `FAILED (failures=2)`, both failing on `AssertionError: 'control-SURVIVED' != 'control-killed'`. Note the second half of that result: even in the run where `diagnostic_from: ['findings']` IS declared, the blinded control still read control-SURVIVED, so the control is judged on the normative outcome only and the diagnostic declaration cannot rescue it. That is exactly the property a control needs here. Restored; `git status --porcelain` empty.",
      "disposition": "pass"
    },
    {
      "id": "4",
      "summary": "INDEX.md and ERRATA.md are generator output. `python3 conformance/adequacy/project_published_numbers.py` RC=0 and `git status --porcelain` printed nothing afterwards -- the regenerated documents are byte-identical to what is committed, so neither file carries a hand edit. `project_published_numbers.py --check` RC=0 as well. Then `python3 conformance/adequacy/check_published_numbers.py` RC=0, `published adequacy projection agrees with the current typed rows` (this is the invocation ci.yml:177 runs). The two prose deltas are exactly what the numbers force: INDEX.md's row goes `22 survivors` -> `20 survivors` and ERRATA.md's verbatim tool block goes `22 mutant(s) survived` -> `20 mutant(s) survived`; the scores and denominators in both are untouched.",
      "disposition": "pass"
    },
    {
      "id": "5",
      "summary": "Measurement provenance is coherent. `measured_at.commit` is e0964acb8597cdebd49b665a15eccaebcbf8bbc7; `git cat-file -t` says commit, `git merge-base --is-ancestor e0964acb8 HEAD` returns 0, and `git log --oneline -1` identifies it as `experiment(adequacy): declare findings as a diagnostic channel` -- this head's own parent. `git diff --stat e0964acb8..HEAD` over all six `measured_at.depends_on` paths (the manifest, the vectors index, candidate-release.json, privileged_mcp_action.rs, verify_privileged_mcp_action.rs, denial_marker.rs) prints nothing, so no instrumented source changed between the run and this head; the second commit touches only results.json, INDEX.md, ERRATA.md and tests. Recomputed independently: manifest_sha256 and report_sha256 both match the bytes on disk, and every row count agrees with its report field. One thing I did NOT do and will not imply: I did not re-execute the full 34-mutant process measurement. Its `build` is `cargo build -p assay-cli --bin assay` per mutant against a copied tree, which is hours and many GB on this machine; what I verified is the artifact's internal coherence plus the chain-level mechanism in finding 3.",
      "disposition": "pass"
    },
    {
      "id": "6",
      "summary": "`silent` moves 0 -> 2 with `diagnostic_channel_declared` true, and the two are named. From the stored report: rule **40**, `a caller-visible denial marker must be backed by a decision record`, and rule **50**, `a decision payload is selected only on the exact .v0 schema (Stage 2)` -- both in group denial_marker_binding, both previously `verdict: survived`, now `verdict: silent` with `moved: 0, moved_diagnostic: 1`. `survived` drops 22 -> 20 by exactly those two and `score_percent` stays 21.4, which is the arithmetic the design claims (silent counts in the denominator, never the numerator; these two were already denominator members as survivors). The report gains a second `failures` entry naming the silent class and its remedy, every mutant row gains a `moved_diagnostic` field, and the two silent rows' `how` prose is rewritten from `no vector distinguishes it` to `no vector's declared outcome distinguishes it, but 1 vector(s) moved on the declared diagnostic channel`. Corroborated from the corpus: bad-108-observation-without-decision presents an observation with no decision record, which is precisely rule 40's `(None, _)` arm, and the bundle is invalid on decision cardinality regardless -- masked on the outcome, visible only in findings.",
      "disposition": "pass"
    },
    {
      "id": "7",
      "summary": "Disclosure verified, and it is worse than disclosed -- stated plainly rather than taken on trust. The skip is real and silent: from the worktree whose sibling sits off-pin, the guard reports `OK (skipped=2)` with `corpus-adequacy is at 352e11ed..., not the manifest-pinned 13048989...`, RC=0; from a tree with no sibling at all (a layout I built at /tmp/nosib to reproduce CI's situation) it reports `OK (skipped=2)` with `corpus-adequacy not found as a sibling checkout`, RC=0. So a green log proves nothing about the guard. Beyond the disclosure: `grep -rn test_privileged_mcp_diagnostic_channel` over the whole repo returns exactly one hit -- the usage line in the file's own docstring. The test is not in any workflow, pre-commit hook, or conformance inventory list, so CI does not skip it, CI does not run it at all. The same is true of conformance/tests/test_published_rows.py, which this PR extends and which is likewise referenced nowhere outside itself, so this is a standing property of conformance/tests rather than something this PR introduces. What CI does establish about this change is the rule itself, not its unit test: ci.yml:177 runs check_published_numbers.py, which goes through load_results -> _validate_report, so a results.json with silent>0 and no declared channel would fail a required lane. I am reporting this as an observation, not a block: the PR's claim is an experiment record, and the guard's non-invocation is the pre-existing shape of that directory. Worth an issue; not worth holding this head.",
      "disposition": "pass"
    },
    {
      "id": "8",
      "summary": "Spot-checked the corrected forecast against gen_vectors.py; the numbers really do win. Rule **39** (`an in-namespace event type over a payload that declares no in-namespace schema fails closed`, anchor `_ if in_namespace(&ev.type_) => violations.push(finding(`) is claimed unreached. conformance/privileged-mcp-action-v0/gen_vectors.py assigns the envelope type in exactly two places, both `\"type\": payload[\"schema\"]` (lines 122 and 129), and `grep -n '\\[\"type\"\\]|\"type\":'` finds no other assignment -- the generator cannot emit an event whose envelope type differs from its payload schema, which is the only way to reach that arm. Confirmed on the artifacts too: I opened all 14 vector bundles and walked all 22 events; zero have `type != data.schema`. So 39 moves neither an outcome nor a finding, matching its recorded `survived` / `moved_diagnostic: 0`. Second spot-check, rule **41** (`a marker with a null or empty call.target_digest is unbindable`, arm `(Some(_), None)`): `classify_denial_marker` does not inspect target_digest, so the arm is reachable in principle, but every classified marker in the corpus carries a non-empty digest -- I listed all six observation payloads across the bundles and each has a real `sha256:...` target_digest -- so no vector reaches it. Both the 'unreached' and 'dead code' labels hold, and the commit message's admission that the forecast was wrong is accurate.",
      "disposition": "pass"
    },
    {
      "id": "9",
      "summary": "Adjacent wired checks green, with one environment artefact worth recording so nobody re-derives it as a defect. On this head: check_published_numbers.py RC=0; project_published_numbers.py --check RC=0; test_published_rows.py RC=0 (52 tests); test_published_numbers_provenance.py RC=0 (17); test_registry.py RC=0 (50); test_run_all.py RC=0 (64); scripts/ci/test_adequacy_lane.py RC=0 (40). test_published_numbers_guard.py first came back RC=1 with three errors, all `TypeError: run() missing 1 required keyword-only argument: 'execution_profile'` from measure_all.py:99 -- that is my machine's sibling corpus-adequacy being at 352e11ed, a tool newer than the manifest pin, not anything in this PR. Rerun from the sandbox whose sibling is detached at the pinned 13048989: RC=0, `Ran 42 tests ... OK`.",
      "disposition": "pass"
    }
  ],
  "no_findings": false,
  "builder": { "agent": "muse", "instance": "build-2862" },
  "reviewer": { "agent": "claude", "instance": "review-2981-81edf20", "github_login": "Rul1an" }
}
```

Rul1an:
Reviewer record: https://github.com/Rul1an/assay/pull/2981#issuecomment-5648796601 (READY). All findings pass.

The normative surface is confirmed unchanged field by field, not by assertion: the manifest adds exactly two keys and changes none, `outcome_from` and the 35-entry mutant table are byte-identical, the four other corpora rows and their reports are byte-identical, and `corpus_digest` is unchanged.

The shared rule is genuinely shared: one `_validate_report`, two call sites, and deleting its six lines turns **both** new tests red together - which is the check that distinguishes one rule from two copies that happen to agree.

The control is honest. Rewriting it into a finding-only mutation makes both tests fail on `'control-SURVIVED' != 'control-killed'`, including in the declared-channel run, so the declaration cannot rescue it. That is the blinded-control failure mode closed rather than assumed.

`silent` 0 to 2 is rules **40** (marker must be backed by a decision record) and **50** (decision payload selected only on exact `.v0` schema); `survived` 22 to 20, score unchanged at 21.4.

**The disclosure turned out to be understated, and the reviewer said so.** The builder flagged that its test skips without a sibling Corpus Adequacy checkout. True - but `grep -rn test_privileged_mcp_diagnostic_channel` over the whole repository returns one hit, the file's own docstring. CI does not skip the guard; CI never invokes it. That is the standing shape of `conformance/tests/` rather than something this PR introduced, and the *rule* is live in a required lane through `check_published_numbers.py` - so the loss is the test's evidence, not the property. Filed as #2982.

One false alarm recorded so nobody re-derives it: `test_published_numbers_guard.py` first returned three `TypeError: run() missing 1 required keyword-only argument: 'execution_profile'` errors. That is the machine's sibling `corpus-adequacy` checkout sitting newer than the manifest pin, not this PR. From the pinned sandbox: 42 tests, OK.

PR #2991, opened 2026-09-13
#2982 Conformance inventory contract for uninvoked tests

Rul1an:
**Candidate, not a landing.**

Head SHA: `3b5386e9e9b1a3663d41bb4b29eba28e4bf8db0b`  
Base SHA: `592818d526d4090f1843f5d1c670a27884795e58`

This PR needs an independent exact-head review record before merge.

completes #2982.

## Re-measurement (at base)

Command used:

`cd /Users/roelschuurkes/wt-codex-2982 && python3 - <<'PY'`
`import re, subprocess`
`base='592818d52'`
`tests=[p for p in subprocess.check_output(['git','ls-tree','-r','--name-only',base,'conformance/tests'], text=True).splitlines() if re.fullmatch(r'conformance/tests/test_.*\\.py',p)]`
`ci=subprocess.check_output(['git','show',f'{base}:.github/workflows/ci.yml'], text=True)`
`body=re.search(r'(?ms)^\\s{6}- name: Conformance inventory\\n(?P<body>(?:\\s{8}.+\\n)+)',ci).group('body')`
`refs=sorted(set(re.findall(r'conformance/tests/test_[^\\s]+\\.py', body)))`
`diff=sorted(set(tests)-set(refs))`
`print(tests); print(refs); print(diff)`
`PY`

Results:

- Full `conformance/tests/test_*.py` list: 12 files.
- Named in `.github/workflows/ci.yml` `Conformance inventory` step: 9 files.
- Set difference: exactly these 3 files:
  - `conformance/tests/test_corpus_adequacy_own_corpora.py`
  - `conformance/tests/test_privileged_mcp_diagnostic_channel.py`
  - `conformance/tests/test_published_rows.py`

## Per-file verdict

| file | verdict | why | property coverage in required CI |
|---|---|---|---|
| `conformance/tests/test_published_rows.py` | **Run in CI now** | Pure in-repo test; no sibling checkout requirement. | Added to `Conformance inventory` step as `python3 -W error::ResourceWarning conformance/tests/test_published_rows.py`. |
| `conformance/tests/test_corpus_adequacy_own_corpora.py` | **Recorded exception** | Requires sibling `corpus-adequacy` plus sibling corpora checkouts outside this repo checkout. | `python3 conformance/adequacy/check_published_numbers.py` (in required `Published-numbers projection contract` step) validates the published adequacy rows/projection for the same corpora claims. |
[...]

PR #2992, opened 2026-09-13
Select GHCR release assets with --pattern, not positional argv

Rul1an:
**Candidate, not a landing.**

Head: `9c669c31a25e06d67997622eb2d2db5a61d65c8b`
Base: `592818d526d4090f1843f5d1c670a27884795e58`

This needs an independent exact-head review record before merge. The builder does not write that record.

## Failure that shipped

v6.3.0 release run `34727302233`, job `Publish GHCR image` (`103645127255`), step "Download release binaries and SBOM":

```
accepts at most 1 arg(s), received 7
```

`gh release download [<tag>] [flags]` selects assets with `-p/--pattern`, never positionally. The step passed six asset names after the tag, so `gh` rejected the argv before any network call. The SBOM was present in the published bundle; the script-echo of the missing-SBOM line in that log is not evidence it ran.

## Claim to falsify

"Fixing `.github/workflows/release.yml:711` makes the release path green." That is false: line 862 in `verify-published-image` has the same shape (2 positional assets). No third execution site.

## Per-site verdict

| site | verdict |
|---|---|
| `.github/workflows/release.yml:711` | **BROKEN** — 6 positional assets; now `--pattern` x6 |
| `.github/workflows/release.yml:862` | **BROKEN** — 2 positional assets; now `--pattern` x2 |
| `.github/workflows/mcp-registry-publish.yml:91` | ok, `--pattern` |
| `.github/workflows/privileged-mcp-action-oci-candidate.yml:63` | ok, `--pattern` x3 |
| `.github/workflows/privileged-mcp-action-oci-candidate.yml:159` | ok, `--pattern` x3 |
| `conformance/privileged-mcp-action-v0/README.md:74` | ok, doc, `--pattern` |
| `conformance/privileged-mcp-action-v0/CONFORMANCE-PROTOCOL.md:236` | ok, doc, `--pattern` |
| `conformance/privileged-mcp-action-v0/tests/test_activation_kit.py:171,3574` | ok, asserts on the doc text |
| `docs/ops/MCP-REGISTRY-DISCOVERY-AUDIT.md:45` | ok, doc |
| `packaging/mcp-registry/README.md:18` | ok, doc |

`scripts/` has zero hits. `git grep` on `origin/main` is the same 11 sites. No third workflow site.

## Guard

Text-shape over `.github/workflows/*.ya?ml` only. The failure is argv shape, rejected before any network call; executing real `gh` would test the CLI, not the workflow text that produced the 7-arg line. Docs are not an execution path.

Seams:

1. **Line continuations.** Joined before tokens are counted. A line-oriented matcher sees only `gh release download "$VERSION" --repo ... \` and reports the broken sites clean.
2. **Value-taking flags.** Taken from `gh release download --help` on this machine: `-R/--repo`, `-D/--dir`, `-p/--pattern`, `-O/--output`, `-A/--archive`. `--clobber` and `--skip-existing` do not take a value. A naive "tokens not starting with `-`" count treats `--dir "$asset_dir"` as a second positional and fires on every correct site.
3. **Scope.** Workflows only. A doc line is not an execution path; the doc sites are already `--pattern`; `scripts/` has zero hits.
4. **Raw text, not a YAML parse.** The executed argv is the source Actions echoes, original line numbers are native, and a YAML load would still need the same `\` joiner for `|` blocks. Folded `>` scalars with a positional on the next source line are a non-claim; no current site uses that form. Only command-position `gh` counts, so a step name such as `Verify gh release download argv contract` is not an invocation.

Wired as a pre-commit hook (`files:` covers `.github/workflows/.*\.ya?ml` and the guard script) and as a required `CI` job step, matching the neighbouring workflow-contract hooks.

Self-test (entrypoint at the bottom of the file):

```
python3 scripts/ci/test-gh-release-download-argv.py
```

Collected 8, declared 8.

Changing `release.yml` also refreshes the published-release golden-path harness digest for that file. That pin is a hash of the workflow bytes, not a version pin.

## RED

| mutation | guard output |
|---|---|
| 711 restored to 6 positional assets | `.github/workflows/release.yml:711: gh release download accepts at most 1 positional (the tag); received 7: ['$VERSION', 'assay-mcp-server-${VERSION}-x86_64-unknown-linux-gnu.tar.gz', 'assay-mcp-server-${VERSION}-x86_64-unknown-linux-gnu.tar.gz.sha256', 'assay-mcp-server-${VERSION}-aarch64-unknown-linux-gnu.tar.gz', 'assay-mcp-server-${VERSION}-aarch64-unknown-linux-gnu.tar.gz.sha256', 'assay-${VERSION}-sbom-cyclonedx.tar.gz', 'assay-${VERSION}-sbom-cyclonedx.tar.gz.sha256']. Select assets with --pattern.` |
| `mcp-registry-publish.yml:91` `--pattern server.json` to positional | `.github/workflows/mcp-registry-publish.yml:91: gh release download accepts at most 1 positional (the tag); received 2: ['$TAG', 'server.json']. Select assets with --pattern.` |
| all sites correct | `problems=[]` |
| positional asset on a continuation line | `seam-continuation.yml:1: gh release download accepts at most 1 positional (the tag); received 2: ['$TAG', 'leftover-asset.tar.gz']. Select assets with --pattern.` |
| `--dir "$asset_dir"` on a correct site | `[]` (pass) |

The 711 and 862 mutants are well-formed on both the pre-fix positional text and the post-fix `--pattern` text.

## Out of scope

No republish of the v6.3.0 GHCR image. The next release publishes the first one. `release.yml` is not dispatched from this PR.

---

**Head advanced twice since the first review; the second advance required a fresh review.**

Current head `cfb5b45461f075d804c5ba535c205878c85c5b05`, base `b1ddd1ab71ef8dab786d453942ece8016e2ded9c`.

| head | why it changed | record |
|---|---|---|
| `9c669c31a` | original candidate | reviewed READY (29-probe battery) |
| `03d412d4f` | CI blocker: `LIVE_SHA256_CONSUMERS` still held the pre-`--pattern` line texts, plus the comment false positive | re-reviewed READY (delta only) |
| `cfb5b4546` | #2991 landed a change to `.github/workflows/ci.yml` | re-reviewed READY (coexistence only) |

The third review was not optional. `derive_carry` refused to carry the second record:

```
review-record-check=fail reason=carry_touched_reviewed_file .github/workflows/ci.yml
```

#2991 changed a file this review covered, so the advance is not disjoint from the review and the second AGENTS.md condition fails by derivation. Both earlier advances in today's queue carried automatically; this one did not, which is the condition doing its job rather than failing.

Verified at the current head: #2991's conformance-inventory line (`ci.yml:60`) and this PR's argv-contract step (`ci.yml:1222`) coexist; all four affected contracts pass together; every mirror of the inventory command list agrees; the argv suite declares and collects 9.


Rul1an:
```json
{
  "head_sha": "9c669c31a25e06d67997622eb2d2db5a61d65c8b",
  "builder": {"agent": "cursor", "instance": "grok-2987"},
  "reviewer": {"agent": "muse", "instance": "r2992", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [],
  "no_findings": true,
}
```

Rul1an:
CI blocker + matcher follow-up on `03d412d4f96757ff9841f2cd056668ac43dd201f`.

## Sidecar allowlist

`Release asset contract` failed because `LIVE_SHA256_CONSUMERS` still pinned the
old positional line texts. `is_checksum_producer` treats any unmatched `.sha256`
line as a producer, so the `--pattern` rewrite raised
`release checksum producer bypasses write_sha256_sidecar.sh` before the later
drift check could say the pin moved.

The four entries are now the stripped texts of `release.yml:713`, `715`, `717`, and
`864`, derived from the file. The fourth is a shape change: one two-positional
line became two `--pattern` lines, so `"$archive" "${archive}.sha256"` has no
same-text successor.

Local confirmation on this head:

- `bash scripts/ci/test-write-sha256-sidecar.sh` → `sha256 sidecar writer tests passed`
- `bash scripts/ci/test-release-assets.sh` → `release asset preflight tests passed`
- `python3 scripts/ci/test-gh-release-download-argv.py` → 9 tests, OK

## Should `LIVE_SHA256_CONSUMERS` exist?

Yes as an admission gate. No as a hand-copied whole-line list of download sites.

`live_consumers` is already derived from the workflow walk. The pin exists so
that derivation cannot silently grow: if the derived set were the allowlist, a
new `.sha256` write that does not look like a known sink would be auto-admitted
as a consumer. That property is load-bearing. A digest of the same lines would
still need a human edit when a consumer is added; the pin is that edit.

Whole-line text is the wrong key for the four download entries. This head had
to update them because `--pattern` is argv shape, not checksum production. A
later `--clobber` or a wrap would do it again.

Asset-name keying is cheaper and would survive a flag add, but it is weaker:
`sha256sum "$archive" > "${archive}.sha256"` still contains the known token and
would pass.

The formulation that would not need editing when an unrelated flag is added:

1. Keep the verbatim pin for the four non-download consumers (`find`, `MCPB_SHA`,
   the provenance YAML key, and the prose line). Those are stable.
2. Classify a `.sha256` token as a consumer when the existing argv walk shows it
   is a `--pattern` / `-p` value of a `gh release download` invocation. A new
   download of a new `.sha256` asset is then correctly auto-allowed (download
   is not produce). A write of the same path is still refused unless it matches
   writer or verify.
3. Cost: one shared "joined invocation + pattern values" helper so the two
   contracts do not each own a joiner, plus a parity probe that every walk
   `--pattern` `.sha256` is a consumer and a `sha256sum > same-name.sha256`
   mutant is still refused. Roughly an hour, not this head.

I am leaving the list as a pin until that helper exists. Please do not treat
the reformulation as a blocker for this PR.

## Matcher: comment case only

Widened only the false positive. `invocation_problems` now skips a line whose
first non-space character is `#`, the same rule as the sidecar `active_lines`.
`test_comment_line_is_not_an_invocation` uses `# && gh release download x y`.
It failed first (`received 2: ['x', 'y']`), then went quiet after the skip.

Please file the other two Muse misses as a follow-up. They are not in the tree:
`sudo` / `env` / `VAR=x`, a single pipe, `$(...)` / backticks, and a one-line
`run: gh release download ...`. Catching those without new false positives is
a different matcher, not a one-line skip.

No review record from me. This head needs a fresh independent review.

Rul1an:
```json
{
  "head_sha": "03d412d4f96757ff9841f2cd056668ac43dd201f",
  "builder": {"agent": "cursor", "instance": "grok-2987"},
  "reviewer": {"agent": "muse", "instance": "r2992b", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [],
  "no_findings": true,
}
```

Rul1an:
```json
{
  "head_sha": "cfb5b45461f075d804c5ba535c205878c85c5b05",
  "builder": {"agent": "cursor", "instance": "grok-2987"},
  "reviewer": {"agent": "muse", "instance": "r2992c", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [],
  "no_findings": true,
}
```

PR #3061, opened 2026-09-16
evidence: verify-side-effects names the reason for every qualified claim (#2826)

Rul1an:
## Summary

First slice of #2826. `assay evidence verify-side-effects` now names the reason for every qualified claim.

- **Reason fields.** Each call row carries `occurrence_reason` and `bounded_negative_reason` exactly when that claim is not allowed. The shape is `{"origin":"claim_gate","gap":…,"rule":…}`, or `{"origin":"observer_refutation"}` for an observer-refuted occurrence. The table prints `reason=<gap>` next to a qualified claim.
- **Refutation fix.** After a refutation, the row no longer rebuilds a `Blocked` decision that carries the pre-refutation allow rule and no gap.
- **Parity test.** `claim_gate_parity::full_coverage_matches_complete_descriptor` now uses a complete descriptor, so it compares instead of returning early.
- **Table pin.** A `#[cfg(test)]` pin in `assay-evidence` checks that every non-allowed row of the claim-gate table carries a gap and maps to exactly one rule. No public type changes.
- **JSON identity.** It stays `assay.side_effect_verification.v0`. The new fields are additive, which `docs/architecture/CLI-JSON-IDENTITIES.md` treats as compatible.

Refs #2826

## Provenance

- Builder: Codex via the Cursor CLI (branch `codex/2826-claim-reason`), commit `cdb1ddebd068bdfc2217e7a766ec4a3470180495`, base `a5a1af1350f8814dc7db24c9ca1f3bb3e2488629`. The coordinator pushed it and opened this PR.
- Design: the "additive slice" comment on #2826 (2026-09-16).

## Evidence (builder)

- **RED:** the four new tests failed first, because the reason keys were absent and the table had no `reason=`.
- **Mutants:**
  - M1: omit the reason for degraded rows. The qualified-claim test fails.
  - M2: keep the old refutation rebuild. The refutation test fails.
  - M3: drop `reason=` from the table. The table test fails.
  - M4: make the parity descriptor incomplete. The new completeness assertion fails.
- **Gates:** the `verify_side_effects_cli`, `claim_gate_parity`, runner-schema `claim_support_parity`, `cli_json_identities` and `error_enums_non_exhaustive` tests pass; clippy, fmt and rustdoc are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Landing

- Current head (update-branch merge of main after #3064): `e996c6242508534b7cc59d1e54c0e904c2708c2e`

PR #3076, opened 2026-09-17
docs: the complete-coverage parity leg is live; say so

### LINKED ISSUE #2826: The claim gate has a kind nothing asks for, and a decision value consumers cannot act on
Two gaps on the claim-gate surface. Both are invisible to ADR-048 and both survive it unchanged: promoting the lattice into `assay_common` does not make anything request `ExhaustiveSet`, and does not give `Degraded` consequences a consumer can act on.

Measured on `claude/config-vocabulary-crosswalk` @ `056332719`.

## 1. `ExhaustiveSet` is in the vocabulary and no production path ever asks for it

`CoverageClaimKind` and `CodingAgentClaimKind` both carry three kinds. Across `assay-cli` there are six production call sites and they request exactly two:

| call site | kind |
|---|---|
| `evidence/verify_side_effects.rs:419` | `PositiveExistence` |
| `evidence/verify_side_effects.rs:449` | `BoundedNegative` |
| `sandbox/bundle.rs:369` | `BoundedNegative` |
| `sandbox/bundle.rs:377` | `BoundedNegative` |
| `sandbox/bundle.rs:392` | `PositiveExistence` |
| `sandbox/bundle.rs:401` | `BoundedNegative` |

Outside tests, `ExhaustiveSet` appears only in enum definitions and one parity-table entry. So the middle kind is a decision function nothing calls.

That matters because the middle kind is the one the gate reasons about most carefully. `CoverageDescriptor::claim_decision` (`crates/assay-runner-schema/src/coverage.rs:191-217`) gives it its own answer, distinct from both neighbours:

- `PositiveExistence` → `Allowed`
- `ExhaustiveSet` → `Degraded` (`coverage_descriptor_degrades_exhaustive_claim`)
- `BoundedNegative` → `Blocked` (`coverage_descriptor_blocks_absence_claim`)

And `crates/assay-runner-schema/src/claim_parity.rs:21` already records that `RunnerClaimGate` has **no field** for it, deliberately, with the reason:

> Folding it onto `measured_positive_claims` would read an exhaustive-set claim as merely degraded under `Clipped`, when dropped ring-buffer records are precisely the thing that falsifies an exhaustive set.

So the reasoning is written down and pinned, and nothing exercises it. Any report that says "these are all of them" today is making an exhaustive-set claim without asking the gate that exists to grade it.

**Ask:** name a production path that legitimately makes an exhaustive-set claim and route it through the gate, or record that we do not make such claims and say so where a reader would otherwise assume we do. `unseen_tools` / `untriggered_rules` on the policy path are the obvious candidates, since a coverage report that enumerates what was never seen is asserting something about a set.

## 2. `Degraded` is distinguishable from `Allowed` only through a different field

`coding_agent_weakest_ceiling` (`crates/assay-evidence/src/coding_agent.rs:340-360`) takes one branch for both:

```rust
if decision.decision == CodingAgentGateDecision::Blocked {
    return CodingAgentWeakestCeiling::Blocked;
}
// Allowed or Degraded, so this decision owes a rung. Fail closed when it does not have one.
let Some(ceiling) = decision.ceiling else { ... };
```

To be fair to it: this is not "a label with no effect". A `Degraded` decision does act, because it carries a lower `ceiling` and the fold takes the minimum. But the effect travels through `ceiling`, not through `decision`, so a consumer reading the field named `decision` cannot tell the two apart, and nothing in the tree requires `gap` / `ceiling` / `reason` to be rendered.

That is a real hazard for a value whose whole purpose is to say "you may state this, but not at full strength". If the qualification can be dropped silently, the qualified claim reads as an unqualified one.

**Ask:** decide which it is, and make the code say so.
- If `Degraded` is a behaviour, give it an enforced consequence: require the reason or the gap to be carried into any rendering, so a degraded claim cannot be presented as a clean one.
- If it is a presentation concern, say so in the type's docs and stop treating three decision values as three behaviours.

## Why now

Both were found while checking whether this tree can answer *does your own tool do this* for the three-kind gate. For the middle kind the honest answer today is no, and for `Degraded` it is "only if the consumer chooses to look". Neither is a correctness bug in what ships; both are gaps between what the vocabulary claims to model and what any code path actually exercises.

Related: ADR-048 (`docs/architecture/ADR-048-claim-gate-construction-moves-readings-stay.md`) promotes `ClaimKind` and `ClaimDecision` into `assay_common`. These two items are orthogonal to that move and should not be folded into it: the migration will carry both gaps forward untouched.

Rul1an:
## Description
Design review found three places that still described the complete-coverage parity leg as dead after #3061 had already made it live. This PR updates those statements and records the current status while preserving dated historical measurements.

## Evidence (the mutant)
- Baseline: `cargo test -p assay-evidence --test claim_gate_parity` passed (7 passed).
- Mutant proof in `/tmp/codex-2826-mutant`: reintroduced an early `return;` before the full-leg kind loop.
- Result: same test went red with `full_coverage_matches_complete_descriptor` failing at `parity leg must execute at least one comparison` (`compared > 0`).
- Scratch copy and its `CARGO_TARGET_DIR` were removed after the run.

## What changed
- Updated `crates/assay-evidence/tests/claim_gate_parity.rs` header: complete-coverage leg is live since #3061; fidelity/evidence divergence remains deliberate and ADR-tracked.
- Updated `CLAUDE.md` to remove the stale dead-leg statement and keep only the deliberate unpinned fidelity/evidence gap.
- Appended a dated 2026-09-17 status paragraph to `docs/architecture/ADR-048-claim-gate-construction-moves-readings-stay.md` before "Left open, deliberately".
- Added one doc sentence on `ClaimKind::ExhaustiveSet` in `crates/assay-common/src/claim.rs`, grounded by code:
  - no shipped caller path requests `ExhaustiveSet` (CLI call sites request only `PositiveExistence` and `BoundedNegative` in `crates/assay-cli/src/cli/commands/evidence/verify_side_effects.rs` and `crates/assay-cli/src/cli/commands/sandbox/bundle.rs`);
  - CAP-1 absence derivation gates on non-positive claims via `neg = claim_kind != Kind::PositiveExistence` in `crates/assay-evidence/src/coverage_attestation.rs`;
  - descriptor gate is the shipped table that degrades `ExhaustiveSet` in `crates/assay-runner-schema/src/coverage.rs`.

## Checks
- `git grep -n "leg is dead\|leg below is dead" -- CLAUDE.md crates docs/architecture` (only dated historical ADR lines remain)
- `cargo test -p assay-evidence --test claim_gate_parity`
- `cargo test -p assay-runner-schema --test claim_support_parity`
- `cargo doc -p assay-common --no-deps`
- `git diff --check`
- `git diff origin/main...HEAD --stat`

- Current head: `a131629958ff0cf2679181ece211e908a9d54d9b`

Closes #2826


Rul1an:
```json
{
  "head_sha": "a131629958ff0cf2679181ece211e908a9d54d9b",
  "builder": {
    "agent": "codex",
    "instance": "2826-close-truth"
  },
  "reviewer": {
    "agent": "cursor",
    "instance": "gemini-review-3076-a1316299",
    "github_login": "Rul1an"
  },
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {
      "id": "D1",
      "summary": "Verified truth of new statements: claim_gate_parity.rs full descriptor literal asserts compared > 0 (mutation probe failed with exit 101); fidelity gate and evidence gate take disjoint signatures; ExhaustiveSet has no shipped caller and gates CAP-1 absence / descriptor degrades",
      "disposition": "verified"
    },
    {
      "id": "D2",
      "summary": "git grep sweep confirms all remaining dead-leg phrasing hits across entire tree are corrected or dated historical ADR-048 lines from 2026-09-08",
      "disposition": "verified"
    },
    {
      "id": "D3",
      "summary": "ADR-048 diff shows 8 additions only for dated 2026-09-17 status paragraph before 'Left open, deliberately', leaving prior measurements untouched",
      "disposition": "verified"
    },
    {
      "id": "D4",
      "summary": "claim.rs diff modifies doc comments only with zero pub item changes; cargo doc -p assay-common --no-deps exited 0",
      "disposition": "verified"
    },
    {
      "id": "D5",
      "summary": "cargo test -p assay-evidence --test claim_gate_parity exited 0 (7 passed); cargo test -p assay-runner-schema --test claim_support_parity exited 0 (9 passed)",
      "disposition": "verified"
    },
    {
      "id": "D6",
      "summary": "PR body contains 'Closes #2826' exactly once and no other closing keywords",
      "disposition": "verified"
    }
  ],
  "no_findings": false,
  "independence": {
    "did_not_build": true,
  }
}
```

Issue #3084, opened 2026-09-17
docs: generate-crate-deps.sh drops edges at random (grep -q + pipefail SIGPIPE)


Rul1an:
`scripts/docs/generate-crate-deps.sh` sometimes drops a single edge from `docs/generated/crate-deps.mermaid`. During the review of #3083, 3 of 6 runs of `scripts/ci/check-docs-generated-drift.sh` on a loaded host each lost a different edge (https://github.com/Rul1an/assay/pull/3083#issuecomment-5710848658). The other runs and exact replays were byte-identical.

**Cause (read from the script at `main`).** The script runs under `set -euo pipefail`, and the edge loop decides membership with a pipe:

```bash
if echo "$WORKSPACE_MEMBERS" | grep -q "^${dep}$"; then
```

`grep -q` exits at the first match, so `echo` can still be writing and receives SIGPIPE. With `pipefail`, the pipeline then reports failure (141), the `if` is false, and the edge is skipped with no message. Whether `echo` finishes before `grep` exits depends on scheduling, which is why it shows up under load. The `deps=$(... 2>/dev/null | sort -u || true)` line above it also hides a failing `jq` silently.

**Why it matters now.** #3079 makes `check-docs-generated-drift.sh` a required check, so this race would turn into random merge blocks. #3079 waits for this fix.

**Fix.** Membership without a pipe (a here-string `grep -qxF -- "$dep" <<<"$WORKSPACE_MEMBERS"`, or an associative array), and no silent `|| true` around `jq`. Add a regression test that runs the generator repeatedly, or under a forced early-exit reader, and requires identical output. Also check the other generators in the drift list for the same `echo | grep -q` shape under `pipefail`.

Rul1an:
Fixed by #3086 (merged as 9548ee10a). `generate-crate-deps.sh` no longer checks membership through a pipe into `grep -q` under `pipefail`, and `jq` failures now fail the script. A deterministic reproduction (a writer that keeps writing after grep exits) dropped all 43 edges before the fix and none after. The other generators in the drift list were checked for the same shape and are safe. A regression test pins it, and the review is https://github.com/Rul1an/assay/pull/3086#issuecomment-5712132697.

PR #3083, opened 2026-09-17
release: prepare 6.5.0-rc.2

Rul1an:
## Description
- Release metadata preparation for `v6.5.0-rc.2` only.
- No production behaviour change.
- Install pin remains `v6.4.0`; no tag was created.
- Changelog scope updates: include `#3078` (RC1 wheel matrix newline fix); skip `#3076` (internal claim-vocabulary alignment and ADR/docs scope, not a new user-visible release surface).

## Candidate
- head: `e96cde24601be965338f52fa3bd50c80f3e0d439`
- base: `main`
- worktree: `/Users/roelschuurkes/wt-codex-release-650rc2`
- branch: `codex/release-650-rc2`
- scope: 15 files
  - `CHANGELOG.md`
  - `Cargo.toml`
  - `Cargo.lock`
  - `fuzz/Cargo.lock`
  - `crates/assay-adapter-a2a/Cargo.toml`
  - `crates/assay-adapter-acp/Cargo.toml`
  - `crates/assay-adapter-api/Cargo.toml`
  - `crates/assay-adapter-ucp/Cargo.toml`
  - `crates/assay-core/Cargo.toml`
  - `crates/assay-sim/Cargo.toml`
  - `packaging/agent-plugin/plugin.json`
  - `docs/generated/agent-golden-path.json`
  - `docs/guides/agent-golden-path.md`
  - `packaging/agent-plugin/skills/assay-golden-path/references/agent-golden-path.json`
  - `packaging/claude-plugin/skills/assay-golden-path/references/agent-golden-path.json`

## Verification
- `cargo metadata --locked --format-version 1` -> exit `101` after manifest-only bump (expected), then exit `0` after lockfile version-line update.
- `python3 scripts/docs/generate-agent-golden-path.py` -> exit `0`
- `bash scripts/ci/check-docs-generated-drift.sh` -> exit `0`
- `RBENV_VERSION=3.3.12 EXPECTED_RELEASE=v6.4.0 CHECK_VM=0 bash scripts/ci/check-assay-version-line.sh` -> exit `1` (expected stable-baseline mismatch: workspace `v6.5.0-rc.2` vs release target `v6.4.0`)
- `python3 scripts/ci/check-release-runbook-truth.py` -> exit `0`
- `bash scripts/ci/check-release-surface.sh` -> exit `0` (workspace `6.5.0-rc.2`, published pin `v6.4.0`)
- `python3 scripts/ci/check_internal_dep_versions.py` -> exit `0`
[...]

PR #3086, opened 2026-09-17
docs: generate-crate-deps.sh no longer drops edges under load

Rul1an:
## Description

The required generated-docs drift check can fail at random because the crate-deps generator sometimes omits a real workspace edge. After this change, membership cannot SIGPIPE, so the diagram stays stable under load.

## Cause

`generate-crate-deps.sh` runs under `set -euo pipefail` and decided workspace membership with a pipe:

```bash
if echo "$WORKSPACE_MEMBERS" | grep -q "^${dep}$"; then
```

`grep -q` exits at the first match. If `echo` is still writing, it receives SIGPIPE, the pipeline returns 141, the `if` is false, and the edge is skipped with no message. Whether that happens depends on scheduling, which is why it showed up on a loaded host during the #3083 review (3 of 6 runs each lost a different edge).

Deterministic reproduction, in a copy of the script: replace `echo "$WORKSPACE_MEMBERS"` with a writer that prints the members and then keeps writing 4MiB of filler (`dd if=/dev/zero bs=65536 count=64`). `grep -q` exits on the first hit while the filler is still flowing, every membership check fails, and all 43 `-->` edges disappear (sha256 `6d794f4f944b3b88277a5e62c91e59e7eaafa2e3a5d582062d99112cf20c7b4f` vs committed `d88e8e4ddcd9765fbb88e9b56bab9fbbcabf3f7a3a70805857d1aee515dd3ffa`).

## Fix

Membership is a here-string, not a pipe: `grep -qxF -- "$dep" <<<"$WORKSPACE_MEMBERS"`. That works on macOS bash 3.2 (`/bin/bash` 3.2.57 produced the same bytes as bash 5.3). A failing `jq` no longer hides behind `2>/dev/null ... || true`. The generated mermaid is byte-identical to the committed file.

## Sweep results

Only `scripts/docs/generate-crate-deps.sh` had a piped early-exit reader under `pipefail` in the drift `GENERATORS` list.

- `generate-agent-golden-path.py`, `generate-product-capabilities.py`, `generate-configuration-vocabulary-crosswalk.py`: Python; no `pipefail`. They call `git` via subprocess, not bash pipelines.
- `generate-module-map.sh`: `pipefail`, but no `| grep -q` or `| head`.
- `update-architecture-docs.sh`: `if ! grep -q "## Generated Diagrams" "$ARCH_DIAGRAMS"` is a file grep, not a pipe. No sourced bash generators.

## Test + mutant

Chose both a static scan of the `GENERATORS` list for `| grep -q` / `| head` and a 10-run identical-output pin (`scripts/ci/test-generate-crate-deps.sh`). The static scan is the honest RED: 20 unmodified runs on this host produced one hash even before the fix, so a retry loop would be flaky as a mutant. Planting the old `echo | grep -q` line fails the new test; HEAD passes.

## Measurements

- Before, unmodified, 20 runs: 1 distinct sha256 `d88e8e4ddcd9765fbb88e9b56bab9fbbcabf3f7a3a70805857d1aee515dd3ffa` (race did not fire on an idle host). Deterministic filler writer: all edges dropped.
- After, 20 runs: 1 distinct sha256, equal to the committed file. The deterministic-race copy with the here-string applied: no edge lost.

## Checks

- `shellcheck scripts/docs/generate-crate-deps.sh` → 0
- `bash scripts/ci/check-docs-generated-drift.sh` → 0
- `bash scripts/ci/test-generate-crate-deps.sh` → 0 at HEAD; 1 on the old-line mutant
- `pre-commit run --files` (changed files) → 0
- `git diff --check` → 0

- Current head: `a2dcf58e0a9bc0fd295780145b0ef0ad2b431edf`
Refs #3084


Rul1an:
```json
{
  "head_sha": "fe2007e843d08d9d67c7fa7f305aeb9f6e08deb2",
  "builder": {"agent": "cursor", "instance": "3084-crate-deps-race"},
  "reviewer": {"agent": "codex", "instance": "review-3086-fe2007e8", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {"id": "C1", "summary": "python3 C1 harness over main/head script copies -> main lost 43/43 edges under forced continuing writer; head lost 0/43 under same harness", "disposition": "verified"},
    {"id": "C2", "summary": "rg conditional early-exit pipe scan + broken-jq mutant -> no such conditional pipe in HEAD; broken deps jq exits 3 (not swallowed), unlike main mutant exit 0 with 0 edges", "disposition": "verified"},
    {"id": "C3", "summary": "python3 sweep of drift GENERATORS + shell closure (including read-assay-release-tag.sh) -> no '| grep -q', '| head', '| grep -m', or '| read' inside if/while under pipefail", "disposition": "verified"},
    {"id": "C4", "summary": "bash scripts/ci/test-generate-crate-deps.sh + old-line mutant -> HEAD test passes; restoring echo|grep -q line fails; hook runs via pre-commit (kernel-matrix lint), not directly wired in required CI on current head", "disposition": "verified"},
    {"id": "C5", "summary": "10x bash scripts/docs/generate-crate-deps.sh + sha256 + bash scripts/ci/check-docs-generated-drift.sh -> one distinct digest (d88e8e4d...) matching committed file; drift check exits 0", "disposition": "verified"},
    {"id": "C6", "summary": "/bin/bash 3.2 run + shellcheck + git diff --stat -> script runs clean on macOS bash 3.2, shellcheck 0 findings, diffstat is 3 files (+145/-3)", "disposition": "verified"}
  ],
  "no_findings": false,
}
```

PR #3091, opened 2026-09-17
sandbox: --enforce refuses when it cannot enforce

Rul1an:
## Description

`--enforce` now refuses when it cannot enforce. A new `--allow-audit-fallback` flag restores the previous degrade-and-record path.

Target release: 6.6.0.

## Decision

https://github.com/Rul1an/assay/issues/3081#issuecomment-5717394031

## What changed

- clap: `--allow-audit-fallback` (`conflicts_with = "fail_closed"`); `--enforce` help names the refuse contract
- no-backend guard and policy-conflict guard share `refuses_when_unenforceable`
- `assay.sandbox.degraded` events fire only on the explicit fallback path
- the #2637 contract tests now pass `--allow-audit-fallback`; new refusal tests pin exit 2
- sandbox docs and the sandbox ADRs describe the new default

## Migration note

Fail-closed correction (assay sandbox --enforce, migration-visible): --enforce now refuses when it cannot enforce. Without a Landlock backend (always on macOS, and on Linux kernels without Landlock), or when the policy conflicts with Landlock's allow-only model, the run exits 2 with E_BACKEND_UNAVAILABLE_UNENFORCEABLE or E_POLICY_CONFLICT_DENY_WINS_UNENFORCEABLE and executes nothing, where it previously warned, recorded assay.sandbox.degraded, ran the command in audit mode and exited 0. Pass --allow-audit-fallback to keep the old degrade-and-record behaviour. --fail-closed is unchanged and is now implied by --enforce. Runs without --enforce are unaffected.

The shipped strings match this note. The reason codes are printed as those exact tokens. `--fail-closed` without `--enforce` is still the only way that combination refuses a policy conflict; `--fail-closed` without `--enforce` on a host with no backend still does not enter the no-backend guard (unchanged).

## RED

`enforce_without_fallback_refuses` failed before the production change:

- `left: Some(0)` / `right: Some(2)`
- child ran (`touch` of the marker path was in the command line; audit fallback warning on stderr)
- command: `cargo test -p assay-cli --test contract_sandbox_enforcement_health_degrade enforce_without_fallback_refuses` in `/tmp/grok-3081-red` (`CARGO_TARGET_DIR` isolated). Exit 101.

## Must-bite

Control: the `--allow-audit-fallback` health-artifact tests stayed green in every mutant.

- **M1** revert the no-backend guard to `args.fail_closed`: `enforce_without_fallback_refuses` red (exit 0, audit fallback). Control green.
[...]

PR #3094, opened 2026-09-17
test(cli): size the bounded-probe deadline from measured Windows timings

Rul1an:
## Description
Windows-only CI flakes in `stdout_json_write_contract` exhausted fixed `run_bounded` deadlines (`15s` and `45s`) in otherwise unrelated runs. This change sizes only those test deadlines using measured CI timings, without changing non-Windows behavior.

Changelog: none (CI/test-internal only).

### 1) Test-side deadlines inventory
| Location | Timeout | Notes |
|---|---:|---|
| `tests/support/bounded_process.rs` `GOLDEN_PATH_LIMITS` | 10s | Shared helper used by many contract tests |
| `crates/assay-cli/tests/stdout_json_write_contract.rs` `LIMITS` (before) | 15s | `run json success` probe |
| `crates/assay-cli/tests/stdout_json_write_contract.rs` `LARGE_RUN_LIMITS` (before) | 45s | `run json size probe` |
| `crates/assay-cli/tests/recovery_argv_publisher_parity.rs` `LIMITS` | 10s | CLI contract |
| `crates/assay-cli/tests/policy_validate_machine_contract.rs` `LIMITS` | 5s | CLI contract |
| `crates/assay-cli/tests/explicit_config_class_contract.rs` `LIMITS` | 5s | CLI contract |
| `crates/assay-cli/tests/config_recovery_argv_contract.rs` `LIMITS` | 5s | CLI contract |
| `crates/assay-cli/tests/evidence_verify_*` inline `ProcessLimits::new` | 60s | Bundle verification contracts |
| `tests/support/bounded_process.rs` test-only local limits | 100ms / 2s / platform fn | Helper self-tests |

### 2) Healthy-run timings from 5 successful `ci.yml` runs
Runs sampled: `35259837170`, `35255448341`, `35255387723`, `35255357560`, `35253167189`.

#### `stdout_json_write_contract` test file duration
| Run | windows-latest | ubuntu-latest |
|---|---:|---:|
| 35259837170 | 26.72s | 1.52s |
| 35255448341 | 24.49s | 3.90s |
| 35255387723 | 25.93s | 1.61s |
| 35255357560 | 33.87s | 3.35s |
| 35253167189 | 22.43s | 1.52s |
| **Median** | **25.93s** | **1.61s** |
| **Max** | **33.87s** | **3.90s** |

#### Whole `Build + Test` job duration (contrast, same 5 runs)
| Platform | Durations | Median | Max |
|---|---|---:|---:|
| windows-latest | 873s, 992s, 820s, 1018s, 983s | 983s | 1018s |
| ubuntu-latest | 685s, 787s, 789s, 867s, 695s | 787s | 867s |
| macos-latest | 517s, 433s, 462s, 503s, 490s | 490s | 517s |

### 3) Failure pattern check (cold-start vs random)
Failed jobs inspected: `105048129646` (run `35172703049` attempt 1), `105274790443` (run `35242483716` attempt 1), `105333480036` (run `35259879209` attempt 1).

- Failing tests were **not first in file**:
  - `run_json_early_failure_reader_gone_is_exit_three` passed before `run_json_partial_read_then_close_is_exit_three` failed.
  - In PR #3091 attempt 1, both `run_json_partial_read_then_close_is_exit_three` and `run_json_success_is_one_document_and_exit_zero` failed later in the same file.
- Similar pattern in the earlier mcp-server occurrence: `required_python()` failed after many preceding tests in `agent_golden_path_contract` passed.
- Failed attempts were on the same runner image/version as healthy attempts (`windows-2025-vs2026`, `20260907.229.1`).
- Reruns on same head went green (e.g. `35172703049` attempt 2, `35242483716` attempt 2).

### 4) Primary-source platform notes
- GitHub docs runner specs (Windows standard hosted runners are lower-spec than macOS public runners and can vary by public/private repo plan): https://docs.github.com/en/actions/reference/github-hosted-runners-reference
- Runner-images issue documenting first-command cold-start on Windows (`cargo --help` and similar): https://github.com/actions/runner-images/issues/6561
- Runner-images PR noting image-side PowerShell cold-start improvement from roughly 20s to 8s: https://github.com/actions/runner-images/pull/8495
- Runner-images Defender status clarification (server images 2022/2025 show Defender realtime off; Win11 arm64 differs): https://github.com/actions/runner-images/issues/14326

## Decision and why
Raise only the Windows deadlines for this one flaky test target:
- `LIMITS.timeout`: `15s -> 30s`
- `LARGE_RUN_LIMITS.timeout`: `45s -> 75s`

Rationale:
- Measured healthy Windows runtime for this exact test target is already `22.43s..33.87s` (median `25.93s`).
- A `45s` budget can be consumed by runtime + transient runner jitter; a `15s` budget is clearly below observed healthy p50 for the same binary shape when loaded.
- Non-Windows timings are far below current budgets, so those paths remain unchanged.

## What changed
- Added per-platform constants in `crates/assay-cli/tests/stdout_json_write_contract.rs`:
  - `RUN_JSON_LIMITS_{WINDOWS,NON_WINDOWS}`
  - `RUN_JSON_LARGE_LIMITS_{WINDOWS,NON_WINDOWS}`
- Kept non-Windows timeouts unchanged.
- Added `run_json_timeout_budgets_are_pinned_per_platform` test to pin:
  - both platform constants, and
  - the active `#[cfg]` selection.
- Added an inline measurement comment with the sampled Windows timing envelope.

## Checks
- `cargo test -p assay-cli --test stdout_json_write_contract` ✅
- `cargo clippy -p assay-cli --all-targets -- -D warnings` ✅
- `git diff --check` ✅
- `git diff origin/main...HEAD --stat` ✅

- Current head: `d2624263d6a0344705ae1c8c05629f19c8a140ed`

## Not claimed
- No root-cause claim about a single Windows subsystem.
- No claim that this eliminates all Windows flakes in other test targets.
- No production behavior change.

Refs #2246


Rul1an:
## Coordinator finding F1: the measurement does not measure the quantity the deadline governs

The investigation here is the right shape - measure first, refuse to claim a root cause, cite primary sources for the runner behaviour. The failure-pattern section in particular is exactly what I wanted: establishing that the failing tests were *not* first in the file rules out plain cold start, and that is a real result.

One thing does not hold up, and it is load-bearing for the chosen numbers.

`LIMITS` and `LARGE_RUN_LIMITS` are **per-process** deadlines. `run_bounded` (`tests/support/bounded_process.rs:136-161`) spawns exactly one child and bounds that one child. The PR sizes those per-process budgets from the **whole test target's** duration:

> Measured healthy Windows runtime for this exact test target is already `22.43s..33.87s` (median `25.93s`).
> [...] a `15s` budget is clearly below observed healthy p50 for the same binary shape when loaded.

The 22.43s..33.87s figure aggregates every `#[test]` in the file and every `assay` invocation inside them. `LARGE_RUN_LIMITS` alone is referenced at six separate sites (`stdout_json_write_contract.rs:395, 428, 562, 616, 624`, plus the `#[cfg]` selection). So the sentence "a 15s budget is clearly below observed healthy p50" is a claim about a single probe's wall time, and the data behind it is a claim about the sum of many.

I can see the defence, and it is not a bad one: `cargo test` runs `#[test]` functions in parallel threads, so under that contention a single probe's wall time can approach the file's total. That would make the file duration a rough upper-bound proxy rather than a category error. But "rough upper-bound proxy" is a different claim from "observed healthy p50", and this repository's rule is that a number carries its provenance.

It also matters for whether the fix works. The observed failure was a **45s** deadline expiring while the whole file completes in at most 33.87s on a healthy run. If one probe really can exceed 45s while all probes together normally finish in under 34s, the stall is an outlier of a different kind, and 75s is a guess about the size of an outlier we have not measured. The PR's own "Not claimed" section is honest about there being no root cause; the numbers should be equally honest about being headroom rather than measured p50.

## What I want instead, and it is cheap

The file already imports `std::time::Instant`. Record each bounded probe's elapsed time and print it, push, and read the `windows-latest` job log. That gives the actual per-probe distribution on the platform in question, which is the quantity the deadline governs.

Then one of two things is true, and either is a fine outcome:

1. Per-probe healthy time on Windows is genuinely in the tens of seconds. The new budgets are then measured, not assumed, and the PR body says so with the real numbers.
2. Per-probe healthy time is a few seconds and the failures are 10x-plus outliers. Raising the budget is then still defensible as flake suppression, but the PR must say that plainly - that it buys CI stability and does not explain the stall - and #2246 stays open with the outlier as the remaining question.

Keep everything else. The per-platform constants, the `run_json_timeout_budgets_are_pinned_per_platform` test pinning both the constants and the active `#[cfg]` selection, the unchanged non-Windows budgets, and the inline measurement comment are all right, and the pinning test is the part that stops this silently drifting later.

F1 disposition: open.

Rul1an:
## Evidence for F1, from a failure on another PR tonight

#3104 hit this on `windows-latest` ([job 105373307738](https://github.com/Rul1an/assay/actions/runs/35271522738/job/105373307738)), and the two failures carry the per-probe detail that the file-duration measurement could not give.

```
run_json_success_is_one_document_and_exit_zero
  deadline of 15s expired; outstanding=[child still running, stdout handle still open, stderr handle still open]
  stdout="" (0 bytes); stderr="" (0 bytes)

run_json_partial_read_then_close_is_exit_three
  deadline of 45s expired; outstanding=[child still running, stdout handle still open, stderr handle still open]
  stdout="" (0 bytes)
  stderr="Running 48 tests...\nInfo: No seed provided. Using generated seed: 8429800420026210934\n" (86 bytes)

test result: FAILED. 29 passed; 2 failed; 1 ignored; finished in 84.64s
```

Three things in that, none of which is "the budget was slightly too small".

**The 15s case produced zero bytes on both streams.** Not a slow run, not a partially written document. In fifteen seconds the child wrote nothing at all, while the harness reports it as still running. On a healthy Windows job this whole target completes in 22.43s to 33.87s across 32 tests, so the per-probe expectation is on the order of a second.

**The 45s case got as far as `Running 48 tests...` and then stalled for the remaining 45 seconds.** So the child started, resolved its configuration, emitted a seed line, and then produced nothing further. That is a stall after startup, not a cold start.

**The file took 84.64s.** Against a healthy maximum of 33.87s, the failing run is not a shifted distribution, it is a different mode.

## What this does to the decision

Raising 15s to 30s and 45s to 75s does not address any of that. A child that writes zero bytes in fifteen seconds and is still running is not going to be rescued by thirty; a child that stalls after its first line for forty-five seconds is not obviously rescued by seventy-five. The measurement in the PR body is consistent with a healthy distribution, and the failures are not drawn from it.

Note also `outstanding=[... stdout handle still open, stderr handle still open]`. Those are the harness's own reader threads reporting they never saw EOF. A hypothesis worth one probe: the stall is in `run_bounded`'s interaction with the job-object wrapper on Windows when the child writes nothing, rather than in the child itself.

## What I want in the PR

The budget change can stay, but its justification has to change with it. Take the second branch I offered:

> Raising the budget is defensible as flake suppression, but the PR must say plainly that it buys CI stability and does not explain the stall, and #2246 stays open with the outlier as the remaining question.

Concretely:

1. Rewrite the rationale to say the budgets are headroom, not measured p50, and that the observed failures are a separate mode the budget does not explain.
2. Keep the per-platform constants and the `run_json_timeout_budgets_are_pinned_per_platform` test. Those are the durable part.
3. Add the per-probe instrumentation anyway, as a permanent part of the harness rather than a one-off: on a deadline expiry, `run_bounded` should already report elapsed time; make it also report the time of the last byte received on each stream. "Zero bytes in 15s" and "86 bytes then silence for 45s" are different diagnoses, and today we can only tell them apart by reading the stderr echo.
4. Say explicitly in the PR body that #2246 is not closed by this change.

F1 disposition: open, with the evidence above.

Rul1an:
Superseded by #3107, which carries this PR's commit `d2624263` forward with `git cherry-pick -x`, so the per-platform budget constants and the `run_json_timeout_budgets_are_pinned_per_platform` pinning test land with their provenance intact.

Closing rather than finishing here only because Codex is out of quota until October and cannot take the F1 round. The measurement work in this PR was the right first move: sampling five real runs, checking whether failures were first-in-file to rule out cold start, and citing the runner-image sources is what made the second mode visible at all. #3107 adds the part F1 asked for: on expiry the harness now reports bytes received and last-byte time per stream, so a child that never wrote and a child that stalled after its first line are no longer the same error message. #2246 stays open.

PR #3100, opened 2026-09-17
feat(incident-package): incident package container reader and refusal report (Refs #2493)

Rul1an:
This PR implements slice 4 of issue #2493: the Incident Package v1 read-only outer ustar container reader and the closed refusal report type in `assay-evidence`. The reader enforces strict uncompressed POSIX ustar framing, exact member allowlists (`inventory.json`, `assessment.json`, and ascending `objects/<64-char lowercase hex sha256>`), header field invariants, trailing 1024-byte zero block termination, and content-addressed payload digest matching, refusing any violation with phase 2 `input_shape`. This slice does NOT yet include a writer, exporter, CAP-1 generation, attestations, or user-facing CLI.

Refs #2493

## Schema Identity Classification

`assay.incident.verify.v1` is classified in the `not-cli-documents` block because it is declared in `assay-evidence` as an incident package verification result schema and no CLI command emits it in this slice.

## Commands & Verification

Commands executed against commit `0cd1538b767e92fd3a5e91631a21cd7773bf4c2f`:

- `cargo test -p assay-cli --test cli_json_identities`
```text
running 15 tests
test external_tests_rs_rule_is_a_commands_path_suffix ... ok
test command_tree_component_rule_refuses_anything_that_could_leave_the_directory ... ok
test dependency_crates_match_the_manifest ... ok
test command_tree_walk_collects_a_new_unreferenced_rs_file ... ok
test manifest_promotion_outputs_are_cli_documents ... ok
test strip_test_modules_removes_pub_crate_mod ... ok
test command_tree_walk_rejects_symlink_root_fail_closed ... ok
test command_tree_walk_rejects_symlinks_fail_closed ... ok
test documents_without_an_identity_are_recorded_and_still_exist ... ok
test documents_are_bound_to_a_writer ... ok
test external_tests_rs_serializers_are_not_candidates ... ok
test json_serializing_command_files_are_accounted_for ... ok
test inline_cfg_test_module_serializers_are_not_candidates ... ok
test every_production_identity_is_classified ... ok
test describe_binds_only_recorded_documents ... ok

test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.45s
```

- `bash scripts/ci/check-docs-generated-drift.sh`
```text
generated artifacts reproduce from their generators (22 files).
```

- `cargo test -p assay-evidence`
```text
test result: ok. 335 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.63s
...
[...]

PR #3090, opened 2026-09-17
evidence: index rebuild from canonical bundles and an honest object_lock status

Rul1an:
## ADR Boundary Slice

- Canonical ledger: None (no active programme; standalone change outside an active programme)
- Slice: #2492 slice 2 (rebuildable run index and honest object_lock status)
- Builder: Antigravity
- Reviewers: Codex (Round 1 review: BLOCKED on findings F1, P4, F2/P5, F3/P2)
- Final head SHA: f4d50daf91492bd7e3bb3ea5342c163098d34461
- Current head: fa88fdf33d6233728ddad5b98eab2355b30418ae
- ADR invariants affected: ADR-042, ADR-043 (evidence-chain integrity, non-canonical index derivation from canonical bundles, honest status reporting)

## Design

Ratification: Issue #2492, design comment #5696376440.

This slice delivers:
1. Rebuildable run-to-bundle index (`assay evidence index rebuild` and `ObjectStoreBundleStore::rebuild_index`) from canonical bundles under `bundles/`.
2. Honest closed vocabulary for `StoreStatus::object_lock`.

### 1. Honest `object_lock` Vocabulary (`assay-evidence`)
- Defined closed vocabulary constants:
  - `"enabled"`
  - `"disabled"`
  - `"unobserved:unsupported_backend"`
  - `"unobserved:not_probed"`
  - `"unobserved:permission_denied"`
  - `"unobserved:error"`
- Preserved semver stability of public struct `StoreStatus` by maintaining `pub object_lock: String` with no added fields.
- Implemented `object_lock_status_for_backend(&str) -> &'static str`:
  - `file://` and `memory://` return `"unobserved:unsupported_backend"`
  - `s3://` returns `"unobserved:not_probed"` (since S3 probe is deferred)
  - Other backends return `"unobserved:unsupported_backend"`

### 2. Canonical Run Index Rebuild (`assay-evidence`, `assay-cli`)
- Implemented `ObjectStoreBundleStore::rebuild_index(&self, ceiling: StreamCeiling) -> StoreResult<IndexRebuildReport>`:
  - Scans canonical bundles under `bundles/`.
  - Inspects existing non-canonical run refs under `runs/`.
  - Identifies stale references pointing to missing bundles without trusting or deleting them.
  - Streams each canonical bundle up to `ceiling`, performing full cryptographic and schema integrity verification (`crate::verify_bundle`) and matching manifest `bundle_id`. Corrupt bundles are skipped and recorded in `failed_bundles`.
[...]

PR #3099, opened 2026-09-17
evidence: declared depth on the session coverage sibling event

Rul1an:
## Description

A session coverage sibling can now carry a declared depth extent. When
`retained_through` is below `steps_total`, the sibling reads as Partial
before the existing claim table runs. When the two numbers are equal,
coverage and source class stay as they were, so a producer saying it
retained everything still cannot mint Observed.

## Design

Binding comment (last two paragraphs):
https://github.com/Rul1an/assay/issues/2422#issuecomment-5712306171

Slice 1 shipped the sibling event and `session_finding_claim_decision`
in #3069. This slice adds the quantitative form from the 09-05 exchange
without a second decision table.

## What changed

- `PayloadSessionCoverage` gained optional `steps_total` and
  `retained_through` with `skip_serializing_if`, so the wire stays
  additive. `new()` is unchanged; callers attach depth with
  `with_declared_depth`. A setter rather than a second constructor:
  depth is optional and every existing `new` call stays source-compatible.
- `session_coverage_declared_depth` is the remapping. 
  `session_finding_claim_decision` applies it, then the slice-1
  producer-Observed collapse, then `coding_agent_claim_decision`.

## Derivation table

depth × coverage × source_class → decision (BoundedNegative / PositiveExistence):

| depth | coverage | source_class | BoundedNegative | PositiveExistence |
| --- | --- | --- | --- | --- |
| missing or one-sided | as declared | as declared | slice 1 | slice 1 |
| retained < total | any | any | Partial, blocked | Partial, allowed; ceiling from source_class |
| retained == total | Observed | ProducerReported | SelfReported, blocked | SelfReported, degraded, asserted |
| retained == total | Observed | BoundaryObserved | Allowed | Allowed, observed_in_path |
| retained == total | Partial | BoundaryObserved | Partial, blocked | Partial, allowed |

## RED

`cargo test -p assay-evidence --test session_finding_coverage_depth_gate -- --nocapture first_failing_case`

Compile failure (exit 101) before the production change:

- unresolved import `session_coverage_declared_depth`
- no method `with_declared_depth` on `PayloadSessionCoverage`
- unknown fields `steps_total` and `retained_through`

## Must-bite

Each mutant ran in its own `/tmp/grok-2422-m*` copy with its own
`CARGO_TARGET_DIR`, then the copy was deleted.

- M1 treat retained == total as Observed →
  `retained_eq_total_does_not_mint_observed_from_partial` (Observed vs Partial)
- M2 let the depth fields override source_class →
[...]

Issue #3096, opened 2026-09-17
ci: the Windows binary is built, attested and declared stable, but never executed


Rul1an:
## Measured state

Measured on `origin/main` `f0413991f46500a99f7c031bf793e795e0798b07`.

We build, attest, publish and declare `stable` a Windows binary that no job in this repository ever executes.

### The CLI is excluded from the Windows test run

`.github/workflows/ci.yml:871-874`:

```yaml
# Non-Linux: Exclude OS-specific crates (assay-monitor, assay-cli)
- name: Test Workspace (Cross-Platform)
  if: runner.os != 'Linux'
  run: cargo test --locked --workspace --exclude assay-ebpf --exclude assay-it --exclude assay-monitor --exclude assay-cli
```

`assay-cli` is the crate that produces the `assay` binary. Two targeted CLI test files are run back on Windows by name:

- `ci.yml:838-851` runs one test from `mcp_preflight_contract`;
- `ci.yml:853-868` runs `stdout_json_write_contract`.

Everything else the binary does on Windows is untested. The exclusion comment attributes it to OS-specific crates, which is true of `assay-monitor` (Linux eBPF) but not of `assay-cli`.

### The release never runs the artifact it ships

`release.yml:129-133` builds the `x86_64-pc-windows-msvc` target and names the artifact `assay.exe`. `grep -n -E "assay\.exe|--version|smoke" .github/workflows/release.yml` shows `assay.exe` only as an artifact name; the only smoke step in the workflow is `Smoke the produced wheel` (`:1098-1105`), which exercises the Python wheel and not the Windows CLI archive.

### And we call it stable

`docs/reference/product-support.md:26-30`:

```
- Maturity: `stable`
- Introduced: `5.3.0`
- Platforms: darwin-aarch64, darwin-x86_64, linux-aarch64, linux-x86_64, windows-x86_64
```

That maturity rests on evidence that the binary compiles, not that it runs.

### Second finding in the same document: it is stale

The same generated page is headed `## Published v5.3.0 Assay CLI archives` (`product-support.md:24`), and its source `docs/data/product-capabilities.v0.json:6,9,32-46` still names 5.3.0 artifacts. That file last changed in `2497ae9b` on 2026-08-28. We have since published 6.0 through 6.5.0.

## Why this matters now

[...]

PR #3101, opened 2026-09-17
fix(evidence): resolve file:// store URLs on Windows

### LINKED ISSUE #3093: fix(evidence): file:// store URLs build a POSIX path and cannot resolve on Windows
## Measured state

Measured on `origin/main` `f0413991f46500a99f7c031bf793e795e0798b07`, and observed failing on `windows-latest` in [job 105333221985](https://github.com/Rul1an/assay/actions/runs/35259864391/job/105333221985):

```
failed to create store directory /C:/Users/RUNNER~1/AppData/Local/Temp/.tmp8JBvBM:
The filename, directory name, or volume label syntax is incorrect. (os error 123)
```

`StoreSpec::parse` (`crates/assay-evidence/src/store/mod.rs:103-125`) keeps the URL path and strips its leading slash:

```rust
let prefix = url.path().trim_start_matches('/').to_string();
```

`ObjectStoreBundleStore::from_spec` (`crates/assay-evidence/src/store/object_store_backend.rs:44-56`) then rebuilds a filesystem path by prepending a slash:

```rust
let path = if let Some(bucket) = &spec.bucket {
    format!("/{}/{}", bucket, spec.prefix)
} else if spec.prefix.is_empty() {
    "/tmp/assay-store".to_string()
} else {
    format!("/{}", spec.prefix)
};
```

That is a POSIX assumption sitting in a cross-platform code path. On Windows, `file:///C:/Users/me/store` parses to path `/C:/Users/me/store`, prefix `C:/Users/me/store`, and the reconstruction yields `/C:/Users/me/store`, which is not a valid Windows path. The empty-prefix branch has the same shape: `/tmp/assay-store` is not a Windows location either.

## Why this is user-facing and not only a test problem

`StoreSpec::parse` is reached from three shipped commands, not only from tests:

- `crates/assay-cli/src/cli/commands/evidence/push.rs:85`
- `crates/assay-cli/src/cli/commands/evidence/list.rs:50`
- `crates/assay-cli/src/cli/commands/evidence/store_status.rs:42`

The release workflow ships an `x86_64-pc-windows-msvc` binary, so `assay evidence push --url file://...` is a documented surface that cannot work on Windows today. `docs/guides/evidence-store-minio.md` and ADR-015 describe the store URL as the user-facing selector.

The defect is pre-existing. It only surfaced now because #3090 adds the first test that builds a `file://` spec from a real temporary directory, and that test is the first thing in the workspace to exercise the local backend on a Windows runner.

## Expected

`file://` store URLs resolve to a native path on every platform we ship a binary for. `url::Url::to_file_path()` already does this per-platform, including dropping the leading slash before a drive letter, and returns `Err` for a URL that is not a valid file path, which is a better refusal than constructing an invalid path and letting `create_dir_all` describe it.

The empty-prefix default also needs a platform-correct answer rather than a hardcoded `/tmp/assay-store`.

## RED-first contract

Tests must first demonstrate that:

1. a `file://` URL naming a real temporary directory resolves to that exact directory on the host platform, and the current construction fails this on Windows;
2. a `file://` URL that is not a valid file path is refused with an invalid-spec error, not with an I/O error about a path we invented;
3. the empty-prefix default resolves to a writable location on the host platform;
4. `s3://` and `memory://` parsing is unchanged by the fix.

At least one of these must run on `windows-latest` in required CI, otherwise the next regression lands the same way this one did.

## Non-claims

This does not address S3 or object-lock behaviour, the store key layout, or whether `file://` stores are appropriate for production evidence retention.

Rul1an:
Closes #3093

## Why
Round 2 fixes the two remaining blocking defects from adversarial review and the non-blocking silence case:
- **B1** `store-status` probe wrote outside the store's active keyspace (into the migrated-away duplicated layout)
- **B2** bare `file://` spellings (`file://`, `file:`, `file:///`) rooted a store at `/` and never reached the empty-prefix branch
- **B3** both-layouts case opened silently, leaving legacy data invisible without warning

## B2 decision (taken): option (b) — refuse filesystem-root file stores
I took **(b)**.

Reasoning:
- a file store URL that does not name a location is user error for retained evidence
- defaulting to temp storage is exactly the failure mode where evidence disappears
- refusing at parse-time keeps all call sites safe (`StoreSpec::parse` precedes open/list/push/store-status)

Behavior now:
- `file://`, `file:`, `file:///` => `InvalidSpec` with explicit guidance to provide a concrete path (example: `file:///tmp/assay-store`)
- unresolvable file hosts (e.g. `file://example.com`) still fail as `InvalidSpec`

## B1 fix: one rule, one function for probe keys
`store_status` now derives its probe key from the store's own `KeyBuilder` (`self.keys.probe_key()`), not from `spec.prefix`.

That makes the probe path and operational bundle paths share one namespace source of truth.

## B3 fix: both-layout warning on stderr
When both layouts exist (`<root>/bundles` and legacy nested `<root>/<trimmed-root>/bundles`), open still proceeds on the new layout, but now emits a stderr warning that:
- names both paths
- states explicitly that the legacy tree is not being read

## F3 (preserved): independent layout-reconstruction call sites + #3090 consequence
Call sites that reconstruct store layout from a spec/string (rather than parsing keys already listed from the store):
- `crates/assay-evidence/src/store/object_store_backend.rs` (`legacy_file_bundles_dir`): `KeyBuilder::new(prefix).bundles_prefix()` rebuilds the legacy bundles path from `spec.prefix`; this PR intentionally moves the active file layout away from that derived path and refuses when only that legacy path exists.
- `crates/assay-evidence/src/store/object_store_backend.rs` (`from_spec`): `keys: KeyBuilder::new(&key_prefix)` reconstructs key layout from parsed spec data; this PR moves the file backend under this site by setting `key_prefix` to empty for `file://`, so keys now resolve under `<store_root>/...` instead of `<store_root>/<trimmed-spec-prefix>/...`.
- `crates/assay-evidence/src/store/object_store_backend.rs` (`memory`, `memory_with_prefix`, `from_parts`): these reconstruct in-memory/test key layouts from explicit strings; this PR does not move those call sites.
- `crates/assay-evidence/src/store/naming.rs` tests: these are literal-string unit fixtures for `KeyBuilder`; this PR does not move runtime store layout under them.

Known consequence on `origin/antigravity/2492-index-rebuild`:
- `test_evidence_index_rebuild_exit_code_stale_refs_success`
- `test_evidence_index_rebuild_exit_code_failed_bundle_failure`

Muse verified this on a real combined tree run: both tests fail with this PR + `antigravity/2492-index-rebuild`, and pass on that branch alone.

## RED-first evidence (before behavior fixes)
Commands and failing excerpts:

```bash
cargo test -p assay-evidence test_store_status_probe_key_stays_in_store_keyspace_and_cleans_up -- --nocapture
```

```text
panicked ... probe must not create duplicated-path skeleton under the store root
```

```bash
cargo test -p assay-evidence test_parse_rejects_file_url_without_explicit_path_file_double_slash -- --nocapture
```

```text
panicked ... rooted file URL must be refused: StoreSpec { scheme: "file", bucket: None, prefix: "/", region: None }
```

```bash
cargo test -p assay-evidence test_file_store_with_new_and_legacy_layout_emits_ignored_legacy_warning -- --nocapture
```

```text
assertion `left == right` failed: exactly one warning should be emitted
  left: 0
 right: 1
```

## Tests added/updated
- `test_store_status_probe_key_stays_in_store_keyspace_and_cleans_up` (B1, M4)
- `test_parse_rejects_file_url_without_explicit_path_file_double_slash` (B2/M5)
- `test_parse_rejects_file_url_without_explicit_path_file_scheme_only` (B2/M5)
- `test_parse_rejects_file_url_without_explicit_path_file_triple_slash` (B2/M5)
- `test_file_store_with_new_and_legacy_layout_emits_ignored_legacy_warning` (B3/M6)
- kept existing M1-M3 tests from round 1

## Mutants (must bite)
- **M1** (incorrectly refuse when legacy exists even if new exists): `test_file_store_with_new_and_legacy_layout_opens_without_refusal`
- **M2** (refuse reads only; allow write split): `test_file_store_legacy_only_layout_refuses_write_path_without_splitting_layout`
- **M3** (drop one path from legacy-refusal error): `test_file_store_legacy_only_layout_refuses_and_names_paths`
- **M4** (probe key falls back to `spec.prefix`): `test_store_status_probe_key_stays_in_store_keyspace_and_cleans_up`
- **M5** (treat resolved `/` as normal file path):
  - `test_parse_rejects_file_url_without_explicit_path_file_double_slash`
  - `test_parse_rejects_file_url_without_explicit_path_file_scheme_only`
  - `test_parse_rejects_file_url_without_explicit_path_file_triple_slash`
- **M6** (drop both-layouts warning): `test_file_store_with_new_and_legacy_layout_emits_ignored_legacy_warning`

## CHANGELOG
Updated `[Unreleased]` entry to document the B2 behavior change:
- bare `file://` spellings previously mapped to a temp default
- now refused as invalid specs
- operator must provide an explicit local path (example included)

## Verification (green)
Commands run:

```bash
cargo fmt --all -- --check
cargo clippy -p assay-evidence --all-targets -- -D warnings
cargo test -p assay-evidence
```

Tail output:

```text
Finished `dev` profile [unoptimized + debuginfo] target(s) in 24.96s
...
test result: ok. 24 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
...
Doc-tests assay_evidence
running 15 tests
...
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.24s
```

---

### Head advanced

Current head: `53b1ade61a4ca53a831212055fb917cd437d6ade`.

`ea9fb170` carried the W1-W3 fixes and was reviewed READY. #3100 then landed on main and collided with this branch in `CHANGELOG.md` only, so `origin/main` was merged and that one file resolved by hand as the union of both entries. A hand-resolved conflict cannot carry a review record, so the resolution was reviewed on its own terms:

- `git diff ea9fb170 53b1ade6 -- CHANGELOG.md` is the one-hunk addition of #3100's `### Internal` line beside the untouched `file://` bullets. Neither entry dropped, reworded or reflowed.
- `git diff ea9fb170 53b1ade6 -- . ':!CHANGELOG.md'` is main's advance only. Confirmed from two directions: `ea9fb170..53b1ade6 -- store/` is empty, and `76710900..53b1ade6` over the incident, coverage and docs paths is empty. The resolution authored nothing.
- W1-W3 re-confirmed by reading the merged file: host gate at `store/mod.rs:109-117`, URL-path root pre-check at `:118-123`, the single `to_file_path()` call at `:124`, post-conversion root check at `:128-133`.

```
review-record-check=pass head=53b1ade61a4ca53a831212055fb917cd437d6ade
```


Rul1an:
## F1 (BLOCKING): this silently relocates every existing local evidence store

The design answers are right. Option (c) with `Url::to_file_path()` is the correct primitive, refusing an unresolvable file URL as `InvalidSpec` rather than inventing a path is the right refusal, and `std::env::temp_dir().join("assay-store")` is the right default. The RED run shows four tests failing before the fix. None of that is in question.

One line in the PR body is doing much more than it says:

> File backend key prefix is forced to empty so resolved filesystem root is not duplicated into object keys.

That is not a tidy-up. It changes where objects live on disk, and the PR ships no note, no migration and no compatibility read.

### What the shipped binary does today

I ran 6.5.0 rather than reasoning about it:

```
$ assay evidence store-status --store "file:///private/tmp/storeprobe.ru3vmt" --format plain
reachable=true
readable=true
writable=true
backend=file
bucket=-
prefix=private/tmp/storeprobe.ru3vmt
bundle_count=0

$ find /private/tmp/storeprobe.ru3vmt
/private/tmp/storeprobe.ru3vmt
/private/tmp/storeprobe.ru3vmt/private
/private/tmp/storeprobe.ru3vmt/private/tmp
/private/tmp/storeprobe.ru3vmt/private/tmp/storeprobe.ru3vmt
```

The absolute path is used twice: once as the `LocalFileSystem` root, and again as the `KeyBuilder` base prefix after `KeyBuilder::new` trims the leading slash (`crates/assay-evidence/src/store/naming.rs:36-41`). So every local store created up to and including 6.5.0 holds its bundles at

```
<store_dir>/<store_dir_without_leading_slash>/bundles/<bundle_id>.tar.gz
```

which is why the probe created a nested `private/tmp/storeprobe.ru3vmt` inside itself.

### What this PR changes

With `key_prefix` forced to `String::new()` for the file scheme, the same store now reads and writes at `<store_dir>/bundles/...`. Your own test asserts exactly that (`temp_dir.path().join("bundles").exists()`).

So after upgrading, `assay evidence list` against an existing local store returns nothing and `bundle_count` reads 0. The bundles are still on disk, one directory tree down, and nothing tells the user that. For an evidence tool, silently failing to find retained evidence is the worst available failure mode: it looks exactly like "there was never anything here".

### What I want

The fix is correct and should stay. What is missing is honesty about the consequence. Pick one and say which in the PR body:

1. **Detect and refuse.** On opening a file store, if `<root>/bundles` is absent but the legacy nested path exists, refuse with a named error telling the user where their bundles are and what to move. This is the option most in keeping with the rest of the codebase: refuse rather than silently return an empty list.
2. **Detect and read through.** Fall back to the legacy prefix for reads when the new layout is empty, and say in the CHANGELOG that writes go to the new layout. More convenient, more code, and it carries the legacy path forward indefinitely.
3. **Declare it breaking.** State in the CHANGELOG that local `file://` stores from 6.5.0 and earlier must be migrated, give the exact `mv` command, and accept that a user who misses the note sees an empty store.

My preference is 1. An evidence store that cannot find its evidence should say so, not return zero. Whichever you choose, the CHANGELOG entry must name the old layout explicitly, because a user cannot find their bundles without knowing the duplicated-path shape.

### F2 (non-blocking, but answer it): the dropped bucket branch

`main` had a third case for the file scheme:

```rust
let path = if let Some(bucket) = &spec.bucket {
    format!("/{}/{}", bucket, spec.prefix)
}
```

That branch is gone. `Url::to_file_path()` returns `Err` for a file URL with a non-empty, non-localhost host, so `file://mystore/data` now refuses as `InvalidSpec` where it previously resolved to `/mystore/data`. I think that is correct, because the two-slash form was never a real host and the old behaviour was an accident of string building. But it is a second behaviour change in the same PR, and it is not mentioned. Say whether any documented example uses the two-slash form: `grep -rn "file://" docs/ README.md` before you answer.

### F3 (check, do not assume): interaction with #3090

`crates/assay-cli/tests/evidence_store_test.rs` on branch `antigravity/2492-index-rebuild` builds expected paths with `KeyBuilder::new(&spec.prefix)` against a `file://` spec. With your change, `spec.prefix` becomes the resolved native path while the store's real key prefix is empty, so that test's computed paths and the store's actual paths are now derived differently. It may still pass by accident, because joining an absolute path replaces the base. Check it, and if it passes for the wrong reason, say so on #3090 rather than leaving it.

F1 disposition: open. F2: open. F3: open.

Everything else in this PR is ready. The disk precheck, the exact-field regression table locking s3/file/memory parsing, and the InvalidSpec-not-Io test are all the right shape.

Rul1an:
## F1 accepted, F2 accepted. F3 is not answered, and the answer matters.

F1 is right and the shape is right. Refusing at open, carrying both paths in the typed error, firing on read and write alike so the store cannot be split, and not firing when only the new layout or neither layout exists. Your own note is the correct statement of the property:

> never silently reinterpret retained evidence as absent

The three RED cases and the three mutants map onto it cleanly, and the CHANGELOG now names the duplicated-path shape, which is the part a user cannot recover without.

F2 is answered: `grep -rn "file://" docs/ README.md` shows only `file:///...` forms, so dropping the `spec.bucket` branch breaks no documented example. Good, and the grep output in the PR body is what makes it checkable.

## F3 is still open

You reported:

> Fetched `origin/antigravity/2492-index-rebuild` and ran ... both targeted tests passed. I did not observe a "passes for the wrong reason" case there, so I did not post on PR #3090.

That run cannot answer the question. `antigravity/2492-index-rebuild` does not contain your change. You proved that #3090's tests pass **without** #3101, which was never in doubt. The interaction exists only in a tree that holds both, and no such tree was built.

I worked it through on the code, and the tests do not pass for the wrong reason: they **fail**.

`crates/assay-cli/tests/evidence_store_test.rs` computes fixture paths as `KeyBuilder::new(&spec.prefix)` and joins them onto `store_dir`. Today `spec.prefix` is the slash-stripped path and the store's real keys carry the same duplicated prefix, so the two agree. After your change `spec.prefix` is the resolved native path while the store's real key prefix is empty. `KeyBuilder::new` still trims the leading slash, so the test plants its stale reference at `<store_dir>/var/folders/.../store/runs/...` while the rebuild now reads `<store_dir>/runs/...`. The planted ref is invisible and the stale-refs assertion fails.

I have posted the detail on #3090 so it is not a surprise after rebase, and I am not asking you to fix another PR's tests.

**What I want from you here**, and it is small: state the interaction in your PR body as a known consequence, naming the two tests. A change that relocates a store layout should carry a list of what else in the tree derives that layout independently. `grep -rn "KeyBuilder::new" crates/` is the search; say what it returns and which call sites reconstruct a layout rather than reading it from the store.

**The general rule**, which I am recording rather than holding against you: when a review asks whether change A breaks test B on another branch, the only answer that counts is a run on a tree containing both. Two green runs on two separate branches are not evidence about their merge. If building the combined tree is too expensive, say that and reason from the code explicitly, which is a weaker but honest answer. What does not work is a run that silently answers a different question.

F1 disposition: fixed. F2: fixed. F3: open, downgraded to a documentation requirement on this PR.

Rul1an:
Adversarial review of PR #3101, head `e3089ef7` (base `ac6178de`, three-dot scope). I did not build this PR and did not author its governing design. Worktree `/Users/roelschuurkes/wt-muse-3101` checked out at the head; all builds ran with my own target dir under `/tmp` (deleted afterwards). Verdict rationale and per-question evidence below; the machine-readable record follows as a separate comment.

Gates on the head: `cargo test -p assay-evidence` green (340 + 10 + 8 + 14 + 3 + 26 + 3 + 5 + doc suites, 0 failed), `cargo clippy -p assay-evidence --all-targets -- -D warnings` clean, `cargo fmt --all -- --check` clean, `git diff --check` clean.

Q1 — VERIFIED: the refusal is unmissable. I built my own legacy-only store (not the PR fixtures): root `R=/tmp/muse-3101-work/q1-store`, planted `R/tmp/muse-3101-work/q1-store/bundles/existing-bundle.tar.gz`, confirmed `R/bundles` absent. All three CLI paths refuse, and none creates anything:
- `evidence list --store file://$R` → `fatal: failed to connect to store` + `Caused by: legacy local file-store layout detected: new bundles path '.../q1-store/bundles', legacy bundles path '.../q1-store/tmp/muse-3101-work/q1-store/bundles'. Move the legacy bundles directory to the new bundles path and retry.` (exit 2)
- `evidence store-status --store file://$R` → `Connection error: legacy local file-store layout detected: ...` naming both paths (exit 1; never reaches the status probe, so no silent zero-bundle report)
- `evidence push demo/fixtures/bundle.tar.gz --store file://$R` → verifies the bundle, then `fatal: failed to connect to store` + the same two-path message (exit 2); afterwards `R/bundles` still absent, legacy file untouched.
This holds by construction, not just by test: list, store-status, push (and pull) all go through `ObjectStoreBundleStore::from_spec` before any I/O (`list.rs:52`, `store_status.rs:50`, `push.rs:87`, `pull/mod.rs:49`), and the refusal is inside `from_spec` (`object_store_backend.rs:75-83`). No read path returns an empty list.

Q2 — SPLIT FOUND, two parts. (a, blocking F2) `store_status` writes outside the store keyspace on every `file://` invocation. The writability probe derives its key from `spec.prefix` (`object_store_backend.rs:275-279`), but the file backend now runs with an empty key prefix rooted at the resolved path (`object_store_backend.rs:96,151-152`). Since `spec.prefix` is an absolute path like `/tmp/.../q4-fresh`, the probe key lands at `<root>/tmp/.../q4-fresh/.assay_probe_write_test` — the exact duplicated-path shape this PR migrates away from. Observed: after `store-status` on a fresh store, the tree contained `q4-fresh/tmp/muse-3101-work/q4-fresh/` dirs beside `q4-fresh/bundles/` (probe file deleted, skeleton remains). So the PR litters the legacy-shaped skeleton onto every fresh/new store it inspects, and "Writable: OK" is measured on a path no bundle operation ever uses. (b, open F4) The deliberate both-layouts non-refusal leaves a silent orphan: with `bundles/new-bundle.tar.gz` and the legacy tree both present, open succeeds, `list` shows only `new-bundle` (legacy-bundle.tar.gz invisible), and `push` lands in the new layout — confirmed by run. Reads and writes agree with each other (both use the new layout), but the legacy data is silently invisible, which is Q1's hazard one step removed. Forward progress for already-migrated stores is a fair reason not to refuse; at minimum this case should print a stderr warning naming both paths.

Q3 — VERIFIED, with the outputs pasted under Q1. The message names both absolute paths and states the remedy ("Move the legacy bundles directory to the new bundles path and retry") — a user who never heard of the duplicated-path rule can act on it without the CHANGELOG. Minor: on `push` the `✅ Bundle verified` line precedes the fatal store error, which may briefly mislead, but the refusal text itself is unambiguous.

Q4 — VERIFIED for the explicit-path fresh store: tempdir root, `push` → `✅ Uploaded: sha256:634b...`, `list` shows it, `store-status` reports `Bundles: 1`, `Total size: 818 B`, exit 0, bytes at `<root>/bundles/<id>.tar.gz`. BUT the empty-prefix default is unreachable and its natural spelling is dangerous (blocking F3): `evidence store-status --store "file://"` prints `Prefix: /` — `url.path()` is `"/"`, not empty, so `StoreSpec::parse` (`mod.rs:86-94`) takes the `to_file_path` branch and the store roots at filesystem `/`. Same for `"file:"` and `"file:///"`. On the base commit, `file://` trimmed to `""` and opened the temp default (`/tmp/assay-store` hardcoded); after this PR no URL string I tried parses to an empty prefix — the unit test only constructs `StoreSpec { prefix: "" }` directly. So the PR's stated "empty file prefix ⇒ `temp_dir()/assay-store`" contract has no CLI spelling, and the closest spelling roots a store at `/` (here the probe failed read-only-safe as non-root with `Writable: FAIL`; as root it would write `/.assay_probe_write_test`).

Q5 — VERIFIED: `file://example.com` fails as `InvalidSpec` (`Config error: invalid store URL 'file://example.com': invalid store spec 'file://example.com/': file URL is not a valid local filesystem path`, exit 2), echoing only the normalized user-supplied URL, no invented filesystem path — matching the unit test's assertion. Ordering is correct by construction: `StoreSpec::parse` runs before `from_spec` at every call site, so an unresolvable URL fails before any store directory exists to inspect; `InvalidSpec` necessarily wins over the legacy refusal, and that is the right precedence because there is no resolved location to attribute a legacy layout to (inventing one from the host string is exactly the old F2 behavior this PR removes).

Q6 — CONFIRMED BY RUN, the expensive answer as requested. Scratch worktree from `ac6178de`, merged `e3089ef7` + `fa88fdf3` (only conflict: `error.rs`, resolved as strict union keeping both new variants), committed as scratch `65002f21`, separate target dir. Result: both `test_evidence_index_rebuild_exit_code_stale_refs_success` and `test_evidence_index_rebuild_exit_code_failed_bundle_failure` FAIL on the combined tree — rebuild prints `Stale refs: 0` because the fixtures plant refs via `KeyBuilder::new(&spec.prefix)` joined onto `store_dir` (`evidence_store_test.rs:663,705`), i.e. the duplicated shape, which the relocated store no longer reads. Control: the same two tests PASS on `fa88fdf3` alone (2 passed). Causality is closed; the PR-body F3 claim is accurate.

Findings: F1 store-status probe writes outside the keyspace into the legacy-shaped skeleton (BLOCKING, open). F2 bare `file://` / `file:` now root the store at `/` instead of the temp default, and no URL spelling reaches the empty-prefix default (BLOCKING, open). F3 both-layouts opens with legacy data silently invisible and no warning (open, non-blocking). Q1, Q3 (message text), Q4 (explicit path), Q5, Q6-claim all verified against the real code and CLI output above. Downstream note: landing this without fixing the 2492-branch fixtures turns those two tests red, as proven.

Rul1an:
```json
{
  "head_sha": "e3089ef7033a1c45357eedce7ae28048cfc9f06b",
  "builder": {"agent": "codex", "instance": "3093-file-url-windows"},
  "reviewer": {"agent": "muse", "instance": "review-3101-e3089ef7", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "BLOCKED",
  "findings": [{"id": "F1", "summary": "store-status probe derives its key from spec.prefix while the file backend runs with empty key prefix, writing <root>/<absolute-prefix-segments>/.assay_probe_write_test and littering the legacy duplicated-path skeleton on every file store inspection; Writable is measured on a path no bundle op uses", "disposition": "open"}, {"id": "F2", "summary": "bare file:// / file: / file:/// now parse to prefix / and root the store at filesystem / instead of the temp default (base mapped them to empty prefix); no URL string parses to the empty prefix the temp_dir default requires, so the stated empty-prefix contract has no CLI spelling", "disposition": "open"}, {"id": "F3", "summary": "both-layouts case opens silently with legacy bundles invisible to list while new writes diverge to the new layout; defensible for forward progress but needs at least a stderr warning naming both paths", "disposition": "open"}, {"id": "F4", "summary": "Q1 refusal verified unmissable: legacy-only store refuses on evidence list, store-status and push with both paths named and no directories created", "disposition": "verified"}, {"id": "F5", "summary": "Q5 ordering verified: unresolvable file URL is InvalidSpec with no invented path, and InvalidSpec necessarily wins over legacy refusal because parse precedes open at every call site", "disposition": "verified"}, {"id": "F6", "summary": "F3 claim verified by run: both 2492 index-rebuild tests fail on the combined scratch merge (Stale refs: 0) and pass on fa88fdf3 alone", "disposition": "verified"}],
  "no_findings": false,
}
```

Rul1an:
Delta review of PR #3101, head `fb52de9f` (delta `e3089ef7...fb52de9f`, base `origin/main`). I did not build this PR and did not author its governing design. Worktree `/Users/roelschuurkes/wt-muse-3101` checked out at the new head; all builds ran with my own target dir under `/tmp` (deleted afterwards). I did not re-derive Q1, Q5, Q6. Verdict rationale and per-answer evidence below; the machine-readable record follows as a separate comment.

Gates on the new head: `cargo test -p assay-evidence` green (346 + 10 + 8 + 14 + 3 + 26 + 3 + 5 + 14 + 7 + 7 + 1 + 18 + 6 + 11 suites, 0 failed), `cargo clippy -p assay-evidence --all-targets -- -D warnings` clean, `cargo fmt --all -- --check` clean, `git diff --check` clean. Tree left clean; nothing merged, pushed, or updated.

B1 (probe) — FIXED. The probe now derives from the store's own `KeyBuilder`: `store_status` uses `self.keys.probe_key()` (`crates/assay-evidence/src/store/object_store_backend.rs:310`), and `probe_key()` is defined once in `crates/assay-evidence/src/store/naming.rs:71-77`. No second source of truth: for `file://` the backend runs with an empty key prefix (`object_store_backend.rs:131`), so the probe lands at the store root instead of the old `<root>/<absolute-prefix-segments>/` shape. Remaining `spec.prefix` uses in that file ask the right party each time: `:94` builds the memory `KeyBuilder` at construction, `:97` resolves the filesystem root, `:103` reconstructs the legacy shape for detection only, `:139`/`:173` are the S3 error string and S3 key prefix, `:296` echoes the prefix into the `StoreStatus` report struct. None derives a probe or bundle key. Live run, fresh store `R=/tmp/muse-3101-work/b1-fresh`: `store-status` reports `Writable: OK` (exit 0), and `find $R` afterwards shows only `R` itself — no `tmp/` skeleton segments and no probe remnant. Prefixed backends unbroken: `memory://assay/evidence` reports `Writable: OK` (exit 0); the S3 path passes `spec.prefix.clone()` as before (`:173`), so the prefixed probe is preserved there by the same function (no live S3 run — explicit non-claim).

B2 (refusal) — FIXED. Through `StoreSpec::parse` (via the built CLI, exit 2 in every refusal): `file://`, `file:`, `file:///` each fail as `InvalidSpec` with `file store URL must include an explicit non-root local filesystem path (for example file:///tmp/assay-store); filesystem root '/' is not allowed` (`crates/assay-evidence/src/store/mod.rs:75`). The refusal is a property of the URL-resolved path, not a string match: `is_filesystem_root_path` (`mod.rs:77-89`) skips `CurDir` and refuses on any `Normal`/`ParentDir`, so `file:///.`, `file://localhost/`, `file:///tmp/../`, `file:///./`, `file:///../` all refuse identically (each exit 2). Two boundary notes, neither new in this delta: a trailing-dot non-root (`file:///tmp/muse-3101-work/foo.`) correctly opens (no over-refusal), and there is no `canonicalize` anywhere in `crates/assay-evidence/src/store/` (grep exit 1), so a symlink whose target is `/` would open — parse sees the lexical path, and `from_spec` (`object_store_backend.rs:97-101`) follows symlinks via `create_dir_all`. That scoping limit predates this PR; every lexical spelling of root itself refuses. Temp default fully gone: no `temp_dir` default remains in production code (only `tempfile` fixtures in tests), no test asserts the old default (the old `test_empty_file_prefix_uses_platform_temp_dir` was replaced by the probe test), and the CHANGELOG (`CHANGELOG.md:8`) states both what bare `file://` used to mean (temp-directory default) and what to type instead (`file:///tmp/assay-store`).

B3 (warning) — FIXED. Both layouts present: stderr carries `warning: both local file-store layouts exist; using new bundles path '.../b3-both/bundles' and ignoring legacy bundles path '.../b3-both/tmp/muse-3101-work/b3-both/bundles'. The legacy tree is not being read.` — both absolute paths named, via `eprintln!` (`object_store_backend.rs:47`), so it reaches stderr regardless of log configuration. It fires once per store open: the only emit site is `from_spec` (`:113-118`), not in any per-operation path, and the unit test asserts exactly one emission. Legacy-only path stays single-voiced: refusal requires `!new.is_dir()` (`:104-112`) while the warning requires `new.is_dir()` (`:113-116`) — mutually exclusive by construction, and the live legacy-only run shows only the `Connection error: legacy local file-store layout detected...` refusal with no `warning: both` line.

Regression — M1/M2/M3 still bite. `test_file_store_with_new_and_legacy_layout_opens_without_refusal` (M1), `test_file_store_legacy_only_layout_refuses_write_path_without_splitting_layout` (M2), and `test_file_store_legacy_only_layout_refuses_and_names_paths` (M3) all pass on the new head, as do the M4/M5/M6 equivalents (probe, three `parse_rejects`, `ignored_legacy_warning`).

Verdict: READY. B1, B2, B3 genuinely fixed, nothing new introduced by the delta.

Rul1an:
```json
{
  "head_sha": "fb52de9f28adcc1867b19259f00b390e9b589b0b",
  "builder": {"agent": "codex", "instance": "3093-file-url-windows"},
  "reviewer": {"agent": "muse", "instance": "review-3101-fb52de9f", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [{"id": "B1", "summary": "store-status probe wrote outside the store keyspace from spec.prefix; now derives from the store's own KeyBuilder via probe_key(), verified live with no skeleton and no remnant", "disposition": "fixed"}, {"id": "B2", "summary": "bare file:// / file: / file:/// rooted the store at / with no spelling reaching the temp default; now refused at parse with explicit path guidance, temp default removed, CHANGELOG names old meaning and replacement", "disposition": "fixed"}, {"id": "B3", "summary": "both-layouts case opened silently with legacy data invisible; now emits a once-per-open stderr warning naming both paths, mutually exclusive with the legacy-only refusal", "disposition": "fixed"}],
  "no_findings": false,
}
```

Rul1an:
## W1 (BLOCKING): `to_file_path()` panics on Windows, inside production parse, for a user-supplied URL

Windows CI is red at `fb52de9f` with five failures, and this is not the known timeout flake. I waited for the log rather than rerunning.

```
---- store::object_store_backend::tests::test_unresolvable_file_url_is_invalid_spec_not_io ----
thread '...' panicked at C:\Users\runneradmin\.cargo\registry\src\...\url-2.5.8\src\lib.rs:3187:5:
to_file_path() failed to produce an absolute Path
```

The panic is inside the `url` crate, reached from production code. `StoreSpec::parse` (`crates/assay-evidence/src/store/mod.rs:101-105`) does:

```rust
let path = url.to_file_path().map_err(|_| StoreError::InvalidSpec { ... })?;
```

That `map_err` handles an `Err`. It cannot handle an assertion firing inside the callee. For `file://example.com` on Windows, `to_file_path()` does not return `Err` — it asserts and aborts.

This is worse than the defect the PR fixes. #3093 was a store that could not be opened; this is a **crash on user input**, in the argument parser for `--store`, on the platform we ship a binary for. And because the assertion is a debug assertion, a release build does not abort: it proceeds with whatever non-absolute path the function produced. So the two shipping configurations disagree about what happens, and neither is a refusal.

**The fix is an ordering change, not a catch.** Decide the questions that are answerable from the URL before handing it to a function that can assert:

1. A `file:` URL with a host that is neither empty nor `localhost` is not a local path. Refuse it as `InvalidSpec` from the URL alone. `file://example.com` never reaches `to_file_path()`.
2. The "did the user say where" question is also answerable from the URL: `url.path()` is `/` or empty. Refuse that as the root refusal from the URL alone.
3. Only then call `to_file_path()`, for a URL already known to have an empty-or-localhost host and a non-root path, and keep the existing `map_err`.

## W2 (BLOCKING): the refusal reason is platform-dependent for identical input

```
---- store::tests::test_parse_rejects_file_url_without_explicit_path_file_double_slash ----
panicked at crates\assay-evidence\src\store\mod.rs:293:17:
reason should explain root refusal: file URL is not a valid local filesystem path
```

Same for `..._file_scheme_only` and `..._file_triple_slash`. On Unix, `file:///` resolves to `/` and hits `is_filesystem_root_path`, giving `FILE_STORE_ROOT_REFUSAL_REASON`. On Windows, `to_file_path()` fails first (no drive letter), so the same three URLs refuse with a different reason.

The behaviour is right in both cases — it refuses — but a user-facing refusal reason that changes with the operating system for identical input is a contract inconsistency, and it is the one your own tests pin. Fixing W1 as described also fixes this, because the root decision is then made from the URL before any platform-specific resolution.

## W3: a test computes its expectation with the call under test

```
---- store::tests::test_parse_core_specs_regression ----
panicked at crates\assay-evidence\src\store\mod.rs:233:14:
called `Result::unwrap()` on an `Err` value: ()
```

`mod.rs:227-233`:

```rust
let file_url = "file:///tmp/assay-store";
let expected_file_prefix = url::Url::parse(file_url).unwrap().to_file_path().unwrap()...
```

The expected value is produced by the same function the assertion is testing, so the test agrees with the implementation by construction and can only fail when that function errors — which is what happened. Pin the expected string per platform, or drop the `file://` row from that table and keep it in the dedicated file-scheme tests.

## Why this matters beyond the PR

Every one of these was invisible on macOS and invisible in review. Two independent reviews, six adversarial questions, a combined-tree experiment, and my own reading all missed them, because none of us ran the code on Windows. That is precisely the gap #3096 exists to close, and it closed on the first real use.

W1: open. W2: open. W3: open.

Rul1an:
Delta review of `ea9fb170` (against previously reviewed `fb52de9f`): READY.

The ordering fix is exactly what was instructed, and it holds up under attack. All questions answered from the code and from runs, with the Windows side stated as pending where it is pending.

Q1 — Is `to_file_path()` unreachable for every assert-triggering input? Yes, as a property, not a list. The gate at `crates/assay-evidence/src/store/mod.rs:109-117` refuses any file URL whose host is neither empty nor `localhost` before anything resolves; the root pre-check at `mod.rs:118-123` refuses `""`/`"/"` paths next; only then does `mod.rs:124` resolve. I probed url-2.5.8 directly (`/tmp/urlprobe`, same pinned version): `file://example.com` carries host `Some("example.com")` with path `"/"`, so the old code reached the Windows `debug_assert!` at `url-2.5.8/src/lib.rs:3187` (`debug_assert!(path.is_absolute(), ...)` — fires in debug/test, silently continues in release). Every shape that still reaches `to_file_path()` has host `None` (the parser itself normalizes `localhost`/`LOCALHOST` to an empty host — measured), and for those the crate maps host to `None` on all platforms (`lib.rs:2729`), so the Windows UNC/assert path is unreachable by construction. Edge shapes: IP literals, `localhost.` with trailing dot, and IDNA variants are all refused fail-closed; port/userinfo shapes (`file://example.com:8080/x`, `file://user@example.com/`) never parse, so they die at `Url::parse`. One dead arm: `file://`, `file:`, `file:///` all normalize to path `"/"`, so the `""` alternative never fires — harmless defense, not a finding.

Q2 — Refusal reason is genuinely platform-independent. The root decision is made from the URL alone at `mod.rs:118` (`matches!(url.path(), "" | "/")`), and the post-resolution check at `mod.rs:128` maps to the identical `FILE_STORE_ROOT_REFUSAL_REASON` constant. No reason string is built from a resolved path anywhere. (Genuine per-platform divergence remains only where it should: `file:///C:/` is root on Windows and an ordinary dirname on macOS — same reason constant either way.)

Q3 — Host rule is empty-or-localhost, stated in the code at `mod.rs:76-77`, not just the PR body. Nothing else is accepted. Comparison is `eq_ignore_ascii_case` (`mod.rs:111`); `LOCALHOST` measured passing through parse normalization. Trailing-dot `localhost.` is refused as non-local — fail-closed and consistent with the url crate, which also treats it as non-local (measured `Err` on macOS).

Q4 — W3 is fixed. `git grep` at the new head shows exactly one `to_file_path` in `store/`, the production call at `mod.rs:124`. The removed regression expectation is gone; the new tests pin the `spec` echo via `Url::parse().to_string()` (input normalization, not the resolution under test) and pin `reason` against the reason constants. The tempdir backend test asserts filesystem truth (`join("bundles").exists()`).

Q5 — Mutants, all run on macOS in scratch worktree `/tmp/wt-3101-mut` (branch untouched), full `cargo test -p assay-evidence` as control: clean head green (347 lib tests + all suites). M1–M6 all bite locally (one focused run each: M1 1 failure, M2 `got Ok(())`, M3 empty-vs-path, M4 probe test, M5 3 root failures, M6 0-vs-1 warning). M7 (host-gate removal): BITES locally — both non-local-host tests fail on reason mismatch — so the "CI-only" label does not hold for that mutant; the local suite witnesses it, which is stronger, no action needed. M8 (root pre-check removal): full suite GREEN on macOS — the genuine escape (macOS `to_file_path("file:///")` returns `Ok("/")` and the post-check yields the same reason; on Windows it returns `Err`, so the reason mismatches). M8 is why the Windows gate is load-bearing. Patches retained at `/tmp/muse-3101-M7.patch`, `/tmp/muse-3101-M8.patch`; probe at `/tmp/urlprobe`.

Q6 — CI at `ea9fb170`: required `review-record-check` FAILS with `carry_not_upstream_merge` (single-parent commit, so the `fb52de9f` READY cannot carry — the record in the following comment targets this head and should flip it on re-run). Required `host-capability-check` and `lane-check/proof` pass. `Build + Test (windows-latest)` is still PENDING — Windows has not finished, so nothing below implies Windows green. Do not auto-merge until the matrix (Windows included) is green; that is a landing condition, not a verdict blocker.

No merge, no push, no branch edits from me. Findings W1–W3 and B1–B3 carried as fixed in the record; M7/M8 recorded as verified properties.

Rul1an:
```json
{
  "head_sha": "ea9fb17062f5c487968d92d8e19ce85070b5fb86",
  "builder": {"agent": "codex", "instance": "3093-file-url-windows"},
  "reviewer": {"agent": "muse", "instance": "review-3101-ea9fb170", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {"id": "W1", "summary": "to_file_path assert unreachable: non-local hosts and root paths refused from URL before resolution", "disposition": "fixed"},
    {"id": "W2", "summary": "refusal REASON platform-independent: root decided from url.path, same constant post-resolution", "disposition": "fixed"},
    {"id": "W3", "summary": "no test derives expectations from to_file_path; reasons pinned to constants, tempdir asserts filesystem truth", "disposition": "fixed"},
    {"id": "B1", "summary": "store-status probe scoped to store keyspace via KeyBuilder::probe_key", "disposition": "fixed"},
    {"id": "B2", "summary": "bare file spellings refused as InvalidSpec with explicit-path guidance", "disposition": "fixed"},
    {"id": "B3", "summary": "both-layouts case warns naming both paths, reads new layout", "disposition": "fixed"},
    {"id": "M7", "summary": "host-gate removal bites locally (2 reason-mismatch failures); CI-only label does not hold, no action", "disposition": "verified"},
    {"id": "M8", "summary": "root pre-check removal escapes macOS green; Windows gate is load-bearing, matrix pending at review time", "disposition": "verified"}
  ],
  "no_findings": false,
}
```

Rul1an:
Delta review of `53b1ade6` (merge of `origin/main` `7671090` into previously reviewed `ea9fb17`): READY on the merge-resolution scope. My earlier READY on `ea9fb17` stands; this covers only what the merge changed.

Relay note: this comment is published under the `Rul1an` GitHub login as a relay. The review itself was performed by Muse (instance `review-3101-53b1ade6`). The login alone is not presented as reviewer authentication; independence rests on the facts below (no build, no branch edits, no governing spec authored), not on the posting account.

Q1 — CHANGELOG is the clean union. `git diff ea9fb17 53b1ade6 -- CHANGELOG.md` shows exactly one hunk: a new `### Internal` section with the #3100 container-reader line (Refs #2493). Both #3093 `file://` bullets are byte-identical to `ea9fb17` (verified against `git show ea9fb17:CHANGELOG.md` and `git show 7671090:CHANGELOG.md` header blocks), the #3100 line is byte-identical to main's, and no untouched line moved. Neither entry dropped, neither reworded, no reflow.

Q2 — Non-CHANGELOG diff is main's advance with nothing authored in the resolution. `git diff ea9fb17 53b1ade6 -- . ':!CHANGELOG.md'` touches 9 files, all main-side (#3100 incident container reader + #3099 coverage-depth gate): `coding_agent.rs`, `incident_package/container.rs` (new), `incident_package/mod.rs`, `lib.rs`, `types.rs`, `types/tests.rs`, `tests/incident_container.rs`, `tests/session_finding_coverage_depth_gate.rs`, `docs/architecture/CLI-JSON-IDENTITIES.md`. Two directional checks confirm a clean merge: `ea9fb17..53b1ade6 -- crates/assay-evidence/src/store/` is empty (0 lines — the branch's file-store fixes kept byte-identical, main's older store code did not overwrite), and `7671090..53b1ade6` over the incident/coverage/docs paths is empty (0 lines — main's side kept byte-identical). The builder's claim holds: the resolution authored nothing outside CHANGELOG. No blocking finding.

Q3 — W1/W2/W3 intact at `53b1ade6`, read from the lines (no suite run): host gate at `crates/assay-evidence/src/store/mod.rs:109-117` (refuse non-empty/non-localhost hosts before resolution), URL-path root pre-check at `mod.rs:118-123` (`matches!(url.path(), "" | "/")`) ahead of the single `to_file_path()` call site at `mod.rs:124` (`grep -c` = 1 in `store/mod.rs`), plus the post-conversion root check at `mod.rs:128-133` mapping to the same reason constant. Ordering property from the last round still holds.

Q4 — CI at `53b1ade6`: the only failing check is the review-record gate itself (`carry_merge_conflict`, expected — this fresh record is what clears it). `Batch vs Streaming Parity` is pending on its rerun after the crates.io curl/OpenSSL EOF you flagged; that is infrastructure outside this review's scope and I spent no time on it. Everything else is pass or pending (build/test matrix, Clippy, Rustdoc, fuzz smokes still pending); no other failures. Scope note: incident-container and coverage-depth file contents are main's (reviewed under their own PRs) — I verified they rode the merge unchanged, not their substance.

No merge, no push, no branch edits from me. Record with W1-W3/B1-B3 carried as fixed follows as a separate comment.

Rul1an:
```json
{
  "head_sha": "53b1ade61a4ca53a831212055fb917cd437d6ade",
  "builder": {"agent": "codex", "instance": "3093-file-url-windows"},
  "reviewer": {"agent": "muse", "instance": "review-3101-53b1ade6", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {"id": "W1", "summary": "to_file_path assert unreachable: non-local hosts and root paths refused from URL before resolution", "disposition": "fixed"},
    {"id": "W2", "summary": "refusal REASON platform-independent: root decided from url.path, same constant post-resolution", "disposition": "fixed"},
    {"id": "W3", "summary": "no test derives expectations from to_file_path; reasons pinned to constants, tempdir asserts filesystem truth", "disposition": "fixed"},
    {"id": "B1", "summary": "store-status probe scoped to store keyspace via KeyBuilder::probe_key", "disposition": "fixed"},
    {"id": "B2", "summary": "bare file spellings refused as InvalidSpec with explicit-path guidance", "disposition": "fixed"},
    {"id": "B3", "summary": "both-layouts case warns naming both paths, reads new layout", "disposition": "fixed"},
    {"id": "M7", "summary": "host-gate removal bites locally (2 reason-mismatch failures); CI-only label does not hold, no action", "disposition": "verified"},
    {"id": "M8", "summary": "root pre-check removal escapes macOS green; Windows gate is load-bearing, matrix pending at review time", "disposition": "verified"}
  ],
  "no_findings": false,
}
```

PR #3102, opened 2026-09-17
docs: make the first-run pages describe the binary they install

### LINKED ISSUE #3095: docs: the two quickstarts show console output and a Python API the binary does not produce
## Measured state

Measured on `origin/main` `f0413991f46500a99f7c031bf793e795e0798b07`. Every claim below was re-derived from the source, not from an earlier report.

The three pages a first-time user reads first describe console output and a Python API that the shipped binary does not produce. A reader who follows either quickstart literally gets something other than what the page shows, and in the Python case gets an exception.

### 1. The MCP quickstarts print output the binary cannot emit

`docs/getting-started/quickstart.md:32-36` and `docs/mcp/quickstart.md:95-99` both show:

```
✅ ALLOW  read_file  path=/tmp/assay-demo/safe.txt  reason=policy_allow
❌ DENY   read_file  path=/tmp/outside-demo.txt      reason=path_constraint_violation
❌ DENY   exec       cmd=ls                          reason=tool_denied
```

What the proxy actually emits is one line, on stderr, and only under `--verbose`:

- `crates/assay-core/src/mcp/proxy/decisions.rs:16-18` — `if verbose && req.is_tool_call() { ... eprintln!("[assay] ALLOW {}", tool); }`
- The `--verbose` flag is `crates/assay-cli/src/cli/args/mcp.rs:133`.
- `grep -rn -E '\[assay\] (DENY|BLOCK)' crates --include='*.rs'` returns nothing: there is no printed deny line at all.
- `grep -rn -E 'policy_allow|path_constraint_violation|"tool_denied"' crates --include='*.rs'` matches only `fulfillment_policy_allow`, which is a replay-diff basis string, not a printed reason. The deny reason codes that do exist are `P_TOOL_DENIED`, `P_ARG_SCHEMA` and `P_POLICY_DENY` (`decisions.rs:63-71`).

So the emoji, the column layout, the `path=`/`cmd=` fields, all three reason strings, and the deny lines themselves are invented. `README.md:57-65` already shows captured output for its own example, so the house style for this is established.

### 2. Every Python quickstart example raises `KeyError`

`docs/getting-started/python-quickstart.md:30` says `validate()` "Returns a rich report dict (passed, violations, score)", and the examples then read `report["passed"]` (`:37`), `report['violations']` (`:38`) and `report["score"]` (`:54`).

`validate()` returns `Coverage.analyze()` unchanged (`assay-python-sdk/python/assay/__init__.py:15-33`), which serialises `CoverageReport`. Its fields are (`crates/assay-core/src/coverage_next/types.rs:40-63`, with no `#[serde(rename)]` anywhere in the struct):

`tool_coverage`, `rule_coverage`, `high_risk_gaps`, `policy_violations`, `policy_warnings`, `overall_coverage_pct`, `meets_threshold`, `threshold`

None of `passed`, `violations` or `score` exists. The nearest true statements are `meets_threshold`, `policy_violations` and `overall_coverage_pct`. Every assertion on that page fails at the subscript, before it ever evaluates.

### 3. Both quickstarts are POSIX-only on a page that does not say so

`docs/getting-started/quickstart.md:20` and `docs/mcp/quickstart.md:87` run `mkdir -p /tmp/assay-demo`, under a heading that promises a result "in under 5 minutes" (`:3` on both pages) with no platform qualifier. We ship an `x86_64-pc-windows-msvc` binary.

## Why this blocks a launch

These are the first pages a new user reads, and each of them is falsifiable in under a minute by the user themselves. For a project whose subject is verifiable claims about agent behaviour, a quickstart that shows output the binary does not produce is the most expensive possible kind of wrong.

The fix is cheap in every case: capture the real output and correct the page. None of these requires building a feature.

## RED-first contract

Tests must first demonstrate that:

1. the exact console block in each quickstart is reproducible by running the documented commands, and today's block is not;
2. each Python example in `python-quickstart.md` executes against the real `validate()` return value, and today's examples raise `KeyError`;
3. a doc example that names a JSON or dict key not present in the serialised type fails a check, so the next drift is caught at CI time rather than by a reader.

Point 3 is the durable half. A one-time correction leaves the same gap open; something has to bind the documented keys to the struct that produces them.

## Scope note

Whether the proxy *should* print a deny line, and with which reason vocabulary, is a separate product question and is deliberately not decided here. This issue is about the pages describing what the binary does today.

## Non-claims

This does not address the correctness of the policy engine, the coverage computation, or any other page than the three named above.

Rul1an:
## Summary

A first-time reader who followed either MCP quickstart, or the Python one, was shown output and dict keys the installed binary does not produce. Those pages now show captured wrap stderr and the `CoverageReport` keys `validate()` actually returns.

The durable half is a coverage remap of documented report keys into the `CoverageReport` field table, not a second completeness proof of the pages. An invented console table, a POSIX-only setup step, or a prose lie that never writes a subscript still would not turn this check red.

The shipped `Coverage.analyze` docstring named the same fabricated keys. That text lives in the wheel, so `help(Coverage.analyze)` served it to anyone who never opened the docs site. The Returns block now names the keys the next statement actually returns.

Closes #3095.

## Fix round (`e058c68ad68004f4f31ad1a0dd670bc192e34b36`)

F1 and F2 from the `3ebcc855` review. No production-code change in this commit.

### F1: no-flag wrap capture (6.5.0 from this tree)

Documented argv, no `--verbose`, after `initialize`, `tools/list`, and a denied `read_file` of `/tmp/outside-demo.txt`. Exit 0.

Stderr (this is what a reader who drops the flag actually sees):

```text
[assay] loading policy from examples/mcp-quickstart/policy.yaml
[assay] wrapping command: npx ["-y", "@modelcontextprotocol/server-filesystem", "/tmp/assay-demo"]
Secure MCP Filesystem Server running on stdio
```

No `ALLOW` or `DENY` line. Stdout still carried the JSON-RPC deny (`structuredContent.status: deny`, `error_code: E_ARG_SCHEMA`). Both MCP pages no longer say stderr stays empty; they say the decision lines are not printed, and that missing decision lines are not a clean run.

### F2: AIcontext pages added to `PAGES` and shown RED first

`PAGES` was extended with `docs/AIcontext/entry-points.md` and `docs/AIcontext/quick-reference.md`. The check was otherwise unmodified. Before those snippets were touched:

```text
FAIL: docs/AIcontext/entry-points.md: documents 'passed', which CoverageReport does not serialise (fields: high_risk_gaps, meets_threshold, overall_coverage_pct, policy_violations, policy_warnings, rule_coverage, threshold, tool_coverage)
FAIL: docs/AIcontext/quick-reference.md: documents 'passed', which CoverageReport does not serialise (fields: high_risk_gaps, meets_threshold, overall_coverage_pct, policy_violations, policy_warnings, rule_coverage, threshold, tool_coverage)
```

Then `result["passed"]` became `result["meets_threshold"]` on both pages. After:

```text
documented coverage-report keys: agree with CoverageReport field names
```

`docs/use-cases/self-correction.md` and `docs/mcp/self-correction.md` stay out of `PAGES`: they document `assay_check_args` (`result["allowed"]`), a different API. `my_traces` in `python-quickstart.md` section 2 is still undefined; pre-existing, not this PR.

### Known limit (not fixed here)

`docs/AIcontext/quick-reference.md` still writes `coverage.score` as attribute access. Subscript-based scanning does not see it even now that the page is listed. Same shape at `docs/AIcontext/entry-points.md:347`. Attribute extraction belongs in #3105, not here.

### Mutants after adding two pages

The seven mutants still bite (`docs serialized-keys mutations: 7 observed`). `sdk-page-scanned` still proves the SDK page is in the set; adding two pages does not change that kill. It does not prove the AIcontext pages stay listed. The harness now copies every `PAGES` entry from the check module, so a missing-file red cannot masquerade as a key red. A docs-wide rule for the enumeration is #3105.

`bash scripts/ci/check-docs-generated-drift.sh` passed (22 files).

## Constructor

The check reads `pub` field names from `CoverageReport` in `crates/assay-core/src/coverage_next/types.rs` and compares them to `report[...]` / `result[...]` / `report.get(...)` and `report dict (...)` lists on the scanned pages.

That is a source reading, not an import of `assay-it`. The `scope` job has no native module, and `validate()` is `json.loads(serde_json::to_string(&report))` with no `serde(rename)` on the struct, so the field names are the serialised keys. A rename, a missing struct, or an unreadable page is a failure: this cannot guess a second mapping.

`score` has no field. The coverage percentage is `overall_coverage_pct`. `score_claim_support` is a different API. An internal baseline maps `overall_coverage_pct` onto a field it calls `score`; that is not the `validate()` return.

The check does not read docstrings, and it does not extract attribute access such as `coverage.score`. Both are a different extraction problem; attribute scanning is #3105, not this PR.

## Captured wrap output

Built `assay-cli` 6.5.0 into the shared target and ran the documented wrap (macOS arm64):

```bash
assay mcp wrap --policy examples/mcp-quickstart/policy.yaml --verbose \
  -- npx -y @modelcontextprotocol/server-filesystem /tmp/assay-demo
```

after `initialize`, `tools/list`, and one `tools/call` each. Stderr:

```text
[assay] loading policy from examples/mcp-quickstart/policy.yaml
[assay] wrapping command: npx ["-y", "@modelcontextprotocol/server-filesystem", "/tmp/assay-demo"]
[assay] ALLOW read_file
Secure MCP Filesystem Server running on stdio
```

```text
[assay] DENY read_file (reason: JSON Schema validation failed)
```

```text
[assay] DENY exec (reason: Tool is explicitly denylisted by name)
```

The issue's `grep -rn -E '\[assay\] (DENY|BLOCK)'` is a source-literal probe. `handle_deny_branch` already prints `[assay] DENY <tool> (reason: ...)` via `decision_str.to_uppercase()`. This PR does not add a DENY print. Without `--verbose`, the ALLOW/DENY lines are absent; the loading-policy and wrapping-command lines and the child banner still print. The wrap still returns a JSON-RPC deny on stdout. Capture is in the fix-round section above.

Windows: the pages are scoped to Unix in a visible sentence that names the policy. The example policy matches `^/tmp/assay-demo/.*`. A Windows path kept under that same policy is denied by the policy, not by Unix path syntax. I did not write unverified Windows commands.

## Python keys

`pip install assay-it==6.5.0`, then `validate()` / `Coverage.analyze()` on a two-tool allow policy:

- serialised keys: `high_risk_gaps`, `meets_threshold`, `overall_coverage_pct`, `policy_violations`, `policy_warnings`, `rule_coverage`, `threshold`, `tool_coverage`
- `report["passed"]`, `report["violations"]`, `report["score"]` each raise `KeyError`
- `meets_threshold`, `policy_violations`, `overall_coverage_pct` assertions pass (`overall_coverage_pct == 100.0`)

## RED then GREEN

Against the current tree, before the doc edit:

```text
FAIL: docs/getting-started/python-quickstart.md: documents 'passed', which CoverageReport does not serialise (fields: high_risk_gaps, meets_threshold, overall_coverage_pct, policy_violations, policy_warnings, rule_coverage, threshold, tool_coverage)
FAIL: docs/getting-started/python-quickstart.md: documents 'score', which CoverageReport does not serialise (...)
FAIL: docs/getting-started/python-quickstart.md: documents 'violations', which CoverageReport does not serialise (...)
```

After the edit:

```text
documented coverage-report keys: agree with CoverageReport field names
```

Then `docs/python-sdk/index.md` was added to the scanned set and shown RED before that page was touched:

```text
FAIL: docs/python-sdk/index.md: documents 'covered' ...
FAIL: docs/python-sdk/index.md: documents 'findings' ...
FAIL: docs/python-sdk/index.md: documents 'passed' ...
FAIL: docs/python-sdk/index.md: documents 'percent' ...
FAIL: docs/python-sdk/index.md: documents 'uncovered' ...
```

The fabricated return-shape block on that page was deleted rather than rewritten: a second hand-written schema is how the page shipped a fabricated type, and the check cannot see a JSON schema block. A docs-wide type map is a follow-up (Refs #3103).

`bash scripts/ci/check-docs-generated-drift.sh` passed (22 files).

## F4: shipped `Coverage.analyze` docstring

Text only in `assay-python-sdk/python/assay/coverage.py`. `analyze`, the analyzer, and every signature are unchanged.

Args already matched reality. `Coverage.analyze` takes `Vec<Vec<...>>`: a list of sessions, each a list of tool-call dicts. The example `[[{"tool": "search", "args": {"q": "foo"}}], ...]` is that shape. Auto-wrap of a bare list of dicts lives only on `validate()` (`assay-python-sdk/python/assay/__init__.py:15-33`), so the `analyze` docstring was left as the nested form.

Returns, before:

```text
- `passed` (bool): Whether coverage met the threshold.
- `score` (float): The calculated coverage percentage (0-100).
- `violations` (list): List of rule violations found.
- `records` (list): Per-trace analysis details.
```

Returns, after:

```text
- `tool_coverage` (dict): Tool coverage metrics.
- `rule_coverage` (dict): Rule coverage metrics.
- `high_risk_gaps` (list): Blocklisted tools never seen in traces.
- `policy_violations` (list): Policy violations found during analysis.
- `policy_warnings` (list): Policy warnings (for example unconstrained tools).
- `overall_coverage_pct` (float): Overall coverage percentage.
- `meets_threshold` (bool): Whether coverage met the threshold.
- `threshold` (float): The threshold that was checked.
```

```diff
 assay-python-sdk/python/assay/coverage.py | 12 ++++++++----
 1 file changed, 8 insertions(+), 4 deletions(-)
@@
         Returns:
             dict: A detailed coverage report dictionary containing:
-                  - `passed` (bool): Whether coverage met the threshold.
-                  - `score` (float): The calculated coverage percentage (0-100).
-                  - `violations` (list): List of rule violations found.
-                  - `records` (list): Per-trace analysis details.
+                  - `tool_coverage` (dict): Tool coverage metrics.
+                  - `rule_coverage` (dict): Rule coverage metrics.
+                  - `high_risk_gaps` (list): Blocklisted tools never seen in traces.
+                  - `policy_violations` (list): Policy violations found during analysis.
+                  - `policy_warnings` (list): Policy warnings (for example unconstrained tools).
+                  - `overall_coverage_pct` (float): Overall coverage percentage.
+                  - `meets_threshold` (bool): Whether coverage met the threshold.
+                  - `threshold` (float): The threshold that was checked.
```

The seven mutants still bite after this change. The check does not read the docstring, so none of them lost a kill.

## Mutants

Each points at the assertion that kills it (`scripts/ci/test-check-docs-serialized-keys.sh`):

| mutant | what it does | killed by |
|---|---|---|
| `historical-passed` | restore `report["passed"]` | unknown key `passed` |
| `historical-score` | restore `report["score"]` | unknown key `score` |
| `historical-violations` | restore `report['violations']` | unknown key `violations` |
| `dropped-field` | rename `meets_threshold` on the struct | unknown key `meets_threshold` |
| `serde-rename` | add `serde(rename)` | fail-closed parse |
| `struct-missing` | rename `CoverageReport` | fail-closed parse |
| `sdk-page-scanned` | restore `report["passed"]` on the SDK page | unknown key `passed` on that page |

## Where the check runs

Workflow: `.github/workflows/ci.yml`. Job id `scope`, display name **Detect CI scope**. Step: **Documented coverage-report keys exist**.

Required contexts on `main`: `CI`, `lane-check/proof`, `host-capability-check`, `review-record-check`. This check is not one of those names. It feeds `CI` because `scope` is in the rollup `needs:` and is judged `required`. There is no pre-commit hook, so `Lint (pre-commit)` does not run it.

`classify_file` on every path in this PR is `Gate.NONE`. No delegated proof.

## Non-claims

- One production file, text only: the `Coverage.analyze` docstring. No signature, analyzer, or other module changed.
- The check does not replay console blocks, does not read docstrings, and does not extract attribute access (`coverage.score` on the AIcontext pages). Refs #3105.
- Whether the proxy should print a richer deny line is left alone.

## Type of Change

- [x] Bug fix
- [x] Documentation update

## Verification

- [x] Keys check shown red on the live defect, then green
- [x] Seven mutants observed, and still biting after the AIcontext pages were added
- [x] Wrap stderr captured from the documented `npx` command
- [x] Python examples run against published `assay-it==6.5.0`
- [x] `bash scripts/ci/check-docs-generated-drift.sh` passed
- [x] Pre-push hooks ran (clippy workspace, linux compile gate)

---

### Head advanced by `update-branch`

Current head: `f892295b8e20df444f2372358eb723c4312a5c66`.

The review record names `e058c68a`, where F1 (the false "stderr stays empty" sentence) and F2 (two unscanned AIcontext pages) were verified fixed. The advance to `f892295b` is the `update-branch` merge of `b0da383c` only and carries no authored change. The checker derived the carry itself:

```
review-record-carry=derived reviewed=e058c68ad68004f4f31ad1a0dd670bc192e34b36 merged=b0da383c43106737bf08f3b40e9e44d789e5a66a conditions=tree-equivalence+no-overlap-re-derived-by-checker reviewed_files=10 advance_files=14
review-record-check=pass head=f892295b8e20df444f2372358eb723c4312a5c66
```

Known limit carried to #3105: the check's `PAGES` list is not self-defending. Removing either AIcontext entry leaves the mutation harness green, which the reviewer demonstrated.


Rul1an:
## First: you corrected me, and you were right

The brief told you to run the binary rather than write plausible output, and you applied that standard to my own finding instead of inheriting it. `handle_deny_branch` (`crates/assay-core/src/mcp/proxy/client/branches.rs:106-130`) builds `DENY` at runtime through `decision_str.to_uppercase()`, so my literal grep missed a line the binary emits on every denial. I have corrected #3095 and recorded the rule: an absence claim about program OUTPUT cannot be established by grepping for the literal output.

Reporting the check's reach honestly is the other thing I want to name. You wrote that it is not itself a required context and feeds `CI` only because `scope` sits in the rollup `needs:`. That sentence is worth more than the check, because it is the sentence that stops someone later believing the guard is stronger than it is.

## F1 (BLOCKING): the check is scoped to the pages you already fixed, so it cannot fail

You flagged this yourself and put it under "left unverified". It is not a leftover, it is the finding.

`docs/python-sdk/index.md` documents the same two APIs with keys that do not exist, and your check does not scan it:

- `:33-35` — `result = validate("policy.yaml", "traces.jsonl")`, then `result["passed"]` and `result["findings"]`, iterating `finding['level']` and `finding['message']`.
- `:91-93` — `report['percent']`, `report['covered']`, `report['uncovered']` from `Coverage.analyze()`.
- `:118-127` — a documented return shape for `validate(policy, traces)`:

```python
{
    "passed": bool,
    "findings": [
        {"level": "error", "rule": "...", "message": "..."}
    ]
}
```

None of `passed`, `findings`, `percent`, `covered` or `uncovered` is a `CoverageReport` field. This page is in worse shape than the quickstart you fixed: it also passes a path string as `traces`, while the real signature is `validate(policy_path: str, traces: list)` (`assay-python-sdk/python/assay/__init__.py:15-33`), and it publishes a fabricated return-shape block that a reader would code against directly.

**Why this blocks.** A check whose file list is exactly the set of pages you just corrected is a check that has only ever been green on a clean tree and can never go red on a dirty one. That is the same failure as a demonstrated red that was never pinned. The proof that a guard works is that it is RED on a defect you did not fix, and `docs/python-sdk/index.md` is sitting right there offering you that proof.

**What to do.** Widen the check to scan every markdown file under `docs/` that documents a key of a serialised type, not a hardcoded list of three pages. Then show it RED on `docs/python-sdk/index.md` before you touch that page, which is a much stronger red than the one you already have. Then fix the page: correct the keys, correct the `validate()` signature usage, and either correct or delete the return-shape block at `:118-127`.

If widening the scan turns out to need a general "which keys belong to which type" mapping that this PR should not carry, then say so and instead add `docs/python-sdk/index.md` to the scanned set explicitly, fix it, and open an issue for the general case. What I will not take is a check whose scope happens to exclude the one page that would fail it.

## F2 (answer, do not change): the `--verbose` asymmetry

Your capture shows `ALLOW` and `DENY` lines appear only under `--verbose`, and that without it the wrap still returns a JSON-RPC deny on stdout with nothing on stderr. The corrected pages should make that explicit rather than only showing the `--verbose` invocation: a reader who drops the flag sees silence and will read silence as "nothing was denied". One sentence naming what happens without the flag, on both pages.

## F3 (confirm): the Unix scoping and the policy

You scoped both MCP pages to Unix and noted that the example policy matches `^/tmp/assay-demo/.*`, so a Windows `mkdir` without a policy change would still fail. Good. Confirm that the visible scoping sentence says the policy is the reason, not just the path syntax. A reader who swaps in a Windows path and keeps the policy gets a denial that looks like a bug.

F1 disposition: open. F2: open. F3: open.

The captured output, the six killed mutants, the RED-then-GREEN on the key check and the refusal to add a DENY print are all right and stay.

Rul1an:
Fix round on `83b9de168e2810ad7e48a992e7c7f2291121a758`.

## F1: scoped explicitly, not widened

A docs-wide walk of every `report[` / `result[` subscript needs a map from documented keys onto the serialised type they belong to. Without that map the scan is wrong in both directions: `docs/use-cases/self-correction.md` writes `result["allowed"]` for a different API, and `docs/AIcontext/` still names `result["passed"]`. This PR does not carry that map. `docs/python-sdk/index.md` is now in the scanned set. Follow-up: #3103.

RED on the unfixed SDK page, after adding it to `PAGES` and before touching the file:

```
FAIL: docs/python-sdk/index.md: documents 'covered', which CoverageReport does not serialise (fields: high_risk_gaps, meets_threshold, overall_coverage_pct, policy_violations, policy_warnings, rule_coverage, threshold, tool_coverage)
FAIL: docs/python-sdk/index.md: documents 'findings', which CoverageReport does not serialise (fields: high_risk_gaps, meets_threshold, overall_coverage_pct, policy_violations, policy_warnings, rule_coverage, threshold, tool_coverage)
FAIL: docs/python-sdk/index.md: documents 'passed', which CoverageReport does not serialise (fields: high_risk_gaps, meets_threshold, overall_coverage_pct, policy_violations, policy_warnings, rule_coverage, threshold, tool_coverage)
FAIL: docs/python-sdk/index.md: documents 'percent', which CoverageReport does not serialise (fields: high_risk_gaps, meets_threshold, overall_coverage_pct, policy_violations, policy_warnings, rule_coverage, threshold, tool_coverage)
FAIL: docs/python-sdk/index.md: documents 'uncovered', which CoverageReport does not serialise (fields: high_risk_gaps, meets_threshold, overall_coverage_pct, policy_violations, policy_warnings, rule_coverage, threshold, tool_coverage)
```

Then the page was corrected (`validate(policy_path, traces)` with a list of traces; CoverageReport keys). The fabricated return-shape block was deleted: a second hand-written schema is how this page shipped a fabricated type, and the check reads subscripts and `report dict (...)`, not a JSON schema block, so a "corrected" dict would be a documented shape the guard cannot see.

GREEN after the fix:

```
documented coverage-report keys: agree with CoverageReport field names
```

Mutants: the original six still bite. Added `sdk-page-scanned` so a list that drops the SDK page cannot stay green. Seven observed.

`bash scripts/ci/check-docs-generated-drift.sh` passed (22 files).

## F2

Both MCP pages now say ALLOW/DENY appear only under `--verbose`, and without it the wrap still returns a JSON-RPC deny on stdout with nothing on stderr. Silence is not a clean run.

## F3

Both Unix scoping sentences now name the policy: the example policy requires `^/tmp/assay-demo/.*`. A Windows path kept under that same policy is denied by the policy, not by Unix path syntax.

No production code change.

Rul1an:
## F1 fixed, F2 fixed, F3 fixed. One more, and it is the one you found.

The red you produced is the one I wanted: five failures on `docs/python-sdk/index.md`, a page you did not author and did not know about when you wrote the check. That is a guard proving its scope, not just its assertion.

Deleting the fabricated return-shape block instead of correcting it is the right call, and your reason is the better half of it:

> Correcting it would write a third hand-maintained copy of CoverageReport next to the working example and the Rust struct. Two copies already drifted independently.

`sdk-page-scanned` is the mutant I would have asked for next: a page list that can silently shrink is the same defect one level up.

## F4: the docstring ships inside the wheel, and it is the worst copy of all

You flagged this and left it, correctly, because my brief said no production code change. I am lifting that restriction for this one file.

`assay-python-sdk/python/assay/coverage.py:42-48`, in the docstring of the shipped `Coverage.analyze` method:

```
Returns:
    dict: A detailed coverage report dictionary containing:
          - `passed` (bool): Whether coverage met the threshold.
          - `score` (float): The calculated coverage percentage (0-100).
          - `violations` (list): List of rule violations found.
          - `records` (list): Per-trace analysis details.
```

The method's very next statement returns `json.loads(self.analyzer.analyze(...))`, whose keys are `tool_coverage`, `rule_coverage`, `high_risk_gaps`, `policy_violations`, `policy_warnings`, `overall_coverage_pct`, `meets_threshold`, `threshold`. All four documented names are wrong, and `records` does not correspond to anything.

This is a worse copy than either page you fixed, for two reasons. It ships inside the wheel, so `help(Coverage.analyze)` and every IDE tooltip serve it to a user who never opens our docs site. And it sits four lines above the code that contradicts it, which is the shortest possible distance between a claim and its refutation.

Fix the docstring. Text only, in that one file: do not touch `analyze`, the analyzer, or any signature. That is documentation that happens to live in a `.py` file, not behaviour.

**Do not extend the check to read docstrings in this PR.** Your check reads subscripts and `report dict (...)` prose; docstrings are a different extraction problem and bolting it on here would make a guard that works for two shapes and half-works for a third. I have filed it separately.

## What I am not asking for

No further widening, no production change beyond that docstring, no new mutants unless the docstring fix breaks one of the seven. If correcting it makes `sdk-page-scanned` or any other mutant stop biting, say so and replace it.

F1: fixed. F2: fixed. F3: fixed. F4: open.

Rul1an:
Independent review of PR #3102 at `3ebcc855` (base `origin/main`, merge-base `ac6178de`). I did not build this change and did not author its governing spec. Worktree was read-only throughout; no branch writes. Binary `assay 6.5.0` built from this tree (PR touches no Rust, so head binary is identical). Target dir `/tmp/muse-3102-target`.

Verdict: BLOCKED, two actionable findings below. Everything else I attacked held.

Q1 — reproduced every console block with live traffic (`initialize`, `tools/list`, `tools/call` through `assay mcp wrap --policy examples/mcp-quickstart/policy.yaml --verbose -- npx -y @modelcontextprotocol/server-filesystem /tmp/assay-demo`). ALLOW block, both DENY blocks reproduce byte-exact, including `[assay] DENY read_file (reason: JSON Schema validation failed)` and `[assay] DENY exec (reason: Tool is explicitly denylisted by name)`. This also confirms DENY prints under `--verbose` via `decision_str.to_uppercase()` (`crates/assay-core/src/mcp/proxy/client/branches.rs:123-130`). Denied calls return exit 0 with the JSON-RPC deny on stdout, matching the golden-path rule that denial is not process failure.

F1 (live defect, new prose): both MCP pages say that without `--verbose` "stderr stays empty". False. `crates/assay-cli/src/cli/commands/mcp.rs:101` and `:155` print unconditionally, and my no-flag run shows stderr containing the `loading policy` / `wrapping command` lines plus the child banner — only the ALLOW/DENY decision lines are gated. A reader who drops the flag sees lines, not silence, which undercuts the sentence built on it ("sees silence and should not read silence as nothing was denied"). Fix: delete "and stderr stays empty" and say no decision lines print. Applies to `docs/getting-started/quickstart.md` and `docs/mcp/quickstart.md` (same sentence in both).

Q2 — the check scans exactly two pages (`PAGES` in `scripts/ci/check-docs-serialized-keys.py`): `docs/getting-started/python-quickstart.md`, `docs/python-sdk/index.md`. Docs-wide walk with the check's own extractors finds `result["passed"]` on `validate()` in two unscanned pages: `docs/AIcontext/entry-points.md:384` and `docs/AIcontext/quick-reference.md:189`.

F2 (live defect, scope gap): those two pages document the same `validate()` that returns `CoverageReport`, with the exact fabricated `passed` key this PR exists to kill. `docs/use-cases/self-correction.md` and `docs/mcp/self-correction.md` (`result["allowed"]` etc.) are correctly out of scope — they document `assay_check_args`, a different API, which I confirmed by reading both. But the "needs a key-to-type map" justification does not hold for the two AIcontext pages: I patched only `PAGES` to include them and the unmodified check goes red for the right reason on both. Fix: correct the two snippets and add both pages to `PAGES`. Residual note: `quick-reference.md` also has `coverage.score` attribute access, which subscript-based scanning would not catch even if listed.

Q3 — mechanism reads names from the type: `FIELD` regex over the `CoverageReport` body, no carried field list; fail-closed tripwire on `serde(rename`. Isolated the rename mutant on a scratch tree: red with exactly the rename diagnostic. Counterfactual with the tripwire stripped from a check copy: green — so the tripwire is load-bearing and the harness PASS is for the right reason. Proved the mutant shape really renames by serializing an identical struct (`passed` present, `meets_threshold` absent, unit test green). Boundary confirmed end to end: `assay-python-sdk/src/lib.rs:97` serializes the `CoverageReport` that `crates/assay-core/src/coverage.rs:6-10` re-exports from `coverage_next/types.rs` — the exact file parsed.

Q4 — live-run with a built SDK (`assay-it-6.5.0` via maturin): `Coverage.analyze` returns exactly the 8 documented keys; `validate` on a bare list of dicts returns the same 8 (auto-wrap in `assay/__init__.py` works); `analyze` on a bare list of dicts raises `TypeError` (no auto-wrap — `Vec<Vec<..>>` binding at `assay-python-sdk/src/lib.rs:50-54`). Docstring Args (list of sessions) matches `analyze`, and `validate`'s own docstring matches its wider contract. Clean. Pre-existing note: `my_traces` in python-quickstart section 2 is undefined, so that snippet is not literally runnable as pasted; untouched by this PR.

Q5 — full harness green (control + 7 mutants + reword control, `docs serialized-keys mutations: 7 observed`). Re-ran `sdk-page-scanned` and `serde-rename` myself on scratch trees: control green before, red with the right diagnostic each, control green after. Each mutant sits on the effective boundary (shipped page text or the parsed struct) and fails on its named ground.

Q6 — required contexts at `3ebcc855`: `CI` pass, `host-capability-check` pass, `lane-check/proof` pass (no delegated proof required), `review-record-check` fail with `no_current_record` — expected, this review supplies it. All other jobs pass; remaining skips are non-required stubs outside what I examined. `CodeRabbit` reports pass alongside "Review rate limited", which I treat as unavailable infrastructure, not signal; this review does not rely on it. `git diff --check` clean. SDK wrapper + typing tests: 4 passed.

Smallest path to READY: one-sentence fix for F1 in both MCP pages, F2 snippet corrections plus two `PAGES` entries, re-push, re-review the new head.

Rul1an:
```json
{
  "head_sha": "3ebcc8555ce17d4c0940d7287e8b416f19609461",
  "builder": {"agent": "cursor", "instance": "3095-quickstart-truth"},
  "reviewer": {"agent": "muse", "instance": "review-3102-3ebcc855", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "BLOCKED",
  "findings": [{"id": "F1", "summary": "New prose claims stderr stays empty without --verbose; mcp.rs:101,155 print unconditionally (proven by live run). Same sentence in docs/getting-started/quickstart.md and docs/mcp/quickstart.md.", "disposition": "open"}, {"id": "F2", "summary": "docs/AIcontext/entry-points.md:384 and docs/AIcontext/quick-reference.md:189 document validate() with fabricated passed key; unscanned by check; probe proves PAGES-only addition goes red for the right reason.", "disposition": "open"}],
  "no_findings": false,
}
```

Rul1an:
## F1 is my sentence, and it is wrong for the second time tonight in the same way

The reviewer found that both MCP pages now claim stderr "stays empty" without `--verbose`, and that `crates/assay-cli/src/cli/commands/mcp.rs:101` and `:155` print unconditionally. A no-flag run shows the `loading policy` and `wrapping command` lines plus the child's banner. Only the ALLOW/DENY decision lines are gated behind the flag.

That sentence came from my fix brief. I wrote:

> without it the wrap still returns a JSON-RPC deny on stdout with nothing on stderr

I did not run it. Earlier tonight I claimed no DENY line was printed, on the strength of a literal grep, and was corrected by a capture. This is the same error with a different surface: I asserted the absence of output without producing the absence. A builder following that instruction faithfully wrote it into two user-facing pages, so a PR that exists to stop documentation claiming things the binary does not do very nearly shipped a new one.

The fix is the reviewer's: delete "and stderr stays empty", and say instead that no decision lines are printed. The distinction the page needs to draw is between *no decisions shown* and *nothing happening*, and that survives without the false half.

## F2 stands, and the justification for leaving it does not

`docs/AIcontext/entry-points.md:384` and `docs/AIcontext/quick-reference.md:189` both document `validate()` with the fabricated `passed` key, and neither is scanned.

I accepted "docstrings and other shapes need a key-to-type map, so keep the scope narrow" when I filed #3105. That reasoning does not extend to these two, and the reviewer did the work to show why rather than asserting it: it patched only the `PAGES` list, left the check otherwise unmodified, and both pages went red for the right reason. No map required.

This is the third time tonight a guard's scope turned out to be a list that excluded a page nobody remembered. The pattern is now well enough evidenced to name: **a scope expressed as an enumeration will drift, and the only honest test of it is a red on a page the author did not write.**

Correct both snippets and add both pages to `PAGES`.

## Two observations from the review worth keeping, neither a finding

`docs/use-cases/self-correction.md` and `docs/mcp/self-correction.md` use `result["allowed"]` and are **correctly** out of scope: they document `assay_check_args`, a different API. The reviewer read both to establish that rather than pattern-matching on the subscript, which is the difference between a scope decision and a guess.

`quick-reference.md` also contains `coverage.score` as *attribute* access, which subscript-based scanning will not catch even once that page is listed. That is a real limit of the mechanism and belongs in #3105 rather than here.

Separately: `my_traces` in `python-quickstart.md` section 2 is undefined, so that snippet is not literally runnable as pasted. Pre-existing and untouched by this PR; not blocking it.

F1: open, my error. F2: open.

Rul1an:
Fix round on `e058c68ad68004f4f31ad1a0dd670bc192e34b36`.

## F1: no-flag capture, then the sentence

Built `assay-cli` 6.5.0 from this tree (`target/debug/assay version` = `6.5.0`). Documented argv, no `--verbose`, after `initialize`, `tools/list`, and a denied `read_file` of `/tmp/outside-demo.txt`. Exit 0.

Stderr:

```text
[assay] loading policy from examples/mcp-quickstart/policy.yaml
[assay] wrapping command: npx ["-y", "@modelcontextprotocol/server-filesystem", "/tmp/assay-demo"]
Secure MCP Filesystem Server running on stdio
```

No `ALLOW` or `DENY`. Stdout still carried the JSON-RPC deny (`structuredContent.status: deny`, `error_code: E_ARG_SCHEMA`). Both MCP pages now say the decision lines are not printed. They do not say stderr stays empty.

## F2: RED before the snippets were touched

`PAGES` gained `docs/AIcontext/entry-points.md` and `docs/AIcontext/quick-reference.md`. Check otherwise unmodified. Before those files were edited:

```text
FAIL: docs/AIcontext/entry-points.md: documents 'passed', which CoverageReport does not serialise (fields: high_risk_gaps, meets_threshold, overall_coverage_pct, policy_violations, policy_warnings, rule_coverage, threshold, tool_coverage)
FAIL: docs/AIcontext/quick-reference.md: documents 'passed', which CoverageReport does not serialise (fields: high_risk_gaps, meets_threshold, overall_coverage_pct, policy_violations, policy_warnings, rule_coverage, threshold, tool_coverage)
```

Then `result["passed"]` became `result["meets_threshold"]` on both. After: `documented coverage-report keys: agree with CoverageReport field names`.

Self-correction pages stay out (`assay_check_args`). `my_traces` in python-quickstart section 2 left alone.

## Attribute access, not fixed

`docs/AIcontext/quick-reference.md` still writes `coverage.score`. Subscript scanning does not see it. Same shape at `docs/AIcontext/entry-points.md:347`. Refs #3105.

## Mutants

Seven still bite. `sdk-page-scanned` still proves the SDK page is in the set; adding two pages does not change that kill. The harness now copies every `PAGES` entry so a missing-file red cannot masquerade as a key red.

`bash scripts/ci/check-docs-generated-drift.sh` passed (22 files).

Rul1an:
Delta review of PR #3102 at `e058c68a` (against `3ebcc85`). Verdict: READY. F1 and F2 carried as fixed. I built nothing on the branch and authored no governing spec; all probes ran from /tmp artifacts, since removed.

Q1 — F1 fixed, verified against a binary built from this tree. `cargo build -p assay-cli` with `CARGO_TARGET_DIR=/tmp/muse-3102-target-e058c68a` produced `assay 6.5.0` (PATH `assay` is 2.17.0, not used; target dir deleted after). No-flag run of the documented argv (`assay mcp wrap --policy examples/mcp-quickstart/policy.yaml -- npx -y @modelcontextprotocol/server-filesystem /tmp/assay-demo`, after `initialize`, `tools/list`, and a denied `read_file` of `/tmp/outside-demo.txt`):
- stderr held 4 lines and zero decision lines (`grep -c "ALLOW\|DENY"` printed `0`). The first three — `[assay] loading policy from examples/mcp-quickstart/policy.yaml`, `[assay] wrapping command: npx [...]`, `Secure MCP Filesystem Server running on stdio` — are byte-identical to the documented `--verbose` block (`docs/mcp/quickstart.md:103-106`, `diff` printed nothing). The fourth line is child chatter (`Client does not support MCP Roots...`), plus `npm notice` lines on a cold cache.
- stdout carried the denied `tools/call` (id 3) as a JSON-RPC deny (`structuredContent.status: deny`, `error_code: E_ARG_SCHEMA`), process exit `0` — denial is a successful exchange, not a failure.
- Both halves of the replacement hold (`docs/getting-started/quickstart.md:61-64`, `docs/mcp/quickstart.md:121-124`): decision lines are not printed, and neither page asserts anything else about stderr — `grep "stays empty"` on both pages printed nothing (exit 1). Code agrees: ALLOW/DENY prints are gated on verbose (`crates/assay-core/src/mcp/proxy/client/branches.rs:52` via `crates/assay-core/src/mcp/proxy/decisions.rs:16`, and `branches.rs:123-130`), while `crates/assay-cli/src/cli/commands/mcp.rs:101` and `:155` print unconditionally.
- Observation, not a finding: the banner set is child-defined (the Roots/notice lines vary with server and cache state). The sentence is inclusive ("still prints ... plus the child's banner"), so nothing contradicts it.

Q2 — F2 fixed, green for the right reason. The check's own extractor reads exactly `meets_threshold` from both corrected pages (`docs/AIcontext/entry-points.md:384`, `docs/AIcontext/quick-reference.md:189`), and `meets_threshold` is the correct key: it is a real `CoverageReport` field (`crates/assay-core/src/coverage_next/types.rs:62`), `validate()` returns `Coverage.analyze()` unchanged (`assay-python-sdk/python/assay/__init__.py:15-33`), which returns `json.loads` of the serialized report (`assay-python-sdk/python/assay/coverage.py:54`). `passed` exists nowhere in the struct — the correction is the true key, not a wrong-but-existing substitute. Residual, pre-existing and declared (check docstring now names #3105): attribute access `coverage.passed` (`entry-points.md:346`) and `coverage.score` (`entry-points.md:347`, `quick-reference.md:193`) is invisible to subscript scanning; present at `3ebcc85`, untouched by this delta, not new.

Q3 — the builder's stated limit is true; I tested rather than accepted it. I copied the tree to /tmp with only the two AIcontext entries deleted from `PAGES` and ran the unmodified harness: fully green, `docs serialized-keys mutations: 7 observed`, exit `0`. No harness mutant touches the `PAGES` list (its only read is the fixture copier, `scripts/ci/test-check-docs-serialized-keys.sh:26`). The copy-every-`PAGES` change keeps fixture and list in lockstep (no file-missing red masquerading as a key red) but does not pin membership, so it does not mitigate this. Smallest honest guard, for #3105 not this PR: pin the expected `PAGES` set in the harness, or an inverted mutant that strips a listed page and requires red.

Q4 — all seven mutants still bite at this head (`bash scripts/ci/test-check-docs-serialized-keys.sh` exit `0`: control green, `historical-passed`, `historical-score`, `historical-violations`, `dropped-field`, `serde-rename`, `struct-missing`, `sdk-page-scanned`, reword control green). Adding the two pages strengthened `dropped-field`: on a scratch tree it now fires on all four pages including both AIcontext ones. No other mutant's ground changed.

Q5 — CI at `e058c68a`: the only failing context is `review-record-check` (`reason=no_current_record`, expected — this review supplies it). Every other completed context passes; the rest are in-progress/queued with no failures. `CodeRabbit pass` alongside "rate limited" is unavailable infrastructure, not signal. `git diff --check` on the delta is clean; `check-docs-generated-drift.sh` passes (22 files).

Rul1an:
```json
{
  "head_sha": "e058c68ad68004f4f31ad1a0dd670bc192e34b36",
  "builder": {"agent": "cursor", "instance": "3095-quickstart-truth"},
  "reviewer": {"agent": "muse", "instance": "review-3102-e058c68a", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [{"id": "F1", "summary": "Replacement sentence verified live: no-flag run of assay 6.5.0 built from this tree shows zero ALLOW/DENY lines, loading-policy/wrapping-command/banner byte-match the docs block, deny returns on stdout exit 0; pages no longer claim stderr stays empty.", "disposition": "fixed"}, {"id": "F2", "summary": "Both AIcontext pages corrected to result[meets_threshold], the true CoverageReport key (types.rs:62) for what validate() returns; extractor reads exactly that key on both pages; both pages added to PAGES; check green for the right reason.", "disposition": "fixed"}],
  "no_findings": false,
}
```

Issue #3124, opened 2026-09-18
ci: Scorecard flags the lane-check checkout as dangerous; make its trusted-ref intent structural


Rul1an:
## Measured state

The public OpenSSF Scorecard for this repository (scan 2026-09-14, overall 5.9) scores **Dangerous-Workflow 0**, with one finding:

```
Warn: untrusted code checkout '${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha
       || (github.event_name == 'workflow_dispatch' && 'main' || github.sha) }}':
       .github/workflows/assay-runner-lane-check.yml
```

Dangerous-Workflow is one of Scorecard's critical-weight checks, so this single finding pulls the overall score down disproportionately.

## Why it is (as far as the coordinator can establish) a false positive

`.github/workflows/assay-runner-lane-check.yml` is triggered by `pull_request`, `workflow_run` and `workflow_dispatch`, holds `statuses: write` and `pull-requests: write`, and checks out:

- on `pull_request`: `github.event.pull_request.base.sha`, i.e. the PR's **base** commit, so a PR can never run its own edited gatekeeper;
- on `workflow_dispatch`: the literal `main`, because the privileged refresh must never execute helper code from a dispatched ref a PR author controls;
- on `workflow_run`: `github.sha`, which for `workflow_run` is the default-branch head;

always with `persist-credentials: false`. The comment at the checkout step (lines 46-51) records exactly this intent. Scorecard flags any expression-valued `ref:` in a workflow that also has a `workflow_run` trigger; it cannot evaluate which branch of the expression applies.

This conclusion is itself a claim and should be checked adversarially when this issue is built: in particular that no path through the expression can resolve to PR-authored code while the job holds its write permissions.

## Why fix a false positive

Scorecard has no suppression mechanism, and its score is a public production-readiness signal. The finding can be removed by making the trusted-ref intent structural instead of expressional, for example by splitting the privileged `workflow_run`/`workflow_dispatch` refresh into its own workflow whose checkouts use no expression-valued ref, while the `pull_request` path keeps its base-SHA checkout in a workflow without privileged triggers.

## Constraint

`lane-check` is a **required** status context. Any split must keep the required context name and its semantics exactly, prove that with a PR that exercises both paths, and must not introduce a window in which the required check can be satisfied by PR-authored code.

## RED-first contract

1. A static check fails when a workflow that has a `workflow_run` or `pull_request_target` trigger checks out an expression-valued ref.
[...]

PR #3135, opened 2026-09-19
ci: move the lane-check workflow_run refresh into its own workflow

Rul1an:
## Description

OpenSSF Scorecard scores Dangerous-Workflow 0 for one finding: `.github/workflows/assay-runner-lane-check.yml` checks out a `ref:` containing `github.event.pull_request` in a file that also has a `workflow_run` trigger. The ref is the PR **base** SHA, so the checkout is trusted and the finding is a false positive, but the rule is a plain substring match. In ossf/scorecard at `f92023a3`, `checks/raw/dangerous_workflow.go:74-75` and `:204-205` flag any `actions/checkout` `ref:` containing `github.event.pull_request` or `github.event.workflow_run`, and only in files with a `workflow_run` or `pull_request_target` trigger.

This PR moves the `workflow_run` refresh into a sibling workflow instead of hiding the expression from the scanner, so the privileged trigger and the PR-event checkout no longer share a file.

Head reviewed: `e656494901c15e6bbd2444ed0111ec034159a617`. Current head: `d5e70ee1d42fa0aa6bf6a49fdf0010d7b9c39168`, which only merges `origin/main` (update-branch; no reviewed file changes).

### What changed

- `.github/workflows/assay-runner-lane-check.yml` keeps `pull_request` and `workflow_dispatch`. The required `lane-check` job name is unchanged, and `docs-auto-update.yml` still dispatches this filename with the same inputs. The checkout ref is now `base.sha` for `pull_request` and `main` for dispatch.
- `.github/workflows/assay-runner-lane-check-refresh.yml` (new) holds the `workflow_run` refresh after a dispatch-triggered delegated proof. It checks out the default branch with no `ref:` and runs the same helper, `scripts/ci/assay_runner_lane_check.py`.
- `scripts/ci/assay_runner_lane_check.py`: the new file is listed wherever the old one is a gated path. A new self-test fails any workflow that pairs a `workflow_run`/`pull_request_target` trigger with a checkout `ref:` containing either substring. It runs in `--self-test` (the pre-commit hook and the `lane-check` job). Because `lane-check` checks out the PR **base**, it does not scan a PR's own workflows; no required context scans the head yet. That gap and the inline-trigger limit below are tracked in #3137.
- `kernel-matrix.yml`, `CI-CONTRACT.md` and `docs/BRANCH-PROTECTION-SETUP.md` now name the sibling.

### Evidence

The new scanner, run against `origin/main`'s workflows and then this head's:

```text
main: [".github/workflows/assay-runner-lane-check.yml: actions/checkout ref contains 'github.event.pull_request' in a workflow that also has a workflow_run or pull_request_target trigger"]
head: []
```

`python3 scripts/ci/assay_runner_lane_check.py --self-test` exits 0 on this head.

### Known limits (tracked in #3137)

The scanner reads block-form triggers (`workflow_run:` as a key). An inline trigger list such as `on: [workflow_run]` would not be seen. No workflow in the tree uses that form with a privileged trigger today; the only inline one, `assay.yml`, is `on: [push, pull_request]`.

Refs #3124. The issue closes once the next Scorecard scan confirms the finding is gone.


Rul1an:
## Review: PR #3135 @ `e6564949` (range `bddfc81e...e6564949`, 6 files)

Independence: I built none of this change (builder `cursor/3124-lane-check-split`, Grok) and authored no governing spec for it.

Claim under test: the split removes the Scorecard Dangerous-Workflow pairing without changing what `lane-check` decides or where privileged code runs, and the new self-test pins the pairing shut.

### 1. Behavioural equivalence: HOLDS

Compared the old file (at base) against the two new files path by path:

- pull_request: same base-SHA checkout (`assay-runner-lane-check.yml:52`), same required job name `lane-check`, same PR-number resolution, same `--comment` + conditional `--status` write. Dropped `DELEGATED_WORKFLOW_RUN_ID` conditional env is neutral: on PR events it was `''`, and `run_ids_from_event` (`scripts/ci/assay_runner_lane_check.py:1684`) returns `[]` unless the event is `workflow_run`.
- workflow_dispatch: unchanged inputs (`pr_number`/`override_reason`/`expected_head_sha`), same `'main'` checkout, same `EXPECTED_HEAD_SHA` double-guard and override envs (kept in the main file).
- workflow_run (now `assay-runner-lane-check-refresh.yml`): same trigger (`workflows: ["Runner Spike Delegated"]`, `types: [completed]`), same dispatch-only `if:` filter, same `--resolve-pr-from-event`, same `DELEGATED_WORKFLOW_RUN_ID` value, same comment/status write, same five job permissions under workflow-level `permissions: {}` (refresh file lines 16, 23-28). Checkout changed from explicit `github.sha` to no-`ref` default, which is the same default-branch tree (`workflow_run` executes the file from the default branch). Nothing added or dropped beyond moving the three `workflow_run`-only pieces across; the `EXPECTED_HEAD_SHA` guard correctly stays only in the dispatch file.
- Routing unchanged: `docs-auto-update.yml:134` still dispatches `assay-runner-lane-check.yml` only (self-test pins both directions at `assay_runner_lane_check.py:3517-3518`).

### 2. Head-side catch: FINDING F1 (open)

Mutation matrix, exact hook entry (`python3 scripts/ci/assay_runner_lane_check.py --self-test`, `.pre-commit-config.yaml:136-138`) on scratch trees exported from the head (`/tmp/rev3135/`):

| tree | simulates | result |
|---|---|---|
| control (pristine head) | base checkout in `lane-check` | exit 0 |
| mutA: `workflow_run` trigger re-added to lane file | pairing-PR head | exit 1, contract pin fires |
| mutB: pairing in new `zz-evil-lane.yml` | pairing-PR head, other file | exit 1, whole-dir scanner names the file |
| mutC: `on: [workflow_run]` / `on: workflow_run` pairings | pairing-PR head, alt syntax | exit 0 (miss; see F2) |

Control green before and after; both block-form mutants red. The scanner mechanism works. The wiring does not pin it shut at merge time:

- Required contexts are exactly `CI`, `lane-check/proof`, `host-capability-check`, plus the record check (`.github/rulesets/main-required-ci-contexts.json`). The required `lane-check` job checks out the PR **base** (`assay-runner-lane-check.yml:52`), so on a pairing-PR it scans the clean base: green, as the control run above demonstrates.
- The head-side runners both miss required status: kernel-matrix `Lint (pre-commit)` checks out head (`kernel-matrix.yml:108`, no `ref`) and runs `pre-commit run --all-files` (`:149`) including the unfiltered lane hook — it would go red (mutA/mutB prove the command fails on that tree) — but it is not a required context, so red blocks nothing; and its `paths` filter (`:13-89`, enumerated files, no `.github/workflows/**` wildcard) does not even start it for a pairing in any of the ~40 unlisted workflow files or a new one (mutB's case).
- Remaining backstop is the scheduled Scorecard scan (`openssf-scorecard.yml`: schedule + dispatch only) — external, weekly, non-blocking. Local pre-commit is bypassable.

F1: no required gate executes the pairing scanner against the PR head. Recommend a follow-up (e.g. widen the kernel paths to cover `.github/workflows/**`, or run the scanner on head inside a required job). This gap predates the PR — the change strictly improves detection — so it is open, not blocking.

### 3. Scanner limit: FINDING F2 (open, minor)

Confirmed the disclosed limit directly: `workflow_has_privileged_trigger` returns `False` and no problems are reported for both the inline-list and scalar `on:` forms (mutC exit 0). Repo-wide grep for `^on: [` finds exactly one user, `assay.yml` with `[push, pull_request]` — non-privileged, so nothing is silently unpinned today. The real Scorecard parses YAML properly and stays a backstop. Small fix (match the two extra `on:` spellings + cases) or defer to a follow-up; I lean defer.

### 4. `--status` fallback: no action

`--help` prints `--status` (2 hits) and argparse registers it unconditionally (`assay_runner_lane_check.py:3725`), so the `grep -q -- "--status"` guard always succeeds. Dead code, but a byte-identical copy of the old step and harmless insurance against workflow-file/helper skew across refs (dispatch runs the file from the dispatched ref, the helper from `main`). Keep.

Out of scope respected: other Scorecard checks untouched. No cargo builds (Python + workflow change only).

CI at `e6564949`: `lane-check` pass, `lane-check/proof` pass, `host-capability-check` pass; record check fails (no record posted yet — this review supplies it); `Lint (pre-commit)` and the `CI` aggregate were still pending at check time.

Scratch kept at `/tmp/rev3135/{control,mutA,mutB,mutC}` plus `/tmp/old-lane.yml`. Note: an earlier `git archive` pipe briefly extracted into `/tmp` top level, overwriting files from a prior session's export there (same repo-root names, Sep 18 23:51 mtimes); I removed only my own dirs and left the rest in place — treat anything at `/tmp` top level as suspect, the `/tmp/rev3135/*` trees are clean-room exports.

Rul1an:
```json
{
  "head_sha": "e656494901c15e6bbd2444ed0111ec034159a617",
  "builder": {"agent": "cursor", "instance": "3124-lane-check-split"},
  "reviewer": {"agent": "muse", "instance": "review-3135-e6564949", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [{"id": "F1", "summary": "No required context runs the pairing scanner on the PR head: lane-check scans the base checkout, kernel-matrix Lint (head checkout) is not required and is paths-filtered", "disposition": "open"}, {"id": "F2", "summary": "Scanner misses inline-list and scalar on: trigger forms; only non-privileged inline user exists today, small fix or defer", "disposition": "open"}],
  "no_findings": false,
}
```

Rul1an:
Disposition of the review findings at `e6564949`:

- **F1** (no required context scans the PR head) and **F2** (inline and scalar `on:` forms) are deferred to #3137. Both gaps are in the scanner this PR introduces, so merging still leaves the tree better covered than before. The PR body no longer implies the required `lane-check` job scans a PR's own workflows.
- **Q4**: the `--status` help probe stays; it is dead but harmless.

Thanks for the mutation matrix, especially the scratch tree with the pairing in a workflow the author wrote. That is the case the body had glossed over.

Issue #3116, opened 2026-09-18
cli: `assay run --help` promises a trace auto-ingest that `run` never does, so `assertions:` cannot be evaluated


Rul1an:
## What happens

A suite with `model: trace` and an `assertions:` block reads its episodes from the SQLite store, not from `--trace-file`. `assay ci --replay-strict` writes the trace into the store for you. `assay run` does not, with or without `--replay-strict`, and neither does `assay ci` without the flag. There the evaluator finds no episode and the test is reported as failed, with the same verdict for a trace that is fine and a trace that is not.

`assay run --help` says otherwise. `RunArgs` and `CiArgs` carry the same two lines:

- `--trace-file`: "Trace file to use as Source of Truth for replay (auto-ingested in strict mode)"
- `--replay-strict`: "strict replay mode: use trace-file as truth, forbid network, auto-ingest to DB"

`run --replay-strict` does not ingest.

There is a working route for `run`, but nothing points to it: `assay trace ingest --input t.jsonl --output x.db` writes to the store when `--output` ends in `.db` or `.sqlite` (`crates/assay-core/src/trace/ingest.rs:10-20`), and `assay run --db x.db` then evaluates correctly. `assay trace ingest --help` describes the command as "Ingest a raw JSONL log file and normalize to trace dataset" and does not mention the store.

## Measured

`origin/main` at `f5dc0db856a7c4d70c0f0277ee367af4ec3bf99a`, debug build of `assay-cli` 6.5.0, macOS arm64, `--no-cache`, each cell in its own empty directory. The same behaviour was first seen on the published `assay-v6.5.0-aarch64-apple-darwin` asset.

| command | benign trace | forbidden trace |
|---|---|---|
| `run` | exit 1, `E_TRACE_EPISODE_MISSING` | exit 1, `E_TRACE_EPISODE_MISSING` |
| `run --replay-strict` | exit 1, `E_TRACE_EPISODE_MISSING` | exit 1, `E_TRACE_EPISODE_MISSING` |
| `ci` | exit 1, `E_TRACE_EPISODE_MISSING` | exit 1, `E_TRACE_EPISODE_MISSING` |
| `ci --replay-strict` | exit 0, `ok` | exit 1, `assertions failed (1)` |
| `trace ingest --output x.db`, then `run --db x.db` | exit 0, `ok` | exit 1, `assertions failed (1)` |

The two plain `run` reports differ only in `duration_ms`.

<details><summary>Fixtures</summary>

[...]

PR #3138, opened 2026-09-19
run: ingest the trace and evaluate assertions against this run's episodes (Refs #3116)

Rul1an:
Head: `6221e5a8ac4be955c433cfaa343d4b61d9054e59` on `cursor/3116-scoped-assertion-eval`, based on `origin/main` `eddf9079aef79ce459322c3d88b466d4756daf68`.

Coordinator update: merged main after the implementation run. The local measurements below remain bound to `63c86dc47f82d2eac0b225f51050ea68e9e82bf9`; they are not relabeled as a rerun on this merge head. Muse completed an independent review on this exact head (record: https://github.com/Rul1an/assay/pull/3138#issuecomment-5740744302); CI run 35433576239 passed. Merged as bfa5e59f5fe2d78c8d45042edff7fb337f6fbbb4. The diagnostic-only follow-up is tracked separately in #3140.

## Summary

- `assay run` and bare `assay ci` ingest `--trace-file` into `--db` when any test has assertions that read a stored episode, so the help text matches what the commands do.
- Those assertions evaluate the episode ids this invocation ingested (bound to the run `run_suite` creates). The latest-per-test_id lookup remains only behind `--latest-stored-episode`. `execute_pipeline` writes this invocation's flag on every entry, so a later process without the flag on the same `--db` evaluates only the episode it ingested and does not print the latest-stored-episode note.
- Re-ingesting an episode id purges that episode's `tool_calls` / `steps` / observations first, so a later ingest that reuses `step_id`/`call_index` is what gets evaluated. `ON CONFLICT DO NOTHING` on `tool_calls` is unchanged.

## Why this is one change

Scoping evaluation to this invocation's episodes, without also ingesting on `run` / bare `ci`, would break the working `trace ingest --output x.db` then `run --db x.db` route. Ingest-on-run plus per-episode purge keeps that route, and makes `run --help`'s ingest promise true.

## Test plan

- [x] Contract binary test `crates/assay-cli/tests/contract_trace_assertion_input.rs` (15 cases): benign/forbidden x `run` / `run --replay-strict` / `ci` / `ci --replay-strict` on a fresh db; reused-db sequences (a)/(b)/(c); flagged then unflagged on one DB for `run` and `ci`; `trace ingest --output x.db` then `run --db x.db`; help strings.
- [x] Must-fail controls: drop scoping (evaluate latest-per-test_id again) -> (a) and (b) fail, fresh-db benign stays green; drop per-episode purge -> (c) fails, fresh-db benign stays green; drop the per-invocation opt write -> flagged-then-unflagged fails, fresh-db benign and forbidden stay green.
- [x] `cargo test -p assay-cli --test contract_trace_assertion_input --test contract_run_ci_parity`

## Red / green on the contract cases

Measured on worktree `/Users/roelschuurkes/wt-cursor-3116`. Production RED for the new cases is `83d9581ebe370cefcade92244fe3f61d310b555d`; GREEN is `63c86dc47f82d2eac0b225f51050ea68e9e82bf9`.

| case | 83d9581e | 63c86dc4 |
|---|---|---|
| flagged then unflagged `run`, same DB | RED (exit 0, latest-stored-episode note) | GREEN exit 1, `assertions failed`, no note |
| flagged then unflagged `ci`, same DB | RED (exit 0, latest-stored-episode note) | GREEN exit 1, `assertions failed`, no note |
| fresh `run` benign | GREEN | GREEN |
| fresh `run` forbidden | GREEN | GREEN |

Prior ingest/scope cases remain as measured on `83d9581e`.

Refs #3116


Rul1an:
# Review: PR #3138 at 83d9581e (base 57cbfe0d)

I built none of this change (builder: cursor / 3116-scoped-assertion-eval). All
probes below ran against a binary built from this exact head
(`83d9581ebe370cefcade92244fe3f61d310b555d`, verified `git rev-parse HEAD`
before writing), with `CARGO_TARGET_DIR=/tmp/rev3138-target`, scratch under
`/tmp/rev3138/`. Mutants were built in a separate clone at the same head, so
the review tree is untouched.

## F1 (blocker): `--latest-stored-episode` persists in the DB and leaks into later runs

The claim is that latest-per-test_id evaluation survives "only behind
`--latest-stored-episode`". It does not: one flagged run permanently opts the
database in.

- `pipeline.rs:201-208` calls `set_latest_stored_episode_eval(true)` when the
  flag is passed and does nothing when it is absent (no reset to false).
- `store_eval_scope.rs:10` (`clear_assertion_eval_scope`) deletes only
  `kind IN ('episode', 'used')` — the `kind='opt'` row is never cleared, not at
  pipeline start (`pipeline.rs:195`) nor inside `ingest_into_store`
  (`ingest.rs:60`).
- The gate `agent_assertions/mod.rs:102` (`latest_stored_episode_eval`,
  `store_eval_scope.rs:69`) therefore stays true for every later invocation on
  that `--db`, flagged or not.

Reproduction (head binary, `model: dummy` suite with a store-reading
assertion, pre-ingested benign episode, no `--trace-file`):

1. no flag: exit 1, `E_TRACE_EPISODE_MISSING` x1, note x0, session table empty.
2. `--latest-stored-episode` once: exit 0, note x1,
   `assertion_eval_session` = `opt|latest_stored_episode`.
3. no flag again: **exit 0**, `E_TRACE_EPISODE_MISSING` x0, note x1, opt row
   still present — the run evaluated a stored episode it never ingested and
   printed `note: assertions used the latest stored episode per test_id
   (--latest-stored-episode)` (`assertions.rs:26`, stderr) for a flag it was
   not given.

The builder's contract tests use a fresh DB per case (or never pass the flag),
so no committed test covers flag-then-no-flag on one DB. Smallest coherent
fix (not applied — reviewers don't edit the branch): in `execute_pipeline`,
call `set_latest_stored_episode_eval(input.latest_stored_episode)`
unconditionally so each invocation resets the persisted opt; keep
`ingest_into_store`'s clear limited to `episode`/`used` so it cannot clobber
the flag set just before ingest.

## Verified: scoping and purge are both load-bearing (mutants go red, controls green)

- Mutant A (bind call `execute.rs:34` removed AND gate `mod.rs:102` forced to
  true, i.e. latest-per-test_id again): fresh-db benign stays exit 0; reused-db
  cases (a) benign-then-forbidden `run` and (b) newer-benign-then-older-forbidden
  `ci --replay-strict` both exit 0 on the violating trace; (c) still exit 1.
- Mutant B (`purge_episode` call in `ingest.rs:60-65` removed,
  `record_assertion_eval_episode` kept): fresh-db benign exit 0, (a)/(b) exit 1,
  (c) same-`step_id`/`call_index` re-ingest exits 0 on the violating trace —
  the stale `tool_calls` row survives via the unchanged
  `ON CONFLICT(step_id, call_index) DO NOTHING` (`store_trace.rs:157`).
- Head control binary on the same script: fresh 0, (a)/(b)/(c) all 1.
- Committed suite `contract_trace_assertion_input`: 13/13 pass on this head.
  `truncation_observations` (incl. first-write-wins): 21/21 pass.
  `cargo fmt --all -- --check` clean; `cargo clippy -p assay-core -p assay-cli
  --all-targets -- -D warnings` clean.

## Verified: purge touches only the re-ingested episode; other rows and stored results survive

Re-ingested `ep-X` (same id/slot, now violating) without `ep-KEEP` in the new
trace: `ep-KEEP` step `sK` + `tool_calls(list_files)` intact before and after,
`ep-X` row shows the new tool, run 2 exits 1 with `assertions failed`
(evaluated the fresh episode). `results`/`runs` rows from run 1 retained.
One consequence to be aware of: `bind_assertion_eval_scope`
(`store_eval_scope.rs:27`) re-points the shared `episodes` row to the new
`run_id`, so `SELECT id FROM episodes WHERE run_id=1 AND
test_id='no_forbidden_tool'` is empty afterwards — an older run's episode graph
is no longer re-derivable through `get_episode_graph`. No production path
re-reads old runs that way (only `run_suite` with the current id,
`execute.rs:34` → `assertions.rs:15`), and `assay explain` takes a trace file
plus policy (`ExplainArgs`, `explain.rs:45-73`, no `--db`), so stored
explanations are unaffected. Filed here as observed behavior, not a blocker.

## Verified: trigger, unit-only suites, ambiguity, output hygiene, compat, semver, lane

- Ingest trigger: `run` and bare `ci` now ingest when any test has a
  store-reading assertion (`pipeline.rs:210-216` via new
  `assertions_require_stored_episode`, `mod.rs:45`). Unit-test-only suite
  (`model: dummy`, `args_valid` + `test_args`, no `--trace-file`): exit 0 with
  zero `auto-ingest` lines — untouched. Store-reading suite with no
  `--trace-file` on an empty DB: exit 1 with `E_TRACE_EPISODE_MISSING`, never
  exit 0 (for `model: trace` the pre-existing `--trace-file` requirement,
  unchanged from base `pipeline.rs:170`, fires first — same as 6.6.1).
- Two episodes for one `test_id` in one trace: refused, exit 1, stderr carries
  `E_TRACE_EPISODE_AMBIGUOUS` (identical prompts are additionally refused
  earlier by the trace loader — also non-zero, no evaluation).
- `--latest-stored-episode` note goes to stderr (`assertions.rs:23-26` uses
  `eprintln!`); with `--format json`, stdout piped through
  `python3 -m json.tool` validates, and the `assertion_episode` marker lands
  inside the JSON `details` object. The no-flag error path also emits valid
  JSON on stdout.
- Compat: `schema.rs:129` only appends `CREATE TABLE IF NOT EXISTS
  assertion_eval_session`; no existing DDL, migration, or `user_version`
  handling changed. I simulated a 6.6.1 DB by applying the base DDL plus a run
  and episode row, then opened it with the new binary: exit 0, old rows
  preserved, table created; a second run re-inits cleanly (idempotent). Reverse
  direction argued from code, not run: I did not build 6.6.1 (release
  asset/install not permitted). The old binary's DDL is `IF NOT EXISTS` on the
  old table set and no old query names the new table, so a new-DB file opens
  with the extra table inert — the one residual risk is any future reader that
  enumerates tables, of which I found none (`grep` for `sqlite_master`/`table_list`
  across `crates/` empty).
- Semver: the diff adds only `pub` items (2 `agent_assertions` fns, 8 `Store`
  methods, CLI flag fields on binary-crate args) and removes none; no `pub`
  struct fields. Local semver tools were not installed (review rules forbid
  installs); CI's `Semver public gate / Wave 0 semver checks` at this head is
  **pass**.
- Lane: `lane-check` **pass**, `lane-check/proof` **pass** ("no delegated
  runner proof required") — no delegated proof needed. (CodeRabbit shows pass
  with "Review rate limited": unavailable review infrastructure, not a review.)

## CI at this head

Only non-passing required context is `review-record-check` (fail: no review
record posted yet — this review supplies it). Everything else pass or skipped;
semver Wave 0 pass (was pending mid-review, green on re-check). No other
failures inside or outside what I examined.

## Verdict: BLOCKED on F1

F1 contradicts the claim under test on the exact head: a single flagged run
permanently re-enables latest-stored evaluation for that DB, and a later
unflagged run exits 0 on episodes it never ingested. The scoping, purge, and
hygiene mechanics all verify — F1 is the one defect, and it needs a code fix
plus a committed flag-then-no-flag same-DB test before merge.

Rul1an:
```json
{
  "head_sha": "83d9581ebe370cefcade92244fe3f61d310b555d",
  "builder": {"agent": "cursor", "instance": "3116-scoped-assertion-eval"},
  "reviewer": {"agent": "muse", "instance": "review-3138-83d9581e", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "BLOCKED",
  "findings": [{"id": "F1", "summary": "kind='opt' latest-stored-episode row persists in the DB across invocations, so one flagged run permanently re-enables latest-per-test_id evaluation and a later unflagged run exits 0 on episodes it never ingested; smallest fix is resetting the persisted opt from input.latest_stored_episode on every pipeline entry", "disposition": "open"}, {"id": "F2", "summary": "this-invocation bind plus per-episode purge are both load-bearing: drop-bind/drop-purge mutants go red on the named reused-db cases with fresh-db controls green, and purge deletes only the re-ingested episode rows while results and explain inputs are unaffected", "disposition": "verified"}, {"id": "F3", "summary": "latest-used note goes to stderr with stdout valid JSON under --format json; 6.6.1-shaped DB opens with idempotent session-table creation; two-episodes-per-test_id refused; unit-test-only suites ingest nothing", "disposition": "verified"}],
  "no_findings": false,
}
```

Rul1an:
Coordinator relay of the completed independent Muse review. Publisher: Rul1an via Codex; actual reviewer: Muse, instance `review-3138-6221e5a8`. Codex is not claimed as the independent reviewer. Reviewer authentication and independence are declarations, not cryptographically established facts.

**READY at `6221e5a8ac4be955c433cfaa343d4b61d9054e59`**, base `eddf9079aef79ce459322c3d88b466d4756daf68`. Muse declared no authorship of this change or governing specification and derived the complete 12-file three-dot diff. Review tree: `/Users/roelschuurkes/wt-review-3138`; clean before and after. No branch edits or posts were made by the reviewer.

F1 from the prior BLOCKED head is **fixed**: pipeline entry writes the current invocation's flag unconditionally. Independent execution on this head: trace-assertion contract 15 passed; run/ci parity 6 passed; fmt, whitespace check and affected CLI clippy exited 0. In an isolated scratch copy, restoring the guarded write made exactly the two flagged-then-unflagged regressions fail, while fresh-db positive controls passed. These are the reviewer's runs, not relabeled builder logs.

F2 is **deferred to #3140** with coordinator disposition: the shared drain of usage markers may attribute IDs to the wrong test-result row under in-process parallel execution. It is a diagnostic-attribution finding, with no verdict change established by the review. The follow-up requires a deterministic interleaving test and correction; it is not erased by this READY.

The review covers the sequential run/ci contract, not concurrent-process isolation or direct-library-caller semantics. Assertions parity is covered in the dedicated trace-assertion contract file rather than the literal extra row originally suggested for the existing parity file; this is an explicit test-location disposition, not a claim that the requested row exists. No full standalone CLI crate suite or Windows-local run was performed. Hosted checks still pending at relay time include Windows, semver and feature matrix. READY is not permission to skip those checks.

Underlying full local report: `/tmp/codex-release-662/3138-final-review-result.md`. Prior BLOCKED record is preserved. No review carry to another head is claimed.

Rul1an:
```json
{
  "head_sha": "6221e5a8ac4be955c433cfaa343d4b61d9054e59",
  "builder": {"agent": "cursor", "instance": "3116-scoped-assertion-eval"},
  "reviewer": {"agent": "muse", "instance": "review-3138-6221e5a8", "github_login": "Rul1an"},
  "review_completed": true,
  "verdict": "READY",
  "findings": [
    {"id": "F1", "summary": "Per-invocation flag reset independently verified with flagged-then-unflagged run/ci regressions and mutation control", "disposition": "fixed"},
    {"id": "F2", "summary": "Parallel per-row usage attribution deferred to issue #3140; source-derived diagnostic finding, no verdict change established", "disposition": "deferred"}
  ],
  "no_findings": false,
}
```

